PR #9, opened 2021-01-11
Added Eurom Mon Soleil 600 Ceiling Wifi Heater support

FeikoJoosten:
Title says it all, basically adds support for the [Eurom Mon Soleil Ceiling panel](https://eurom.nl/en/product/permanent-heating/infrared-heaters-permanent/mon-soleil-600-ceiling-wifi/).

It's pretty much a copy pasta of the code for the geco heater. But with different DPSs.

I've never written any code for anything Home Assistant related. So If I forgot something or something is not correct. Please let me know.


FeikoJoosten:
errr I have no clue why it's failing. Something about a `Set the SONAR_TOKEN env variable.` not being assigned. Is this something for you @make-all?

custom_components/tuya_local/device.py
```diff
@@ -98,7 +99,7 @@ async def async_inferred_type(self):
         if "8" in cached_state:
             _LOGGER.info(f"Detecting {self.name} as Goldair Fan")
             return CONF_TYPE_FAN
-        if "5" in cached_state and "3" not in cached_state:
+        if "5" in cached_state and "3" not in cached_state and "2" not in cached_state:
```
make-all, line 102: There are two variants of the Kogan switch, and one of them uses DPS 2.

If you write tests for the newly added device and run the unit tests, you would have detected that this change interfered with detection of  one variant of the Kogan Switch.

make-all (changes requested):
The Sonar integration in github doesn't work well for pull requests, it uses the submitting user's SONAR_TOKEN but the target repository's Sonar account. I wouldn't worry about the failure, except that the main thing it is checking is test coverage, and I see you have not written any unit tests for the added functionality.

custom_components/tuya_local/device.py
FeikoJoosten: Would you care to explain how I run these unit tests? First PR for something Homeassistant related 😅.

make-all: The tests are run automatically when you submit the pull request. See https://github.com/make-all/tuya-local/runs/1730523686

If you want to run them locally, I use an Ubuntu VM and basically copied the commands from the .github/worflows/tests.yml file.

FeikoJoosten: Yeah after some fiddeling around I managed to setup vs code with the correct plugins to run them through there. Next commit _should_ actually fix it this time 😓

FeikoJoosten: When running the reformat code task locally I receive the following error:
```
At line:1 char:21
+ isort --recursive . && black .
+                     ~~
The token '&&' is not a valid statement separator in this version.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : InvalidEndOfLine
```

Changing the task to `"isort --recursive . ;  black ."` allows me to run it locally. Could this be a misconfiguration?

FeikoJoosten: Finally got everything to pass (apart from the sonar test). 😅

make-all: The non-recognition of && in shell commands is probably because you are using vs-code on Windows (maybe powershell), and not a Linux VM.  Unix shells will recognize && to mean only run the second command if the first succeeds.  ; will run the second command always and ignore the return code from the first.

FeikoJoosten: Ah okay, thanks for the explanation 😄! (running through windows command line btw)

PR #26, opened 2021-07-08
Create electriq_dehumidifier.yaml

lapy:
Support for Electriq Dehumidifier CD25PRO-LE-V2

Tested on my home assistant and working very well. Only the icons don't really change based on the state for some reason.

Suggested Lovelace layout:

```
type: vertical-stack
cards:
  - type: humidifier
    entity: humidifier.electriq_dehumidifier
  - type: entity
    entity: fan.electriq_dehumidifier
    name: Settings
    attribute: preset_mode
  - type: entity
    entity: humidifier.electriq_dehumidifier
    attribute: current_humidity
    name: Current humidity
    unit: '%'
  - type: entity
    entity: humidifier.electriq_dehumidifier
    name: Oscillate
    attribute: oscillate
  - type: entity
    entity: light.electriq_dehumidifier
    name: UV sterilization
  - type: entity
    entity: switch.electriq_dehumidifier
    name: Ionizer
  - type: entity
    entity: lock.electriq_dehumidifier
    name: Child lock

```

### DIFF
--- a/custom_components/tuya_local/devices/electriq_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/electriq_dehumidifier.yaml
@@ -1,0 +1,121 @@
+name: "Electriq Dehumidifier CD25PRO-LE-V2"
+legacy_type: dehumidifier
+primary_entity:
+  entity: humidifier
+  class: dehumidifier
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+      mapping:
+        - dps_val: false
+          icon: "mdi:air-humidifier-off"
+          icon_priority: 1
+        - dps_val: true
+          icon: "mdi:air-humidifier"
+          icon_priority: 3
+    - id: 2
+      name: mode
+      type: string
+      mapping:
+        - dps_val: "auto"
+          value: "Auto"
+        - dps_val: "low"
+          value: "Low"
+        - dps_val: "high"
+          value: "High"
+          icon: "mdi:tshirt-crew-outline"
+          icon_priority: 2
+        - dps_val: "fan"
+          value: "Air clean"
+          icon: "mdi:air-purifier"
+          icon_priority: 2
+    - id: 3
+      type: integer
+      name: current_humidity
+      readonly: true
+    - id: 4
+      type: integer
+      name: humidity
+      range:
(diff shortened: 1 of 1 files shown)


make-all:
Did you test this on top of release 0.8.4, or latest HEAD?  Icons are only working in latest head, I haven't yet released that change.

lapy:
Hi, I am running latest from HACS, so 0.8.4. Ok it makes sense then. PS thanks for maintaining this integration, it made it so easy to integrate this dehumidifier! :)

make-all:
I made some changes to the config installed. You can try it out now by selecting "Reinstall" from the integration's menu in HACS, and selecting "main" as the version to install.

I used preset_mode rather than oscillate for the fan's 45/90/0_90 setting, as that allows all three selections to be available. in the UI. The previous preset_mode was a duplicate of the humidifier's mode, so not needed in both places.

lapy:
Oh yes, it was bugging me that I could not find a way to fit in the 45deg setting for the louvre. Also for the UV and Ionizer I had to make the UV a light entity as it's not possible to have 2 switches entities.

Testing now, thank you!

make-all:
Yes duplicate entities of the same type are on my TODO list.  I will need them when I implement the sensor entity and break out the current_temperature and current_humidity into their own sensors to make them easier to work with.

lapy:
Oh yes, I haven't thought about that, it will be so nice!

lapy:
By the way, tested the changes in main, I had to delete integration and add it again, and it works flawlessly, also the louvre aperture, and the icons are working. I noticed you removed the t-shirt icon for the high setting, even though that setting is for drying clothes as the manual says, but it's not that important.

make-all:
I've added the T-shirt icon back, though I won't urgently make another release for that.

The reason you needed to delete and add it again is that you included `legacy_type: dehumidifier` in your config file. This made it shadow the goldair dehumidifier since electriq sorts before goldair and broke some of the tests.  The legacy_type is only for backwards compatibility with the upstream project that only supports goldair devices, so shouldn't be used for new devices.  After removing that line from the config file, your setup was now pointing back at the goldair_dehumidifier config.

PR #28, opened 2021-07-10
Add support for Poolex Vertigo FI heatpump

lperez31:
Inspired from previous commit for Poolex Silverline FI heatpump.
The preset names and error code of Poolex Vertigo FI are different from those of the Poolex Silverline FI.


make-all:
Thanks. I've added this in as well, and renamed the other one to better distinguish them.

Issue #76, opened 2021-11-20
Add support for Madimack Elite V3


petrkotek:
Hi @make-all,

first of all, thank you so much for this fantastic HA integration!

Few days back, I got a brand new heat pump "Madimack V3 Elite Series 13.8kw 2022 Model" installed and tried to set it up with this integration.

Looks like the DPS mapping is different from the previous versions as this Integration detected it only as a `simple_switch`.

Good thing is, that turning it on & off works.

I also tried "Local Tuya" integration and was able to manually add all the sensors and map them to various the sensors & states. I'll try to create the device yaml definition myself (incl. tests), however, for the reference I'm sharing the mappings in the meantime.

If there is anything else I should know, please do let me know. Thanks!

## Mapped Sensors/States
- ID1 - Power On [`True`/`False`]
- ID2 - Mode [`heating`, `cold`, `auto`]
- ID4 - Target temperature (in °C or °F)
- ID5 - Work Mode [`silence`,`power`,`boost`] which slightly confusingly map to "Silence", "Perfect" and "Power" in the Smart Life app UI.
- ID6 - Temperature unit [`c` or `f`]
- ID20 - "Compressor speed" aka "Frequency percentage" aka "Running speed percentage" [0-100]
- ID21 - maximum temperature for current Mode [40 or 30 or equivalent in °F]
- ID22 - minimum temperature for current Mode [18 or 12 or equivalent in °F]
- ID23 - (C4) Evaporator coil pipe temperature (in °C or °F)
- ID24 - (C3) Exhaust gas temperature (in °C or °F)
- ID25 - (C1) Outlet water temperature (in °C or °F)
- ID26 - (C2) Ambient temperature (in °C or °F)
- ID102 - (C0) Inlet water temperature (in °C or °F)
- ID103 - (C5) Return gas temperature (in °C or °F)
- ID104 - (C6) Cooling coil pipe temperature (in °C or °F)
- ID105 - (C9) Cooling plate temperature (in °C or °F)
- ID106 - EEV opening [`480` in my case]

## Unknown DPS/States
- ID15 - Unknown - my state is currently `0` and doesn't change
- ID101 - Unknown - my state is currently `0` and doesn't change
- ID107 - Unknown - my state is currently `False` and doesn't change

## Device firmware versions
- Main Module: `V2.1.6`
- MCU Module: `V1.0.0`

PR #77, opened 2021-11-20
Add support for Madimack Elive V3 heat pump (work in progress)

petrkotek:
Adds support for Madimack Elive V3 heat pump.

There's still quite a few things to be resolved:

- [ ] add tests
- [ ] confirm name & filename
- [ ] fix ranges for temperature
  - is it possible to have it dynamic (based on min_temperature (dps 22) and max_temperature (dps 21))?
  - or, is it possible to support different ranges based on two conditions - 1/ mode (heat/cool/auto) and 2/ temperature_unit? (c/f)
- [ ] add additional sensors as `secondary_entities` (nice to have)

This addresses https://github.com/make-all/tuya-local/issues/76.

### DIFF
--- a/custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
+++ b/custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
@@ -1,0 +1,151 @@
+name: Madimack Elite V3 Pool Heatpump
+primary_entity:
+  entity: climate
+  dps:
+    - id: 1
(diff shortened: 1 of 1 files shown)


make-all:
Don't worry about the sonarcloud failure, it doesn't seem to work for pull requests.

When you add more sensors, you'll need to also add strings in translations/en.json.  The unit tests will pick up any missing strings.

PR #80, opened 2021-11-24
Add support for Aspen smart fan

irakhlin:
Hello,

I have added a yaml device configuration for a window smart fan I use currently available in the US, [here](https://www.amazon.com/ASPEN-Controlled-Multi-Function-Detachable-Reversible/dp/B08KTQVN93)

The manual can be seen [here](https://static1.squarespace.com/static/5f7ce77e9aaa054d65fed97d/t/5f99a03c6dbd107a6123ede2/1603903564723/ASP-200+User+Manual.pdf)

Features:

- Turn On/OFF
- Set display brightness 1/2/3
- Set fan speed (High/Med/Low)
- Set fan direction ( in, out, exchange)
- View current temperature
- Enable/Disable 'temperature set point mode' and setting a temperature if enabled

The device allows control from a simple fan entity or a climate entity for easier integration with lovelace climate entity cards.

### DIFF
--- a/custom_components/tuya_local/devices/aspen_fan.yaml
+++ b/custom_components/tuya_local/devices/aspen_fan.yaml
@@ -1,0 +1,95 @@
+name: Aspen Fan
+primary_entity:
+  entity: fan
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+    - id: 2
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: in
+          value: cool
+        - dps_val: out
+          value: exhaust
+        - dps_val: exch
+          value: exchange
+    - id: 3
+      type: integer
+      name: speed
+      mapping:
+        - dps_val: 1
+          value: low
+        - dps_val: 2
+          value: medium
+        - dps_val: 3
+          value: high
+secondary_entities:
+  - entity: climate
+    name: Bedroom Fan
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 2
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: in
+            value: intake
+          - dps_val: out
+            value: exhaust
+          - dps_val: exch
+            value: exchange
+      - id: 101
+        type: boolean
+        mapping:
+          - dps_val: false
+            value: "fan_only"
+            icon: "mdi:fan-off"
+          - dps_val: true
+            value: "cool"
+            icon: "mdi:fan"
+        name: hvac_mode
+      - id: 3
+        type: integer
+        name: fan_mode
+        mapping:
+          - dps_val: 1
+            value: low
+            icon: "mdi:fan-speed-1"
+          - dps_val: 2
+            value: medium
+            icon: "mdi:fan-speed-2"
+          - dps_val: 3
+            value: high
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/aspen_fan.yaml
```diff
@@ -0,0 +1,95 @@
...
+      name: speed
+      mapping:
+        - dps_val: 1
+          value: low
+        - dps_val: 2
+          value: medium
+        - dps_val: 3
+          value: high
+secondary_entities:
+  - entity: climate
```
make-all, line 29: I think its better to have either Fan + current temperature (sensor) + temperature (number), or a climate entity.

custom_components/tuya_local/devices/aspen_fan.yaml
```diff
@@ -0,0 +1,95 @@
...
+        mapping:
+          - dps_val: in
+            value: intake
+          - dps_val: out
+            value: exhaust
+          - dps_val: exch
+            value: exchange
+      - id: 101
+        type: boolean
+        mapping:
```
make-all, line 47: Should it be "off" and "fan_only" rather than "fan_only" and "cool"?  Or is this an evaporative cooler or something?  In that case, climate as primary entity would be better than fan.

make-all: It looks like this should interact with dps 1, so add `hidden: true` to dps 1, and here, add
```
    constraint: switch
    conditions:
       - dps_val: false
         value: "off"
       - dps_val: true
          value: fan_only
```
for both the fan_only and cool mappings.

irakhlin: > 

id: 101, this controls the "temperature/climate set point function. It has two modes, the way I have them named right now:
fan_only: the fan runs regardless of the room temperature
cool: works like the 'cool' function on an AC, as the room reaches the set temperature the fan will turn off, and turn back on if the room temperature rises. I can rename the values but I was worried "off" may be confusing as this does not turn off the fan. This is also the reason I had added it as a climate device and a fan device, this way it was easier to integrate with climate lovelace cards. Let me know what you think the best way to do this.

custom_components/tuya_local/devices/aspen_fan.yaml
irakhlin: I mentioned this in the response below but the reason for this is the functionality of id:101. If this is on it functions like a "cooling device" it allows you to set a "desired" temperature and the fan will run until reaches this temperature, to me this seemed more like the functionality of a climate device. At the same time this is not a feature I used much so I wanted an alternative for control. Is it an issue to have both? If so which one do you think I should remove?

custom_components/tuya_local/devices/aspen_fan.yaml
make-all: I think because of the thermostat feature, it is a better fit with the climate entity rather than fan. But I think it is the switch (dps 1) that is HVAC mode, with off and fan_only as values. Dps 101 should be preset mode, I think, selecting between "constant" and "auto" (or similar names), with temperature setting invalid when it is false.
The fan direction should be a separate select entity I think.

PR #83, opened 2021-11-26
add TMWF02 Wi-Fi Fan Module

vampywiz17:
add TMWF02 Wi-Fi Fan Module. It support On/Off and speed via percentage. (same like in tuya app)

### DIFF
--- a/custom_components/tuya_local/devices/tmwf02_fan.yaml
+++ b/custom_components/tuya_local/devices/tmwf02_fan.yaml
@@ -15,17 +15,6 @@ primary_entity:
     - id: 3
       type: string
       name: fan_level
-      mapping:
-        - dps_val: level_1
-          value: level_1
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/tmwf02_fan.yaml
```diff
@@ -12,6 +12,21 @@ primary_entity:
       range:
         min: 1
         max: 100
+    - id: 3
+      type: string
+      name: fan_level
+      mapping:
```
make-all, line 18: No need for the mapping if it is hidden.  The mapping is only to 1) map to different values, 2) to make the list of possible values known to the UI.  If it is hidden, there is no need for either of these.

vampywiz17: Ok, I delete it.

PR #88, opened 2021-11-29
Add  timer function to simple switch and add Woox R4028 + DIGOO DG-PS01 + DIGOO DG-SP202

vampywiz17:
I checked 4-5 brand of simple plug (without led or measuring) and it almost all only contain this two dp_id. I think this config help to use this switches all available local function. But also usefull that the original simple switch config is also remain, so i add it to a different name.


custom_components/tuya_local/translations/en.json
```diff
@@ -61,6 +61,8 @@
 		    "number_timer": "Include timer as a number entity",
 		    "number_timer_1": "Include timer 1 as a number entity",
 		    "number_timer_2": "Include timer 2 as a number entity",
+		    "number_timer_3": "Include timer 2 as a number entity",
```
make-all, line 64: Don't forget to change the numbers in the translation strings as well.

custom_components/tuya_local/translations/en.json
```diff
@@ -61,6 +61,8 @@
...
+		    "number_timer_4": "Include timer 2 as a number entity",
```
make-all, line 65: I think this should be `number_timer_usb`

custom_components/tuya_local/translations/en.json
```diff
@@ -101,6 +103,8 @@
 		    "switch_open_window_detector": "Include open window detect as a switch entity",
 		    "switch_outlet_1": "Include outlet 1 as a switch entity",
 		    "switch_outlet_2": "Include outlet 2 as a switch entity",
+		    "switch_outlet_3": "Include outlet 3 as a switch entity",
+		    "switch_usb_1": "Include USB 1 as a switch entity",
```
make-all, line 107: `switch_usb`

custom_components/tuya_local/devices/woox_r4028_powerstrip.yaml
```diff
@@ -0,0 +1,83 @@
+name: Woox R4028 + DIGOO DG-PS01
+primary_entity:
+  entity: switch
+  name: switch 1
```
make-all: To be consistent with the other supported multi-switch device and avoid adding new translations unnecessarily, the name should be `Outlet 1` (similarly for 2 and 3)

custom_components/tuya_local/devices/woox_r4028_powerstrip.yaml
```diff
@@ -0,0 +1,83 @@
...
+        name: switch
+  - entity: switch
+    name: switch 3
+    class: switch
+    dps:
+      - id: 3
+        type: boolean
+        name: switch
+  - entity: switch
+    name: switch USB
```
make-all: I think just USB is better here.  Natural English word order would be "USB switch", but the switch is redundant since the name is attached to a switch entity.

vampywiz17:
I hope fix this things and try to standardize.

vampywiz17:
@make-all 

It need to change any other thing? I tested it, but not work, only the existing strings (outlet 1 2, timer 1 2....)

make-all:
The unit tests have some clue, but only one string is missing, I don't know why the others are not showing.

vampywiz17:
@make-all 

strange. Today morning i re-add the device to HA (because change the device IP address), and now show this:

262049596_419224349894658_5628095656015943213_n

It possible that is a HA cache?

vampywiz17:
all devices are tested and all working without any problem. Not plan to do any changes. If you found any bug, please write it to me.

Issue #84, opened 2021-11-27
Add support for Beca BAC-002-ALW


shakin89:
Hi,
first of all thank you for this fantastic integration! 
If there is a way i can help to support this new device let me know how i can help.

I've bought a brand new [Beca Smart BAC-002-ALW](http://www.becaenergy.com/index.php/Home/New/productsinfo/id/76.html)

I've been able to obtain the key and add it to local tuya, but the thermostat is not recognized. 
The log reports this

2021-11-27 10:35:06 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches arlec_fan with quality of 62%. DPS: {'1': True, '2': 44, '3': 41, '4': '1', '5': False, '6': False, '102': '1', '103': '2', 'updated_at': 1638005704.286728}
2021-11-27 10:40:17 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches arlec_fan with quality of 62%. DPS: {'1': True, '2': 44, '3': 43, '4': '1', '5': False, '6': False, '102': '1', '103': '2', 'updated_at': 1638006015.372264}
2021-11-27 10:41:33 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches arlec_fan with quality of 62%. DPS: {'1': True, '2': 44, '3': 43, '4': '1', '5': False, '6': False, '102': '1', '103': '2', 'updated_at': 1638006090.981737}

the DPS is
{
  '1': true,
  '2': 39,
  '3': 45,
  '4': '1',
  '5': false,
  '6': false,
  '102': '1',
  '103': '2'
}

### **the mapped sensor states:**
1: powered 'true'/'false' -> sting
2: target temperature -> int
3: actual temperature  -> int
4: manual mode / program mode 0=program, 1= manual
5: eco mode: 'true'/'false' -> sting
6: lock mode 'true'/'false' -> sting
102: operation '0'=cool, '1'=heat, '2'=fan (string)
103: fan speed '0'=auto, '1'=high, '2'=med, '3'=low (string)

vampywiz17:
@shakin89 

You able to control it via tuya-cli?

vampywiz17:
@shakin89 you able to answer me that what type of 102 and 103? integer maybe...

vampywiz17:
@shakin89 it is a rough test. Please try it!

https://pastebin.com/4nNedWTG

shakin89:
[...]

PR #97, opened 2021-12-19
Fixed current_temperature and temperature

zigul:
current_temperature is 3
actual temperature is 2


zigul:
#84

make-all:
I fixed this before seeing this PR.  Basically the same with an additional change for the unit test.

PR #128, opened 2022-02-03
Created Himox H06 Air Purifier .yaml

dieantu:
Hi! I just added the Himox H06 Air Purifier device to the list of supported devices.

I discovered the _dps_ with an older Smart Life apk and a rooted android. All functions seems to work fine with my device. All dps variants are covered and there's no log warnings or errors (during my testing I received quality warnings but seems to work smoothly now).

I need to say that this is my ever first contribution to a Home Assistant Integration. Actually, this is my **first** ever "pull request" in Github of my life. Thank you for your work and for reviewing this little YAML file. I am very excited to be here and to be alive. This fact is really very important to me. 

So please, I appreciate any type of feedback. Hope I'm doing this in a correct way.

Thank you so much <3

### DIFF
--- a/custom_components/tuya_local/devices/himox_h06_purifier.yaml
+++ b/custom_components/tuya_local/devices/himox_h06_purifier.yaml
@@ -72,7 +72,7 @@ secondary_entities:
         name: switch ##!! Device matches himox_h06_purifier with quality of 44%. !!##
   - entity: switch
     dps:
-      - id: 8
+      - id: 11
         type: boolean
         name: switch ##!! Device matches himox_h06_purifier with quality of 44%. !!##
 
--- a/custom_components/tuya_local/devices/himox_h06_purifier.yaml
+++ b/custom_components/tuya_local/devices/himox_h06_purifier.yaml
@@ -1,0 +1,105 @@
+---
+## @antudev custom addition to tuya_local ##
+
+#{dps.id.101};{name.switch};{type.boolean}#{'1': True,
+#{dps.id.104};{name.fan_speed};{type.string}#'4': 'low',
+#{dps.id.105};{name.filter};{type.integrer};{unit.'%'};#'5': 47,
+#{dps.id.108};{name.light_switch};{type.boolean}#'8': False,
+#{dps.id.111};{name.filter_reset};{type.boolean}#'11': False,
+#{dps.id.119};{name.countdown_left};{type.integrer};{unit.min}#'19': 0,
+#{dps.id.122};{name.string}#'22': 'severe',
+
+  # 2022-02-02 15:55:31 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches himox_h06_purifier with quality of 33%. DPS:
+
+  # {'1': True,
+  # '4': "property":{"range":["low","mid","high"];"type":"enum", rw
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/himox_h06_purifier.yaml
```diff
@@ -0,0 +1,97 @@
+name: himox h06
+products:
+ - id: 80256465f4cfa2c5ec84
```
make-all, line 3: Product ID should be alphanumeric.  This looks like a device id, which is specific to your device.

custom_components/tuya_local/devices/himox_h06_purifier.yaml
```diff
@@ -0,0 +1,97 @@
...
+      mapping:
+        - dps_val: "low"
+          value: 33.34
+        - dps_val: "mid"
+          value: 66.67
+        - dps_val: "high"
+          value: 100
+secondary_entities:
+  - entity: sensor
+    name: active_filter_life
```
make-all, line 22: Entity names should be as they should appear in the UI.  The names will be automatically downcased and slugified for use in the default entity ID.  So for this, Active filter life (is it an active filter?)

custom_components/tuya_local/devices/himox_h06_purifier.yaml
```diff
@@ -0,0 +1,97 @@
...
+        type: string
+        mapping:
+          - dps_val: cancel
+            value: "Off"
+          - dps_val: 4h
+            value: "4 hours"
+          - dps_val: 8h
+            value: "8 hours"
+  - entity: sensor
+    name: countdown_left
```
make-all, line 64: To be consistent with other devices, I suggest using Timer for the name.

custom_components/tuya_local/devices/himox_h06_purifier.yaml
```diff
@@ -0,0 +1,97 @@
...
+        readonly: true
+        unit: "min"
+        class: measurement
+        range:
+          min: 0
+          max: 480
+        scale: 0
+        step: 1
+  - entity: sensor
+    name: aq_indicator
```
make-all, line 78: For consistency with other devices, use Air Quality as the name here.

custom_components/tuya_local/devices/himox_h06_purifier.yaml
```diff
@@ -0,0 +1,97 @@
...
+    dps:
+      - id: 22
+        type: string
+        name: sensor
+        readonly: true
+  - entity: select
+    name: air_debugger
+    category: diagnostic
```
make-all, line 86: Controls that you can set cannot be diagnostic, they should be config.  As a general comment, I would review all the secondary entities, things that are seldom used config should be marked as category: config, sensors and binary sensors that are not providing primary information should be marked as diagnostic.  Those that are main controls or sensors for the device should be left unmarked.

custom_components/tuya_local/devices/himox_h06_purifier.yaml
```diff
@@ -0,0 +1,97 @@
...
+        scale: 0
+        step: 1
+  - entity: sensor
+    name: aq_indicator
+    dps:
+      - id: 22
+        type: string
+        name: sensor
+        readonly: true
+  - entity: select
```
make-all, line 84: I think the fan entity's preset_mode is still available, which might be better for this.

PR #157, opened 2022-04-11
Support for Wilfa Haze Humidifier

daitj:
Temperate unit needs some fixes.
I have no clue about the diagnostics sensors dps 20 and 22, I will add them once I run into them.

### DIFF
--- a/custom_components/tuya_local/devices/wilfa_haze_hu400bc_humidifier.yaml
+++ b/custom_components/tuya_local/devices/wilfa_haze_hu400bc_humidifier.yaml
@@ -1,0 +1,186 @@
+name: Wilfa Haze
+products:
+  - id: 7naqdm8whbezn28v
+primary_entity:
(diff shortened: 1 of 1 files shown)


make-all:
To take care of some of the translation issues, and after consulting the manual here: https://www.wilfa.fi/wp-content/uploads/2021/08/HU-400BC_20_01.pdf, I made the following naming changes:

light: Sleep -> Display (this already exists as a name, so it easier to support - according to the manual, this turns off both the display and the water tank indicator, the latter has a separate control also that seems to be the "Mood" light in your config)

switch: Heater -> Air Clean (this already exists as a name, and according to the manual, although the mechanism is by heating, the purpose is for "Degerming" the air)

Also, I moved the intensity setting to a fan entity.  As an unsupported attribute on the humidifier it cannot be set.

make-all:
Rebased and merged

PR #158, opened 2022-04-11
Fixed some dps value for Wilfa Haze

daitj:
In my initial PR dps and value shown were flipped, so I have fixed that. 

Also I have added step to the humidity, I tried under id 3, under range and under mapping but I do not know if I did something wrong, it doesn't seem to work, is there some doc on how to add new device?


daitj:
also about the mood, I do not have it in my device, it was just something I saw in Smart Life app, it did change dps 5 but did nothing.

make-all:
Has been rebased and merged

PR #169, opened 2022-05-09
add support for QS-WIFI-C01(BK) Curtain Module

choppedpork:
Adds support for a very no-name curtain / blind controller (I opted for "Curtain Module" as that's what the packaging and unit itself say) as found on [ebay](https://www.ebay.co.uk/itm/384641995828). The device belongs to the [`clkg` category](https://developer.tuya.com/en/docs/iot/category-clkg?id=Kaiuz0gitil39) but is very simple and supports only four of the instructions (w/o any read-only status DPS). Do let me know if I missed anything!

The translations file had a mix of tabs and spaces, my editor decided to clean that up via autoformat on save but also expanded a couple of objects across more lines, happy to revert that if it's a step too far 🙂 

Thanks for a great integration!

edit: just noticed that the diff view of the translation file doesn't actually show the added entries amongst formatting changes, the entries are `select_motor_reverse_mode` and `number_travel_time_configuration`.


make-all:
Merged

make-all:
The final version renames the config to follow the existing convention of having the type of device last in the name rather than first.  So any devices you have configured will need reconfiguring after the update.

PR #181, opened 2022-06-14
Add Jiahong ET-72W Thermostate

craibo:
- Add Jiahong ET-72W Thermostate C and F configs
- Tested

### DIFF
--- a/custom_components/tuya_local/devices/jiahong_et72w_thermostat_f.yaml
+++ b/custom_components/tuya_local/devices/jiahong_et72w_thermostat_f.yaml
@@ -1,0 +1,225 @@
+name: Jiahong ET-72W Thermostat (F)
+primary_entity:
+  entity: climate
+  dps:
+    - id: 101
+      type: boolean
+      name: power
+      mapping:
+        - dps_val: false
+          value: "off"
+      hidden: true
+    - id: 102
+      type: integer
+      name: temperature
+      range:
+        min: 120
+        max: 750
+      mapping:
+        - step: 30
+          scale: 10
+    - id: 103
+      type: string
+      name: hvac_mode
+      mapping:
+        - dps_val: "Smart"
+          icon: "mdi:calendar-sync"
+          constraint: power
+          conditions:
+            - dps_val: true
+              value: auto
+            - dps_val: false
+              value_redirect: power
+              value: "off"
+        - dps_val: "Manual"
+          icon: "mdi:fire"
+          constraint: power
+          conditions:
+            - dps_val: true
+              value: heat
+            - dps_val: false
+              value_redirect: power
+              value: "off"
+        - dps_val: "Anti_frozen"
+          icon: "mdi:snowflake"
+          constraint: power
+          conditions:
+            - dps_val: true
+              value: cool
+            - dps_val: false
+              value_redirect: power
+              value: "off"
+    - id: 105
+      type: integer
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/jiahong_et72w_thermostat_f.yaml
```diff
@@ -0,0 +1,225 @@
...
+      name: power
+      mapping:
+        - dps_val: false
+          value: "off"
+      hidden: true
+    - id: 102
+      type: integer
+      name: temperature
+      range:
+        min: 120
```
make-all, line 16: Does this range come from a manual or tuya cloud api, or did you calculate it yourself?
It doesn't seem right, as it equates to a range of -12C - 24C with a step of 2C.  If you calculated yourself, I can understand that you might have forgotten to add the 32 degree offset at the end, and also may have inverted somewhere in the step calculation (normally devices with 0.5C step have 1F steps).  If it came from docs, then maybe the device designers made these mistakes, making the device practically unusable in F, but in that case the config should reflect the actual device limits.
My guess would be a 400-1000 range with 10 step, which gives 4-38, not quite exactly matching the C range but with round numbers that are very close.

craibo: Hi @make-all

The range on the C configuration is correct. 
I messed up the conversion to F. 
I'll raise another PR with the corrected range and include the test cases when I have a chance.
Thanks for catching this

craibo: Hi @make-all 

I just reviewed your PR's to merge the configs into a single one. I had tried that but couldn't get the secondary entities to work correctly. I see how you did it an know what I did wrong.

make-all:
Is the energy sensor for this device returning useful values?

In a previous change I pessimistically removed the energy class from all devices that were not smart meters, as in general (at least for smart switches), the energy readings are not reliable enough to be useful, in many cases reporting 1 for an increment then immediately auto-resetting to 0.

craibo:
Hi @make-all 

I found a different issue with the energy sensor but haven't had a chance to fix it. 

The energy sensor is actually tracking the time (in minutes) that the device is on and active. 
It is a continually increasing value and doesn't reset to 0.
There is a second sensor that allows choosing the resistance for the under-floor heating cable.
When you combine those 2 sensors you get the approximated energy consumption. 

It would be good to be able to fix this so that we can get the energy usage. However just fixing the sensor to be a unit of time sensor instead of an energy sensor and using a template to create the energy sensor is a work around that will work

make-all:
The "second sensor that allows choosing the resistance" is dp 116, or something else?

craibo:
Hi @make-all 

Ok I have confirmed the sensors values correctly now

```
dp 116: Is the `Sensor Type` it has 4 options `[0,3950, 3600, 3700]`, Default is `3950`
dp 117: `Run Time` in (secs)
dp 112: `Floor Sensor Calibration` values `[-5:5]` degrees Celsius , increments `1`
dp 113: `Room Sensor Calibration` values `[-5:5]` degrees Celsius , increments `1`
```

Energy is calculated by `(dp 116) * (dp 117) / 1000` in kWh 

I will try get to creating a PR with these changes over the Christmas period while I have a little downtime.

Thanks

PR #208, opened 2022-07-22
Add support for Nexxt 220v Smartplug

x-keita:
Features: 
- Switch
- Timer


make-all:
I've chosen to rename this to a generic name, as dp 9 is commonly used for timer by other devices, so I think it will be likely to match other switches.

Issue #216, opened 2022-08-04
New device: Catit Pixi Smart Fountain 


alexmaras:
Device: Catit Pixi Smart Cat Fountain

Note: Tuya isn't documented for this fountain anywhere - they use a branded app for this - but pairing works perfectly when configuring in the app, and all relevant settings work. 

The functionality includes:
- switching the fountain on and off
- seeing the water level
- time until filter needs replacing
- resetting the filter life after replacement
- time until pump needs cleaning
- resetting the pump cleaning notification after cleaning
- running a UV cycle
- time that the UV cycle has been running for
- "energy saving" mode - though I'm not sure if this is perhaps implemented server-side to send off/on events to the device?
- timer mode - only runs between time x and y - not sure if implemented client side or server-side

I'm not sure what "water time" and "water reset" are - there's no notifications to change the water as far as I can tell, just one that tells you when the water level is low. 

Instructions:
https://www.catit.com/shared/Product-Manual_PIXI-Smart-Fountain_NA.pdf

API Explorer result:
```
{
  "result": {
    "category": "cwysj",
    "functions": [
      {
        "code": "switch",
        "desc": "{}",
        "name": "开关",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "water_reset",
        "desc": "{}",
        "name": "用水重置",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "filter_reset",
        "desc": "{}",
        "name": "滤芯复位",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "pump_reset",
        "desc": "{}",
        "name": "水泵重置",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "uv",
        "desc": "{}",
        "name": "UV杀菌",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "uv_runtime",
        "desc": "{\"unit\":\"s\",\"min\":0,\"max\":10800,\"scale\":0,\"step\":1}",
        "name": "UV运行时间",
        "type": "Integer",
        "values": "{\"unit\":\"s\",\"min\":0,\"max\":10800,\"scale\":0,\"step\":1}"
      }
    ],
    "status": [
      {
        "code": "switch",
        "name": "开关",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "water_time",
        "name": "用水时长",
        "type": "Integer",
[...]

PR #217, opened 2022-08-04
Add support for Catit Pixi Smart Fountain

alexmaras:
Adds initial support for the Catit Pixi Smart Fountain.

I've managed to successfully test the power switch and UV functionality, which turn on and off appropriately and report their status. I'm not 100% sure how I'm supposed to handle the runtime numbers and stuff - I'd appreciate some guidance on that - and I'm not sure whether there's existing support for the `Enum` tuya type, which this needs for the Water Level sensor

In relation to: https://github.com/make-all/tuya-local/issues/216

### DIFF
--- a/custom_components/tuya_local/devices/catit_pixi_smart_fountain.yaml
+++ b/custom_components/tuya_local/devices/catit_pixi_smart_fountain.yaml
@@ -29,7 +29,7 @@ secondary_entities:
           min: 0
           max: 43200
         mapping:
-          - scale: 0
+          - scale: 1
             step: 1
         unit: min
   - entity: switch
@@ -52,6 +52,6 @@ secondary_entities:
           min: 0
           max: 10800
         mapping:
-          - scale: 0
+          - scale: 1
             step: 1
         unit: s
--- a/custom_components/tuya_local/devices/catit_pixi_smart_fountain.yaml
+++ b/custom_components/tuya_local/devices/catit_pixi_smart_fountain.yaml
@@ -1,0 +1,57 @@
+name: Catit Pixi Smart Fountain
+products:
+  - id: z3rpyvznfcch99aa
+primary_entity:
+  entity: switch
+  icon: "mdi:water-pump"
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+secondary_entities:
+  - entity: switch
+    name: Filter Replaced
+    icon: "mdi:air-filter"
+    category: config
+    dps:
+      - id: 7
+        type: boolean
+        name: switch
+  - entity: number
+    name: Filter Life
+    icon: "mdi:air-filter"
+    category: config
+    dps:
+      - id: 4
+        type: integer
+        name: value
+        range:
+          min: 0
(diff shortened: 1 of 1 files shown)


alexmaras:
```
homeassistant    | 2022-08-04 21:58:09 ERROR (MainThread) [homeassistant.components.number] Error while setting up tuya_local platform for number
homeassistant    | Traceback (most recent call last):
homeassistant    |   File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 258, in _async_setup_platform
homeassistant    |     await asyncio.gather(*pending)
homeassistant    |   File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 397, in async_add_entities
homeassistant    |     await asyncio.gather(*tasks)
homeassistant    |   File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 540, in _async_add_entity
homeassistant    |     capabilities=entity.capability_attributes,
homeassistant    |   File "/usr/src/homeassistant/homeassistant/components/number/__init__.py", line 233, in capability_attributes
homeassistant    |     ATTR_MIN: self.min_value,
homeassistant    |   File "/usr/src/homeassistant/homeassistant/components/number/__init__.py", line 264, in min_value
homeassistant    |     return self._convert_to_state_value(self.native_min_value, floor_decimal)
homeassistant    |   File "/config/custom_components/tuya_local/generic/number.py", line 40, in native_min_value
homeassistant    |     r = self._value_dps.range(self._device)
homeassistant    |   File "/config/custom_components/tuya_local/helpers/device_config.py", line 432, in range
homeassistant    |     return _scale_range(r, scale)
homeassistant    |   File "/config/custom_components/tuya_local/helpers/device_config.py", line 41, in _scale_range
homeassistant    |     return {"min": r["min"] / s, "max": r["max"] / s}
homeassistant    | ZeroDivisionError: division by zero
```

I'm having a look at this issue at the moment

alexmaras:
Fixed the issue with scale of 1. The data suggested it should be zero, but I don't think that makes sense - that's probably just their way of ignoring the scale. 

image

alexmaras:
I suspect I need to set the number values to read-only, looking at that next

alexmaras:
image
This is working better now, but I'm a little annoyed by my integers with `step: 1` + `scale: x` showing up as non-integers. Looking for a nice way to make them round at the moment.

make-all:
Tuya's idea of scale seems to be a number of decimal places to shift (actual_val = dps_val * 10 ^ scale).  But the first device I encountered that needed scale was a thermostat that doubled the deg C values to allow for 0.5 steps.  I'm not sure how exactly Tuya represents that case as iot.tuya.com didn't exist back then.  If the numbers are read-only, they should probably be sensor entities rather than number entities.  And an enum for a read-only value can just be a string - if it is read-write, add a mapping with all the possible dps_val listed, and mapped to the values you want to see in the UI (because tuya or the device manufacturer doesn't always choose the best names for things due to machine translation or whatever).

Issue #176, opened 2022-06-09
DPS update speed


myhomeiot:
Thanks for great integration!

I'm thinking to move few my devices from [localtuya](https://github.com/rospogrigio/localtuya) integration. Everything works much better with your integration but I found that localtuya integration updates device dps much faster. For example if I turn on/off or change target temperature of my dehumidifier using Tuya mobile application, or PM25 sensor was updated, the localtuya integration get the new data almost instantly in your integration it's takes around 20-25 seconds. Can you explain this? Is it possible to make this faster?

make-all:
I have not looked closely at how localtuya is handling updates, but they have their own forked version of pytuya which they have customised themselves (this integration is using tinytuya, which is another derivative of the original pytuya).  There are a couple of things that might improve the speed 
 
 1. as an early design decision, the upstream developer (nicole-ashley/homeassistant-goldair-climate) chose to poll the devices rather than use a persistent connection to allow co-existence of the mobile app and home assistant.  Co-existence doesn't really work that well, since the app doesn't similarly co-operate, so a persistent connection might help. Though I did try it once, and any improvement didn't seem to be automatic, since the poll commands still need to be sent over the persistent connection, but it allow the poll rate to be increased without overloading the tuya device.
 2. Tuya devices send out broadcast messages. I have thought about using these for discovery, but I suspect that they may also help with getting push updates from the device without polling.

make-all:
Related to #116

myhomeiot:
Thanks for explanation, polling explains the slow updates.

About co-existence with Tuya mobile app, I know from the forum that many Tuya compatible bulbs can handle only one connection but fortunately for me all my Tuya device can handle more than one connection and I doesn't have this problems and unfortunately I can't test this case.

fabianluque:
I second this request. I have a fan I have configured that definitely updates instantly under Local Tuya and takes a little longer with this integration. 

Another issue I’m experiencing that might be related is that a quick succession of commands like turn on/off a light sometimes results in some of them to be delayed.

make-all:
It seems the way that localtuya does it is to use a persistent connection with asyncio, but it isn't as simple as making the connection persistent, the way the connection is handled also needs to change.  tinytuya has added support for persistent connections, so refactoring device.py to use it should give a big improvement to update speed for external status changes, though it still probably will not be sufficient to support battery powered sensors.

Fannangir:
@make-all, just updated to the new version.
[...]

PR #231, opened 2022-09-10
Add support Dongguan Garage Door

myhomeiot:
It's looks like noname Chinese Door Controller, one which I have and tested is produced by Dongguan Pinjia Technology Co. and has model number PJ-WGDOOR01 can be found as [Tuya Smart Life WiFi Motorized Swing Gate Sliding Gate Opener Dry Contact 110V 240V Voice Control with Alexa Google Home](https://www.aliexpress.com/item/1005003913566298.html)


myhomeiot:
Looking forward for fix #176 without it status update works very slooow.

PR #251, opened 2022-10-24
Add support for water heater devices and the Hydrotherm DYNAMIC/X8 heat pump hot water system

fsevilla3:
Hi @make-all. Thanks for your work on tuya-local. I love how clean and abstracted your code is. Please accept this pull request, which adds generic support for `water_heater` devices and support for the Hydrotherm DYNAMIC/X8 heat pump hot water system. I've tested this and it works for me.

### DIFF
--- a/custom_components/tuya_local/const.py
+++ b/custom_components/tuya_local/const.py
@@ -11,5 +11,6 @@ CONF_LIGHT = "light"
 CONF_LOCK = "lock"
 CONF_SWITCH = "switch"
 CONF_HUMIDIFIER = "humidifier"
+CONF_WATER_HEATER = "water_heater"
 API_PROTOCOL_VERSIONS = [3.3, 3.1, 3.2, 3.4]
 SCAN_INTERVAL = timedelta(seconds=30)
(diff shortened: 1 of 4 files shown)


make-all:
Thank you, I had thought about water_heater support for the pool heaters that are supported but hadn't yet gotten around to figuring out the differences compared to climate.

Spalfnaganalf1:
Out of curiosity - does your Hydrotherm still show 50 degrees when its actually reached the set temp of 60?

fsevilla3:
> Out of curiosity - does your Hydrotherm still show 50 degrees when its actually reached the set temp of 60?

Yes, my reading also stop at 50C when actual is at or above 50C. This is the same behaviour with the Tuya mobile app, so it’s not an issue with the HACS integration. I’ve reported it to the Hydrotherm team but I don’t know if it’s something they can or will fix (eg via OTA update).

Spalfnaganalf1:
> Yes, my reading also stop at 50C when actual is at or above 50C. This is the same behaviour with the Tuya mobile app, so it’s not an issue with the HACS integration. I’ve reported it to the Hydrotherm team but I don’t know if it’s something they can or will fix (eg via OTA update).

No worries - For now i've just re-entered the 'scale' part of the code that was removed and i am using 0.8333333333333 as the scale. Not sure if it is accurate for anything outside of the 60 degrees, but at least i'm not being triggered seeing 50 degrees anymore.

curik8:
Hi! Having the same issue here regarding max 50 deg. Is there a solution for this yet?

make-all:
You would have to ask Hydrotherm.  There is nothing that a Home Assistant integration can do about faulty devices.

PR #256, opened 2022-10-28
Add the ElectriQ CD12PRO-LE dehumidifier

cooperaj:
Found that my new dehumidifier wasn't in the repo so added it :) 

Didn't know if I was supposed to add the contribution line but can remove if it shouldn't be a part of this PR.


cooperaj:
I don't supposed I could have a label added for "that yearly github event" please ;) I'm missing one last thing.

cooperaj:
Mmm. Not sure how to fix this sonarcloud issue? Looks to be some sort of permissions check on it.

make-all:
Don't worry about the sonarcloud issue, it is due to test coverage in the newly added water_heater implementation that was basically copied from climate so has a lot of code that isn't actually needed for the one device that uses it.

make-all:
Merged

PR #280, opened 2022-11-19
Add Beok TOL47 wifi thermostat to supported devices

kekkutya:
https://www.beok-controls.com/Content/upload/PDF/2021837153/TOL47-Heating-Thermostat-Manual.pdf

### DIFF
--- a/custom_components/tuya_local/devices/beok_tol47_thermostat.yaml
+++ b/custom_components/tuya_local/devices/beok_tol47_thermostat.yaml
@@ -24,11 +24,11 @@ primary_entity:
       type: string
       name: preset_mode
       mapping:
-        - dps_val: cold
+        - dps_val: "cold"
           value: Manual mode
-        - dps_val: hot
+        - dps_val: "hot"
           value: Program mode
-        - dps_val: wind
+        - dps_val: "wind"
           value: Temporary program mode
     - id: 5
       type: string
@@ -70,6 +70,18 @@ secondary_entities:
       - id: 13
         type: boolean
         name: switch
+  - entity: binary_sensor
+    name: Fault alarm
+    category: diagnostic
+    class: problem
+    dps:
+      - id: 16
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: False
+          - value: True
   - entity: number
     name: Calibration offset
     category: config
@@ -101,16 +113,16 @@ secondary_entities:
     category: config
     dps:
       - id: 25
-        type: integer
+        type: string
         name: option
         mapping:
-          - dps_val: 1
+          - dps_val: "1"
             icon: "mdi:gauge"
             value: Internal
-          - dps_val: 2
+          - dps_val: "2"
             icon: "mdi:thermometer-chevron-down"
             value: External
-          - dps_val: 3
+          - dps_val: "3"
             icon: "mdi:thermometer"
             value: Both
   - entity: switch
@@ -126,19 +138,19 @@ secondary_entities:
     category: config
     dps:
       - id: 41
-        type: integer
+        type: string
         name: option
         mapping:
-          - dps_val: 0
+          - dps_val: "0"
             icon: "mdi:lightbulb-on-10"
             value: None
-          - dps_val: 1
+          - dps_val: "1"
             icon: "mdi:lightbulb-on-40"
             value: Low
-          - dps_val: 2
+          - dps_val: "2"
             icon: "mdi:lightbulb-on-70"
             value: Middle
-          - dps_val: 3
+          - dps_val: "3"
             icon: "mdi:lightbulb-on"
             value: High
   - entity: select
@@ -147,14 +159,14 @@ secondary_entities:
     icon: "mdi:calendar-clock"
     dps:
       - id: 42
-        type: integer
+        type: string
         name: option
         mapping:
-          - dps_val: 1
+          - dps_val: "1"
             value: "Weekday+Weekend"
-          - dps_val: 2 
+          - dps_val: "2" 
             value: "Mon-Sat+Sun"
-          - dps_val: 3
+          - dps_val: "3"
             value: "Daily"	 
   - entity: switch
     name: Output main reverse
@@ -179,7 +191,7 @@ secondary_entities:
         mapping:
           - scale: 10
   - entity: number
-    name: External Sensor temperature limit
+    name: External sensor temperature limit
     category: config
     icon: "mdi:thermometer"
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/beok_tol47_thermostat.yaml
```diff
@@ -0,0 +1,192 @@
...
+          step: 5
+    - id: 3
+      type: integer
+      name: current_temperature
+      mapping:
+        - scale: 10
+    - id: 4
+      type: string
+      name: preset_mode
+      mapping:
```
make-all, line 26: This seems like a strange mapping.  Are you sure you got preset_mode and hvac_mode around the right way?

custom_components/tuya_local/devices/beok_tol47_thermostat.yaml
```diff
@@ -0,0 +1,192 @@
...
+          - dps_val: 3
+            icon: "mdi:lightbulb-on"
+            value: High
+  - entity: select
+    name: Schedule
+    category: config
+    icon: "mdi:calendar-clock"
+    dps:
+      - id: 42
+        type: integer
```
make-all, line 150: Normally enum types in tuya are strings, even if they use numbers as the value (they will have quotes around them in the raw Tuya protocol).  This integration is lenient with string/integer distinction, so it probably works either way but better for the config to use the correct one.

custom_components/tuya_local/devices/beok_tol47_thermostat.yaml
kekkutya: Yes, a little bit strange for me too. Debugged dps are:

```
"device_config": {
"friendly_name": "Living Room Thermostat",
"local_key": "xxxxxxxxxxxxxxx",
"host": "192.168.0.55",
"device_id": "xxxxxxxxxxxxxxx",
"protocol_version": "3.3",
"scan_interval": 1,
"dps_strings": [
	"1 (value: True)",	# Switch
	"2 (value: 205)",	# Set Temperature *10
	"3 (value: 202)", 	# Current Temperature *10
	"4 (value: cold)",	# Mode "cold": Manual "hot": Program "wind": Temporary Program 
	"5 (value: 1)",		# Working status 0: off 1: on
	"8 (value: False)",	# Child lock
	"13 (value: True)",	# Sound
	"16 (value: 0)",	# Fault alarm
	"20 (value: -7)",	# Temperature calibration *10
	"21 (value: 30)",	# Max setting temperature
	"25 (value: 1)",        # Sensor selection 1: internal 2: external 3: both
	"26 (value: False)",    # Frost protection
	"31 (value: False)",	# Reset
	"41 (value: 3)",       	# Backlight brightness 0: Close 1: Low 2: Middle 3: High
	"42 (value: 1)",      	# Working day settings 1: 5+2 2: 6+1 3: 7
 	"101 (value: False)",	# Output main reverse
 	"105 (value: 5)",     	# Temperature control switch difference *10
 	"107 (value: 30)"       # External Sensor temperature limit
```
- on dp4, i can change value with the Tuya Smart app "Mode" settings, and if i change it from HA, the app settings reflect correctly.
- on dp5, value 0 if dry contact is open, 1 if closed (=boyler heating)
        		
Or i completely wrong interpreting preset_mode and hvac_mode?

custom_components/tuya_local/devices/beok_tol47_thermostat.yaml
kekkutya: Thanks for your comment (very beginner with this), modified variable types (+ add dp16 Fault alarm settings)

custom_components/tuya_local/devices/beok_tol47_thermostat.yaml
make-all: In that case, preset_mode is correct for dp 4, but I would use hvac_action for dp 5, and hvac_mode directly for dp 1.

kekkutya:
> I'll fix up the hvac_action and hvac_mode later

Thank you,
i finished at now.

feel free to use it or modify. Thanks for the revision and help (it was my first activity on github)
  
name: Beok TOL47 thermostat
primary_entity:
  entity: climate
  dps:
    - id: 1
      type: boolean
      name: hvac_mode
      mapping:
        - dps_val: true
          value: "heat"
        - dps_val: false
          value: "off"
    - id: 2
      type: integer
      name: temperature
      mapping:
        - scale: 10
          step: 5
    - id: 3
      type: integer
      name: current_temperature
      mapping:
        - scale: 10
    - id: 4
      type: string
      name: preset_mode
      mapping:
        - dps_val: "cold"
          value: Manual
        - dps_val: "hot"
          value: Program
        - dps_val: "wind"
          value: Temporary
    - id: 5
      type: string
      name: hvac_action
      mapping:
        - dps_val: "0"
          icon: "mdi:radiator-disabled"
          constraint: hvac_mode
          conditions:
            - dps_val: true
              value: idle
            - dps_val: false
              value: off
        - dps_val: "1"
          value: heating
          icon: "mdi:radiator"

Regards, Zoli

make-all:
Sorry, it seems I did not quite wait long enough.  Was there anything else you changed  besides the hvac_mode and hvac_action I suggested above?

escoand:
@kekkutya Do you know what this sound switch is useful for? I have the same device and always wonder if this could be used for something smart.

make-all:
My understanding is that the wall panel emits a beep if you press a button, and this switch can be used to turn that on and off.

kekkutya:
> My understanding is that the wall panel emits a beep if you press a button, and this switch can be used to turn that on and off. I guess one smart thing you could do is turn it off at night and on again in the morning with automations.

seems logical, but my device does not make sound at any position of the switch 
@make-all: may you check issue #289 by escoand pls.? Seems some incompatibility with awatto wt100, when add a new device.

make-all:
A fix for #289 was pushed about an hour ago.  There was a syntax error in the config for Avatto WT100, which was not picked up by unit tests.

escoand:
> seems logical, but my device does not make sound at any position of the switch

That's why I was asking...

make-all:
This thermostat seems to share a lot in common with the Avatto WT100.  There may be a wider range by the same OEM manufacturer, with slight differences in spec.  It is possible that the beep hardware was left off the Beok TOL47, even though it is still exposed in the firmware.  It was mentioned in the Avatto manual that I read, and the original config came from the iot.tuya.com info for this device I suppose.  Does the official app expose this switch?  Maybe it is best moved to an attribute on the climate entity where it can still be observed, but the switch removed to prevent confusion.

kekkutya:
> Does the official app expose this switch? 

Yes. As i see, the app screens exactly the same for both devices. And Beok's (not implemented yet) dp31 boolean is the "Reset" like in Avatto.

PR #307, opened 2022-12-11
Create tellur_usb_power_strip.yaml

escoand:
This one is similar to the already existing "4 outlet + USB powerstrip" in `logicom_powerstrip.yaml`.

Is there a way to add this general switch? My selected way is not working...
Screenshot_20221128-172558_Tuya Smart

### DIFF
--- a/custom_components/tuya_local/devices/tellur_usb_power_strip.yaml
+++ b/custom_components/tuya_local/devices/tellur_usb_power_strip.yaml
@@ -1,0 +1,93 @@
+name: 3 outlet + USB powerstrip
+primary_entity:
+  entity: switch
+  dps:
+    - id: 1
+      type: boolean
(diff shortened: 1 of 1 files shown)


escoand:
Thanks but the primary entity is not workig in the current version. I wanted to add all the buttons independently and one switch for all at once.

make-all:
The device doesn't seem to support the "one switch for all at once".  After merging your change, I edited the config to remove the non-working entity (moving Outlet 1 to the "primary" slot).

escoand:
But the official app is able to do so. See the screenshot above.

make-all:
If your information about the available dps is complete, then the official app must be doing this at the UI level.  There is no provision in this integration to configure additional features which are not implemented by the device itself.

PR #308, opened 2022-12-11
Add support for Benexmart Wi-Fi Blind Motor

cjones813:
Add support for the Benexmart WiFi Blind Motor.

Only 3 dps are being exposed to HA even though cloud debug shows more available.

  },
  "data": {
    "name": "Right Curtain",
    "type": "benexmart_blind_motor",
    "device_id": "**REDACTED**",
    "local_key": "**REDACTED**",
    "host": "**REDACTED**",
    "api_version": 3.3,
    "status": {},
    "cached_state": {
      "updated_at": 1670793080.225058,
      "1": "stop",
      "2": 0,
      "5": true
    },

Screenshot 2022-12-11 at 21 34 07

IMG_4252
IMG_4251

https://amzn.eu/d/drCsbO4

### DIFF
--- a/custom_components/tuya_local/devices/benexmart_blind_motor.yaml
+++ b/custom_components/tuya_local/devices/benexmart_blind_motor.yaml
@@ -1,0 +1,26 @@
+name: benexmart blind motor
+primary_entity:
+  entity: cover
+  class: blind
+  dps:
+    - id: 1
+      name: control
+      type: string
+      mapping:
+        - dps_val: open
+          value: open
+        - dps_val: close
+          value: close
+        - dps_val: stop
+          value: stop
+    - id: 2
+      name: position
+      type: integer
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/benexmart_blind_motor.yaml
```diff
@@ -0,0 +1,26 @@
...
+          value: stop
+    - id: 2
+      name: position
+      type: integer
+      range:
+        min: 0
+        max: 100
+      mapping:
+        - invert: true
+    - id: 5
```
make-all, line 24: I think if you call this "reversed", you can remove the invert: true mapping from position.  You may need to add a mapping to swap true and false if the "reverse" direction is the opposite of what is expected.

Issue #75, opened 2021-11-19
Eurom Heater is recognized as mirabella_genio_usb


mvroosmalen1970:
Hi
My Heater is not correctly recognized:
This error originated from a custom integration.

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:73
Integration: Tuya Local (documentation, issues)
First occurred: 8:34:20 PM (1 occurrences)
Last logged: 8:34:20 PM

Device matches mirabella_genio_usb with quality of 40%. DPS: {'1': False, '2': 21, '3': 19, '4': 'off', '7': False, 'updated_at': 1637350453.9208913}

Any help is appreciated

make-all:
Which model Eurom heater is this?

mvroosmalen1970:
Sani-Wall-Heat 2000 Wifi with swing function

https://eurom.nl/product/verwarming/badkamerkachels/sani-wall-heat-2000-wifi/

make-all:
I made a guess at the values for dps 4 being "off", "low" and "high" based on the manual saying that the mode can be switched between fan only, half and full heating, combined with the known value of "off" above.  Another guess is that dps 7 controls the swing function (otherwise, there are no other dps left, so the swing would not be controllable if it is not that).

mvroosmalen1970:
The device is recognized. temperature is functional. The on/off, swing and power doesn;t seem to match with values expected as I get status:
off - fan
- auto
- 50_perc
- 100_perc
no clue how to resolve this or what info could be helpful for you?

make-all:
I'm not sure what is going on with the swing, but it may just be a side effect of the presets being incorrect.

mvroosmalen1970:
Everything seems to function :-). Thanks!!! Would be nice to improve the 
<img width="251" alt="Capture" src="https://user-images.githubusercontent.com/36565907/142840804-91c83499-00fc-4d27-976c-9746954cae95.PNG">
with swing function and that you can click in the icon starting/stopping the swing.  Maybe this is for future improvements

Good work.

make-all:
Thanks for confirming.  The UI is a general Home Assistant UI, so any enhancement reqests should go there.  I think though that will not want to clutter it with the full functionality of climate devices, and will suggest adding a row of buttons below to control additional functionality.

PR #310, opened 2022-12-12
Eurom wall Designheat 2000 heater fix heat modes

markwellis:
The eco & boost mode were not working because of a spelling mistake in the yaml.

Attached is a screenshot from Tuya IoT showing the correct values

image

### DIFF
--- a/custom_components/tuya_local/devices/eurom_walldesignheat2000_heater.yaml
+++ b/custom_components/tuya_local/devices/eurom_walldesignheat2000_heater.yaml
@@ -19,10 +19,10 @@ primary_entity:
             - dps_val: auto
               value: heat
               icon: "mdi:radiator"
-            - dps_val: "50_perc"
+            - dps_val: "50_per"
               value: heat
               icon: "mdi:radiator"
-            - dps_val: "100_perc"
+            - dps_val: "100_per"
               value: heat
               icon: "mdi:radiator"
     - id: 2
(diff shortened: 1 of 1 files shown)


make-all:
@mvroosmalen1970 provided the original values, and reported "everything working" when the device was initially added in #75.

So it is likely that there are different firmware versions using different values.  So rather than changing the existing config, it will probably be better to add another one.  But for users to select the correct config, they will need some identifying information. Firmware version may work, but is not very visible.  Is it possible that these are different year models and have some other cosmetic differences too?

markwellis:
I thought that might be the case.

According to the Tuya Smart app, the firmware is v1.0.1
image

All I can find in tuya iot is this
image

Looking at eurom.nl, it's not a Wall Designheat 2000 Wifi ( https://eurom.nl/en/product/heating/ceramic-heaters/wall-designheat-2000-wifi/ )

The label says it's a Sani Wallheat 2000 Wifi ( https://eurom.nl/en/product/heating/bathroom-heaters/sani-wall-heat-2000-wifi/ )

So I guess there's a slight difference between the models?

But I can see the yaml used to be for a sani wallheat https://github.com/make-all/tuya-local/commit/cddfcdf691c1fd4dbba9a63dde3cc2970c9f9174

So all that said, I'll add the new yaml file & tests later

make-all:
If they are slightly different model names, then that will probably be better than firmware versions for identifying which config to select.  I remember when adding other Eurom devices finding their webpage with a lot of confusingly similar named devices, and often found myself trying to figure out what dps ranges were based on a manual which turned out to be for a different model.

markwellis:
replaced by #313

PR #321, opened 2022-12-25
Support for AlecoAir D12 ECO and AlecoAir D16 PURIFY dehumidifiers

mihsu81:


mihsu81:
Is it possible to set a certain speed and target humidity when a operating mode is selected? I.e. Set humidity to 60% and speed low when the mode Sunny_Day is selected?
Thanks in advance.

make-all:
This is up to the device's firmware.  If it is not implemented there, then it should be done by your automations.  This integration should not contain logic beyond what is required for the device to function.

PR #341, opened 2023-01-15
Translation to Spanish

david1111967:
Implementation to the Spanish language


make-all:
Thanks!  I haven't yet looked at the new translation features in 2023.1, but the config dialogs that the current translations cover is a good start.

PR #351, opened 2023-01-17
Support for GE Jasco Ultra Pro Dimmer, Toggle, and TH16 Temp sensor

jjlauer:
* GE Jasco Ultra Pro Toggle Dimmer v2
 * GE Jasco Ultra Pro Toggle Switch
 * TH16 TH06WB3S Temperature and Humidity Sensor


make-all:
The test failures are all from the main branch hotfix last night.  Auto version rollover was updated to be async, so the tests need updating.

Issue #282, opened 2022-11-23
New device request: Light Motion Sensors request(DET100HA / DET102HA)


isarib01:
Hi,

I seem to have a fair bit of light sensors which are not yet supported, there are two different models and wondering if anyone has some fixes to the below.  Many thanks.

***************************************************************
Sensor Model 1 - Deta 360 Motion Sensor, Grid Connect(DET102HA).
{
  "result": {
    "active_time": 1666953679,
    "category": "pir",
    "category_name": "Motion Detector",
    "create_time": 1664431066,
    "gateway_id": "",
    "icon": "smart/icon/ay1536534747810ekKBj/33e050831a2f667fcae9bff8aa5ec73b.png",
    "id": "67506371e868e784b0d7",
    "ip": "1234567890",
    "lat": "1234567890",
    "local_key": "1234567890",
    "lon": "115.9807",
    "model": "Starlux DET102HA",
    "name": "Garage Light",
    "online": true,
    "owner_id": "65067975",
    "product_id": "oapo5mmfbcr2y7ox",
    "product_name": "DET102HA",
    "sub": false,
    "time_zone": "+08:00",
    "update_time": 1666953686,
    "uuid": "1234567890"
  },
  "success": true,
  "t": 1669133304844,
  "tid": "e4a17bd36a7f11edadcaa64798946735"
}

{
  "result": {
    "category": "pir",
    "functions": [],
    "status": [
      {
        "code": "pir",
        "lang_config": {
          "none": "",
          "pir": ""
        },
        "name": "人体感应状态",
        "type": "Enum",
        "values": "{\"range\":[\"pir\",\"none\"]}"
      }
    ]
  },
  "success": true,
  "t": 1669134803744,
  "tid": "620c535f6a8311edadcaa64798946735"
}

***************************************************************
Sensor Model 2 - Deta Light Motion Sensor, Grid Connect(DET100HA).

{
  "result": {
    "active_time": 1666953283,
    "category": "pir",
    "category_name": "Motion Detector",
    "create_time": 1659200434,
    "gateway_id": "",
    "icon": "smart/icon/ay1558866774161DZ6MV/1566282712ae29041f3ee.png",
    "id": "58607270e868e782f909",
    "ip": "58.167.237.241",
    "lat": "1234567890",
    "local_key": "1234567890",
    "lon": "1234567890",
    "model": "Starlux DET100HA",
    "name": "Driveway Light Sensor",
    "online": true,
    "owner_id": "65067975",
[...]

PR #356, opened 2023-01-23
Create arlec_motion_sensor_light.yaml

raphaelcouzet:
HI,
Thanks for this great project!
I have been playing around with it for a few days and find it to be more stable than other integrations I've used for Tuya devices. I really like that I can create  device.yaml files to fine tune how devices are handled.

Here is one that I have put together and provides the same functionality as the manufacturer's app (e.g. Grid Connect; or Tuya apps) Product is sold by Bunnings in Australia
https://www.bunnings.com.au/arlec-20w-grid-connect-smart-security-light-with-movement-activated-sensor_p0223575

Entities & DPs:
This smart motion sensor light is a bit odd as it doesn't have a DP to represent a switch but instead has a select entity to choose between modes (On, Off, or Auto)  This means that the generic motion_sensor_light.yaml doesn't work with this device. There is a binary sensor that is read only and represents the state of the light's relay. I haven't quite figure out how to best configure that and currently have it as the primary entity. This works well with the exception that the HA UI provides a switch that can be toggled but doesn't actually have any effect. The UI does correctly represent when the switch is on/off based on the select.sensor_mode entity. The only other DP that I have not configured correctly yet  is DP106 : Auto Rest. This is supposed to be a switch to turns on a 1 hr timer for the mode to switch back to Auto if it has been turned On or Off. The native app does not use this feature either, so it is not really necessary and is better handled by an automation anyway :)

Entity type | ID | Friendly Name
Select | 101 | Sensor Mode
Binary sensor | 102 | Relay Status
Number | 103 | Sensitivity
Number | 104 | Duration
Number | 105 | Light Threshold
Switch | 106 | Auto Reset(unused)

### DIFF
--- a/custom_components/tuya_local/devices/arlec_motion_sensor_light.yaml
+++ b/custom_components/tuya_local/devices/arlec_motion_sensor_light.yaml
@@ -1,0 +1,76 @@
+name: Arlec Grid Connect Smart Security Light
+products:
+  - id: pxyhz90upglpoedu
+    name: MAL315HA
+#https://www.bunnings.com.au/arlec-20w-grid-connect-smart-security-light-with-movement-activated-sensor_p0223575
+primary_entity:
+  entity: light
+  dps:
+    - id: 102
+      type: boolean
+      name: switch
+secondary_entities:
+  - entity: select
+    name: Mode
+    category: config
+    dps:
+      - id: 101
+        type: string
+        name: option
+        mapping:
+        - dps_val: mode_off
+          value: "Off"
+        - dps_val: mode_auto
+          value: Auto
+        - dps_val: mode_on
+          value: "On"
+  - entity: number
+    name: Sensitivity
+    icon: "mdi:human-greeting-proximity"
+    category: config
+    dps:
+      - id: 103
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 4
+        mapping:
+          - invert: true
+  - entity: number
+    name: Duration
+    icon: "mdi:camera-timer"
+    category: config
+    dps:
+      - id: 104
+        type: integer
+        name: value
+        unit: s
+        range:
+          min: 10
+          max: 900
+        mapping:
+          - step: 10
+  - entity: number
+    name: Light Level
+    icon: "mdi:theme-light-dark"
+    category: config
+    mode: slider
+    dps:
+      - id: 105
+        type: integer
+        name: value
+        unit: lx
+        range:
+          min: 0
+          max: 3900
+        mapping:
+          - invert: true
+  - entity: switch
+    name: Auto Reset
+    category: config
+    dps:
+      - id: 106
+        type: boolean
+        name: sensor
+


make-all:
This appears to be the same as the Deta 360 motion light that is already supported.  I don't think there is justification for another config if these are the same.

raphaelcouzet:
The Deta 360 motion light supported by the motion_light_sensor.yaml configuration is not compatible with this Arlec motion sensor light. I tried.
The key difference seems to be that the Arlec light does not have a switch and can only be turned on using the select mode entity.
The secondary entities are very similar but there are also minor differences in the ranges allowed which will also cause issues by either going out of bound or just not being able to use the full range allowed by the device.

make-all:
I'm not sure that the Deta light is working properly, I tried to guess ranges based on the info I could find online, and have not received any feedback on #282.  Given that Arlec and Deta are made by the same company and the dps mapping is the same, I strongly suspect they are actually the same.  So probably it is better to modify the existing motion_light_sensor.yaml config.

raphaelcouzet:
I think you might be right. They look the same and the states provided by isarib01 are within the allowed ranges.

A bit more info is available here:
https://templates.blakadder.com/arlec_MAL315HA.html
This describes how to replace the Tuya chip with ESP and how to create a switch since the device doesn't really have one.
Useful info as it confirms the DPs, gives us a clue on how to control the on/off state, and gives the rough ranges to expect. 

I used the device logs from iot.tuya.com to get the exact ranges and created a template switch in my configuration.yaml to control the on/off state. (Example below in case it is useful)
I wonder if there is a way to create a switch with mapping rules to get the same functionality in the config?

I should also mention that I have tested Arlec_motion_sensor_light.yaml with my HA instance and it has been working well and stable for the last 2 days.
I would be happy if you used this config and product Id in motion_sensor_light.yaml config.

Thanks again for this integration! :)



----Example for Template Switch---------------
#Add switch for floodlights
switch:
  - platform: template
    switches:
      laundry_sensor_light:
        value_template: "{{ is_state('select.laundry_motion_sensor_light_mode','On') }}"
        turn_on:
            - service: select.select_option
              data:
                option: "On"
              target:
                entity_id: select.laundry_motion_sensor_light_mode
        turn_off:
            - service: select.select_option
              data:
                option: Auto
              target:
                entity_id: select.laundry_motion_sensor_light_mode

make-all:
I think the issue with creating a switch by mapping in the config is that the actual selection is tri-state, so what to do with the "Auto" option?
The current config uses what appears to be a master switch as the switch, then puts "Off", "Auto" and "On" as effects.  In your config, you moved them to a separate select, which should give the same result.  I don't think you can create an actual switch without either templating, or losing functionality of the light.

raphaelcouzet:
The "master switch" or DP 102 is actually a binary sensor which reflects the status of the relay for the light inside the actual physical device. It is read only and can't be directly controlled. 
Unfortunately, there is no actual switch in this device. 

I've had a look at how the localtuya project handles this device and they do a similar thing as I do. But for some reason it is not very stable in that implementation. When i tried it the devices were becoming unavailable very frequently. This is what led me to your integration....
localtuya


Anyway... This integration and new config file work well for this device and just needs the additional Template Switch.
What do you think is the best way to document the need for Template Switch if you were to commit the new config file?

make-all:
I've modified motion_sensor_light.yaml based on this, and also made a change to the light implementation so that if the switch dp is readonly, and "on" and "off" are available in effects, it will use those to turn on and off.  Setting back to auto will still require sending the "effect" property in the turn_on service (which can be linked to a button in the UI for convenience).

raphaelcouzet:
Thanks, Will give it a whirl.

isarib01:
Just tried the data 360 motion sensor and the other data motion sensor tonight.

Just tried the "Auto Reset", what this does is just puts the light in the ON mode and counts down timer to turn off in 1hr. I checked and compared the behaviours what each of the functionality did against the official Tuya or vendor app.

So in hind sight, the entity Auto Reset should be called Sensor Off/Auto/On/On(Auto Reset).

That should be the only bit missing and is probably the most critical function of it.

Hope that makes sense.
Cheers

make-all:
The suggested name change is difficult to parse and means nothing to me.  Maybe "On for 1 hour" is a better name?

isarib01:
> The suggested name change is difficult to parse and means nothing to me. Maybe "On for 1 hour" is a better name?

That would also work, "On for 1 hour".

Do you think its difficult to sort out the "Sensor Off/Sensor Auto" function?

make-all:
I don't understand the last comment "Do you think its difficult to sort out the "Sensor Off/Sensor Auto" function?"

isarib01:
Hi,

Thanks for your reply.  I have included a screenshot of the only missing bits which is the Sensor feature to Off(disable) and Auto(enable the sensor).   

From what I have tested, the new features you had added is of the Master Switch On Toggle, and also the Auto On Toggle - which is just the Auto Off 1hour schedule.

image


Thank you.

make-all:
I don't think thosse options are missing.  If you open the menu on a light card, or long press on the light in an entities card, you should get a window that exposes the full functionality of the light, which contains a select box with all three options.  You can also create buttons for them if you wish, the action to assign to the button should be "Call service" with the "light.turn_on" service with parameter "effect" set to "on", "off" or "auto".  "Off" is also accessible by the light.turn_off service.

raphaelcouzet:
Quick message just to confirm that this configuration works really well. 
Has been running for a week without any issues or any dropouts. Also really like the light switch you implemented.
Thanks again and great work!

isarib01:
> Quick message just to confirm that this configuration works really well. Has been running for a week without any issues or any dropouts. Also really like the light switch you implemented. Thanks again and great work!

Can you screenshot how you had created the switch for the Motion Sensor enable function?

make-all:
There are multiple ways of creating buttons in the UI.  You can create a button card (it may look a little big unless you constrain it inside a grid card or similar), or use the Buttons setting of the header of an entities card.  Both these methods allow you to set the action on the button directly to Call Service, without linking the button to a button entity, but you might need to switch to yaml mode to use Call Service in those places.  For more flexibility, you can also create an Button Helper (input_button), and set the action in an Automation triggered by the button press, then you can add the button to an entities card body or other places where an arbitrary button cannot be created (and Automations also let you configure Call Service actions in the UI).

Which ever way you do it, your action yaml ends up looking like:

```yaml
    tap_action:
      action: call-service
      service: light.turn_on
      data:
        entity_id: light.motion_sensor_light
        effect: auto
```

make-all:
It will also accept "on" and "off" for effect, but these are the same as calling light.turn_on without the effect parameter, and calling light.turn_off, respectively.

isarib01:
Sorry to be a pain, but I have attached a screenshot of what I can see for the device light sensor which was added.

Is it meant to have a entity  "light.turn_on"?

image

make-all:
light.turn_on is a Service, not an entity.  It is used in the "Call Service" action of the button, or an automation triggered by the button press if you use a Button Helper.  The entity you would apply it to is light.garage_light_sensor in your screenshot, that should replace light.motion_sensor_light in my example above.

Sorry for confusing you with all the options.  The easiest method will be:
1. Go to Settings / Devices & Services, Helpers tab, press Create Helper and create a Button with your desired name and icon.
image

2. Go to Settings / Automations & Scenes, create an Automation, starting with an empty automation.  Add a trigger based on State, and select the button you just created as the entity.  Leave everything else blank.
image

3. Scroll down and Add an Action, choosing "Call Service".
image
 I tried
4. Scroll to the bottom, select Effect and type auto, then press Save
image

5. In your UI, add a row in an entities card and set the entity to the button you created.

6.  You can alternatively add it to a Buttons header for the entities card, which will show it as an oval button at the top of the card rather than the entity name on the left with "PRESS" to the right as the button. Since the automation is defined, the buttons header config just needs to list the entity id: input_button.garage_motion_auto (or whatever you chose as the name)
image

Here is what the two options look like together (of course, you would just choose one);
image

I tried also an Entity card, but it seems it wants to show the state of the entity, and for buttons that is always "Unknown".  I also tried the Button card, but it seems that can't leverage the input_button and automation set up in the early steps, so it is a completely separate manual yaml config like I posted above in my previous comment, since it doesn't use the UI for Call Service actions.

isarib01:
Thank you for the detailed info!

I have tried it, and looks like I do not need the helper!  Just creating the automation, action, light device and effect "auto" sets works for me, setting the sensor mode to "auto" as I need it...

make-all:
Correct, you only need the helper if you want to create a UI button without using the Button card, or to simplify the config of using Entities card button header.  If you're looking at automating it, then just straight to step 2 with a different trigger condition will also work.

isarib01:
After creating those automations, the only way to use the automation is if I click on Run.  Other than that, making buttons which toggles the Automation ON just does nothing.  Something I have to figure out....

image

Appreciate your effort in getting this code sorted out for these devices.  My home is very much Tuya Cloud free soon.

cg6:
Hi, I just commented about the on-off-auto option in the issue https://github.com/make-all/tuya-local/issues/282#issuecomment-1413394796

I saw your post about using Call Service to set the effect to Auto. However this still doesn't give us visibility of if the current state is Off or Auto. I don't know how much work it is to implement the way I suggested, but I think it would be more usable.

Thanks.

isarib01:
> After creating those automations, the only way to use the automation is if I click on Run. Other than that, making buttons which toggles the Automation ON just does nothing. Something I have to figure out....
> 
> image
> 
> Appreciate your effort in getting this code sorted out for these devices. My home is very much Tuya Cloud free soon.

All sorted.

Got rid of the Automation entries, and created one help button to use for everything in the cards.  Created a button for each Sensor Mode: Auto, Off, On which calls all my Deta Sensor devices.  Works a treat!  Thanks again.

PR #357, opened 2023-01-23
BlitzWoolf BWSH5 Humidifier

spuljko:
Added support for BlitzWoolf BWSH5 Humidifier

### DIFF
--- a/custom_components/tuya_local/devices/blitzwoolf_bwsh5-humidifier.yaml
+++ b/custom_components/tuya_local/devices/blitzwoolf_bwsh5-humidifier.yaml
@@ -1,0 +1,193 @@
+# sample DPS {'updated_at': 1674438037.778799, '1': True, '4': 'health', '5': True, '10': 24, '12': 75, '13': 55, '14': 77, '18': 'c', '19': 'cancel', '20': 0, '21': True, '22': 0, '23': 'level_1', '26': True, '35': True} 
+name: BlitzWoolf BWSH5 Humidifier
+products: 
+  - id: lqzme7o0kzhumdxr
+primary_entity:
+  entity: humidifier
+  class: humidifier
+  dps:
+    - id: 1
+      name: switch   
+      type: boolean
+      mapping:
+        - dps_val: true
+          icon: "mdi:air-humidifier"
+          value: ON
+        - dps_val: false
+          icon: "mdi:air-humidifier-off"
+          value: OFF
+    - id: 5
+      name: led
+      type: boolean
+    - id: 10
+      name: temperature 
+      type: integer
+      readonly: true
+      unit: C
+    - id: 12
+      name: temperature_f
+      type: integer
+      readonly: true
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/blitzwoolf_bwsh5-humidifier.yaml
```diff
@@ -0,0 +1,193 @@
...
+        name: sensor
+        class: measurement
+        readonly: true
+        unit: C
+      - id: 12
+        name: temperature_f
+        type: integer
+        unit: F
+        readonly: true
+        hidden: true
```
make-all, line 109: There's not really any point in adding these as entities then hiding them if you are not using them to switch temperature units.  To do this, you need specific naming on dp 18, but if both are always correct, it is optional, since HA will do unit conversion anyway.

custom_components/tuya_local/devices/blitzwoolf_bwsh5-humidifier.yaml
```diff
@@ -0,0 +1,193 @@
...
+        - dps_val: true
+          icon: "mdi:air-humidifier"
+          value: ON
+        - dps_val: false
+          icon: "mdi:air-humidifier-off"
+          value: OFF
+    - id: 5
+      name: led
+      type: boolean
+    - id: 10
```
make-all, line 22: The attributes that have sensors defined below do not need to be added here as well, since they aren't a standard part of the humidifier entity interface so have no special meaning.

custom_components/tuya_local/devices/blitzwoolf_bwsh5-humidifier.yaml
spuljko: I thought device is better recognised if all data points are specified, that is why I left it hidden. Should i remove it?
Temperature in dps 10 is always in C and dps 12 is in F, no matter what is dps 18. That dps seems to only affects temperature on display of device, as I've written in comment.

make-all: If they are elsewhere in the config (like the unit dp), they are already covered for detection.  Otherwise, yes, including them will help avoid false detections, but there is no need to set them hidden, someone might find it useful to be able to read them as extra attributes.  I generally only make dp hidden that are already visible elsewhere, for example when merging a power switch and mode into hvac_mode in climate devices, or using the a dp for conditional mapping on another dp, when the dp itself is available on another entity.

make-all:
dps 20, 22 and 23 are missing from the config.  

Due to proximity to 19 (Timer set), I suspect 20 is time remaining.  I will add it on the Timer as an optional attribute.

22 is likely a fault code, which most devices of this type have.  Likely one of the codes means "Tank Empty" probably a shared code with "Tank removed".

23 looks like it might be controlling a fan speed or spray volume, perhaps linked with the manual mode.

PR #358, opened 2023-01-23
Update yym_805SW_aroma_nightlight.yaml

spuljko:
Added missing dpses

### DIFF
--- a/custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
+++ b/custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
@@ -15,15 +15,15 @@ primary_entity:
         - dps_val: false
           icon: mdi:scent-off
           value: OFF      
-    - id: 2
-      name: speed
-      type: string
-      hidden: true
-      mapping:
-        - dps_val: large
-          value: 100
-        - dps_val: small
-          value: 50
+		- id: 2
+			name: speed
+			type: string
+			hidden: true
+			mapping:
+				- dps_val: large
+					value: 100
+				- dps_val: small
+					value: 50
 secondary_entities:
   - entity: light
     name: Nightlight
@@ -92,10 +92,10 @@ secondary_entities:
     icon: "mdi:timer"
     category: config
     dps:
-      - id: 3
-        name: option  
-        type: string
-        mapping:
+			- id: 3
+				name: option  
+				type: string
+				mapping:
           - dps_val: "cancel"
             value: "Off"
           - dps_val: "1"
--- a/custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
+++ b/custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
@@ -23,7 +23,7 @@ primary_entity:
         - dps_val: large
           value: 100
         - dps_val: small
-value: 50
+          value: 50
 secondary_entities:
   - entity: light
     name: Nightlight
--- a/custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
+++ b/custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
@@ -22,18 +22,8 @@ primary_entity:
       mapping:
         - dps_val: large
           value: 100
-          icon: "mdi:sprinkler-variant"
         - dps_val: small
           value: 50
-          icon: "mdi:sprinkler"
-    - id: 4          
-      name: "Left time"
-      readolny: true
-      optional: true
-      type: integer
-      range:
-        min: 0
-        max: 360
 secondary_entities:
   - entity: light
     name: Nightlight
@@ -42,14 +32,7 @@ secondary_entities:
       - id: 5
         name: switch
         type: boolean
-      - id: 6 
-        name: color_mode  #effect
-        type: string
-        hidden: true
-        mapping:
-          - dps_val: colour
-            value: rgbw 
-      - id: 8   # not working
+      - id: 8   
         name: rgbhsv  
         type: hex     
         format:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -7,18 +8,114 @@ primary_entity:
...
+          icon: mdi:scent-off
+          value: OFF      
     - id: 2
       name: speed
       type: string
+      hidden: true
       mapping:
         - dps_val: large
           value: 100
+          icon: "mdi:sprinkler-variant"
```
make-all, line 25: If you are setting the icon from two dps on the same entity, you need to define the icon_priority to get consistent behaviour.  In this case I do not see the point, because there is full coverage of the options on both dps, so only one will take effect (but without priorities, it will be random which).

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -7,18 +8,114 @@ primary_entity:
...
+            bytes: 1
+            range:
+              min: 0
+              max: 100
+  - entity: select
+    name: "Color mode"
+    category: config
+    icon: "mdi:palette"
+    dps:
+    - id: 6
```
make-all, line 82: Indentation does not look right

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -7,18 +8,114 @@ primary_entity:
...
+        max: 360
 secondary_entities:
   - entity: light
     name: Nightlight
+    icon: "mdi:light-recessed"
     dps:
       - id: 5
         name: switch
         type: boolean
+      - id: 6
```
make-all, line 45: Not really needed if a separate select is used.  Alternatively you could move the "colourful1" here, then select it as an effect on the light instead of a separate select..

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -7,18 +8,114 @@ primary_entity:
...
+        - dps_val: colour
+          value: Color
+        - dps_val: colourful1
+          value: Colorful          
+  - entity: select
+    name: "Spray mode"
+    category: config
+    icon: "mdi:sprinkler-variant"
+    dps:
+    - id: 2
```
make-all, line 95: Indentation looks wrong again.  But this is already covered by the fan speed, so redundant.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -7,18 +8,114 @@ primary_entity:
...
       type: string
+      hidden: true
       mapping:
         - dps_val: large
           value: 100
+          icon: "mdi:sprinkler-variant"
         - dps_val: small
           value: 50
+          icon: "mdi:sprinkler"
+    - id: 4
```
make-all, line 29: If this is related to the Timer, you should put it on that entity, or make a separate sensor entity for it.

make-all (changes requested):

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -7,18 +8,114 @@ primary_entity:
...
+          - name: h
+            bytes: 2
+            range:
+              min: 0
+              max: 360
+          - name: s
+            bytes: 1
+            range:
+              min: 0
+              max: 100
```
make-all, line 71: It appears from the comment at the top of the file to have a range of 0 to 255.  The RGB and HSV values appear to be a close match, so I think the format is correct apart from that, and this would be the same as the atomi_string_lights.yaml format.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
spuljko: Since this is not fan I find this approach better. This way simple button can be used and it looks much better.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
spuljko: Fixed (this was copied from other integration). Anyway even with this selecting color doesn't work because device only supports hs, and not rgbw mode.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
spuljko: When it is in effect under light it doesn't change. Having separated select works fine.
I don't know if it is related to supported mode of light because hv is not supported - on Tuya integration mode is hv: supported_color_modes: hs.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
spuljko: Moved to as attribute timer. Update is not very frequent, and even reports "nullmin" sometimes, so I don't find it useful enough to be separate sensor.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
spuljko: I have explained this in comment below.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -1,24 +1,114 @@
...
+        - dps_val: true
+          icon: mdi:scent
+          value: ON
+        - dps_val: false
+          icon: mdi:scent-off
+          value: OFF
     - id: 2
       name: speed
       type: string
+      hidden: true
```
make-all, line 21: Speed should not be hidden, it is a basic part of the fan interface.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -1,24 +1,114 @@
...
       mapping:
         - dps_val: large
           value: 100
         - dps_val: small
-          value: 50
+value: 50
```
make-all, line 26: Indentation error.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -1,24 +1,114 @@
...
+            bytes: 1
+            range:
+              min: 0
+              max: 255
+          - name: v
+            bytes: 1
+            range:
+              min: 0
+              max: 255
+  - entity: select
```
make-all, line 60: I think this should be "color_mode" on the light entity instead of a separate select.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -1,24 +1,114 @@
...
+    dps:
+      - id: 6
+        name: option
+        type: string
+        mapping:
+          - dps_val: colour
+            value: Color
+          - dps_val: colourful1
+            value: Colorful
+  - entity: select
```
make-all, line 73: As previously mentioned, this is duplicating the speed attribute of the fan.  If you fix the indentation error and remove the hidden attribute above, that should work.

custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
spuljko: This is not actually fan so it doesn't have speed But fan device requires speed. Since that just doesn't make a sense that is why I've hide this and created new property for diffuser mode.

make-all: It does make sense to model it as a fan though.   The fan entity displays in the entities card as a slider, that lets you adjust from off to maximum output, which is quite an intuitive way to control such a device.  Otherwise you are just left with a bundle of switches and select entities which cannot leverage any specialised UI features.

PR #362, opened 2023-01-29
fix(devices): instruct home assistant to fetch device state if unavailable

larueli:
Hello,

Thanks for the great work, this module is really helpful to me.

I have one issue, when my devices go offline (bulb with no power anymore, or tuya switch unplugged), they are never reported as unavailable to HA, we can still try to control them.

After digging through the code, I discovered that we never tell HA to pull the state if we cannot connect to the device. Or, I think we should do so, like that HA we'll have the "available" property set to false, and the bulbs will display as greyed out until the power comes back.

I have tested this in HA 2023.1.7 with multiple tuya lights and switches.

Feel free to ask any questions or provide suggestions.

### DIFF
--- a/custom_components/tuya_local/device.py
+++ b/custom_components/tuya_local/device.py
@@ -403,6 +403,8 @@ class TuyaLocalDevice(object):
                 self._api_protocol_working = True
                 return retval
             except Exception as e:
+                for entity in self._children:
+                    entity.async_schedule_update_ha_state()
                 _LOGGER.debug(
                     f"Retrying after exception {e} ({i}/{connections})",
                 )


custom_components/tuya_local/device.py
```diff
@@ -403,6 +403,8 @@ async def _retry_on_failed_connection(self, func, error_message):
                 self._api_protocol_working = True
                 return retval
             except Exception as e:
+                for entity in self._children:
```
make-all, line 406: I think this would be better below where we call self._reset_cached_state() after reaching the retry limit, as that is what actually causes the side effect of making the devices go unavailable.  Otherwise there could be a race condition where HA updates the state before continuing on to that line.

larueli: I thought of that at first, but I wanted the device to go unavailable since the first network error (if the error is temporary, the device would show available later) but if that leads to unexpected consequences (I trust you, you know better than me !), I will move it below, it will be a good start !

make-all: It anyway won't go unavailable until the conditions for it to be unavailable exist, which is after the call to self._reset_cached_state()

larueli: Well, I've got another idea, can't we reset the state on every failure ? You tell me.

```python
          except Exception as e:
                self._reset_cached_state()
                for entity in self._children:
                    entity.async_schedule_update_ha_state()
                _LOGGER.debug(
                    f"Retrying after exception {e} ({i}/{connections})",
                )
                if i + 1 == connections:
                    self._api_protocol_working = False
                    _LOGGER.error(error_message)
                if not self._api_protocol_working:
                    await self._rotate_api_protocol_version()
```

larueli: And yeah you(re right it was stupid at first, calling HA to make update before resetting state would not make the device unavailable since the unavailable property depends on the size of the state...

make-all: I don't think it is a good idea, one off failures are actually quite frequent and the devices do not always return full state so it might take some time to get all the cached data back.  The retry limit is 3 once the device is communicating (and before it is communicating it is already unavailable), so it should not be too much delay.   Whatever you are seeing after your change is what you will get, just moving it avoids the possiblity of a race condition where it might take longer.

larueli: Change is done, I let you resolve this conversation if it's ok for you.

PR #363, opened 2023-01-30
Add support for generic lightbulbs.

austinhodak:
There does not seem to be a configuration file for generic white Tuya bulbs.

This file is working with 2 of my basic Merkury bulbs. I'm sure it works with other white Tuya bulbs also such as Feit.

> DP 1: `switch`
> DP 2: `brightness`

I don't like making it a generic config, but I was surprised when I was setting up the integration that none of the included ones worked with these basic bulbs.

**Suggestion**: Add an "override" option on stage two of the configuration in HA. It wouldn't have made a difference in my case as there isn't a config for these DPs, but it would have saved me some troubleshooting time.

MI-BW911-999W

MI-BW932-999W

### DIFF
--- a/custom_components/tuya_local/devices/simple_lightbulb.yaml
+++ b/custom_components/tuya_local/devices/simple_lightbulb.yaml
@@ -1,0 +1,13 @@
+name: Simple lightbulb
+primary_entity:
+  entity: light
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+    - id: 2
+      name: brightness
+      type: integer
+      range:
+        min: 10
+        max: 255


make-all:
Sorry, I don't understand what you mean by "add an override option on stage two of the configuration in HA".  Ovrride what, and with what?

austinhodak:
Sorry, what I meant by that was when I got to stage 2 of adding these light bulbs, I was only given 2 options to select from, an outlet and some other random device. My original thought was to allow users to "bypass" the device selection and pick from a list of all the devices, however now that I think about it that is probably not the best option.

Maybe something like `local-tuya` does where it displays the DPs found and lets the user manually select what each DP does.

make-all:
You are right, it has been requested before, but it is not the best option, because all it is doing is allowing the user to waste their time trying devices at random that you already know are not going to be compatible with the device due to mismatches in the dps available.

austinhodak:
> For the option of letting the user manually configure the device, localtuya already exists to do that, so I have no plan to duplicate that effort.

Unfortunately local-tuya is consistently broken for me, and the past few releases have bricked a lot of installations, judging by the 720 open issues. I've found your integration to be much more stable, and for the most part after the initial setup, I haven't had to touch anything, I can't say the same for LT.

Issue #328, opened 2022-12-30
AlecoAir D16 not detected correctly


culdaraul:
Hello there.
I've tried today to setup my Alecoair D16 dehumidifier - known to be supported but seems that I do have some issues. Is there something wrong (maybe a newer FW on my unit) which makes the identification not working?

![image](https://user-images.githubusercontent.com/8573177/210113757-a8258a6a-3bb3-40d5-969a-f9c3cc08c98c.png)

![image](https://user-images.githubusercontent.com/8573177/210113578-0209049f-e0a9-4b4e-8344-74a5197a3f11.png)

I'm running 2022.12.8 on my VM server. Any help would be appreciated.

make-all:
It seems to be using a completely different command/data mapping than the AlecoAir D16 that was submitted previously.
Please go through the steps in the README.md file regarding finding info for new devices.

make-all:
OK, that is useful information, as it gives us something obvious to differentiate the two configurations.  But I still need more info about the dps values to make a config.  With just the list from the log they are nothing but a list of random numbers and booleans.

masterminds84:
> OK, that is useful information, as it gives us something obvious to differentiate the two configurations. But I still need more info about the dps values to make a config. With just the list from the log they are nothing but a list of random numbers and booleans.

I am newbie in the Home Assistant. I will help you with the DPS values, but please tell me, how can I do this values get. The purify is in the Tuya IOT platform, and the Tuya Smart app can controll it now.
Thank you your answer.

make-all:
There are details here https://github.com/make-all/tuya-local#device-support

masterminds84:
There are the values:

```
Power - 1
Mode - 2
Humidity - 3
Set Humidity - 4
Anion - 5
Inside Drying - 101
Wind Speed - 6
Water Pump - 102
Child Lock - 7
Swing - 8
Fault - 11
Countdown - 12
Remaining Time - 13
```

make-all:
If you can post the functions and status info from any of the API Explorer functions that list the above (don't worry about finding the one with dps, because you already posted them, as long as it has the names), that will contain ranges and indicate which are read-only and which writable.

masterminds84:
I found only just this, but I know, that the Set Humidity values is: 30-35-40-45-50-55-60-65-70-75-80. 

`{
  "result": {
    "category": "cs",
    "functions": [
      {
        "code": "anion",
        "lang_config": {
          "false": "OFF",
[...]

PR #386, opened 2023-02-10
Add AlecoAir D12 Home dehumidifier

romicaiarca:
Please take a look over the file since is first device that I'm adding. 

P.S.: It would be helpfully a readme with description how to use these files.

### DIFF
--- a/custom_components/tuya_local/devices/alecoair_d12_home_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/alecoair_d12_home_dehumidifier.yaml
@@ -1,7 +1,7 @@
 name: AlecoAir D12 Home dehumidifier
 products:
   - id: hp1sjssyokokbrsg
-#    name: AlecoAir D12 Home
+   name: AlecoAir D12 Home
 primary_entity:
   entity: humidifier
   class: dehumidifier
@@ -11,21 +11,21 @@ primary_entity:
       type: boolean
       mapping:
         - dps_val: false
-          icon: "mdi:power"
+          icon: "mdi:air-humidifier-off"
           icon_priority: 1
         - dps_val: true
-          icon: "mdi:power"
+          icon: "mdi:air-humidifier"
           icon_priority: 4
     - id: 2
       name: mode
-      type: integer
+      type: string
       mapping:
         - dps_val: Continuous
-          value: 1
+          value: "1"
           icon: mdi:fan
           icon_priority: 3
         - dps_val: Auto
-          value: 0
+          value: "0"
           icon: "mdi:fan-auto"
           icon_priority: 3
     - id: 4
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/alecoair_d12_home_dehumidifier.yaml
```diff
@@ -0,0 +1,235 @@
...
+      type: integer
+      range:
+        min: 30
+        max: 80
+      mapping:
+        - step: 5
+    - id: 11
+      name: error
+      type: bitfield
+      # Docs mention E1 and E2, it isn't clear which is tank full, so for now
```
make-all, line 42: It should be fairly easy to find which error code is related to the tank.  Normally the same error will trigger when you remove the tank.  To check it is the same error code, you can fill the tank up before you put it back in (or see what error appears in the app).

custom_components/tuya_local/devices/alecoair_d12_home_dehumidifier.yaml
```diff
@@ -0,0 +1,235 @@
...
+        name: option
+        type: integer
+        mapping:
+          - dps_val: Cancel
+            value: 0
+            icon: mdi:cancel
+            icon_priority: 3
+          - dps_val: 1 hour
+            value: 1
+            icon: "mdi:store-clock"
```
make-all, line 85: I think you have misunderstood the icons here.  These icons do not appear in the list of options, they are applied to the entity when the specified value is selected.  I think just the "mdi:timer" icon is enough to convey the information that this entity is for the timer, but "mdi:store-clock" would work too but does not need setting for every value, just replace the "mdi:timer" above.  In particular the "mdi:cancel" icon is ambiguous, so I would leave that out.

custom_components/tuya_local/devices/alecoair_d12_home_dehumidifier.yaml
```diff
@@ -0,0 +1,235 @@
...
+            icon_priority: 3
+          - dps_val: 23 hours
+            value: 23
+            icon: "mdi:store-clock"
+            icon_priority: 3
+          - dps_val: 24 hours
+            value: 24
+            icon: "mdi:store-clock"
+            icon_priority: 3
+  - entity: select
```
make-all, line 179: This should be a fan entity.  Make this the speed (mapped to values 50 and 100), and make dp 1 the switch.  Using standard entities where appropriate like this will help with things like voice control.

custom_components/tuya_local/devices/alecoair_d12_home_dehumidifier.yaml
make-all: Based on the info on #328, which is for D16 and D14 models which seem to match this same layout, there may also be a dp id 13 for the remaining time, which I guess needs to be optional, and is maybe only sent by the device when the timer is running.

PR #394, opened 2023-02-12
Add Stadler Form Eva humidifier

bergdahl:


bergdahl:
I opted to set the fan speed to a dropdown as the speeds are 1-5. I couldn't find a way to use a slider that did not start with 0.

Issue #397, opened 2023-02-13
AlecoAir D12 Home is not detelcted


romicaiarca:
**Describe the bug**
AlecoAir D12 Home is not detelcted properly.

**To Reproduce**
Try to add new device setting: Device id = dehumidifier_id, IP address = dehumidifier_ip_address, local key = dehumidifier_local_key and protocol=auto.

Here is the error that I can see:
```
Device matches garage_door_opener with quality of 18%. DPS: {'updated_at': 1676254822.5699, '1': False, '2': '1', '3': 68, '4': 30, '5': False, '6': '1', '7': True, '11': 0, '12': '0', '13': 0, '101': False}
Report this to https://github.com/make-all/tuya-local/issues/
```

```
Logger: homeassistant
Source: custom_components/localtuya/discovery.py:67
Integration: LocalTuya ([documentation](https://github.com/rospogrigio/localtuya/), [issues](https://github.com/rospogrigio/localtuya/issues))
First occurred: 4:20:21 AM (2 occurrences)
Last logged: 4:20:26 AM

Error doing job: Exception in callback _SelectorDatagramTransport._read_ready()
Traceback (most recent call last):
  File "/config/custom_components/localtuya/discovery.py", line 65, in datagram_received
    data = decrypt_udp(data)
  File "/config/custom_components/localtuya/discovery.py", line 30, in decrypt_udp
    return _unpad(decryptor.update(message) + decryptor.finalize()).decode()
  File "/usr/local/lib/python3.10/site-packages/cryptography/hazmat/primitives/ciphers/base.py", line 186, in finalize
    data = self._ctx.finalize()
  File "/usr/local/lib/python3.10/site-packages/cryptography/hazmat/backends/openssl/ciphers.py", line 222, in finalize
    raise ValueError(
ValueError: The length of the provided data is not a multiple of the block length.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/local/lib/python3.10/asyncio/selector_events.py", line 1027, in _read_ready
    self._protocol.datagram_received(data, addr)
  File "/config/custom_components/localtuya/discovery.py", line 67, in datagram_received
    data = data.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa4 in position 10: invalid start byte
```

**Expected behavior**
To detect alecoair d12 home as option

[...]

PR #398, opened 2023-02-13
Fix detection alecoair d12 home

romicaiarca:
Fixing https://github.com/make-all/tuya-local/issues/397

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
-labels: ''
+labels: 'bug'
 assignees: ''
 
 ---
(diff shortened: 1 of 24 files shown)


make-all:
Please try the latest main HEAD before making further reversions.

romicaiarca:
> Please try the latest main HEAD before making further reversions.

Ok, Now should be fine. I've tested with this version and works fine :)

make-all:
1. The advantage of using the standard HA fan entity is that it will just work with voice assistants and other external integrations.  The fan uses a slider, I don't think it is a major issue, as it should snap to the closest setting: right for high, middle for low, left for off.
2. This seems to be an issue, but I cannot see any difference between the humidity in your config and in the installed one.  Are there any messages in the log?  It may be that recent changes to scale handling have messed up some cases where scale is not specified and should default to 1.  Or the device may not be returning any value for some values, or some dehumidifiers return "CO" when in constant humidifying mode, which could be causing this.

romicaiarca:
> 



> 1. The advantage of using the standard HA fan entity is that it will just work with voice assistants and other external integrations that tend to use the fan.increase_speed and fan.decrease_speed commands rather than setting a specific value.  The fan uses a slider, I don't think it is a major issue, as it should snap to the closest setting: right for high, middle for low, left for off.    You can manually configure your UI to use "High" and "Low" buttons that call fan.set_percentage (percentage=50 and percentage=100 respectively) if you don't like the default UI.
> 2. This seems to be an issue, but I cannot see any difference between the humidity in your config and in the installed one.  Are there any messages in the log?  It may be that recent changes to scale handling have messed up some cases where scale is not specified and should default to 1.  Or the device may not be returning any value for some modes, or some dehumidifiers return "CO" when in constant humidifying mode, which could be causing this.

1. Ok, then there is other issue. It looks like when I slide to left to 0, the dehumidifier is not turning off. 
2. After multiple restarts of HA, now the slider/display of value are ok.

make-all:
Its a while since I had a fan that works with HA (now I only have an air conditioning unit, but that works differently, not with a separate fan entity). I thought I remembered them working like that, but maybe they have changed, or I am misremembering.

romicaiarca:
@make-all , should I close the pull request since you are saying that is better to keep the current configuration for this new device?

make-all:
If the device is working in the current release, then please close this pull request.

If there is a problem that needs addressing still, maybe make a new branch based off the current HEAD, because this is saying it cannot be rebased due to conflicts.

PR #406, opened 2023-02-13
extended support for mh1823d

apoplavs:
* data from internal sensor
* Set temperature entity

### DIFF
--- a/custom_components/tuya_local/devices/minco_mh1823d_thermostat.yaml
+++ b/custom_components/tuya_local/devices/minco_mh1823d_thermostat.yaml
@@ -252,6 +252,40 @@ secondary_entities:
             value: C
           - dps_val: f
             value: F
+  - entity: number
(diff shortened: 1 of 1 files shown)


make-all:
These changes just duplicate information that is already accessible through the climate entity.

apoplavs:
> These changes just duplicate information that is already accessible through the climate entity.

@make-all Without these changes user can build automation using only external sensor.

apoplavs:
file:///home/andrii/screen.png

make-all:
You are mistaken.  I can't see enough of the screen behind there to tell what you are looking at and why you have gotten this impression, but it is entirely possible to use the Current Temperature attribute of climate entities in a Numeric State trigger, and to call the Set Temperature service on a climate entity as an action.

apoplavs:
ok, thanks lot

PR #431, opened 2023-02-17
Updated W'eau pool heatpump 13kW

abroeders:
Renamed entities of secondary section.
Moved unknown dps entities to primary climate entity and renamed them to unknown_xxx


make-all:
This is already included in the merged PR

PR #445, opened 2023-02-19
Add energy DP ID for Generic smartplug with energy monitoring (newer models)

liondag:
It would be great to add the energy DP ID for smartplug devices. The ID is called "Add Electricity" and the code = 17
Capture d’écran 2023-02-19 à 20 30 50


make-all:
This PR is not against tuya-local/main.  It needs to be rebased to be acceptable.

But there are plenty of smartplugs that contain the energy reading already.  In 2023.2.3, smartplugv2_energy was modified to make most of the attributes optional, so it should match most/all such plugs now.

PR #465, opened 2023-02-25
Add support for Breville Smart Air Connect Purifier

hebenon:
This PR adds support for the Breville Smart Air Connect Purifier (model lap308). I've modeled this off of the Breville Easy Air Purifier. The key differences between the models is the addition of a PM2.5 sensor, Air Quality indicator and automatic/manual fan mode.

### Reference device information:
Manufacturer website: https://www.breville.com/au/en/products/air-purifier/lap308.html

Product ID: tltdxtehttjbkjni

#### Device Attributes
```
{
  "result": {
    "category": "kj",
    "functions": [
      {
        "code": "switch",
        "dp_id": 1,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "mode",
        "dp_id": 3,
        "type": "Enum",
        "values": "{\"range\":[\"manual\",\"auto\"]}"
      },
      {
        "code": "light",
        "dp_id": 8,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "uv",
        "dp_id": 9,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "filter_reset",
        "dp_id": 11,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "countdown_set",
        "dp_id": 19,
        "type": "Enum",
        "values": "{\"range\":[\"cancle\",\"2h\",\"4h\",\"8h\"]"
      },
      {
        "code": "fan_speed_enum",
        "dp_id": 4,
        "type": "Enum",
        "values": "{\"range\":[\"low\",\"mid\",\"high\"]}"
      }
    ],
    "status": [
      {
        "code": "switch",
        "dp_id": 1,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "pm25",
        "dp_id": 2,
        "type": "Integer",
        "values": "{\"unit\":\"\",\"min\":0,\"max\":9999,\"scale\":1,\"step\":1}"
      },
      {
        "code": "mode",
        "dp_id": 3,
        "type": "Enum",
        "values": "{\"range\":[\"manual\",\"auto\"]}"
      },
      {
        "code": "light",
        "dp_id": 8,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "uv",
        "dp_id": 9,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "filter_reset",
        "dp_id": 11,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "filter_days",
        "dp_id": 16,
        "type": "Integer",
        "values": "{\"unit\":\"day\",\"min\":0,\"max\":1000,\"scale\":0,\"step\":1}"
      },
      {
        "code": "countdown_set",
        "dp_id": 19,
        "type": "Enum",
        "values": "{\"range\":[\"cancle\",\"2h\",\"4h\",\"8h\"]}"
      },
      {
        "code": "countdown_left",
        "dp_id": 20,
        "type": "Integer",
        "values": "{\"unit\":\"分钟\",\"min\":0,\"max\":720,\"scale\":0,\"step\":1}"
      },
      {
        "code": "air_quality",
        "dp_id": 22,
        "type": "Enum",
        "values": "{\"range\":[\"great\",\"good\",\"medium\",\"severe\"]}"
      },
      {
        "code": "fan_speed_enum",
        "dp_id": 4,
        "type": "Enum",
        "values": "{\"range\":[\"low\",\"mid\",\"high\"]}"
      }
    ]
  },
  "success": true,
  "t": 1677236178619,
  "tid": "ddca0830b43111edad0316dc0085482e"
}
```

### DIFF
--- a/custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
+++ b/custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
@@ -13,7 +13,7 @@ primary_entity:
         - dps_val: false
           icon: "mdi:air-purifier-off"
     - id: 3
-      name: mode
+      name: preset_mode
       type: string
       mapping:
         - dps_val: "manual"
@@ -41,19 +41,6 @@ secondary_entities:
         class: measurement
         type: integer
         unit: µg/m3
-  - entity: select
-    name: Fan Mode
-    icon: "mdi:auto-mode"
-    category: config
-    dps:
-      - id: 3
-        type: string
-        name: option
-        mapping:
-          - dps_val: "auto"
-            value: "Auto"
-          - dps_val: "manual"
-            value: "Manual"
   - entity: switch
     name: Night mode
     category: config
@@ -105,7 +92,7 @@ secondary_entities:
       - id: 20
         name: sensor
         type: integer
-        unit: min
+        unit: d
   - entity: sensor
     name: Air quality
     class: enum
--- a/custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
+++ b/custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
@@ -1,4 +1,4 @@
-name: Breville Smart Air
+name: Breville Smart Air Connect Purifier
 products:
   - id: tltdxtehttjbkjni
 primary_entity:
--- a/custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
+++ b/custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
@@ -40,7 +40,7 @@ secondary_entities:
         name: sensor
         class: measurement
         type: integer
-        unit: µg/m3
+        unit: ugm3
   - entity: switch
     name: Night mode
     category: config
@@ -92,7 +92,7 @@ secondary_entities:
       - id: 20
         name: sensor
         type: integer
-        unit: d
+        unit: min
   - entity: sensor
     name: Air quality
     class: enum
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -112,7 +112,7 @@
 ### Air Purifiers
 
 - Breville Easy Air purifier
-- Breville Smart Air purifier
+- Breville Smart Air Connect purifier
 - essentials portable air purifier
 - Himox H05 and H06 air purifiers
 - Hosome air purifier
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -112,6 +112,7 @@
 ### Air Purifiers
 
 - Breville Easy Air purifier
+- Breville Smart Air purifier
 - essentials portable air purifier
 - Himox H05 and H06 air purifiers
 - Hosome air purifier
--- a/custom_components/tuya_local/devices/breville_smartair_purifier.yaml
+++ b/custom_components/tuya_local/devices/breville_smartair_purifier.yaml
@@ -1,0 +1,125 @@
+name: Breville Smart Air
+products:
+  - id: tltdxtehttjbkjni
+primary_entity:
+  entity: fan
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+      mapping:
+        - dps_val: true
+          icon: "mdi:air-purifier"
+        - dps_val: false
+          icon: "mdi:air-purifier-off"
+    - id: 3
+      name: mode
+      type: string
+      mapping:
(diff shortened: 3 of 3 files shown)


custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
```diff
@@ -0,0 +1,125 @@
...
+    name: PM2.5
+    icon: "mdi:air-filter"
+    class: pm25
+    dps:
+      - id: 2
+        name: sensor
+        class: measurement
+        type: integer
+        unit: µg/m3
+  - entity: select
```
make-all, line 44: This is duplicating the mode under fan.  A separate entity for it is not required.

custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
```diff
@@ -0,0 +1,125 @@
...
+    name: Micro shield
+    category: config
+    icon: "mdi:shield-bug"
+    dps:
+      - id: 9
+        name: switch
+        type: boolean
+  - entity: sensor
+    name: Filter days left
+    category: diagnostic
```
make-all, line 75: Suggest adding `class: duration` . unit (below) should be `d` (see https://github.com/home-assistant/core/blob/dev/homeassistant/const.py#L595)

custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
hebenon: When it's placed under the primary entity, it seems to be informational only and doesn't allow the user to change the value - but does show the current setting on the main Lovelace panel for the device (see below). Having it as a sub entity also allows for configuration of the option. I'm more than happy to deduplicate it though. Do you have a preference on where the dp should be set?
smart_air_connect_panel

make-all: Ah, I missed that.  Under fan, the name should be preset_mode, not mode.

hebenon: Ah, cool, I understand how that element works now. I've updated the name to preset_mode and removed the secondary entity.

custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
hebenon: Done, the unit is now `d`.

custom_components/tuya_local/devices/breville_smartairconnect_purifier.yaml
```diff
@@ -105,7 +92,7 @@ secondary_entities:
       - id: 20
         name: sensor
         type: integer
-        unit: min
+        unit: d
```
make-all, line 95: It looks like you changed the wrong one here.  This was correct at "min", I think.  The one to change to "d" should be the "days" one.  Also, I suggested adding class: duration there as well (this one already has it).

hebenon: Whoops, you're right, days make no sense there. I tried setting the Filter days left with `class: duration` and the `d` unit. The result is technically true (4128:00:00 is 172 days), but may not be immediately interpretable to the user (see screenshot below).
smart_air_connect_filter_days_w_duration

The config for this was:
```
  - entity: sensor
    name: Filter days left
    category: diagnostic
    class: duration
    icon: "mdi:air-filter"
    dps:
      - id: 16
        name: sensor
        type: integer
        unit: d
```

Are there other advantages to specifying the class as duration (e.g. in different themes)? Or is it better to drop `class: duration` and specify the unit as `days` for readability?

make-all: I wasn't aware that duration only displayed up to hours. I thought it would use a sensible unit for the scale of the current value, which in this case might be months or weeks. If it only displays h:m:s, then it would be better without the class.

hebenon: No worries, I've reverted the `unit: d` change on the Time left sensor.

hebenon:
The most recent commit also changes the unit of the PM2.5 sensor to be `ugm3`. Looking through other similar devices, this was the unit being used.

PR #500, opened 2023-03-05
loratap is now cover

pasqualehun:
Loratap is now workign as a cover, not just 3 buttons
before
after


make-all:
Actually this is how I originally implemented it, but the reporter said it is not working.

pasqualehun:
For my loratap switch, my version works like a charm. Should I create a brand new yaml, like loratap_cover.yaml?

make-all:
No, the issues reported were not with the device itself, but the HA cover interface.  I made some changes to try to fix it, and maybe the reporter was not testing the updated version when they reported the fixes did not work.

PR #540, opened 2023-03-16
Added supporting for Avatto WT(ME)81 floor thermostat

yarosman:
It were cases when some dps are missing, for resolving it you should change corresponding via tuya mobile app

https://www.aliexpress.com/item/4000398480147.html?spm=a2g0o.order_list.order_list_main.5.33351802Yh6acE

### DIFF
--- a/custom_components/tuya_local/devices/avatto_wt81_floor_thermostat.yaml
+++ b/custom_components/tuya_local/devices/avatto_wt81_floor_thermostat.yaml
@@ -1,0 +1,213 @@
+name: Avatto WT81 floor thermostat
+products:
+  - id: gogb05wrtredz3bs
+    name: Avatto WT81(ME81) floor thermostat
+primary_entity:
+  entity: climate
+  dps:
+    - id: 1
+      name: hvac_mode
+      type: boolean
+      mapping:
+        - dps_val: true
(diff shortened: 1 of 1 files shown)


yarosman:
> As the functionality appears to be almost identical to Beok TGR81, I will merge the additional dps 101 and 102 as optional dps on that config. They are so close, down to the 81 in the model number, that I suspect they are just rebrandings of the same device, and the original submitter missed those two dps for some reason.

Hello, @make-all 
It make sense. It strange that it didn't see beok tgr81 as similar devices. 
For `dp:45` I have extended information with enum, maybe  we can add it under another name ?

make-all:
Are the codes for dp 45 confirmed?  Often these bitmap types are not simply assigned from bit 1 increasing, they often skip bits, and have additional undocumented codes mixed in.  Normally I will add the fault_code as an attribute on the primary entity for more info, and have a binary_sensor for indicating that there is an error.  As the beok tgr81 was a pull request, the author doesn't seem to have aed the full into attribute, so I will add it as part of the change.

 If there are codes that are expected in normal use, (like tank full codes on dehumidifiers, or tank empty on humidifiers and aroma diffusers) then another binary sensor for just those is justified.

PR #581, opened 2023-03-31
Create gx_aroma_diffuser.yaml

MackoMici:


make-all:
I merged this with the almost identical YYM device that has already been added.

PR #582, opened 2023-03-31
Add support for Silvercrest Sous Vide Stick SSVSS 1200 A1 (100358082)

eugenelitvinov:
Device is similar to Inkbird sous vide, but doesn't support id: 102 for current status.


make-all:
I modified the inkbird config to also cover this, since they are almost the same.

PR #589, opened 2023-04-03
Added Boolean Switch (ID 2 part of primary entity

thewolf-oz:


custom_components/tuya_local/devices/arlec_thermostat_smartplug.yaml
```diff
@@ -6,6 +6,10 @@ primary_entity:
   entity: climate
   dps:
     - id: 2
+      type: boolean
+      name: switch
```
make-all, line 10: 'switch' is not part of the climate spec, so this just ends up being a read-only attribute that duplicates the switch entity.

I think the only change here is to change the id from 2 to 9.

thewolf-oz: It's working... and isn't read-only. Not sure if it's undocumented. Alternatively, could be moved down to the Config/Secondary Entities

thewolf-oz: I never removed it from below and it was in the version published. Removing it from the secondary entities causes it to remove the switch from the Controls area (without it showing in the Configuration

image

I'm about to update the file to move it only to the configuration as a "Manual Switch". Is it at all possible to add the option to the Category control to place items in the "Control" section? I'm not sure if the method I've used in the current version is allowing it to work and place it in the Control section even though Climate Spec doesn't otherwise allow it?

make-all: Just adding additional commits to your repository will update this PR.

If you remove 'category: config' from the master switch, it will appear in the controls section instead of config. Also, you can remove the other switch for dp 9, as that can now be controlled by turning the climate entity on and off.

thewolf-oz: OK! Thanks! That seems to have done it. File updated in my repo. DP9 is required for the auto mode, the switch is just an override (the equivalent of pressing the button on the device). I'll only use the switch/button however others will use DP9 that does the automated action based on the setting

image

Issue #578, opened 2023-03-30
Arlec 2200W Grid Connect Smart Convection Panel Heater


Silvrav:
# Log Message

```
Please paste logs here
```

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/HANNES~1/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/HANNES~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

DP ID | Code | example
-- | -- | --
switch | 1 | OFF
Set   Temperature | 2 | 20'C
Current   Temperature | 3 | 17'C
Countdown | 19 | 1h
Mode | 4 | eco, comfort,   anti-frost
childlock | 7 | OFF

</body>

</html>

```
{
  "result": {
    "category": "qn",
    "functions": [
      {
        "code": "switch",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":5,\"max\":50,\"scale\":0,\"step\":1}"
      },
      {
        "code": "lock",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "countdown_set",
        "type": "Enum",
        "values": "{\"range\":[\"cancel\",\"1h\",\"2h\",\"3h\",\"4h\",\"5h\",\"6h\"]}"
      }
    ],
    "status": [
      {
        "code": "switch",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":5,\"max\":50,\"scale\":0,\"step\":1}"
      },
      {
        "code": "temp_current",
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":-20,\"max\":50,\"scale\":0,\"step\":1}"
      },
      {
        "code": "lock",
        "type": "Boolean",
[...]

PR #590, opened 2023-04-03
Arlec 2400W Panel Heater

thewolf-oz:
After seeing https://github.com/make-all/tuya-local/issues/578, I figured I'd have a go at creating the file for my Arlec 2400W Panel Heater based on the andersson_gsh_heater. I've tested this with my heater and it now has the child lock and timer function working where the anderson file didn't


make-all:
I manually merged this, to untangle the conflicting other changes that were bundled in with it.

PR #599, opened 2023-04-08
Hungarion translation, new energy monitoring device

erelke:
I added hungarian translation, and a new smart energy monitoring device RDCBC 1P+N


make-all:
I merged the translations. The switch I have dealt with by modifying an existing config which was almost covering this - by making two of the dps optional, and adding one more optional dp, it was able to match smartplugv2_energyv2

Issue #504, opened 2023-03-07
Request support for doorbell WGH-ML-001


nijel:
# Log Message

```
Device matches None with quality of 0%. DPS: {"updated_at": 1678177158.7740445, "2": 1, "3": 42, "6": false}
Report this to https://github.com/make-all/tuya-local/issues/
```

# Information about DPS mappings

```json
{
  "result": {
    "category": "wxml",
    "functions": [
      {
        "code": "doorbell_ring_value",
        "dp_id": 2,
        "type": "Integer",
        "values": "{\"unit\":\"\",\"min\":0,\"max\":40,\"scale\":0,\"step\":1}"
      },
      {
        "code": "doorbell_volume_value",
        "dp_id": 3,
        "type": "Integer",
        "values": "{\"min\":0,\"max\":100,\"scale\":0,\"step\":1}"
      },
      {
        "code": "alarm_propel_switch",
        "dp_id": 6,
        "type": "Boolean",
        "values": "{}"
      }
    ],
    "status": [
      {
        "code": "doorbell_ring_value",
        "dp_id": 2,
        "type": "Integer",
        "values": "{\"unit\":\"\",\"min\":0,\"max\":40,\"scale\":0,\"step\":1}"
      },
      {
        "code": "doorbell_volume_value",
        "dp_id": 3,
        "type": "Integer",
        "values": "{\"min\":0,\"max\":100,\"scale\":0,\"step\":1}"
      },
      {
        "code": "alarm_propel_switch",
        "dp_id": 6,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "doorbell_call",
        "dp_id": 10,
        "type": "Integer",
        "values": "{\"unit\":\"\",\"min\":1,\"max\":255,\"scale\":0,\"step\":1}"
      }
    ]
  },
  "success": true
}
```

Additional DPs:

5 - alarm message - it triggers when button is pressed
1 - doorbell list data - buttons configuration
7 - disturb data - DND configuration

# Product ID

Product id: 26xtgamy3tv1awhl

# Information about how the device functions

Sold under many brands, for example https://www.aliexpress.com/item/1005004524467677.html

DP 2 / 3 are configuration options, and they trigger the bell as well. Having DP 2 as a service to trigger via automation is useful (even when the value doesn't change, the bell rings). Integers.

DP 6 toggles notifications via Tuya app. Boolean.

[...]

PR #625, opened 2023-04-19
Improve WHM-04 doorbell integration

nijel:
- make the ringing sensor actually trigger (needs to be based on DP 5)
- add it attribute with trigger name
- add names to siren tones

Followup for #504.

### DIFF
--- a/custom_components/tuya_local/devices/whm04_doorbell.yaml
+++ b/custom_components/tuya_local/devices/whm04_doorbell.yaml
@@ -10,88 +10,86 @@ primary_entity:
       type: integer
       name: tone
       mapping:
-        - dps_val: 0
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/whm04_doorbell.yaml
```diff
@@ -117,31 +120,21 @@ secondary_entities:
...
+      - id: 5
+        type: base64
         name: sensor
         optional: true
         persist: false
         mapping:
           - dps_val: null
             value: false
           - value: true
+      - id: 5
```
make-all, line 136: This will not work - the same dp cannot be repeated within the same entity.  That is why the sensor and binary_sensor were originally separated.

PR #651, opened 2023-04-26
Add support for gateway subdevices.

burn1879:
Add subdevices integration.
Can link device with parent.
device_cid = node_id or uuid from tuya iot device information


tests/test_config_flow.py
```diff
@@ -57,8 +58,9 @@ async def test_init_entry(hass):
             CONF_HOST: "hostname",
             CONF_LOCAL_KEY: "localkey",
             CONF_POLL_ONLY: False,
-            CONF_PROTOCOL_VERSION: "auto",
```
make-all, line 60: Do all these changes from "auto" to 3.3 mean that the auto protocol version detection is no longer working after this change?

burn1879: According to the tinytuya documentation, there is no auto protocol version detection.
If no protocol version is passed when creating the device, tinytuya initializes it to 3.1
`########################################################
#             Core Classes and Functions
########################################################

class XenonDevice(object):
    def __init__(
            self, dev_id, address=None, local_key="", dev_type="default", connection_timeout=5, version=3.1, persist=False, cid=None, node_id=None, parent=None # pylint: disable=W0621
    ):`

custom_components/tuya_local/config_flow.py
```diff
@@ -188,12 +194,14 @@ async def async_test_connection(config: dict, hass: HomeAssistant):
         await asyncio.sleep(5)
 
     try:
+        subdevice_id = config[CONF_DEVICE_CID] if CONF_DEVICE_CID in config else None
```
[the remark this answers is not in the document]
burn1879: i'll fix that

PR #655, opened 2023-04-27
Fix open/close for some covers and fixes for zemismart

xZetsubou:
There is a common issue in some covers where is some of them action is stuck on opening or closing and as for my motor is stuck on "opening " and the open button always disabled in HA in case it stuck on "closing" the close button will always be disabled.
This PR contains:
- Fix typo in motor mode continuation -> contiuation ```Probably Tuya typo```
  I don't know for sure if this typo's only in my motor model ```[Blind] 192.168.1.4 - [Off] - DPS: {'1': 'open', '2': 100, '3': 80, '5': 'forward', '7': 'opening', '11': 'fully_open', '12': 0, '103': True, '104': False, '105': True, '106': 'contiuation'}```
- Add new device ```zemismart_roller_shade_v2.yaml``` fixed the shade that have stuck action
- added _action_stuck in cover.py for stuck covers ```This the best way I found without breaking anything```

### DIFF
--- a/custom_components/tuya_local/cover.py
+++ b/custom_components/tuya_local/cover.py
@@ -42,6 +42,7 @@ class TuyaLocalCover(TuyaLocalEntity, CoverEntity):
         self._currentpos_dp = dps_map.pop("current_position", None)
         self._control_dp = dps_map.pop("control", None)
         self._action_dp = dps_map.pop("action", None)
+        self._action_stuck_dp = dps_map.pop("action_stuck", None)
         self._open_dp = dps_map.pop("open", None)
         self._reversed_dp = dps_map.pop("reversed", None)
         self._init_end(dps_map)
@@ -115,6 +116,11 @@ class TuyaLocalCover(TuyaLocalEntity, CoverEntity):
     @property
     def is_opening(self):
         """Return if the cover is opening or not."""
+        # this fix an issue where is some of shades actions is stuck
+        if self._action_stuck_dp:
+            pos = self.current_cover_position
+            if pos < 100:
+                return self._action_stuck_dp.get_value(self._device) == None
         # If dps is available to inform current action, use that
         if self._action_dp:
             return self._action_dp.get_value(self._device) == "opening"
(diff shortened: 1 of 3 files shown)


make-all:
I think this is better handled by modifying the action handling to skip trying to use "None" values without introducing a new "action_stuck" state.

xZetsubou:
> I think this is better handled by modifying the action handling to skip trying to use "None" values without introducing a new "action_stuck" state.

Thanks but, I tried your way first but I didn't fix the issue because the shade is returning some value not None.
I also tried to remove action DP from device configs and didn't fix it as well.

xZetsubou:
I'll try to find better solution for this asap

make-all:
If you can capture the dp values when this problem occurs it might help me to understand the problem.  Probably switching logging to debug for custom_components.tuya_local.device will help, if not, the same for tinytuya (but there is a lot of noise at that level).

xZetsubou:
after digging more into this issue it seems when HA started the open button is disabled, by clicking on close button the issue will be fixed and open will return to works and detect work state aswell, but when I use the set position command the issue will show up again.  [GIF to clarify more](https://user-images.githubusercontent.com/46300268/235278000-0e32e8a4-b9ed-4a82-8d1f-e4787d934383.gif)

xZetsubou:
Ok I tried to test as you said,
by trying to detect values by tinytuya I noticed that every time I run the wizard the issue show up again open button disabled and state stay on opening.

make-all:
Using another local protocol tool at the same time will cause the connection to be lost, so it is not surprising you see problems start then.

xZetsubou:
> Using another local protocol tool at the same time will cause the connection to be lost, so it is not surprising you see problems start then.

But the issue happen either way, at the start of HA the cover state always is opening.
Also when I tested tiny wizard while HA was running and it just confused me that while HA is running and trying to run tiny wizard the cover state goes opening. same goes delete/add the device the state will be opening.
so what I think the issue is each time tuya-local reconnect with cover the state always set to opening.

make-all:
There is no state kept by the integration.  So any discrepency is caused by a difference in the dps values reported (most likely dps not being reported until a change).
Were your above tests with the latest change ([0e0c26f](https://github.com/make-all/tuya-local/commit/0e0c26f3276a6014faa60e0e8c6bddaa792b1727))?  If not, then it might be worth checking again.

xZetsubou:
Yes I tested with latest changes same issue. also if ```is_opening``` is false the issue would be fixed but it won't detect opening action so. I'll keep testing more hoping for better solution.

make-all:
Sorry, I accidentally merged this, then reverted.  I don't see an option to reopen the same PR, so if you find a better way of fixing the issue, please file a new PR.

PR #661, opened 2023-04-29
Revert "Fix open/close for some covers and fixes for zemismart"

make-all:
Reverts make-all/tuya-local#655


make-all:
Accidentally merged wrong PR.  This one was already partially fixed in different ways, and still under investigation how to best fix it without introducing new "action_stuck" variant of "action" dp

PR #667, opened 2023-05-01
Add support for Antela smart plug (Italian socket)

pmazz:
Additional device info here: https://www.amazon.it/dp/B09NBN31HW


make-all:
What is the difference between this and existing configs?  Specifically "smartplugv2_energy", which seems to match all of the dps.

I don't want to end up with thousands of configs that differ just by their name.

pmazz:
Yes, "smartplugv2_energy" is the most similar config which I found, except for the 46 dp (overcharge cutoff) which I don't seem to have.
I understand you don't want to have many similar configs, so let me know how to proceed.
Thanks.

make-all:
I think you can just ignore the overcharge_cutoff switch, as it is already marked as optional.

pmazz:
Great, thank you.
👍

PR #672, opened 2023-05-04
Fix cover actions - Add zemismart roller

xZetsubou:
Hello, after many tests to fix actions and #655 didn't really fully fix the issues.
By far this the best way I found to detect cover actions, this PR fixing and improving that.

# Summery:
- Added device Zemismart Ignoring the action DP.
- Fix open/close stuck states.
- Detect whether the state is opening/closing even with set position.

### DIFF
--- a/custom_components/tuya_local/cover.py
+++ b/custom_components/tuya_local/cover.py
@@ -45,7 +45,7 @@ class TuyaLocalCover(TuyaLocalEntity, CoverEntity):
         self._open_dp = dps_map.pop("open", None)
         self._reversed_dp = dps_map.pop("reversed", None)
         self._init_end(dps_map)
-
+        self._current_action = "stop"
         self._support_flags = 0
         if self._position_dp:
             self._support_flags |= CoverEntityFeature.SET_POSITION
@@ -111,7 +111,23 @@ class TuyaLocalCover(TuyaLocalEntity, CoverEntity):
         if self._action_dp:
             state = self._action_dp.get_value(self._device)
             return self._state_to_percent(state)
-
+        
+    def update_action(self, action, position=None):
+        """Updating state action"""
+        # Set Commands func,
+        if self._current_action and self._control_dp and position is None:
+            self._current_action = action
+            return action
+        # Set position cmd, check if target position weither close or open
+        if action is "_set_position" and position is not None and self._current_action and self._position_dp:
+            pos_diff = self._maybe_reverse(position) - self.current_cover_position #self._maybe_reverse(self._position_dp.get_value(self._device))
+            if pos_diff > 0:
+                self._current_action = "set_opening"
+                return self._current_action
+            elif pos_diff <= 0:
+                self._current_action = "set_closing"
+                return self._current_action
+        
     @property
     def is_opening(self):
         """Return if the cover is opening or not."""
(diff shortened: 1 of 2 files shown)


make-all:
It seems you have broken three other devices with this change

xZetsubou:
These test devices doesn't handle None or false values which is weird because it's not necessary ```is_opening``` or ```is_closing``` need to return True value all the time if it's stopped both of them will return false or None, At least it should be like that.
**I might be wrong** either way, I tweaked the codes to these handle devices.

make-all:
I have cleaned up the logic of is_opening, is_closing and is_closed, and renamed the "action" dp on zemismart roller to reflect its unreliability.

I hope this can address the issues you are facing.

I did this rather than accept the PR as it is, because I do not think that keeping internal state based on previous commands will be a workable solution, as the device can also be operated from its own control panel or remote control, and the cloud app, so the likelihood of internal state getting out of sync is high.

xZetsubou:
I agree with you that's not the best solution, but this's the only solution as far as I know to detect the actual state, most of Tuya covers won't return the current state if it's not all of them, your changes [4aad42f](https://github.com/make-all/tuya-local/commit/4aad42fe0814ccf02c8ffe5b389fcc46da4e2b00) returning [unknown value](https://user-images.githubusercontent.com/46300268/236665695-8b3c5f7f-f09e-40ff-9619-ea4ed6293e1d.png) and when I use set position the state will stuck again whether [opening or closing](https://user-images.githubusercontent.com/46300268/236665723-0c7f49a4-766c-4e58-9217-a05110ff8aae.png) so set position and open/close actions won't works as it should be. 
we can sync states with current pos value in case device begin controlled by out of HA.
I won't force my solution I can rewrite cover.py locally it's fine, but I just wanted to fix this issue since cover state might be important for some people making automations.
Thanks for trying to fix this if you want me to try other fixes I wouldn't mind to test ^^

make-all:
Maybe it will help if you can explain what the exact state of all the dp are in various situations (both working and not working).

Part of the difficulty for me is that I do not have this device to see how it is acting.  Maybe we can use the action, just not in the direct way we are using it now, for example.  I see other configs use the current_position to allow it to also represent opened and closed states, so I made that change.  But depending on how dp 11 is working, it may be a better candidate for detecting opened/closed states.

Also, it may be worth adding `persist: false` to the command dp, maybe it will then get cleared instead of outdated commands staying and interfering with the open/close/opening/closing detection.

xZetsubou:
Detecting closed or opened isn't big deal, but the issue is with `opening` and `closing`,
 `is_opening` shouldn't return true unless the motor moving up `if pos < 95 and cmd is open` that means the state will always be `opening` as default state which will lead HA to disable `open` button.  same goes for `closing` will lead HA to disable `close` button. HA have `is_opening and is_closing` assuming that there's an action DP return whether `opening or closing or stopped` probably tuya covers doesn't have this?

- Open/Close won't update `DP 1` value to new states on the cloud so you can't relay on last cmd to detect action
  `when device reconnected cmd is always is open thats why it's stuck in opening at start this the main issue.`
  `this also why I set the default value of _current_state is stop`
- `DP 2` is the actual current pos and would be updating after open, close and set position.
- `DP 11` Won't change at all its probably broken as well.
- Open/Close won't update _currentpos_dp `DP 3` this only updated using set position.
- Using set position `DP 3` won't always update current pos `DP 2` unless the cover moved by 10 steps at least for my cover, 
  so it will only move the if inserted position `DP 2` > 5 OR  `DP 2` < -5  `This is how I detect action by set position`

So to fully relay on Tuya cloud states to detect current action perfectly all the time is I don't know if there's good way, using open and close won't change any DP till the cover stops then it will only update current pos `DP 2` to the new position.
Also this issue's in official Tuya integration the opening action won't be detected.

This Comment after I tested https://github.com/make-all/tuya-local/commit/546c0b013054abfb484a999946f637e5e09e54b1 https://github.com/make-all/tuya-local/commit/ba6edb54ff3b39d171fcb5a0d5036c29e82b4b9d

PR #690, opened 2023-05-08
Add Klarstein DryFy Connect support

syepes:
This MR adds full support for the Klarstein DryFy Connect dehumidifier (not the PRO version).

Note: The [PRO version ](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/klarstein_dryfy_pro_connect_dehumidifier.yaml) benefit from this update it just has some additional features.

@make-all Thanks for creating this integration this one is way more flexible then the OOTB one, hope to see this get integrated into the HACS.


syepes:
If you want you can adjust this PR so you that is suites your standards and I can give it a test run. With the actual device.

make-all:
The shinco config has been updated in main.  It isn't really a case of updating this PR, since the main change I want to make is not unnecessarily creating a new config file.

PR #753, opened 2023-06-01
Add 9in1 Air Quality Monitor

syepes:
Add support for the `ZN-2C09` 9in1 Air Quality Monitor

https://fr.aliexpress.com/item/1005005418667771.html
CleanShot 2023-06-02 at 12 17 16

### DIFF
--- a/custom_components/tuya_local/devices/9in1_airquality_monitor.yaml
+++ b/custom_components/tuya_local/devices/9in1_airquality_monitor.yaml
@@ -41,9 +41,6 @@ secondary_entities:
         name: sensor
         unit: "%"
         class: measurement
-        range:
-          min: 0
-          max: 100
   - entity: sensor
     name: CO2
     class: carbon_dioxide
@@ -74,9 +71,6 @@ secondary_entities:
         name: sensor
         unit: µg/m³
         class: measurement
-        range:
-          min: 0
-          max: 1000
         mapping:
           - scale: 1
             step: 1
@@ -90,9 +84,6 @@ secondary_entities:
         name: sensor
         unit: µg/m³
         class: measurement
-        range:
-          min: 0
-          max: 999
         mapping:
           - scale: 1
             step: 1
@@ -106,9 +97,6 @@ secondary_entities:
         name: sensor
         unit: µg/m³
         class: measurement
-        range:
-          min: 0
-          max: 999
   - entity: sensor
     name: PM 10
     class: pm10
@@ -119,9 +107,6 @@ secondary_entities:
         name: sensor
         unit: µg/m³
         class: measurement
-        range:
-          min: 0
-          max: 999
   - entity: sensor
     name: Battery
     class: battery
@@ -132,9 +117,6 @@ secondary_entities:
         name: sensor
         unit: "%"
         class: measurement
-        range:
-          min: 0
-          max: 100
   - entity: sensor
     name: Charge State
     class: enum
@@ -150,7 +132,6 @@ secondary_entities:
           - dps_val: true
             value: "Charging"
             icon: "mdi:battery-charging"
-            default: true
   - entity: sensor
     name: Alarm Volume
     category: diagnostic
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/9in1_airquality_monitor.yaml
```diff
@@ -0,0 +1,161 @@
...
+    name: HCHO
+    class: volatile_organic_compounds
+    icon: "mdi:factory"
+    dps:
+      - id: 5
+        type: integer
+        name: sensor
+        unit: µg/m³
+        class: measurement
+        range:
```
make-all, line 77: Again, range is not needed, as this is read-only and it will be ignored.

custom_components/tuya_local/devices/9in1_airquality_monitor.yaml
```diff
@@ -0,0 +1,161 @@
...
+      - id: 6
+        type: integer
+        name: sensor
+        unit: µg/m³
+        class: measurement
+        range:
+          min: 0
+          max: 999
+        mapping:
+          - scale: 1
```
make-all, line 97: Same as above for `range`, `scale` and `step`

custom_components/tuya_local/devices/9in1_airquality_monitor.yaml
```diff
@@ -0,0 +1,161 @@
...
+    name: PM 2.5
+    class: pm25
+    icon: "mdi:factory"
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
+        unit: µg/m³
+        class: measurement
+        range:
```
make-all, line 109: Same as above for `range`

custom_components/tuya_local/devices/9in1_airquality_monitor.yaml
```diff
@@ -0,0 +1,161 @@
...
+    name: PM 10
+    class: pm10
+    icon: "mdi:factory"
+    dps:
+      - id: 9
+        type: integer
+        name: sensor
+        unit: µg/m³
+        class: measurement
+        range:
```
make-all, line 122: Same as above for range

custom_components/tuya_local/devices/9in1_airquality_monitor.yaml
```diff
@@ -0,0 +1,161 @@
...
+    name: Battery
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 22
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+        range:
```
make-all, line 135: Same as above for range

custom_components/tuya_local/devices/9in1_airquality_monitor.yaml
```diff
@@ -0,0 +1,161 @@
...
+        type: boolean
+        name: sensor
+        mapping:
+          - dps_val: false
+            value: "Battery"
+            icon: "mdi:battery"
+          - dps_val: true
+            value: "Charging"
+            icon: "mdi:battery-charging"
+            default: true
```
make-all, line 153: If the intention is to make this apply in the case where the dp is missing, or has an unknown value, then removing `dps_val: true` is the way to do that. 

The `default` attribute is used only in one specific use case for siren tones, where it is not intended to appear as the current value, but when activating the siren, a tone is needed even if none is being returned as the .current value.

custom_components/tuya_local/devices/9in1_airquality_monitor.yaml
```diff
@@ -0,0 +1,161 @@
...
+      - id: 5
+        type: integer
+        name: sensor
+        unit: µg/m³
+        class: measurement
+        range:
+          min: 0
+          max: 1000
+        mapping:
+          - scale: 1
```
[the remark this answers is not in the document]
syepes: Fixed all the scale by double checking on the device and doing the conversion.

syepes:
Actual values on the device:
CleanShot 2023-06-03 at 7 55 58

PR #767, opened 2023-06-06
Add support for NovaDigital Switch (single/quad)

xgustavoh:
Add support for [NovaDigital Switch](https://www.novadigitalsmart.com.br/produtos/interruptor-touch-wi-fi-ws-us)

- Single:
image

- Quad:
image


make-all:
I have merged the single switch into somgom_single_switch, which is identical (dp 14 was missing from the config, but that is because it did not show up in logs, and the optional dps support was not available yet when it was originally added).

Please resubmit only the quad switch (and no .DS_Store file - you should add that to your global .gitignore if your OS automatically generates such files all over the place).

make-all (changes requested):
As above, please resubmit just the quad switch config.

xgustavoh:
@make-all 

Perfect!
I rebased it with just the `Novadigital quad switch` and removed the `.DS_Store`.

--

Thanks a lot for the global .gitignore tip.

make-all (approved):

PR #768, opened 2023-06-06
Fix Kogan LX10 Vacuum config

YodaDaCoda:
Sorry for not getting back to you on the OG PR - life got in the way.

Some DPs should have been optional, start/pause needed mapping in command. Some error codes mapped.

### DIFF
--- a/custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
+++ b/custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
@@ -6,15 +6,19 @@ primary_entity:
   entity: vacuum
   dps:
     - id: 101
-      name: power # clean_switch
+      name: start # clean_switch
       type: boolean
     - id: 102
-      name: activate # pause_switch
+      name: pause # pause_switch
       type: boolean
     - id: 104
       name: command # clean_mode
       type: string
       mapping:
+        - value: start
+          value_redirect: start
+        - value: pause
+          value_redirect: pause
         - dps_val: smart
           value: smart
         - dps_val: pose
@@ -108,6 +112,7 @@ primary_entity:
       type: string
     - id: 128
       name: message_report
+      optional: true
       type: string
     - id: 130
       name: sn
@@ -193,6 +198,7 @@ secondary_entities:
       - id: 115
         type: string
         name: button
+        optional: true
         mapping:
           - dps_val: gettime
             value: true
@@ -204,6 +210,7 @@ secondary_entities:
       - id: 115
         type: string
         name: button
+        optional: true
         mapping:
           - dps_val: resetsidebrush
             value: true
@@ -215,6 +222,7 @@ secondary_entities:
       - id: 115
         type: string
         name: button
+        optional: true
         mapping:
           - dps_val: resetmainbrush
             value: true
@@ -226,6 +234,7 @@ secondary_entities:
       - id: 115
         type: string
         name: button
+        optional: true
         mapping:
           - dps_val: resetfilter
             value: true
@@ -264,6 +273,7 @@ secondary_entities:
     dps:
       - id: 119
         name: sensor
+        optional: true
         type: integer
         unit: s
   - entity: sensor
@@ -274,6 +284,7 @@ secondary_entities:
     dps:
       - id: 120
         name: sensor
+        optional: true
         type: integer
         unit: s
   - entity: sensor
@@ -284,6 +295,7 @@ secondary_entities:
     dps:
       - id: 121
         name: sensor
+        optional: true
         type: integer
         unit: s
   - entity: button
@@ -293,10 +305,5 @@ secondary_entities:
     dps:
       - id: 129
         name: button
+        optional: true
         type: boolean
-
-
-
-
-
-
--- a/custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
+++ b/custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
@@ -87,9 +87,16 @@ primary_entity:
       type: boolean
     - id: 122
       name: error # robot_fault
-      # TODO: correctly parse the bitfield
-      type: integer
-      # "values": "{\"label\":[\"1\",\"2\",\"3\",\"4\",\"5\",\"6\",\"7\",\"8\",\"9\",\"10\",\"11\",\"12\",\"13\",\"14\",\"15\",\"16\",\"17\",\"18\",\"19\",\"20\",\"21\",\"22\",\"23\",\"24\",\"25\",\"26\",\"27\",\"28\",\"29\",\"30\"],\"maxlen\":30}"
+      type: bitfield
+      mapping:
+        - dps_val: 0
+          value: 0
+        - dps_val: 32
+          value: Drop sensor abnormal
+        - dps_val: 524288
+          value: Machine trapped
+        - dps_val: 1048576
+          value: Was picked up at work
     - id: 123
       name: path_comm
       optional: true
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/helpers/device_config.py
```diff
@@ -128,21 +128,43 @@ def secondary_entities(self):
...
-        return True
+        missing_dps = [dp for dp in required_dps if dp.id not in dps.keys()]
+        if len(missing_dps) > 0:
+            _LOGGER.debug(
+                "Not match for %s, missing required DPs: %s",
+                self.name,
+                [{dp.id: dp.type.__name__} for dp in missing_dps],
+            )
+
+        incorrect_type_dps = [
```
make-all, line 141: This changes the behaviour of the function, as it no longer checks for a typematch of optional dps that are present, which could lead to some false detections.

How about submitting two PRs, one for the config, and one for improvements to logging and variable naming.

custom_components/tuya_local/helpers/device_config.py
```diff
@@ -432,7 +454,8 @@ def values(self, device):
         """Return the possible values a dps can take."""
         if "mapping" not in self._config.keys():
             _LOGGER.debug(
-                "No mapping for %s, unable to determine valid values",
+                "No mapping for dpid %s (%s), unable to determine valid values (if this dpid doesn't require mapping, this is not an error)",
```
make-all, line 457: This is just a debug message, so I don't think it needs a note to say it might not be an error.  I don't think there is any circumstance where values() is called where there is not a list of values expected in response though, except maybe from logging and diagnostics code.

make-all (changes requested):

custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
```diff
@@ -32,19 +36,19 @@ primary_entity:
...
           value: mop
         - dps_val: fault
-          value: fault
+          value: error
         - dps_val: pause
-          value: pause
+          value: paused
         - dps_val: chargring
-          value: charging
+          value: docked
```
make-all, line 49: "charging" is also recognised as a status which translates to a docked state (status and state are separate in HA's vacuum  entity - state has very limited set of fixed states to choose from, status can be anything)

custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
```diff
@@ -293,10 +308,38 @@ secondary_entities:
...
         type: boolean
-
-
-
-
-
-
+  # add additional binary_sensor for each fault code
+  # add fault codes to mapping
+  - entity: sensor
```
make-all, line 315: I don't think this catch-all fault sensor is needed, since it will show up under the error attribute of the vacuum.  Also you can add a mapping to that to translate states into something more readable, rather than having binary sensors for each error, as there are typically dozens of possible errors for vacuums.

In any case, this sensor can only have a class of enum if there is a mapping to a list of possible values it can have (and that list must be  complete, or HA starts throwing errors when it gets unexpected values).

make-all (changes requested):

custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
YodaDaCoda: Ah making this an enum was accidental, thanks for picking that up.

I want this to report the name of the fault, or just the number if there wasn't a mapping - e.g. 'Trapped' (for a known fault code) or e.g. '12345' (for an unknown fault code). This allows an error state to surface, and further details of the error to be visible.

Regardless of the `type`, mappings for the `name: error` sensor are more-or-less ignored - an error state is surfaced but the details of the fault code are obscured. My thinking was to then have a 'fault code' sensor, which would surface known errors by name, and unknown errors by number. This didn't seem to work - if a mapping is specified, HASS wants it to be an enum and throws an error if it's not (not sure why, or if i'm doing something wrong - see error below). If i make it an enum, it throws an error if an unknown fault code is encountered (which makes sense). My thought from there is to create a single sensor to expose the error code, and separate binary sensors (as hinted in the docs RE bitfield) for each error code.

Without `class: enum` specified, any `mapping` produces the following error:
```hass           | 2023-06-13 10:14:15.488 DEBUG (MainThread) [custom_components.tuya_local.device] Retrying after exception <class 'ValueError'> Sensor sensor.kogan_lx10_robot_vacuum_fault_code is providing enum options, but is missing the enum device class (2/3)```

Comment (make-all) [reply custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml:315]: You can see the fault code on the error attribute of the vacuum.  A mapping there won't get caught up in HA's overly strict policing of sensor values.

Comment (YodaDaCoda) [reply custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml:315]: Thanks, so I can get the error via `state_attr('vacuum.kogan_lx10_robot_vacuum', 'error')` but as far as I can tell it's not otherwise surfaced to the user. I don't particularly care about the logbook, I care about being able to see _why_ there's an error, without having to go find the vacuum - this information is surfaced in the Tuya app & IMO it should be visible in HASS without additional work required by the end-user.

Comment (make-all) [reply custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml:315]: You can also surface it in most general cards (not the Glance card IIRC).  The Entity card has a UI for selecting attributes, with the Entities card, I think you still have to edit the yaml.  Or you can see it if you open the "Attribute" drawer in the more info window for the entity.

Comment (YodaDaCoda) [reply custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml:317]: 🤦 that's exactly what I wanted, thanks

File: custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml [hunk 5]
```diff
@@ -6,15 +6,19 @@ primary_entity:
   entity: vacuum
   dps:
     - id: 101
-      name: power # clean_switch
+      name: start # clean_switch
       type: boolean
```
make-all, line 10: I think for the value_redirect to work, a mapping is needed from the string value of the command to a boolean true:
```
mapping:
   - dps_val: true
     value: start
```
and likewise for pause.

custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
```diff
@@ -6,15 +6,19 @@ primary_entity:
...
     - id: 102
-      name: activate # pause_switch
+      name: pause # pause_switch
       type: boolean
     - id: 104
       name: command # clean_mode
       type: string
       mapping:
+        - value: start
```
make-all, line 18: I never tried using a value in a mapping without a dps_val before.  I am not sure if it will work without side effect, as it makes it the default mapping (and there are two, which makes it indeterminate which is default).

custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
```diff
@@ -83,9 +87,16 @@ primary_entity:
...
+      type: bitfield
+      mapping:
+        - dps_val: 0
+          value: 0
+        - dps_val: 32
+          value: Drop sensor abnormal
+        - dps_val: 524288
+          value: Machine trapped
+        - dps_val: 1048576
+          value: Was picked up at work
```
make-all, line 99: Is this a description of the error, or your note?

YodaDaCoda: Those values are the messages displayed by the Tuya Smart app when the robot flags the given dps_val - this seemed the most sensible value to use, rather than attempting to distill it down to a single word or something else myself.

custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
YodaDaCoda: For the record, it works as-is. 

Without these entries in the 'command' mapping, the glance card doesn't have start/pause buttons. With these entries, the buttons appear, and work as expected. Happy to test & adjust if you recommend.

PR #775, opened 2023-06-08
Add Support for PC311TY Energy Meter

Doudou14:
Add pc311ty_energy_meter.


Doudou14:
Update pc311ty_energy_meter.yaml
Add missing entity
Fix missing class Energy sensor
Fix "Generation B"

custom_components/tuya_local/devices/pc311ty_energy_meter.yaml
```diff
@@ -0,0 +1,222 @@
+name: PC311-TY power clamp
+products:
+  - id: vfwxetkavqrvg02u
+primary_entity:
+  entity: sensor
+  category: diagnostic
+  class: power
+  name: Total active power
```
make-all, line 8: For an energy meter, would it make more sense for the Total Energy reading to be the primary sensor?

make-all:
Please separate the translations into a separate PR

custom_components/tuya_local/devices/pc311ty_energy_meter.yaml
Doudou14: I changed the primary sensor

custom_components/tuya_local/devices/pc311ty_energy_meter.yaml
```diff
@@ -2,18 +2,6 @@ name: PC311-TY power clamp
...
-  name: Total active power
-  dps:
-    - id: 9
-      type: integer
-      name: sensor
-      unit: W
-      force: true
-      class: measurement
-secondary_entities:
   - entity: sensor
```
make-all, line 5: You'll need to remove the `-` here, as primary entity is not a list.

make-all (changes requested):

custom_components/tuya_local/devices/pc311ty_energy_meter.yaml
Doudou14: I think this is good, sorry I am beginner to home assistant

make-all (approved):

make-all:
Changes have been merged

PR #792, opened 2023-06-13
Core changes for improved device match debugging

YodaDaCoda:
Split from #768 

Rename a variable to help readability

WIP device matching logging - based on comment from #768 this needs to output information for optional sensors with mismatched types too

### DIFF
--- a/custom_components/tuya_local/helpers/device_config.py
+++ b/custom_components/tuya_local/helpers/device_config.py
@@ -140,7 +140,7 @@ class TuyaDeviceConfig:
 
         incorrect_type_dps = [
             dp
-            for dp in required_dps
+            for dp in self._get_all_dps()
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/helpers/device_config.py
```diff
@@ -128,21 +128,43 @@ def secondary_entities(self):
...
+        if len(missing_dps) > 0:
+            _LOGGER.debug(
+                "Not match for %s, missing required DPs: %s",
+                self.name,
+                [{dp.id: dp.type.__name__} for dp in missing_dps],
+            )
+
+        incorrect_type_dps = [
+            dp
+            for dp in required_dps
```
make-all, line 143: This loop needs to be over all dps to catch any optional ones that have incorrect types.

make-all (changes requested):

make-all (approved):

PR #801, opened 2023-06-15
Update wistar_roller_blind_nopos to match Tuya and TuyaLocal pre-2023…

mihsu81:
For the primary entity cover: Inverted dp_id 102, re-added dp_id 103 with inverted mapping.
Inverted the mapping of secondary entity switch dp_id 103.

### DIFF
--- a/custom_components/tuya_local/devices/wistar_roller_blind_nopos.yaml
+++ b/custom_components/tuya_local/devices/wistar_roller_blind_nopos.yaml
@@ -22,7 +22,15 @@ primary_entity:
         min: 0
         max: 100
       mapping:
-        - invert: true
+        - invert: false
+    - id: 103
+      type: boolean
+      name: reversed
+      mapping:
+        - dps_val: true
+          value: false
+        - dps_val: false
+          value: true
     - id: 105
       type: boolean
       name: curtain_hand
@@ -138,3 +146,8 @@ secondary_entities:
       - id: 103
         type: boolean
         name: switch
+        mapping:
+          - dps_val: true
+            value: false
+          - dps_val: false
+            value: true


custom_components/tuya_local/devices/wistar_roller_blind_nopos.yaml
```diff
@@ -138,3 +146,8 @@ secondary_entities:
       - id: 103
         type: boolean
         name: switch
+        mapping:
```
make-all, line 149: I don't think it is necessary to flip the meaning of the tuya dp, as this no longer has a special mapping within the cover entity, so it does not affect the behaviour of the cover like it used to.

custom_components/tuya_local/devices/wistar_roller_blind_nopos.yaml
```diff
@@ -22,7 +22,15 @@ primary_entity:
         min: 0
         max: 100
       mapping:
-        - invert: true
+        - invert: false
```
make-all, line 25: If this does not need inverting, then it is the first device encountered that doesn't.  Normally Tuya uses the opposite definition of 0 and 100 than Home Assistant does.

custom_components/tuya_local/devices/wistar_roller_blind_nopos.yaml
```diff
@@ -22,7 +22,15 @@ primary_entity:
...
+    - id: 103
```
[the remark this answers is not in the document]
mihsu81: Reverting it made the directions be recognized without toggling the Reversed switch.

custom_components/tuya_local/devices/wistar_roller_blind_nopos.yaml
mihsu81: In my case Tuya reports the meaning of Reversed opposite of how you'd display it and makes it appear correctly.

mihsu81:
I really don't understand why my PR upset you so much. I was only trying to match the Tuya layout in a device profile which apparently nobody else uses, but me.

PR #802, opened 2023-06-15
Update fr.json

Doudou14:
Fix French translation


make-all:
While making these corrections might also be a good time to add the new translations for climate preset_mode, swing_mode and fan__mode and humidifier mode strings in en.json.

make-all:
And based on the message that the branch cannot be rebased due to conflicts, you maybe need to rename your local branch, and start again from a new branch from the latest make-all/tuya-local main.  I think the fr.json should not be conflicting, but maybe some of the other changes in your local branch are conflicting with what was merged.

make-all:
I think the deletion of files, which were then added back via another pull request has confused git into thinking the histories are different even though there is a merge commit that should work as a new starting point for the new changes.

But generally it is better to do pull requests on branches, and keep your main branch in sync with the upstream, then each pull request goes on a different branch that is abandoned and optionally deleted after the PR is merged, and the main history is always clean to avoid conflicts.

make-all:
As you will start over, I will close this PR and wait for its successor.

Issue #759, opened 2023-06-02
Request support for Petlibro feeder with camera


kshepard339:
# Log Message

```
Device matches None with quality of 0%. DPS: {"updated_at": 1685733132.0985677, "101": true, "103": false, "105": false, "106": "0", "108": "0", "109": "31243360|3030192|28213168", "110": 1, "111": true, "117": 100, "134": false, "139": false, "140": "0", "150": true, "151": "2", "231": "7f060002017f101e04010011000200", "233": "standby", "235": 100, "236": 0, "237": "{\"value\":2,\"time\":1685700019}", "238": 2, "239": "record_stop", "240": "no_voice", "241": "full", "242": "record_end", "244": true, "247": "{\"value\":1,\"time\":1685666184}"}
Report this to https://github.com/make-all/tuya-local/issues/
```

# Information about DPS mappings

```
Device Specification Attribute from iot.tuya:
{
  "result": {
    "category": "sp",
    "functions": [
      {
        "code": "basic_indicator",
        "dp_id": 101,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "basic_flip",
        "dp_id": 103,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "basic_private",
        "dp_id": 105,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "motion_sensitivity",
        "dp_id": 106,
        "type": "Enum",
        "values": "{\"range\":[\"0\",\"1\",\"2\"]}"
      },
      {
        "code": "basic_nightvision",
        "dp_id": 108,
        "type": "Enum",
        "values": "{\"range\":[\"0\",\"1\",\"2\"]}"
      },
      {
        "code": "sd_format",
        "dp_id": 111,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "motion_record",
        "dp_id": 113,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "motion_switch",
        "dp_id": 134,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "decibel_switch",
        "dp_id": 139,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "decibel_sensitivity",
        "dp_id": 140,
        "type": "Enum",
        "values": "{\"range\":[\"0\",\"1\"]}"
      },
      {
        "code": "record_switch",
        "dp_id": 150,
        "type": "Boolean",
        "values": "{}"
[...]

PR #810, opened 2023-06-17
[BUGFIX] Feit-dimmer: Removing optional and DPS null value to make it work

alexeyatbluescape:
WIthout this fix, I can add the device, it reads the switch status, but when trying to control it dows not work.

### DIFF
--- a/custom_components/tuya_local/devices/feit_dimmer.yaml
+++ b/custom_components/tuya_local/devices/feit_dimmer.yaml
@@ -9,10 +9,6 @@ primary_entity:
     - id: 1
       name: switch
       type: boolean
-      optional: true
-      mapping:
-        - dps_val: null
-          value: false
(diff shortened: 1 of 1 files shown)


make-all:
I think this change should be unnecessary after the change from 3 days ago:

commit 3488430bad3ca2d415fcd8554d100c3b6fd82e56
Date:   Wed Jun 14 22:54:31 2023 +0900

    Avoid null mappings being applied in reverse.
    
    When a default is specified using a null mapping, the default value
    should not be reverse mapped back to null.  Also the logic should stop
    considering the mapping when it detects it as a default mapping, as
    the default should only.
    
    Issue #759

alexeyatbluescape:
@make-all Can confirm, added these changes locally, no issues. Should be good in the next release then! Thx for the great and simple integration!

PR #811, opened 2023-06-17
Add Immax NEO LITE VENTO Smart

LesTR:
DPS:
```
{'dps': {'1': False, '2': 'normal', '3': 2, '8': 'forward', '15': False, '22': '1hour'}}
```

### DIFF
--- a/custom_components/tuya_local/devices/immax_neo_light_vento.yaml
+++ b/custom_components/tuya_local/devices/immax_neo_light_vento.yaml
@@ -1,0 +1,80 @@
+name: Immax NEO LITE VENTO Smart
+products:
+  - id: ea6vpac5hfe5rqw4
+    name: Immax NEO LITE VENTO Smart
+primary_entity:
+  entity: fan
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+      mapping:
+        - dps_val: true
+          value: ON
+        - dps_val: false
+          value: OFF
+    - id: 2
+      type: string
+      name: preset_mode
+      category: config
+      mapping:
+        - dps_val: normal
+          value: Normal
+        - dps_val: sleep
+          value: Sleep
+        - dps_val: nature
+          value: Natural
+    - id: 3
+      name: speed
+      type: integer
+      category: config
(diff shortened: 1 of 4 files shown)


custom_components/tuya_local/devices/immax_neo_light_vento.yaml
```diff
@@ -0,0 +1,80 @@
...
+products:
+  - id: ea6vpac5hfe5rqw4
+    name: Immax NEO LITE VENTO Smart
+primary_entity:
+  entity: fan
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+      mapping:
```
make-all, line 11: This mapping is unnecessary

custom_components/tuya_local/devices/immax_neo_light_vento.yaml
```diff
@@ -0,0 +1,80 @@
...
+          conditions:
+            - dps_val: nature
+              step: 5
+            - dps_val: sleep
+              step: 5
+    - id: 8
+      name: direction
+      type: string
+      category: config
+      mapping:
```
make-all, line 46: I don't think this mapping is needed either, because direction has two fixed values in HA, and does not need the device to declare its possible values.

custom_components/tuya_local/devices/immax_neo_light_vento.yaml
```diff
@@ -0,0 +1,80 @@
...
+          value: forward
+        - dps_val: reverse
+          value: reverse
+secondary_entities:
+  - entity: light
+    dps:
+      - id: 15
+        type: boolean
+        name: switch
+        mapping:
```
make-all, line 57: This mapping is unneccesary

make-all (approved):

custom_components/tuya_local/devices/immax_neo_light_vento.yaml
make-all: Since the things I pointed out are not problems, I will merge this without waiting for modification.  The comments are just for your future reference.

PR #816, opened 2023-06-18
Add Elesion Outdoor Siren

mharrend:
Adding outdoor siren sold by Pearl and controlled by Elesion App. More information about device:

Model Name: NX-4980
Distributor: Pearl
Creator: VisorTech
https://www.pearl.de/a-ZX5035-3110.shtml

Siren is working with this configuration perfectly for myself.

Code has been copied from orion_outdoor_siren.yaml file and id adjusted.


make-all:
There don't appear to be any adjustments compared with the Orion outdoor siren.
I think this can be handled just by adding the product id and name to the Orion siren.

PR #824, opened 2023-06-21
Fixed asyncio.sleep() not being awaited

mild-delirium:
Those yellow warnings in HA logs are annoying.


make-all:
Thanks, I don't actually use polling,so I had not encountered that error.

Issue #698, opened 2023-05-09
Help with support for Smart Bird Feeder Camera


szupi-ipuzs:
Ok, I'm going to experiment a bit with creating a yaml that will make tuya-local enable using my Birdfeeder. Even without video feed, it will be very useful, so I will try to decode as many entities as possible.
I think I've got all the info - the dpids, the actual values and the jsons from iot.tuya. But I really hope you can help me make heads and tails of all of this...
So, the mappings (found from "debugging" the device on iot.tuya cloud back when my birdfeeder was able to work with it):
DpIds:
```
101: Indicator
103: Vision Flip
104: OSD Watermark
106: Motion Sensitivity
108: Night Vision
109: Get Memory Card Capacity
110: Memory Card Status
111: Format Memory Card
114: Schedule
117: Formatted State
126: Battery Level Report Cap.
134: Motion Alert
135: Motion Detection Mode
137: Record Switch
145: Battery
146: Power Supply Mode
147: Low Battery Alarm
149: Device State
150: Record Switch
159: Siren Switch
160: Device Volume Control
168: Motion Area Switch
169: Motion Area
185: Alarm Report
188: Anti Flicker
212: Message
234: Alarm Interval
235: Recording Duration Setting
245: Event Early End Switch
246: Alarm Linkage
```

```
{
  "result": {
    "category": "sp",
    "functions": [
      {
        "code": "basic_indicator",
        "lang_config": {
          "false": "Off",
          "true": "On"
        },
        "name": "Indicator",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "basic_flip",
        "lang_config": {
          "false": "Off",
          "true": "On"
        },
        "name": "Vision Flip",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "basic_osd",
        "lang_config": {
          "false": "Off",
          "true": "On"
        },
        "name": "OSD Watermark",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "motion_sensitivity",
        "lang_config": {
          "0": "Low",
          "1": "Medium",
          "2": "High"
        },
        "name": "Motion Sensitivity",
        "type": "Enum",
        "values": "{\"range\":[\"0\",\"1\",\"2\"]}"
      },
      {
        "code": "basic_nightvision",
        "lang_config": {
          "0": "Auto",
          "1": "Off",
          "2": "On"
        },
        "name": "Night Vision",
        "type": "Enum",
[...]

PR #828, opened 2023-06-21
Support for Pinelake Birdfeeder Camera

szupi-ipuzs:
As mentioned in #698 I'm trying to add support for my tuya-based Birdfeeder.
I've managed to make it work somewhat, but:
- It only works as long as the device is open in the Birdfeeder app. When it's closed, the device immediately stops responding over wifi... Is there any remedy for this?
- Most of the dpids are not reported when not set (hence the "force" param).
- The camera doesn't show anything, most likely because the video feed is cloud-only.
- There are 2 "Record" switches actually reported on tuya iot (and 2 separate dpids), not sure yet which one is for what.

Care to iron this out?

### DIFF
--- a/custom_components/tuya_local/devices/pinelake_birdfeeder.yaml
+++ b/custom_components/tuya_local/devices/pinelake_birdfeeder.yaml
@@ -1,0 +1,157 @@
+name: Smart Bird Feeder Camera
+products:
+  - id: bk9uhkngejjo8wup
+    name: Pinelake BF002
+primary_entity:
+  entity: camera
+  dps:
+    - id: 150
+      name: record
+      type: boolean
+      optional: true
+      force: true
+secondary_entities:
+  - entity: switch
+    name: Siren
+    dps:
+      - id: 159
+        type: boolean
+        name: switch
+        optional: true
+        force: true
+  - entity: switch
+    name: Light indicator
+    category: config
+    icon: "mdi:led-on"
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+        optional: true
+        force: true
+  - entity: sensor
+    name: Battery level
+    class: battery
+    dps:
+      - id: 145
+        type: integer
+        name: sensor
+        unit: "%"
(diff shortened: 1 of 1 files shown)


make-all:
Is it battery powered?  Maybe it is being woken by a cloud message, and fully active during that period only, otherwise only periodically polling the cloud for messages to save battery?

At least one of the dps must be non-optional for detection to work.

Video streaming requires a cloud connection to get the URL and session token to connect directly to the camera (I don't know why they didn't provide this in the encrypted local protocol as well).  Some cameras will provide a still image locally, but it depends on the camera.

To iron it out for you, I_'m going to need some logs of the dps returned by the device.

custom_components/tuya_local/devices/pinelake_birdfeeder.yaml
```diff
@@ -0,0 +1,157 @@
...
+        mapping:
+          - dps_val: "0"
+            value: "Auto"
+          - dps_val: "1"
+            value: "Off"
+          - dps_val: "2"
+            value: "On"
+        optional: true
+        force: true
+  - entity: switch
```
make-all, line 106: Motion detection can move inside the camera entity I think.

custom_components/tuya_local/devices/pinelake_birdfeeder.yaml
```diff
@@ -0,0 +1,157 @@
...
+primary_entity:
+  entity: camera
+  dps:
+    - id: 150
+      name: record
+      type: boolean
+      optional: true
+      force: true
+secondary_entities:
+  - entity: switch
```
make-all, line 14: I'm not sure a siren is useful if it only contains a switch (no tone/volume_level/duration settings).  It might as well be a switch or a button.

make-all:
The logs needed seem to be available on #698, I will try to tidy this up and make it workable, though the need to open the app to activate it is likely to remain, as the device is battery powered, so will rely on the cloud for battery management when idle.

PR #829, opened 2023-06-22
Adding support for Tontine electric blanket using the 039-WIFI control

prunes-git:
I did consider using the climate entity but it didn't seem suitable, however this configuration works.

The only annoyance is sometimes it's slow to update the countdown when the value of the timer is changed and vice versa.
The controller always resets to 1hr whenever turned on or off.

### DIFF
--- a/custom_components/tuya_local/devices/tontine_039-WIFI_blanket.yaml
+++ b/custom_components/tuya_local/devices/tontine_039-WIFI_blanket.yaml
@@ -40,6 +40,7 @@ secondary_entities:
           max: 10
   - entity: sensor
     name: Countdown
+    class: duration
     icon: "mdi:timer"
     dps:
       - id: 10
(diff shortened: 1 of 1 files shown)


make-all:
Yes, the coarse nature of the temperature setting and the lack of a current temperature sensor make this not a good fit for the climate platform. I think this is the correct choice.

custom_components/tuya_local/devices/tontine_039-WIFI_blanket.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - id: 9
+        type: integer
+        name: value
+        unit: h
+        range:
+          min: 1
+          max: 10
+  - entity: sensor
+    name: Countdown
+    icon: "mdi:timer"
```
make-all, line 43: Add `class: duration`, it may give a better display.

prunes-git: If you want it displayed that way that's fine, although it shows seconds which is always 00, and personally I feel it looks better as minutes, which is how the Tuya app displays it.

prunes-git: You are probably right looking at them side by side, I think more people would be happier with duration. I just updated it.

PR #830, opened 2023-06-22
Fixed brightness scaling issue for CCT lightbulb + added 2 product details

bfg100k:
By chance I noticed that setting full brightness in HA for my 2 tuya based CCT lightbulbs from different brands is showing up as 25% brightness in the tuya app. More importantly, when I turned up the brightness in tuya app, the bulbs did get brighter accordingly. I compared the model definition returned by tuya's API with the yaml file and found them to be the same hence I figured this is a scaling issue instead. With this change, both HA and tuya app now shows the same brightness level.

### DIFF
--- a/custom_components/tuya_local/devices/cct_lightbulb.yaml
+++ b/custom_components/tuya_local/devices/cct_lightbulb.yaml
@@ -4,6 +4,10 @@ products:
     name: A60 lightbulb
   - id: znfxaaxc89zz6m5a
     name: A65 lightbulb
+  - id: c59bbbt5xsm6fkkk
+    name: Laser 10W Smart White Bulb E27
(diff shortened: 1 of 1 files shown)


make-all:
Thanks, but it is not quite right, the scale factor needs to be 3.92 to fit the 0-255 range.  And there were other recently added lights that had the same issue, so I created a different patch for the issue.

For the scale, the HA UI does not take this into account, is it actually needed?  It seems a bit strange that they would make the range 10-1000 and then only use steps of 100.

bfg100k:
ah ok. I thought it was 4 as when HA is set to 100%, tuya app shows 25%. I
just checked again and there was a 1% difference between HA and tuya app
for some values in between. Changing it to 3.92 seems to fix it. Cheers

Issue #834, opened 2023-06-23
Request support for Aquatech Rapid/X6 Heat Pump Water Heater


jrotenstein:
This device is VERY SIMILAR to the [Hydrotherm DYNAMIC/X8](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/hydrotherm_dynamic_x8_water_heater.yaml) device. It has the same manufacturer and the same functionality (eg same operating modes).

For example, the logic used in DPS 1 for Hydrothem can be reused on Aquatech.

# Log Message

```
{
    "updated_at": 1687501244.629134,
    "1": true,
    "2": "Stand",
    "4": 60,
    "7": false,
    "10": 0,
    "15": 0,
    "16": 60,
    "18": 0,
    "21": 60,
    "22": 0,
    "30": false,
    "101": 0,
    "102": 0,
    "103": 0,
    "104": 0
}
```

# Information about DPS mappings

I added translations in the `name` fields:

```
{
    "modelId": "00000459ez",
    "services": [
        {
            "actions": [],
            "code": "",
            "description": "",
            "events": [],
            "name": "默认服务 (Default service)",
            "properties": [
                {
                    "abilityId": 1,
                    "accessMode": "rw",
                    "code": "switch",
                    "description": "",
                    "extensions": {
                        "iconName": "icon-dp_power2",
                        "attribute": "1152"
                    },
                    "name": "开关 (Switch)",
                    "typeSpec": {
                        "type": "bool",
                        "typeDefaultValue": false
                    }
                },
                {
                    "abilityId": 2,
                    "accessMode": "rw",
                    "code": "mode",
                    "description": "",
                    "extensions": {
                        "iconName": "icon-dp_mode",
                        "attribute": "1152"
                    },
                    "name": "模式 (Model)",
                    "typeSpec": {
                        "range": [
                            "Stand",
                            "ECO",
                            "HYB",
                            "HYB1",
                            "Ele"
                        ],
                        "type": "enum",
                        "typeDefaultValue": "Stand"
                    }
                },
                {
                    "abilityId": 4,
                    "accessMode": "rw",
                    "code": "temp_set",
                    "description": "",
                    "extensions": {
                        "iconName": "icon-set",
                        "attribute": "1152"
                    },
                    "name": "温度设置 (temperature setting)",
                    "typeSpec": {
                        "max": 75,
                        "min": 15,
                        "scale": 0,
                        "step": 1,
                        "type": "value",
[...]

PR #836, opened 2023-06-24
Create aquatech_x6_water_heater.yaml

jrotenstein:
This matches issue #834 (Request support for Aquatech Rapid/X6 Heat Pump Water Heater).

It is correctly retrieving the temperatures, but I haven't figured out how to get 'defrost' as an attribute in HA. I presume it would need a secondary entity?

I have also excluded any DPs that are just returning zero for me. (See examples in the Issue showing values when the Heat Pump is ON and OFF.)

### DIFF
--- a/custom_components/tuya_local/devices/aquatech_x6_water_heater.yaml
+++ b/custom_components/tuya_local/devices/aquatech_x6_water_heater.yaml
@@ -1,0 +1,68 @@
+name: Aquatech RAPID/X6
+primary_entity:
+  entity: water_heater
+  dps:
+    - id: 1
+      type: boolean
+      name: operation_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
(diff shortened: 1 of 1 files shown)


make-all:
I'm not sure what functionality you need for "defrost".  If it is a switch for an anti-frost vacation mode, then renaming it to "away_mode" is probably the correct thing.  If it is just an indicator then it should be visible as an attribute as-is, or to make it more prominent, you can make it a binary_sensor of its own (`class: cold` probably fits).

jrotenstein:
Hi Jason!

I'm okay leaving it out. While I have seen my system go into Defrost mode, it's rather rare. It happens when the unit has been running in cold temperature.

make-all:
An entities card can show the state of an attribute, you just need to edit the yaml to add `type: attribute` and `attribute: defrosting`  to it.  With a single entity card, I think you can select attributes in the UI.

PR #839, opened 2023-06-25
Added support for Windcalm Fans

hanzwo:
Added support for the Create Windcalm Fan with and without led light.

### DIFF
--- a/custom_components/tuya_local/devices/windcalm_fan.yaml
+++ b/custom_components/tuya_local/devices/windcalm_fan.yaml
@@ -8,19 +8,11 @@ primary_entity:
     - id: 62
       type: integer
       name: speed
+      range:
+        min: 0
+        max: 6
       mapping:
-        - dps_val: 1
-          value: 20
-        - dps_val: 2
-          value: 36
-        - dps_val: 3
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/windcalm_fan.yaml
```diff
@@ -0,0 +1,33 @@
+name: Windcalm Ceiling Fan
+primary_entity:
+  entity: fan
+  dps:
+    - id: 60
+      type: boolean
+      name: switch
+    - id: 62
+      type: integer
```
make-all, line 9: Is it really an integer, or is it a string?
If an integer, it might be better to map using a range and scale instead of listing all the options.  The scale is a divisor, so since the HA target is 0 - 100, the scale for 6 speeds should be 0.06.

hanzwo: I'll look into it.

hanzwo: A bit late, but I was enjoying a little holiday.

I updated the speed mappings. Works like a charm.

PR #843, opened 2023-06-26
Woods Milan AC 7 + Norwegian translation

bskjon:
Device: Woods Milan AC 7
Translation: Norsk Bokmål

### DIFF
--- a/custom_components/tuya_local/devices/woods_milan_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/woods_milan_airconditioner.yaml
@@ -1,0 +1,69 @@
+name: Woods Milan Air Conditioner
+products:
+  - id: i5rv3ykmtfcagw1t
+    name: Wood's Milan 7K
+primary_entity:
+  entity: climate
+  dps:
+    - id: 1
+      name: hvac_mode
+      type: boolean
+      persists: true
+      mapping:
+        - dps_val: false
+          value: "off"
+          icon: "mdi:hvac-off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: auto
+              icon: mdi:thermostat-auto
+              value: auto
+            - dps_val: cold
+              icon: "mdi:snowflake"
+              value: cool
+            - dps_val: wind
+              icon: "mdi:fan"
+              value: fan_only
+            - dps_val: wet
+              icon: "mdi:water"
+              value: dry
+            - dps_val: hot
+              value: heat
+              icon: "mdi:heat-wave"
+    - id: 2
+      name: temperature
+      type: integer
+      range:
+        min: 15
+        max: 31
+      persists: true
+    - id: 3
+      name: current_temperature
+      type: integer
+      readonly: true
+    - id: 4
+      name: mode
+      type: string
+      hidden: false
+    - id: 5
+      name: fan_mode
+      type: string
+      mapping:
+        - dps_val: "0"
+          value: low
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/woods_milan_airconditioner.yaml
```diff
@@ -0,0 +1,69 @@
...
+      mapping:
+        - dps_val: false
+          value: "off"
+          icon: "mdi:hvac-off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: auto
+              icon: mdi:thermostat-auto
+              value: auto
```
make-all, line 21: Normally "auto" maps to "heat_cool" in HA terminology - the mode that automatically switches between heat and cool according to whether abbient temperature is above or below the target temperature.  In HA terminology "auto" is reserved for a mode that follows a schedule or monitors activity, or something else smarter than the standard A/C modes..

custom_components/tuya_local/devices/woods_milan_airconditioner.yaml
```diff
@@ -0,0 +1,69 @@
...
+            - dps_val: hot
+              value: heat
+              icon: "mdi:heat-wave"
+    - id: 2
+      name: temperature
+      type: integer
+      range:
+        min: 15
+        max: 31
+      persists: true
```
make-all, line 40: persist is singular, but the default is true so it is not needed here.  It is only required when you want dps to report as None when the device stops sending them.

custom_components/tuya_local/devices/woods_milan_airconditioner.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - id: 3
+      name: current_temperature
+      type: integer
+      readonly: true
+    - id: 4
+      name: mode
+      type: string
+      hidden: false
```
make-all, line 48: False is the default - I guess you had true here and changed it for debugging?  If you are not sure you have captured all the possible values, it is OK to leave visible for now.

custom_components/tuya_local/devices/woods_milan_airconditioner.yaml
```diff
@@ -0,0 +1,69 @@
...
+      name: fan_mode
+      type: string
+      mapping:
+        - dps_val: "0"
+          value: low
+        - dps_val: "1"
+          value: high
+      persists: true
+secondary_entities:
+  - entity: switch
```
make-all, line 59: This can be swing_mode on the climate entity.  Standard values are "off", "vertical", "horizontal" and "both", but I also see "on" in the strings.json file for climate if the swinging is not obviously either "horizontal" and "vertical".

custom_components/tuya_local/devices/woods_milan_airconditioner.yaml
bskjon: I believe this is correct as the device itself has a auto mode. Both in the dp and functionally on the device

make-all: I am talking about the list of modes here https://developers.home-assistant.io/docs/core/entity/climate/#hvac-modes

95% of air conditioners have a mode labelled auto which corresponds to heat_cool in HA.  It is rare to find what HA calls auto, especially in a portable unit.  In the HA UI you will mostly only see icons anyway - the auto icon is a calendar, the heat_cool icon is a loop with arrows (mdi:sync).

PR #844, opened 2023-06-27
Added goldair tower fan config

horse-pants:
Tested with two fans at my home

### DIFF
--- a/custom_components/tuya_local/devices/goldair_platinum_tower_fan_with_wifi.yaml
+++ b/custom_components/tuya_local/devices/goldair_platinum_tower_fan_with_wifi.yaml
@@ -1,0 +1,53 @@
+name: Goldair Platinum Tower Fan with Wifi
+products:
+  - id: hgsp7xfajllkqujw
+primary_entity:
+  entity: fan
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+    - id: 2
+      type: string
+      name: preset_mode
+      mapping:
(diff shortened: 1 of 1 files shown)


horse-pants:
The product name coming out of the tuya APIs is all upper case, so I copied it in the same because I wasn't sure if they had to match. 

If the casing doesn't matter I can make it normal casing

horse-pants:
Ah true, I forgot about the temp sensor.

I'll add this tonight, thanks

custom_components/tuya_local/devices/goldair_platinum_tower_fan_with_wifi.yaml
```diff
@@ -0,0 +1,61 @@
...
+      name: oscillate
+      mapping:
+        - dps_val: "on"
+          value: true
+        - dps_val: "off"
+          value: false
+    - id: 6
+      type: integer
+      name: timer
+    - id: 19
```
make-all, line 42: I would make this a standalone sensor, as temperature is not part of the fan entity standard API.

Under secondary_entities:
```
  - entity: sensor
    class: temperature
    name: Temperature
    dps:
      - id: 19
        type: integer
        name: sensor
        class: measurement
        unit: C
```

horse-pants: Good call, I figured there had to be a way to make them their own sensor

Thanks

PR #851, opened 2023-06-28
Expand Water Timer valve definitions

fcrozat:
Both Diivoo and Johgee water timer valve are identical on their DPS mapping and report.

Some dps should be set to optional, they are not reporting on fresh devices.

There is one remain issue in this PR: 
Fault code and Fault sensors being optional, if not fetched, default value will be set to true and UI will report there is a problem. I don't know how to change that.


fcrozat:
it is probably possible to also merge HCT-611 definition but I don't have any way to test if it would cause any regression on HCT-611.

make-all:
The missing fault code can be dealt with by adding a mapping from dps_val: null to value: false

PR #861, opened 2023-07-02
New devices and support for manufacturer / model specification

timlaing:
Added support for:

Diivoo Dual Smart Water Timer
Generic Tuya Smart Water Timer
Specification of Manufacturer & Model in yaml configuration

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -4,7 +4,7 @@
     {
       "label": "Install requirements",
       "type": "shell",
-      "command": "pip3 install -r requirements.txt;pip3 install -r requirements-dev.txt",
+      "command": "pip3 install -r requirements-dev.txt",
       "problemMatcher": []
     },
     {


custom_components/tuya_local/devices/water_timer_sop10.yaml
make-all, line 1: Check whether the recently added ble_water_valve.yaml works with this device.
Some of the entity types are different - number/select vs sensor, and switch vs binary sensor, but the types match 100% and the device type is the same, with descriptions of each function for the dps seeming to match.

make-all:
I think the manufacturer and model support is premature, as we need to be able to recognise devices based on the product id first, otherwise the chances of getting the wrong details is high for devices that are generic with multiple products sharing the same config.  The details in the config file itself are OK, as that is how I already planned to extend the config once product id is implemented, but the code to use it based on first match is incorrect.

A lot of these changes are just unnecessary whitespace changes and reordering of lines, probably by some automated tool.  While it might improve the compliance to HA coding standards, it distracts from the actual changes, so should be split into a separate PR.  The translation changes all seem to be left over from a change that was reverted after my first round of comments.

I think the changes to add new devices and improve existing ones would best be split from any remaining python code changes that you want to submit, and do them as separate PRs.

timlaing:
Many thanks for your comments, I'll split this into separate PR's.

PR #882, opened 2023-07-11
Diivoo Dual water timer & updated ble_water_valve

timlaing:
- Added Diivoo Dual Smart Water Timer WT-05
- Updated Generic Water Valve to support SOP10


timlaing:
@make-all - please find first split out PR for new/updated devices

PR #883, opened 2023-07-12
Device Aubess Irrigation System  RainPoint TTP106W

spuljko:
Added support for Aubess Irrigation System  RainPoint TTP106W

### DIFF
--- a/custom_components/tuya_local/devices/aubess_rainpoint_irrigation_system.yaml
+++ b/custom_components/tuya_local/devices/aubess_rainpoint_irrigation_system.yaml
@@ -1,0 +1,140 @@
+name: Aubess Irrigation System 
+products:
+  - id: 2ak7r2culspkc7hx
+    model: RainPoint TTP106W
+    manufacturer: Aubess
+primary_entity:
+  entity: switch
+  icon: "mdi:pipe-valve"
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+
+secondary_entities:
+  - entity: sensor
+    name: Work State
+    class: enum
+    category: diagnostic
+    dps:
+      - id: 106
+        name: sensor
+        type: string
+        icon_priority: 1
+        mapping:
+          - dps_val: "0"
+            value: unknown
+          - dps_val: "1"
+            value: idle
+          - dps_val: "2"
+            value: auto_run
+          - dps_val: "3"
+            value: manual_run
+
+  - entity: switch
+    name: Auto Run
+    icon: "mdi:timer"
+    category: config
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/aubess_rainpoint_irrigation_system.yaml
```diff
@@ -0,0 +1,140 @@
+name: Aubess Irrigation System
```
make-all, line 1: Home Assistant naming convention is to only capitalise the first word in a name, unless it is a trademark or abbreviation that needs to be capitalised.  I have also started to prefer not using brand/model names in the name, as there are often other brands of the same device sold in other markets or by different retailers in the same market. So just "Irrigation system" may be better here, and the model and manufacturer will be picked up later  from the products entry once discovery is implemented properly.

custom_components/tuya_local/devices/aubess_rainpoint_irrigation_system.yaml
```diff
@@ -0,0 +1,140 @@
...
+  entity: switch
+  icon: "mdi:pipe-valve"
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+
+secondary_entities:
+  - entity: sensor
+    name: Work State
```
make-all, line 16: The same comment about naming convention applies here, but also "Work state" is just following the Tuya protocol naming, normally these will be labelled as something else in the app.  Maybe just "Status"?

The values below should be how they should appear in the UI (human readable) so generally capitalised and with spaces rather than underscores, since there is currently no translation done for enum sensor values.

custom_components/tuya_local/devices/aubess_rainpoint_irrigation_system.yaml
```diff
@@ -0,0 +1,140 @@
...
+    icon: "mdi:timer"
+    category: config
+    dps:
+      - id: 109
+        name: switch
+        type: boolean
+        optional: true
+
+  - entity: number
+    name: How long
```
make-all, line 45: Duration?

custom_components/tuya_local/devices/aubess_rainpoint_irrigation_system.yaml
```diff
@@ -0,0 +1,140 @@
...
+        unit: s
+        range:
+          min: 6
+          max: 1440
+        mapping:
+        - scale: 1
+          step: 1
+
+  - entity: number
+    name: How often
```
make-all, line 62: Interval? Frequency?

PR #885, opened 2023-07-12
Added devcontainer support for development ease

timlaing:
VS Code devcontainer support to allow to easier development / updates / testing

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -4,7 +4,7 @@
     {
       "label": "Install requirements",
       "type": "shell",
-      "command": "pip3 install -r requirements.txt;pip3 install -r requirements-dev.txt",
+      "command": "pip3 install -r requirements-dev.txt",
       "problemMatcher": []
     },
     {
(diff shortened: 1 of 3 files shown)


.vscode/tasks.json
```diff
@@ -4,31 +4,19 @@
     {
       "label": "Install requirements",
       "type": "shell",
-      "command": "pip3 install -r requirements-first.txt;pip3 install -r requirements-dev.txt",
+      "command": "pip3 install -r requirements.txt;pip3 install -r requirements-dev.txt",
```
make-all, line 7: I think just requirements-dev.txt is enough, as the requirements in requirements.txt are already included in there.  Years ago there used to be a version mismatch between the pycryptodome library required for tinytuya and one of the other dependencies, so we used to install that first to ensure the correct version was installed, but the versions are in sync now which is why requirements-first.txt was removed to simplify the process.

timlaing: Thanks, I'll update this when I back from a work trip.

PR #889, opened 2023-07-13
feat: added support for the Fantasy 3 HDMI Syncbox by Lytmi

jafar-atili:
Full control of the Lytmi Fantasy 3 HDMI Syncbox

Screenshot 2023-07-14 at 17 08 38

### DIFF
--- a/custom_components/tuya_local/devices/lytmi_fantasy3_hdmi_syncbox.yaml
+++ b/custom_components/tuya_local/devices/lytmi_fantasy3_hdmi_syncbox.yaml
@@ -50,14 +50,26 @@ primary_entity:
     - id: 107
       type: boolean
       name: unknown_107
-    - id: 108
-      type: integer
-      name: unknown_108
     - id: 109
       type: boolean
       name: unknown_109
 
 secondary_entities:
+  - entity: number
+    category: config
+    name: Degree of diffusion
+    type: integer
+    dps:
+      - id: 108
+        name: value
+        type: integer
+        precision: 0
+        range:
+          min: 10
+          max: 1000
+        mapping:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/lytmi_fantasy3_hdmi_syncbox.yaml
```diff
@@ -93,6 +95,7 @@ secondary_entities:
           max: 1000
         mapping:
           - scale: 10
+            step: 1
```
make-all, line 98: This one is not needed, the step defaults to 1.  If your intention was to make the UI step 1.0 instead of 0.1 after scaling then you would need to make the step 10.  That would actually fit with the precision of 0.

jafar-atili: You are correct, I fixed that locally and now it behaves as expected

jafar-atili:
Not sure why I get this error in HA upon restarts

This the state currently set

```
This error originated from a custom integration.

Logger: custom_components.tuya_local.helpers.device_config
Source: custom_components/tuya_local/helpers/device_config.py:393
Integration: Tuya Local (documentation, issues) 
First occurred: 12:14:04 (60 occurrences) 
Last logged: 12:24:06

HDMI sync backlight sent invalid hex 'Sync - Game' for option
```

make-all:
> HDMI sync backlight sent invalid hex 'Sync - Game' for option

It looks like the hex decoding is being done after the mapping.  This is probably a bug, but since the hex decoding isn't necessary for this select to work, the type: hex could just be changed to type: string.

PR #892, opened 2023-07-15
Add support for Sencor MT7048C Air Conditioner

WallyCZ:
https://www.sencor.com/mobile-air-conditioner/sac-mt7048c
Maybe others similar Sencor A/C's can work the same way, but I can't test them.

### DIFF
--- a/custom_components/tuya_local/devices/sencor_air_conditioner.yaml
+++ b/custom_components/tuya_local/devices/sencor_air_conditioner.yaml
@@ -1,7 +1,8 @@
-name: Sencor Air Conditioner
+name: Air conditioner
 products:
   - id: ypgwqvhjrl5chq6h
     model: SAC MT7048C Air Conditioner
+    manufacturer: Sencor
 primary_entity:
   entity: climate
   translation_key: aircon_extra
@@ -52,9 +53,6 @@ primary_entity:
           conditions:
             - dps_val: true
               invalid: true
-    - id: 19
-      name: temperature_unit
-      type: string
     - id: 22
       type: bitfield
       name: fault_code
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/sencor_air_conditioner.yaml
```diff
@@ -0,0 +1,85 @@
...
+      mapping:
+        - dps_val: low
+          value: low
+        - dps_val: high
+          value: high
+          constraint: preset_mode
+          conditions:
+            - dps_val: true
+              invalid: true
+    - id: 19
```
make-all, line 55: You should have either a temperature_unit dp, or a fixed unit attribute on the temperature, not both.  The temperature unit dp must return either "C" or "F" to Home Assistant (case sensitive), it the device is using lower case, this requires a mapping.

custom_components/tuya_local/devices/sencor_air_conditioner.yaml
```diff
@@ -0,0 +1,85 @@
...
+      type: boolean
+      mapping:
+        - dps_val: true
+          value: sleep
+        - dps_val: false
+          value: comfort
+    - id: 103
+      name: type
+      type: string
+      hidden: true
```
make-all, line 72: If you make things hidden, it is generally because they are exposed indirectly by a conditional mapping.  This one does not seem to be used anywhere, so it may be better to leave it visible in case anyone has a use for it.

custom_components/tuya_local/devices/sencor_air_conditioner.yaml
```diff
@@ -0,0 +1,85 @@
+name: Sencor Air Conditioner
```
make-all, line 1: HA naming convention is to only capitalise the first word of a name.

PR #897, opened 2023-07-16
Add support for GauTone PG-103 alarm

luyzfernando08:
image

https://pt.aliexpress.com/item/1005001700651666.html

### DIFF
--- a/custom_components/tuya_local/devices/gautone_pg103_alarm.yaml
+++ b/custom_components/tuya_local/devices/gautone_pg103_alarm.yaml
@@ -1,7 +1,7 @@
-name: GauTone PG-103 alarm
+name: Alarm security system
 products:
   - id: eb26b0818666ecd284o4qe
-    name: GauTone PG-103 alarm security system
+    name: GauTone PG-103
 primary_entity:
   entity: alarm_control_panel
   dps:
@@ -52,7 +52,7 @@ secondary_entities:
           min: 0
           max: 1000
   - entity: number
-    name: Siren Duration
+    name: Siren duration
     category: config
     icon: "mdi:timer"
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/gautone_pg103_alarm.yaml
```diff
@@ -0,0 +1,100 @@
...
+    icon: "mdi:timer"
+    dps:
+      - id: 28
+        type: integer
+        name: value
+        unit: sec
+        range:
+          min: 0
+          max: 1000
+  - entity: number
```
make-all, line 54: This can be made part of a siren entity. But if there is no way to trigger the siren on demand, maybe that is not useful.

custom_components/tuya_local/devices/gautone_pg103_alarm.yaml
```diff
@@ -0,0 +1,100 @@
...
+        name: switch
+  - entity: switch
+    name: Mode App Sound
+    category: config
+    dps:
+      - id: 10
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Alarm Message propel
```
make-all, line 95: I think this is for notifications, which may be a more obvious name than "propel".

custom_components/tuya_local/devices/gautone_pg103_alarm.yaml
```diff
@@ -0,0 +1,100 @@
...
+    dps:
+      - id: 3
+        type: integer
+        name: value
+        unit: min
+        range:
+          min: 0
+          max: 1000
+  - entity: switch
+    name: Alarm Sound
```
make-all, line 67: HA naming convention is to only capitalise the first word of names, unless the following words are abbreviations or trademarks etc that require capitalising.

PR #899, opened 2023-07-16
Ultenic K10 fixes

blakadder:
Fixes for Ultenic K10 air fryer config

### DIFF
--- a/custom_components/tuya_local/devices/ultenic_air_fryer.yaml
+++ b/custom_components/tuya_local/devices/ultenic_air_fryer.yaml
@@ -13,9 +13,6 @@ primary_entity:
     - id: 12
       type: bitfield
       name: error_code
-    - id: 104
-      type: string
-      name: cooking_report
 secondary_entities:
   - entity: switch
     name: Cooking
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/ultenic_air_fryer.yaml
```diff
@@ -181,30 +184,47 @@ secondary_entities:
...
+        type: boolean
+  - entity: binary_sensor
+    name: Shake
+    icon: "mdi:car-brake-alert"
+    dps:
+      - id: 104
+        name: sensor
+        type: string
+        mapping:
+          - dps_val: 0
```
make-all, line 221: numeric values of string type need double quotes.

custom_components/tuya_local/devices/ultenic_air_fryer.yaml
```diff
@@ -86,25 +85,25 @@ secondary_entities:
         name: sensor
         mapping:
           - dps_val: "0"
-            value: Standby
+            value: standby
```
make-all, line 88: These were capitalised as there is no translation for sensor values, so they will appear in the UI as-is.

custom_components/tuya_local/devices/ultenic_air_fryer.yaml
```diff
@@ -122,19 +121,18 @@ secondary_entities:
...
         mapping:
-          - dps_val: "0"
+          - dps_val: c
             value: C
-          - dps_val: "1"
+          - dps_val: f
             value: F
   - entity: sensor
-    name: Temperature
+    name: Internal Temperature
```
make-all, line 129: HA naming convention is that only the first word in a name gets capitalised, so if changing this the t also needs changing.

blakadder:
thanks for the review and comments, those are helpful in making future PRs better

PR #904, opened 2023-07-17
Create makegood_double_power_point_with_energy.yaml

iotinkognito:
Added configuration for Makegood Double Power Points MG-AUWF01 sold as Kogan, Oz Smart Things, Aus Electronics Direct, Witron, Qubino. 

I got mine for Oz Smart Things who have got them certified for Australia with RCM. 
https://equipment.erac.gov.au/Registration/RSDetails.aspx?atn=public&uid=a6738ea3-1ebb-4650-a4bd-67bf93ee41a4&bk=false

### DIFF
--- a/custom_components/tuya_local/devices/makegood_double_power_point_with_energy.yaml
+++ b/custom_components/tuya_local/devices/makegood_double_power_point_with_energy.yaml
@@ -1,4 +1,4 @@
-name: Makegood double power point
+name: Double power point
 primary_entity:
   entity: switch
   class: outlet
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/makegood_double_power_point_with_energy.yaml
```diff
@@ -0,0 +1,84 @@
+name: Makegood double power point
```
make-all, line 1: If they are sold under different brands, remove the brand from here.   If you can find the product ids for the various brands you have access to, then listing them here under a `products` section would help with future plans for discovery.

PR #910, opened 2023-07-18
Inkbird IBS-M1S support

mbakhtin:
Added support for IBS-M1(S) support

Product:
https://inkbird.com/products/wi-fi-gateway-ibs-m1

Related thread:
https://community.home-assistant.io/t/inkbird-ibs-m1-with-ibs-p01/255050

Had to add "endianness" configuration property, there were no way to extract temperature instead.

Integration creates 6 sensors, however the device can handle more, but 6 is the maximum mentioned in the above thread.
It would be great to make it dynamic, but I don't see any options for that at the moment.

Protocol is the following:

Data received as a base64-encoded string. Data consists of multiple (up to 30 or so) 10-bytes fragments, each fragment is a dedicated sensor connected to the gateway. First byte is the sensor type (have seen 3 and 8). It is zero if there is no sensor. Second and third bytes are the temperature in Celsius multiplied by 10. 10th byte is the battery status in percents. Bytes in the middle are probably meaningful for other sensor types (humidity, ambient temp, etc.), but I don't have any of these.

Example: CAwB////////LQD//////////wAA//////////8AAP//////////AAD//////////wAA//////////8AAP//////////AAD//////////wAA//////////8AAP//////////AAD//////////wAA//////////8AAP//////////AAD//////////wAA//////////8AAP//////////AAD//////////wAA//////////8AAP//////////AAD//////////wAA//////////8AAP//////////AAD//////////wAA//////////8AAP//////////AA==

### DIFF
--- a/custom_components/tuya_local/devices/inkbird_ibs_m1s.yaml
+++ b/custom_components/tuya_local/devices/inkbird_ibs_m1s.yaml
@@ -10,11 +10,9 @@ primary_entity:
     - id: 110
       type: integer
       name: dummy
-      persist: false
     - id: 111
       type: boolean
       name: dummy2
-      persist: false
     - id: 114
       optional: true
       type: base64
@@ -53,6 +51,7 @@ secondary_entities:
           - endianness: little
             mask: "FF"
       - id: 114
+        optional: true
         type: base64
         name: sensor
         unit: "%"
@@ -105,6 +104,7 @@ secondary_entities:
           - endianness: little
             mask: "FF00000000000000000000"
       - id: 114
+        optional: true
         type: base64
         name: sensor
         unit: "%"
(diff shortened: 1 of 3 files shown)


custom_components/tuya_local/devices/inkbird_ibs_m1s.yaml
```diff
@@ -0,0 +1,326 @@
...
+    name: IBS-M1S
+primary_entity:
+  entity: sensor
+  name: Temperature probe 1
+  class: temperature
+  dps:
+    - id: 110
+      type: integer
+      name: dummy
+      persist: false
```
make-all, line 13: `persist: false` should only be needed when you want to trigger events off the dp being reported without any change in value since the last report, and should always be paired with `optional: true`.  In this case, it seems unnecessary.

mbakhtin: Ok, removed persist. I can't make them optional since they define how this device is detected, if they are optional then this device type will pop up for every discovered device.

PR #918, opened 2023-07-22
Added Garza Wi-Fi water timer

jamarju:
This is a water timer commonly found in Leroy Merlin shops.

It consists of a hub that doubles as a smart switch and controls the actual water timer via a 433 MHz custom protocol. 

Added basic support:

- Switch on/off
- Manual irrigation timer
- Battery report (0/50/100%)
- Current working mode (Off/Manual/Auto)
- Current alarm status (untested, actually)

Not supported:

- Timer scheduling: just use HA automations instead! this uses a binary protocol that should not be hard to reverse using Tuya cloud's "debug device" feature if anyone cares.

Also included directions to get the `node_id` AKA the "sub device id" using tinytuya as I couldn't find it anywhere in the developer portal.

### DIFF
--- a/custom_components/tuya_local/devices/garza_irrigation_system.yaml
+++ b/custom_components/tuya_local/devices/garza_irrigation_system.yaml
@@ -1,0 +1,87 @@
+name: Garza WiFi garden irrigation system
+product:
+  - id: xnajzurgqo1ugohn
+    model: ITV103W
+primary_entity:
+  entity: switch
+  icon: "mdi:watering-can"
+  name: Switch
+  dps:
+    - id: 108
+      type: boolean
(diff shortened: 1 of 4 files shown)


custom_components/tuya_local/devices/garza_irrigation_system.yaml
```diff
@@ -0,0 +1,87 @@
...
+        unit: L
+  - entity: number
+    name: Timer
+    icon: "mdi:timer"
+    category: config
+    dps:
+      - id: 107
+        type: integer
+        name: value
+        unit: m
```
make-all, line 56: unit should be min for minutes (m is metre)

Issue #867, opened 2023-07-05
Request support for Tuya Smart Life EU WiFi Double Curtain Blind Switch


Levanterman:
# Log Message

```
Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:49
Integration: Tuya Local ([documentation](https://github.com/make-all/tuya-local), [issues](https://github.com/make-all/tuya-local/issues))
First occurred: 19:22:06 (2 occurrences)
Last logged: 19:22:06

Device matches qs_c01_curtain with quality of 44%. DPS: {"updated_at": 1688581326.0668912, "1": "stop", "2": 100, "4": "stop", "5": 100, "8": "forward", "9": "forward", "10": 2, "11": 2, "14": "relay"}
Report this to https://github.com/make-all/tuya-local/issues/

Using https://github.com/blakadder/tuya-uncover

    "2 Way Smart Curtain": {
        "id": "",
        "local_key": "",
        "category": "clkg",
        "uuid": "",
        "product_id": "zqypumy8ot4clzyi",
        "dps": {
            "1": "stop",
            "10": 2,
            "11": 2,
            "14": "relay",
            "2": 0,
            "3": "start",
            "4": "stop",
            "5": 0,
            "6": "start",
            "8": "back",
            "9": "forward"
        }
    }
```

# Information about DPS mappings

```
{
  "result": {
[...]

PR #920, opened 2023-07-23
Create loratap_wifi_curtain_switch_double.yaml

Levanterman:
I am still not sure why it does show the state unknown when  the state is fully it is Open or Close

### DIFF
--- a/custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
+++ b/custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
@@ -47,7 +47,7 @@ secondary_entities:
           max: 100
           min: 0
       - id: 5
-        name: currrent_position
+        name: current_position
         type: integer
   - entity: number
     name: Timer 1
--- a/custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
+++ b/custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
@@ -1,0 +1,175 @@
+name: Loratap Wifi Double Curtain Switch
+products:
+  - id: zqypumy8ot4clzyi
+primary_entity:
+  entity: cover
+  name: Curtain
+  class: blind
+  dps:
+    - id: 1
+      name: control
+      type: string
+      mapping:
+        - dps_val: open
+          value: open
+        - dps_val: close
+          value: close
+        - dps_val: stop
+          value: stop
+    - id: 2
+      name: position
+      type: integer
+      unit: "%"
+      range:
+        max: 100
+        min: 0
+      mapping:
+        - step: 1
+secondary_entities:
+  - entity: cover
+    name: Curtain 2
+    dps:
+      - id: 4
+        name: control
+        type: string
+        mapping:
+          - dps_val: open
+            value: open
+          - dps_val: close
+            value: close
+          - dps_val: stop
+            value: stop
+      - id: 5
+        name: position
+        type: integer
+        unit: "%"
+        range:
+          max: 100
+          min: 0
+        mapping:
+          - step: 1
+  - entity: number
+    name: Position 1
+    category: config
+    icon: "mdi:blinds-horizontal-closed"
+    dps:
+      - id: 2
+        name: value
+        type: integer
+        unit: "%"
+        range:
+          max: 100
+          min: 0
+        mapping:
+          - scale: 0
+          - step: 1
+  - entity: number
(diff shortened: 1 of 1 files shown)


make-all:
As noted on #867

If the position reliably indicates the current position of the blind, including when it is opened/closed by the control dp, then you can duplicate that dp as current_position as is done in the benexmart_blind_motor config. Due to a huge variety of Tuya curtain/blind controllers with different behaviours, the position dp is not considered reliable as an indicator of the true position of the cover.

custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
```diff
@@ -0,0 +1,175 @@
...
+    dps:
+      - id: 2
+        name: value
+        type: integer
+        unit: "%"
+        range:
+          max: 100
+          min: 0
+        mapping:
+          - scale: 0
```
make-all, line 64: This mapping is incorrect.  It contains two separate entries without a dps_val, but there can only be one default, and a scale of 0 will result in a division by zero error (tuya-local does not define scale in the same way as iot.tuya.com).   But since the whole entity is not needed, it can all be removed.

custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
```diff
@@ -0,0 +1,175 @@
...
+          - step: 1
+  - entity: number
+    name: timer
```
make-all, line 82: First word of name should be capitalised, according to HA naming conventions.

custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
```diff
@@ -0,0 +1,175 @@
...
+    category: config
+    icon: "mdi:timer"
+    dps:
+      - id: 10
+        name: value
+        type: integer
+        range:
+          max: 120
+          min: 2
+        mapping:
```
make-all, line 92: There should be a unit here (min, or h?)  Or if the unit is s, this is probably not a simple timer, but the setting for how long the curtains take to open/close fully, which implies a different name for the entity.

And remove the mapping, because it is wrong as stated above, and should work OK with the defaults of no scaling and step of 1.

custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
```diff
@@ -0,0 +1,175 @@
...
+          - scale: 0
+          - step: 1
+  - entity: number
```
make-all, line 95: Same comments as for timer above

custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
```diff
@@ -0,0 +1,175 @@
...
+    name: Light Mode
+    category: config
+    icon: "mdi:toggle-switch"
+    dps:
+      - id: 14
+        name: option
+        type: string
+        mapping:
+          - dps_val: "relay"
+            value: Mode 1
```
make-all, line 145: Mode 1, Mode 2 is not very descriptive.  Suggest you look at similar devices for the mappings used there.  "relay" normally means the light follows the switch position, "pos" means always on as a locator for the switch when it is dark, and "none" means always off.  But I'm not entirely sure how "relay" maps to curtain switches.

Levanterman:
Thanks for review :-) I'll make the ammendments .

custom_components/tuya_local/devices/loratap_wifi_curtain_switch_double.yaml
Levanterman: I just copied the text from the app, they call it Mode1 and Mode2, but even when i change them in the App, the do change in tuya-local but i see no behaviour change on the switch itself. Not entire sure what they do. I wondered if it just disables the relay in the switch to stop it clicking (ie if you're not using the physical electrical connections)

make-all: They may have taken a standard wall-switch template from Tuya, and left off the LEDs for button backlights. Normally in a typical Tuya wall switch with touch-sensitive buttons, there will be an LED backlit ring around each button, and this controls whether this is always lit, never lit, or lights up when the switch is on.  If it does nothing, just add it as an attribute on the corresponding cover entity, as it isn't useful to have as a separate entity.

PR #928, opened 2023-07-26
Update its_45hd_heatpump.yaml

nicopret1:
Updated initial file with the changes listed in #866.  Also removed Linkage sensor id=106

### DIFF
--- a/custom_components/tuya_local/devices/its_45hd_heatpump.yaml
+++ b/custom_components/tuya_local/devices/its_45hd_heatpump.yaml
@@ -67,7 +67,7 @@ primary_entity:
       hidden: true
     - id: 30
       type: boolean
-      name: Chlorination
+      name: chlorination
       attribute: true
     - id: 32
       type: boolean
@@ -78,27 +78,27 @@ primary_entity:
       name: model_code
     - id: 101
       type: integer
-      name: Current
+      name: current
       attribute: true
     - id: 102
       type: integer
-      name: Voltage
+      name: voltage
       attribute: true
     - id: 103
       type: integer
-      name: Power
+      name: power
       attribute: true
     - id: 104
       type: integer
-      name: Energy
+      name: energy
       attribute: true
     - id: 105
       type: boolean
-      name: Electricity monitor
+      name: electricity_monitor
       attribute: true
     - id: 106
       type: boolean
-      name: Linkage
+      name: linkage
       attribute: true
     - id: 108
       type: boolean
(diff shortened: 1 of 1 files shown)


nicopret1:
Herewith the pull request for the following updates: 

1. Changed current temperature id from 26 to 21. The best temperature sensor to use for the current water temperature is the "Inlet temperature" (ID=21) due to the location of the sensor. ID=26 is the "Ambient temperature"
2. Added Suction temperature as secondary sensor (ID=20)
3. Added Ambient temperature as secondary sensor (ID=26)
4. Renamed "Exhaust temperature" to "Vent temperature" (ID=24)
5. Removed all electricity related sensors for Voltage, Current and Energy (i.e. ID's=18,101, 101, 104, 105,). The ITS heat pump currently do not have sensors for these aspects and all values are therefore=0.
6. Removed Chlorination (ID=30) since this type of heat pump is not a swimming pool heat pump. It is used as a water heater for the house (i.e geyser element alternative).
7. Removed Linkage sensor (ID=106)

make-all:
Please add attributes to the climate entity for the sensors you are removing.  There may be other related models that do have those sensors, so users may still want to access the data.

nicopret1:
Added attributes to the climate entity for the removed sensors in case other related models that do have those sensors.

custom_components/tuya_local/devices/its_45hd_heatpump.yaml
```diff
@@ -58,20 +58,48 @@ primary_entity:
...
+      type: integer
+      name: Power
+      attribute: true
+    - id: 104
+      type: integer
+      name: Energy
+      attribute: true
+    - id: 105
+      type: boolean
+      name: Electricity monitor
```
make-all, line 97: One further note: at attribute (dps) level, the name is not a UI exposed name (Home Assistant automatically capitalises and removes underscores - the opposite of what it does to entity names) so it should have underscores in place of spaces and by convention be all lower case.

make-all: The lower case part of that also applies to the other attributes converted from sensors by the last commit.

nicopret1: Thank you for your patience and educating me. Its still new to me. Will submit update shortly

nicopret1: I noticed that I previously removed "Electricty today" sensor, but forgot to add the attribute to the climate entity. Will add it now and send upadte

PR #935, opened 2023-07-27
Updated support for Tellur WiFi Aroma Diffuser

lageorgem:
Added support for device 58107061e868e7c462da - Tellur WiFi Aroma Diffuser.
It already works great, with one caveat - when setting fan to OFF, or setting spray level when it's already OFF, nothing happens. I hope these changes will help with that.

DP code `1` log:
```
            {
                "devId": "58107061e868e7c462da",
                "eventType": "Report",
                "sourceDetail": "",
                "eventTime": 1690485440832,
                "eventName": "Spray",
                "eventTimeStr": "2023-07-27 19:17:20:832",
                "eventDetail": "ON",
                "requestFrom": "device itself"
            },
```

DP code `2` log:

```
            {
                "devId": "58107061e868e7c462da",
                "eventType": "Report",
                "sourceDetail": "",
                "eventTime": 1690486518944,
                "eventName": "Mode",
                "eventTimeStr": "2023-07-27 19:35:18:944",
                "eventDetail": "Small",
                "requestFrom": "device itself"
            }
```

image

### DIFF
--- a/custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
+++ b/custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
@@ -5,6 +5,8 @@ products:
     name: YYM-805SW
   - id: 4870500398f4abbfbae4
     name: GX
+  - id: 58107061e868e7c462da
+    name: Tellur WiFi Aroma Diffuser
 primary_entity:
   entity: fan
   dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/yym_805SW_aroma_nightlight.yaml
```diff
@@ -5,19 +5,36 @@ products:
...
       name: switch
       type: boolean
       mapping:
-        - dps_val: true
-          icon: mdi:scent
-          value: ON
-        - dps_val: false
-          icon: mdi:scent-off
-          value: OFF
+        - conditions:
```
make-all, line 17: true, on and ON are all the same value in yaml, likewise false, off and OFF.   This change just complicates the config, especially with the added condition that applies to both true and false cases.

lageorgem: Yeah, I was hoping I could address this bug: https://github.com/make-all/tuya-local/issues/936

But then I did more research and I realized the mappings are fine

PR #952, opened 2023-07-31
Added Loginovo TV Sync backlight support

lageorgem:
Added support for the TV synchronization backlight
The model name is not exactly known, I wasn't able to find it anywhere. The [product listing](https://www.aliexpress.com/item/1005004879862763.html) indicates it's produced by Loginovo, but no specific model number is anywhere, including the packaging.

Entities:
* Light switch 
  * on/off switch
  * color picker
  * brightness for all modes
  * display modes selectable as light effects
    * Video mode - responds to HDMI input changes
    * Music mode - responds to sounds
    * Wide film - responds to HDMI input changes, adjusting for wide film formats
    * Static color - changes light to the color picked on this entity
    * Ambient scene - changes light effect to the one chosen in the "Scene select" entity
* Intensity picker - allows you to choose how sensitive the backlight should be to changes
* Scene select - you can choose between 12 light effect presets
* Other diagnostic info - mostly added to increase match %, I did not make some of the setup settings writable, since these wouldn't usually be modified after the original setup.


custom_components/tuya_local/devices/tv_backlights.yaml
```diff
@@ -0,0 +1,223 @@
...
+      type: boolean
+      name: switch
+    - id: 104
+      name: brightness
+      type: integer
+      range:
+        min: 0
+        max: 100
+      mapping:
+        - step: 1
```
make-all, line 20: step: 1 is not needed.

custom_components/tuya_local/devices/tv_backlights.yaml
```diff
@@ -0,0 +1,223 @@
...
+    mode: slider
+    icon: mdi:speedometer-medium
+    dps:
+      - id: 103
+        name: value
+        type: integer
+        range:
+          min: 1
+          max: 3
+        mapping:
```
make-all, line 89: This whole mapping is not needed

custom_components/tuya_local/devices/tv_backlights.yaml
```diff
@@ -0,0 +1,223 @@
...
+          - step: 1
+            scale: 1
+  - entity: select
+    icon: mdi:shimmer
+    name: Scene select
+    format:
```
make-all, line 95: This does not belong here.  It needs to be specified on a specific dp (if at all)

custom_components/tuya_local/devices/tv_backlights.yaml
```diff
@@ -0,0 +1,223 @@
...
+      - name: scene_num
+        bytes: 2
+        range:
+          min: 1
+          max: 12
+    dps:
+      - id: 113
+        name: option
+        type: integer
```
make-all, line 104: In tuya, numeric values for selecting from a fixed list are often actually strings surrounded by quotes.  Although the integration is forgiving about this it is better to match the protocol exactly in case of future changes that require more strictness.

custom_components/tuya_local/devices/tv_backlights.yaml
```diff
@@ -0,0 +1,223 @@
...
+        name: sensor
+        type: integer
+  - entity: sensor
+    category: diagnostic
+    icon: mdi:speedometer-slow
+    name: Frame delay
+    dps:
+      - id: 120
+        name: sensor
+        type: integer
```
make-all, line 223: Text files should always end with a newline

custom_components/tuya_local/devices/tv_backlights.yaml
```diff
@@ -0,0 +1,223 @@
...
+          - dps_val: 11
+            icon: mdi:flower-outline
+            value: Sunflower
+          - dps_val: 12
+            icon: mdi:looks
+            value: Rainbow
+  - entity: sensor
+    category: diagnostic
+    icon: mdi:information-outline
+    name: Video Output
```
make-all, line 145: Home Assistant naming convention is to only capitalise the first word, unless following words are proper names (including trademarks) or abbreviations.

custom_components/tuya_local/devices/tv_backlights.yaml
lageorgem: I am reading this data using `tinytuya.Device`, and it seems the data is returned as an integer

```json
{
	"dps": {
		"101": "color_mode",
		"102": true,
		"103": 1,
		"104": 100,
		"105": 0,
		"108": 64,
		"111": "ff2806",
		"112": 100,
		"113": 1,
		"114": 100,
		"115": false,
		"116": 12,
		"117": 7,
		"118": 150,
		"119": 150,
		"120": 0,
		"122": "ffffff",
		"125": "BRG",
		"126": false,
		"127": "RD200 _HDMI2.0_230318           V08.08.08.3840*2160p 60Hz 02. 02.296.M.SDR1921.1079.0. 38."
	}
}
```

Issue #957, opened 2023-08-01
Request support for MakeGood/Witron 2 gang smart wall switch with power monitoring


jaymunro:
# Log Message

```
Report this to https://github.com/make-all/tuya-local/issues/
Device matches simple_dual_switch_timer with quality of 50%. DPS: {"updated_at": 1690878119.2272758, "1": false, "2": true, "7": 0, "8": 0, "101": 0, "102": 93, "103": 213, "104": 2380}
Device matches None with quality of 0%. DPS: {"updated_at": 1690879302.549306, "104": 2380}
Device matches simple_dual_switch_timer with quality of 50%. DPS: {"updated_at": 1690879374.1637967, "1": false, "2": true, "7": 0, "8": 0, "101": 0, "102": 93, "103": 212, "104": 2380}
Device matches simple_dual_switch_timer with quality of 50%. DPS: {"updated_at": 1690879689.2542176, "1": false, "2": true, "7": 0, "8": 0, "101": 0, "102": 93, "103": 213, "104": 2360}
```

# Information about DPS mappings

```
{
  "result": {
[...]

PR #958, opened 2023-08-03
Generic switch dual energy

jaymunro:
As per #957, file is tested and running on MG-MD04W
Screenshot 2023-08-03 at 5 04 00 PM

### DIFF
--- a/custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
+++ b/custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
@@ -1,4 +1,4 @@
-name: Dual switch with timers and power monitoring
+name: Dual switch
 products:
   - id: XdGCESvXTI5Mvfx2
     name: MakeGood 2 Gang Switch (MG-MD04W)
--- a/custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
+++ b/custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
@@ -1,7 +1,7 @@
-name: Generic dual switch with timers and power monitoring
+name: Dual switch with timers and power monitoring
 products:
   - id: XdGCESvXTI5Mvfx2
-    name: 2 Gang Switch (MG-MD04W)
+    name: MakeGood 2 Gang Switch (MG-MD04W)
 primary_entity:
   entity: switch
   name: Switch 1
--- a/custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
+++ b/custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
@@ -47,7 +47,6 @@ secondary_entities:
             step: 60
         unit: min
   - entity: sensor
-    name: Current
     class: current
     category: diagnostic
     dps:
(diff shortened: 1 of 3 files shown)


make-all:
I don't think this is generic.  Tuya has a template for switches which contains energy, current, power and voltage sensors on dps 17, 18, 19, 20, and the vast majority of switches follow that assignment.  This one seems to use a custom assignment so I think the name can be manufacturer specific, as it is unlikely there are many other switches following a compatible layout.

custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
```diff
@@ -0,0 +1,87 @@
+name: Dual switch with timers and power monitoring
```
make-all, line 1: I think name would be better shortened to just Dual switch.  The entity names will get added to the end of this, so it will become quite long and annoying to use in automations etc where the distinguishing part might scroll off the end of the field in the UI, or if you are manually editing be a lot to type.

custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
```diff
@@ -0,0 +1,87 @@
...
+        name: value
+        range:
+          min: 0
+          max: 86400
+        mapping:
+          - scale: 60
+            step: 60
+        unit: min
+  - entity: sensor
+    name: Current
```
make-all, line 50: Remove the name when it matches the class (same for power and voltage).

This is a new feature in the next release, so you'll need to be using the main branch HEAD for testing.

custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
jaymunro: Sure, perhaps consider updating `simple_dual_switch_timer.yaml` and `simple_dual_switch_timer_v2.yaml`
```
name: Dual switch with timers
products:
...
```

custom_components/tuya_local/devices/makegood_switch_dual_timer-energy.yaml
jaymunro: There is a difference in case `Current` vs. `current`. The name should be in `Title` case. Is this handled?

make-all: If the name is omitted, then Home Assistant will use its standard translations (the translations take care of capitalising the name).

jaymunro: I do love the structure of HA!

jaymunro: I copied those lines from other devices so there will be a lot to update.
While updating, the yaml files could do with a standard structure such as:
<manufacturer>_<type>_<subtype>_<features>.yaml

jaymunro: Maybe you should be marking these as resolved??

make-all: You must have copied the file before Sunday, as the change was made here 927f7c779b

Renaming config files is a lot more complicated, as the names are stored in the config for devices that have been set up.

PR #959, opened 2023-08-03
Update README.md

jaymunro:
Added guide for yaml file naming convention.


make-all:
Sorry, this change does not seem appropriate. There isn't such a strict convention, and your examples don't really follow the stated convention anyway.  I don't think this clarifies, but rather causes confusion that may hold someone back from contributing due to fear their contribution will be rejected because they don't understand the naming convention.

Issue #773, opened 2023-06-07
Request support for VIVAX COOL ACP-12CH35REWI aircon


Rain-ZG:
# Log Message

```
Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:49
Integration: Tuya Local ([documentation](https://github.com/make-all/tuya-local), [issues](https://github.com/make-all/tuya-local/issues))
First occurred: 20:38:36 (2 occurrences)
Last logged: 20:38:36

    Device matches mirabella_genio_usb with quality of 10%. DPS: {"updated_at": 1686163115.8559422, "1": false, "2": 23, "3": 49, "4": "cold", "5": "auto", "8": false, "12": false, "13": true, "18": 0, "19": "c", "102": 0, "105": "0", "106": false, "107": false, "108": false, "111": false, "113": "cancel", "114": false, "116": 0, "117": "ud_00", "118": "lr_00"}
    Report this to https://github.com/make-all/tuya-local/issues/
```

# Information about DPS mappings

```
{
  "result": {
    "category": "kt",
    "functions": [
      {
        "code": "switch",
        "dp_id": 1,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "dp_id": 2,
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":16,\"max\":30,\"scale\":0,\"step\":1}"
      },
      {
        "code": "mode",
        "dp_id": 4,
        "type": "Enum",
        "values": "{\"range\":[\"auto\",\"cold\",\"wet\",\"hot\",\"wind\"]}"
      },
      {
        "code": "fan_speed_enum",
        "dp_id": 5,
        "type": "Enum",
        "values": "{\"range\":[\"high\",\"low\",\"auto\",\"strong\"]}"
      },
      {
        "code": "mode_eco",
        "dp_id": 8,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "heat",
        "dp_id": 12,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "light",
        "dp_id": 13,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "beep",
        "dp_id": 16,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_unit_convert",
        "dp_id": 19,
        "type": "Enum",
        "values": "{\"range\":[\"c\",\"f\"]}"
      },
[...]

PR #964, opened 2023-08-04
add Thermor Nagano

latargere:
I have integrated my heat pump and it was recognized as a vivax heatpump but the temperature scale was wrong and I was missing the quiet switch.
I used the vivax heatpump yaml as a baseline to fix the configuration

Tested and everything works perfectly


make-all:
After reviewing the information on https://github.com/make-all/tuya-local/issues/773, I think these changes are correct for the Vivex device as well, so I made the changes in there instead.

latargere:
Great thanks, I tested and it seems to work perfectly

Issue #777, opened 2023-06-09
Poolex Q line adds as domestic water heater with different commands 


WilcoE:
My Poolex Q-line heatpump is discovered as "hydrotherm_dynamic_x8_water_heater", which is a domestic water heater.
current and set temperatures are there, but the possible mode settings are all wrong.

# Log Message

2023-06-08 22:38:23.114 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches hydrotherm_dynamic_x8_water_heater with quality of 71%. DPS: {"updated_at": 1686256702.6255624, "1": true, "2": 31, "3": 30, "4": "ECO_Heat", "11": "heating", "13": "c", "21": 0}

```
Please paste logs here
```

# Information about DPS mappings

```
Please paste the output here.
```
{
  "result": {
    "category": "qn",
    "functions": [
      {
        "code": "switch",
        "dp_id": 1,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "dp_id": 2,
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":5,\"max\":80,\"scale\":0,\"step\":1}"
      },
      {
        "code": "temp_unit_convert",
        "dp_id": 13,
        "type": "Enum",
        "values": "{\"range\":[\"c\",\"f\"]}"
      },
      {
        "code": "temp_set_f",
        "dp_id": 14,
        "type": "Integer",
        "values": "{\"unit\":\"℉\",\"min\":41,\"max\":176,\"scale\":0,\"step\":1}"
      }
    ],
    "status": [
      {
        "code": "switch",
        "dp_id": 1,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "dp_id": 2,
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":5,\"max\":80,\"scale\":0,\"step\":1}"
      },
      {
        "code": "temp_current",
        "dp_id": 3,
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":-30,\"max\":100,\"scale\":0,\"step\":1}"
      },
      {
        "code": "work_state",
        "dp_id": 11,
        "type": "Enum",
        "values": "{\"range\":[\"heating\"]}"
      },
      {
        "code": "temp_unit_convert",
        "dp_id": 13,
        "type": "Enum",
        "values": "{\"range\":[\"c\",\"f\"]}"
      },
[...]

PR #966, opened 2023-08-05
Add support for POOLSANA InverPOWER Heatpump

VadimKraus:
This PR adds a config for POOLSANA InverPOWER (bought 2022) heatpumps.


make-all:
Rather than create a new device, I edited the poolex_q7_heatpump config, which seems compatible, but was missing the presets as there was insufficient info in #777 to support them.  Also changed the primary entity from water_heater to climate entity now that it is known that this device can also cool.

nest000:
Hey @VadimKraus, thank you very much for this implementation! I have also a poolsana inverpower and i want to ask you if you see a chance to get the values from the internal menu like in and out temp, current, power consumption etc.

Many thanks!

Issue #954, opened 2023-07-31
2023.8: Vacuum battery_level will be deprecated


make-all:
Vacuum battery_level is being deprecated, to be replaced by a battery sensor.

Update all vacuum devices to follow this change and remove the deprecation warnings.

https://developers.home-assistant.io/blog/2023/07/10/vacuum-updates

PR #969, opened 2023-08-07
Proscenic M9 battery entity fix

blakadder:
Proscenic M9 fix for #954


make-all:
Updates for all vacuum devices were added

PR #970, opened 2023-08-07
8in1 air sensor - add missing names to sensor entities

blakadder:
Added names to entities since with latest 2023.9dev sensors without name: fields inherited only the device name as entity name. 

Not sure what's actually correct with all the new naming changes


make-all:
There seem to be some unwritten requirements for this feature to work.  I would prefer to figure out how to get it working than to undo the changes at this point.

PR #981, opened 2023-08-12
Fix Kogan LX10 Vacuum 'start' command no longer working after 2023.6.1

YodaDaCoda:
Bug introduced by 3488430 in #759 / [2023.6.1](https://github.com/make-all/tuya-local/releases/tag/2023.6.1) - `value_redirect` mappings need a non-null `dps_val`.

If there's a preferred `dps_val` other than `true` I'll happily change this. I think you may have commented on the original PR something to the effect that a `dps_val` would be necessary but at the time it worked without.

### DIFF
--- a/custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
+++ b/custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
@@ -16,8 +16,10 @@ primary_entity:
       type: string
       mapping:
         - value: start
+          dps_val: true
           value_redirect: start
         - value: pause
+          dps_val: true
           value_redirect: pause
         - dps_val: smart
           value: smart


custom_components/tuya_local/devices/kogan_lx10_vacuum.yaml
```diff
@@ -16,8 +16,10 @@ primary_entity:
       type: string
       mapping:
         - value: start
+          dps_val: true
```
make-all, line 19: The type is string, it cannot take a value of true.

It seems you actually want a conditional mapping here, but there are two other dps that you want to be conditional on, which is beyond the capabilities of the mapping logic.

"start" can probably be renamed to "activate", which is already handled in code.  Then you only have "pause" that you need to be conditional on.

make-all:
I have rearranged things in commit a7820d8 so that this mingling of commands and separate start and pause dps should now work by design, and not by chance as it previously did.

PR #982, opened 2023-08-12
Position of DPS ID4 (Poolex Q7) moved upwards

TheBigNeo:
Hi
I don't understand why the position of the entry is important. But with this change, the heat pump can be controlled correctly. Without this adjustment, the pump can neither be deactivated nor can the mode be changed.

Except the problem is that I have the `Q-Line 9` and not `Q-Line 7`
https://www.poolstar.fr/de/produit/warmepumpe-pool-poolex-q-line-7

image

image

### DIFF
--- a/custom_components/tuya_local/devices/poolex_q7_heatpump.yaml
+++ b/custom_components/tuya_local/devices/poolex_q7_heatpump.yaml
@@ -43,6 +43,22 @@ primary_entity:
                 min: 68
                 max: 104
               value_redirect: temp_set_f
+    - id: 4
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: Boost_Heat
+          value: quick_heat
+        - dps_val: Silent_Heat
+          value: quiet_heat
+        - dps_val: ECO_Heat
+          value: smart_heat
+        - dps_val: ECO_Cool
+          value: smart_cool
+        - dps_val: Silent_Cool
+          value: quiet_cool
+        - dps_val: Boost_Cool
+          value: quick_cool
     - id: 3
       type: integer
       name: current_temperature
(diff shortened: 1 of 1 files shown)


make-all:
I don't think it is this change that makes the difference, it is the addition of the climate entity in a previous change, which is not yet released.

make-all:
It should probably be reordered for aesthetic reasons, but after dp 3, not before.  I installed that change.

TheBigNeo:
> I don't think it is this change that makes the difference, it is the addition of the climate entity in a previous change, which is not yet released.

That would make more sense. But why did the change work anyway? 

And yes, def. after the number 3...

make-all:
In the first screenshot, you don't have a climate entity, because you were using the 2023.7.0 release.
Then when you went to make your changes, you started from the latest version, which had already been modified.

In addition to adding the climate entity and covering all the presets, including cooling ones, Those modifications also changed the water_heater operation_mode `off` to `"off"`, as without the quotes, off gets translated to false when the yaml is parsed, and prevented the off/on from working. 

I guess moving the mode was the first thing you tried, so you thought it was mysteriously fixing your issue, when the actual fix was in moving from 2023.7.0 to the later (at that time unreleased) version.

TheBigNeo:
I was very confused at first, because according to `manifest.json` I am still using version `2023.7`.
But I found my mistake.
In HA I am currently using version `2023.7`.
My change in `poolex_qline_heatpump.yaml` file I made in PyCharm and there I used the `main` branch. 
Afterwards I copied only this file to HA. So of course my changes and the changes of the main branch are used.

Thanks a lot for your answer and this great project!
Coffee is on the way!

PR #840, opened 2023-06-25
Add support for Monzana Klimaanlage air conditioner.

stikonas:
Tested with my air conditioner MZK1000.

PR #985, opened 2023-08-13
Update monzana_air_conditioner to add heat function

declanthebritton:
I have been using this integration for a while for this AC and wondered why it didn't have the heat function, so changed the files locally to add it in and everything worked so thought I would put a PR in for it. 

Simple 2-line addition, if making a new device with my specific AC model number is better then I will happily modify this PR but it would seem that all Monzana ACs that have multi-function also can act as a heater.

### DIFF
--- a/custom_components/tuya_local/devices/monzana_klimaanlage_air_conditioner.yaml
+++ b/custom_components/tuya_local/devices/monzana_klimaanlage_air_conditioner.yaml
@@ -20,6 +20,8 @@ primary_entity:
               value: dry
             - dps_val: Fan
               value: fan_only
+            - dps_val: Heat
+              value: heat
     - id: 2
       type: integer
       name: temperature


make-all:
It might be better with a new config.  The original PR #840 mentioned the model MZK1000.  I can only find [MZKA1000](https://www.amazon.de/Monzana-Klimaanlage-Fernbedienung-Abluftschlauch-Klimager%C3%A4t/dp/B096TDRP4G) online, and the description says it supports 4 modes: cooling, dehumidifier, fan and sleep mode.

declanthebritton:
Just went to check my model number and it is infact the MZKA1000 
PXL_20230814_060705669
The new Amazon listing doesn't mention heating but when I bought the AC it was listed as a 5 in one not a 4 in one. But the sticker says 4 in one. 

I think this means they just aren't advertising that it can do heat. 

What would you like to do about this PR? A new file with maybe a name like MZKA1000-5-in-1 ?

make-all:
If it is the same as the original device, then I guess adding the hidden heating mode to the existing config is OK.  I hope the unadvertised nature of this is not because of a safety recall though.

stikonas:
It's not because they stopped advertising. The new 2023 batch doesn't seem to have any heat function. If I switch heat function via tuya, it just jumps back to one of the other modes. Out of curiosity, do you have 4 mode LEDs on top? My device only has cool, fan and dehumidifier options.

declanthebritton:
That’s super odd. Yeah my ac had 4 buttons on the top, I bought mine back in 2021 so might have just got lucky at the time. I think I should reopen this PR and move it to a new file. Thoughts @make-all?

stikonas:
Yes, perhaps we should split into MZKA1000_v1 and MZKA1000_v2 devices. Bad marketing but not unheard of... Router manufactures also often reuse exactly the same model number for different devices.

Issue #268, opened 2022-11-09
New Device request "AGL Ultra Magic" - Lock with contact door status


raphaeleduardo42:
Brand: AGL
Model: Ultra Magic
Country availability: Brazil

I've tried to retrieve as much important options as I could. From what I gather some share the same information, like the 3 and 124.
This model have a alarm when the door keeps opened for a defined time lenght, maybe this can be setted as well.

**https://www.aglbrasil.com/ultra-magic**

It has other models that share the same configuration, but with little differences how it behaves. Some with RFID options.

Thanks in advance.

`{'1': False, '3': False, '5': 0, '101': 1, '102': 1, '103': 'Silencioso', '104': True, '105': True, '107': True, '108': True, '109': True, '110': 0, '111': False, '112': True, '113': True, '114': 'Selecionar', '115': 3, '116': True, '117': '2', '120': False, '121': False, '122': False, '123': False, '124': False, '125': False, '126': False, '127': False, '128': False, '129': False, 'updated_at': 1667941093.3233285}`

```
{
  "result": {
    "category": "ckmkzq",
    "functions": [
      {
        "code": "switch_1",
        "desc": "{}",
        "name": "车库门开关",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "countdown_alarm",
        "desc": "{\"unit\":\"\",\"min\":-1,\"max\":2000,\"scale\":0,\"step\":1}",
        "name": "推送告警延时倒计时",
        "type": "Integer",
        "values": "{\"unit\":\"\",\"min\":-1,\"max\":2000,\"scale\":0,\"step\":1}"
      }
    ],
    "status": [
      {
        "code": "switch_1",
        "name": "车库门开关",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "doorcontact_state",
        "name": "门磁状态",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "countdown_alarm",
        "name": "推送告警延时倒计时",
        "type": "Integer",
        "values": "{\"unit\":\"\",\"min\":-1,\"max\":2000,\"scale\":0,\"step\":1}"
      }
    ]
  },
  "success": true,
  "t": 1667947492526,
  "tid": "f5685c9b5fb611ed8475060dd053b58c"
}
```

Opener: 1
Door contact open: 124
Door contact: 3
Time lenght for lock after closing: 102 (number from 1 to 5)
Auto close option: 107 (bolean)
[...]

PR #1004, opened 2023-08-19
Fix behavior for AGL Ultra Magic

raphaeleduardo42:
Still lack of configurations to expose as Lock or Cover

More information: https://github.com/make-all/tuya-local/issues/268#issuecomment-1685025907


custom_components/tuya_local/devices/agl_ultramagic_lock.yaml
```diff
@@ -1,99 +1,56 @@
-name: AGL Ultra Magic lock
+name: AGL Ultra Magic Lock
```
make-all, line 1: Move the branding into the name under products, and make this something generic.  Maybe "Gate opener" would be more suitable, since it doesn't contain a lock entity any more.

custom_components/tuya_local/devices/agl_ultramagic_lock.yaml
```diff
@@ -1,99 +1,56 @@
...
-        unit: s
+      - id: 119
+        type: boolean
+        name: button
+        optional: true
+
+
+## Sensors
+  - entity: binary_sensor
+    name: Door Contact
```
make-all, line 27: I think the name is unnecessary, as the class should provide a default name (which will come with translations) already.  But if a cover can be used per the comment above, then this is not needed, since it will be covered by the cover entity.

custom_components/tuya_local/devices/agl_ultramagic_lock.yaml
```diff
@@ -1,99 +1,56 @@
-name: AGL Ultra Magic lock
+name: AGL Ultra Magic Lock
+products:
+  - id: xrqkoapmq1rwjvms
+    name: Ultra Magic
 primary_entity:
-  entity: lock
+  entity: button
```
make-all, line 6: I think it should be possible to combine the open and close using constraint and conditions into either a lock or cover entity.
```yaml
entity: cover
class: gate
dps:
  - id: 118
    type: boolean
    name: control
    mapping:
       - dps_val: true
          value: "open"
       - constraint: close_ctrl
          conditions:
            - dps_val: true
              value: "close"
  - id: 119
    type: boolean
    name: close_ctrl
    hidden: true
  - id: 3
    type: string
    name: open
    mapping:
      - dps_val: true
        value: false
      - dps_val: false
         value: true
```

Lock is maybe more difficult due to the separate dp 117 to indicate current state, and 118 and 119 to lock and unlock.

custom_components/tuya_local/devices/agl_ultramagic_lock.yaml
```diff
@@ -1,99 +1,56 @@
...
+        type: boolean
+        name: sensor
+
+        mapping:
+          - dps_val: true
+            value: false
+          - dps_val: false
+            value: true
+  - entity: binary_sensor
+    name: Gate Contact
```
make-all, line 40: I think maybe better to use the default naming from the lock class here, as  the Tuya naming of "Gate contact" vs "Door contact" above is not very obviously distinct - this is the lock state, and the above is the open/closed state.

custom_components/tuya_local/devices/agl_ultramagic_lock.yaml
```diff
@@ -113,15 +71,66 @@ secondary_entities:
             value: High
   - entity: switch
     category: config
-    name: Auto close
+    name: Allow Intercom Unlock
```
make-all, line 74: Are these configuration switches?  The naming "approve_intercom" etc suggests that they are to be used to unlock the door in response to a "request_intercom" signal, which is what the lock entity is able to handle if the simpler "lock" dp is left out.

make-all:
Sorry it has taken me a while to review this, as the changes were quite extensive and seem to completely change the function of the device from how the Tuya documentation describes it.

custom_components/tuya_local/devices/agl_ultramagic_lock.yaml
```diff
@@ -1,99 +1,29 @@
-name: AGL Ultra Magic lock
+name: AGL Ultra Magic Lock
+products:
+  - id: xrqkoapmq1rwjvms
+    name: Ultra Magic
 primary_entity:
-  entity: lock
+  entity: cover
+  class: garage
```
make-all, line 7: It seems like gate would be a more appropriate class for this device.

PR #1005, opened 2023-08-19
Add support for AGL utracontato R2

robsonmantovani:


custom_components/tuya_local/devices/agl_ultracontato.yaml
```diff
@@ -0,0 +1,84 @@
...
+        - dps_val: true
+          value: "open"
+        - dps_val: false
+          value: "close"
+    - id: 101
+      name: open
+      type: boolean
+secondary_entities:
+  - entity: select
+    name: Switch type
```
make-all, line 19: Maybe sensor type.

custom_components/tuya_local/devices/agl_ultracontato.yaml
```diff
@@ -0,0 +1,84 @@
+name: AGL ultracontato r2
```
make-all, line 1: Judging by the dps map values below being in Portuguese, this is probably specific to this device and not a whitelabel device, so it is OK like this, but in general, branding should go under a products section as whitelabel devices are common in the Tuya ecosystem, and the top level name should be generic.

custom_components/tuya_local/devices/agl_ultracontato.yaml
```diff
@@ -0,0 +1,84 @@
...
+    icon: "mdi:pulse"
+    dps:
+      - id: 109
+        name: option
+        type: string
+        mapping:
+          - dps_val: "Pulsado"
+            value: Pulsed
+          - dps_val: "Retencao"
+            value: Retencion
```
make-all, line 43: I think name should be Output type, since pulsed is just one of the options.  I think the first two are what is commonly referred to as Pulse and Level in electrical engineering terms, but I'm not sure where the ONOFF option fits in.  Maybe the Installation manual has enough details to work out what these really mean.  Retention would be the literal translation of Rentenção in English (Retencion looks more like a Spanish spelling), but doesn't make much sense in this context.

custom_components/tuya_local/devices/agl_ultracontato.yaml
```diff
@@ -0,0 +1,84 @@
...
+          - dps_val: "ONOFF"
+            value: ON - OFF
+  - entity: select
+    category: config
+    name: Autoclose
```
make-all, line 48: This can be a switch since there are only two options.  Name should be "Auto close"

custom_components/tuya_local/devices/agl_ultracontato.yaml
```diff
@@ -0,0 +1,84 @@
...
+        name: option
+        type: string
+        mapping:
+          - dps_val: "habilitado"
+            value: Enabled
+          - dps_val: "desabilitado"
+            value: Disabled
+  - entity: number
+    category: config
+    name: During time
```
make-all, line 61: This name is not clear.  What is the actual timing controlled by this?  Unlock duration?
Proabably `class: duration` should be added, but name should be kept, as just "Duration" is not clear enough as to the function.

custom_components/tuya_local/devices/agl_ultracontato.yaml
```diff
@@ -0,0 +1,84 @@
...
+    dps:
+      - id: 107
+        name: value
+        type: integer
+        range:
+          min: 1
+          max: 65
+        unit: s
+  - entity: binary_sensor
+    name: Door Opened
```
make-all, line 72: Replace name with `class: door` to get translated name and state.

custom_components/tuya_local/devices/agl_ultracontato.yaml
```diff
@@ -0,0 +1,84 @@
...
+    category: diagnostic
+    dps:
+      - id: 111
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
```
make-all, line 78: It seems strange to have door opened and door closed as different sensors.  Is one of them the lock state?  In which case the class should be "lock" and no name is needed.

But there is also an additional open sensor on dp 101 used in the cover entity.

robsonmantovani:
I fixed a few names for better comprehend

PR #1017, opened 2023-08-22
Add Sonoff SNZB-02 ZigBee Temperature & Humidity Sensor

LesTR:
It uses the same DPS as `MultiIR Zigbee TE100-TY`, just without scale. I'm using it for some time already, just don't have time to contribute the change back.

DPS:
```json
{"101": 51, "102": 83, "103": 7, "full_poll": true}
```

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -496,6 +496,7 @@ port and password.
 - Nedis ZBSC10WT temperature and humidity sensor
 - Temperature and humidity sensor with alarm feature
 - ZTH08ZTU temperature and humidity sensor
+- Sonoff SNZB-02 ZigBee temperature and humidity sensor
 
 NOTE: this project does not intend to expand the scope to support non-Tuya
 devices via Tuya hubs.  Though it may be techincally feasible to do such a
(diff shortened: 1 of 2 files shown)


make-all:
This is what I feared with the addition of Zigbee hub support.  The scope of the project has expanded beyond Tuya devices to whatever zigbee devices users want to add to a Tuya zigbee hub.

I think this needs to be rejected to set some boundaries.

OleksandrChekmez:
@LesTR I added zigbee_sonoff_snzb-02-climate-sensor.yaml to custom_components/tuya_local/devices/. Did a full restart of HA and tried to add Sonoff SNZB-02 (with the selection of the appropriate Zigbee gateway). Device ID and sub device ID were recognized correctly, when I click Submit, I get this error:

LesTR:
> Any suggestions on how to fix it?

To be honest, buy a regular ZB gateway :D 

I'm not using that Tuya GW anymore. There were many issues, like slow updates, bad precision, device disconnect, etc. I replaced it with Skyconnect, and everything works out of the box and is reliable.

OleksandrChekmez:
@LesTR 
I use zig star with HA for 99% devices. 
I wanted to use this temperature sponsor in my garret but zigbee signal was too weak. I have plenty of router devices on the lower level but it does not help. I suppose the foil insulation between garret and lower lever can be the reason. 
this is why I tried to use Tuya zigbee gateway with WiFi. WiFi signal is strong enough in the garret and it works. Now I want to add the data from the sensor to HA. 

I’ll try cloud Tuya integration, but hopped to do everything locally.

LesTR:
Understand, but I think adding just a definition should not break the connection, it's just a DPS mapping. The reason should be in the logs or something might change in the subdevices code.

PR #1047, opened 2023-08-31
#930 add carro ceiling fan device

whit-tj:


whit-tj:
I created a devices file for this ceiling fan and light combo based on the guidance in the readme. I have the reported codes in here and compared to other fan/light combos to ensure I followed the correct format. This is my first pull request so any input is appreciated.

If I manually add this file into the /custom_components/tuya_local/devices/ will I be able to add and test the device, or are there more steps to test this?

MaximeNagel:
Yes you can test it on your own if you put your file in the correct folder and try to add your device. It should be display in the select field after you give the device information :)

whit-tj:
Well after adding my file and toggling some features on the fan to ensure the codes were all set, it now matches to a "circuit breaker" at 25% but not my fan. Is it only trying to match on this: `DPS: {"updated_at": 1693507546.1336124, "1": true, "3": "5", "6": "4", "7": 234, "9": true, "10": 500, "101": true, "102": 500}`? I have every one of those values mapped in the device file (except "updated_at"). So you'd think it could match a higher percent to my device.

But there are 2 missing codes in that DPS, fan_direction and mode. There is no mode set, obviously there is a fan_direction which is in its default direction, so perhaps it doesn't send if its in default. 

I thought it was supposed to used the product_id as well to match? Any suggestions on how to troubleshoot why its not matching my new device? I feel like my syntax looks correct, and I'm not getting any errors with the file.

whit-tj:
Working:

- Fan on/off
- Fan mode (readonly)
- Warm light on/off
- White light on/off
- Fan direction forward/reverse
- Fan countdown hours (readonly)
- Fan countdown left, how many minutes of those hours (readonly)

Not working fully:

- Fan speed (not sure if this is a UI issue, even though range is 10-100 with a step of 10, a 10% is full speed (100), anything 11% or above goes to speed lowest speed 1 (value 10). When I first brought it in it was defaulting to speed 5 (which should be 50), but the UI showed that as 5%. Not quite sure if there is an issue with the speed being set/read from the device or all UI. Because if I try to change the slider to a range of 1-10, it errors saying it cant call the fan service, it needs to be in steps of 10.
- Warm light brightness. Similar as above, the device says its range is 10-1000 (step 1). The initial value was coming in as 500. Which shows up in the UI as 200% brightness, but if I slide a slider to max 100%, thats only 250 to the light. So technically I'm only controlling brightness from 0 to 25%.
- White light brightness (same issue as above)

Anyone have some thoughts on if the data needs to change on the device or set a setting I'm missing. Or is there some sort of conversion or template needed on the home assistant side to be able to set the correct full range of values.

make-all:
HA uses fixed ranges of 0-100 for fan speed and 0-255 for brightness, so you will need to scale the device values to make them within that range.
In the case of the fan, the device range is 1-10, so you need a scale of 0.1 to convert from HA's range.
In the case of the brightness, the device range is 10-1000, so you need a scale of 3.92.

The step should be in device range (after scaling), so the fan speed does not need a step set (after scaling the default step of 1 should work).

whit-tj:
> HA uses fixed ranges of 0-100 for fan speed and 0-255 for brightness, so you will need to scale the device values to make them within that range. In the case of the fan, the device range is 1-10, so you need a scale of 0.1 to convert from HA's range. In the case of the brightness, the device range is 10-1000, so you need a scale of 3.92.
> 
> The step should be in device range (after scaling), so the fan speed does not need a step set (after scaling the default step of 1 should work).

Brilliant! Setting the scale for brightness fixed the lights. And setting the scale to .1 and step back to default seems to read in the info to match the UI. Fan speed 3 is 30% on a slider, speed 7 is 70%. 

However when I try to slide the UI to 50% (speed 5) I get this error in the service call:

`Failed to call service fan/set_percentage. speed (50) must be between 100.0 and 1000.0` 

Do I need to scale something on the home assistant side to? It reads it correctly, so I would have thought its in the right format to make that service call back to change it.

make-all:
Your range should also be changed to be the device range (1 - 10) rather than the HA UI range.

whit-tj:
Thank you! Everything is working smoothly in home assistant now. I appreciate the help and the work on this project.

make-all (changes requested):

make-all:
Regarding earlier issues: you can mark dps which are not consistently sent with `optional: true` to improve device detection.  Product id is not currently used, as it is not possible to query from the devices once connected, it requires either network discovery or cloud access to get, neither of which are currently implemented (I hope to add network discovery soon, but as an optional feature since it clashes with localtuya's use of the same feature, and long term limited cloud functionality - but again, optional as many users use this integration because they explicitly do not want to use the cloud).

make-all:
Merged

PR #1057, opened 2023-09-02
Add support to a new zigbee presence sensor

felipejfc:
Added support to the zigbee variant of the mw100 presence sensor.

### DIFF
--- a/custom_components/tuya_local/devices/zigbee_dv7vcs_human_presence_sensor.yaml
+++ b/custom_components/tuya_local/devices/zigbee_dv7vcs_human_presence_sensor.yaml
@@ -1,7 +1,7 @@
 name: human presence sensor
 products:
   - id: sxm7l9xa
-    name: Human presence sensor
+    name: Loginovo mmWave human presence sensor
 primary_entity:
   entity: binary_sensor
   class: occupancy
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/zigbee_dv7vcs_human_presence_sensor.yaml
```diff
@@ -0,0 +1,115 @@
+name: human presence sensor
+products:
+  - id: sxm7l9xa
+    name: Human presence sensor
```
make-all, line 4: This product specific name should have the brand and model name.

custom_components/tuya_local/devices/zigbee_dv7vcs_human_presence_sensor.yaml
```diff
@@ -0,0 +1,115 @@
...
+        name: value
+        unit: m
+        range:
+          min: 0
+          max: 950
+        mapping:
+          - scale: 10
+            step: 10
+  - entity: sensor
+    name: Target distance
```
make-all, line 60: I think the name is not needed, as the class: distance should have suitable translated names, and this is the only sensor of that class.

felipejfc:
updated! ptal

PR #1073, opened 2023-09-08
Added device support wf520d_wifi_dual_touch_dimmer

suyashrai:
Tuya Smart WiFi Smart Touch Panel Dimmer Light Retrofit for Anchor Roma Modular Switch 2 Gang Modular switch (WF520D)

https://www.sonoff.in/product/smart-wifi-smart-touch-panel-dimmer-light-switch-2-gang-modular-switch-works-with-alexa-google-2/

### DIFF
--- a/custom_components/tuya_local/devices/wf520d_wifi_dual_touch_dimmer
+++ b/custom_components/tuya_local/devices/wf520d_wifi_dual_touch_dimmer
@@ -1,4 +1,4 @@
-name: Dual Dimmer
+name: Dual dimmer
 primary_entity:
   entity: light
   name: Dimmer 1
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/wf520d_wifi_dual_touch_dimmer
```diff
@@ -0,0 +1,59 @@
+name: WF520D Dual Touch Dimmer
```
make-all, line 1: HA naming convention is for devices and entities to only capitalise the first word of the name unless the words are trademarks or abbreviations that need to be capitalised.

If you can access the product id (from tuya developer portal, or by a network scan using tinytuya), then it is better to use a generic name above like "Dual dimmer" and add a products section with id and name that includes branding.

suyashrai: Sure, I've changed it to "Dual dimmer".

suyashrai (commented):
Name change.

PR #1079, opened 2023-09-09
Add support for new devices

almirus:
Add support for Royal Clima fresh, neatsvor x600 vacuum, ble Smart Curtain, ble smart plant moisture, dimming track light

### DIFF
--- a/ACKNOWLEDGEMENTS.md
+++ b/ACKNOWLEDGEMENTS.md
@@ -409,3 +409,4 @@ Further device support has been made with the assistance of users.  Please consi
 - [lyatam](https://github.com/lyatam) for contributing support for ZX-DB11B doorbell alarms.
 - [urgelaszlo](https://github.com/urgelaszlo) for contributing support for Fisher Summer air conditioners.
 - [felipecrs](https://github.com/felipecrs) for assisting with support for SmarDeer smart locks.
+- [rbalaev](https://github.com/rbalaev) for assisting with support for Thermex Lima 80V water heaters.
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -98,6 +98,7 @@
 - Hydrotherm Dynamic/X8 heat pump hot water systems
 - ITS-4.5HD super water heatpump
 - Thermex IF series V pro hot water systems
+- Thermex Lima 80V
 
 ### Thermostats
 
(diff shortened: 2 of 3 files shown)


custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,414 @@
...
+    dps:
+      - id: 102
+        name: switch
+        type: string
+        mapping:
+          - dps_val: "off"
+            value: false
+          - dps_val: "auto"
+            value: true
+  - entity: select
```
make-all, line 164: Based on the dp name, I think this should have not been a separate select entity, but the preset_mode attribute of the fan above.  in that case, the preset names should come from the ventilation key if possible.
Maybe circulate, anti-condensation, eco, fresh, auto?

make-all:
I suggest splitting these into separate pull requests.  Some are ready to merge, but are being held up by the issues in the other configs.

MaBeniu:
I would also like to request to add one more product id.
"model": "X600",
 "name": "Neatsvor X600",
 "product_id": "zcpce4xrxhx4ep05",
 "product_name": "Neatsvor X600"
this is information from iot.tuya platform. I have reviewed all the configured ID and they all perfectly match.

MaBeniu:
Is this going to any release or what happened?
also maybe someone could help me to set this up. I used neatsvor yaml and placed it under tuya_local/devices/
however it is not mapped to my device. I did went through all the DP's in tuya developer platform all is defined in the yaml correctly however it is not maching this configuration at the point of adding device.
instead i get suggestion to other few devices:
image

in the system log I get this error:
Device matches asc_wifi_circuit_breaker with quality of 23%. DPS: {"updated_at": 1702208651.6489308, "1": false, "4": "chargego", "5": "charge_done", "6": 0, "7": 0, "8": 100, "9": "gentle", "10": "closed", "17": 1, "19": 101, "23": 149, "26": 8, "28": 0}

make-all:
The original author closed it, I presume to work on separating it into individual PRs for each device.

MaBeniu:
I will test the file until I reach better quality. I was finally able to get it running by commenting most of DPs and finally reaching 30% match.
Already altered the file based on the comments (bug list) above, removing unsupported battery DP and other stuff

PR #1085, opened 2023-09-10
Create AlecoAir_D25_Traditio.yaml

b1be:


custom_components/tuya_local/devices/AlecoAir_D25_Traditio.yaml
```diff
@@ -0,0 +1,226 @@
...
+      - id: 10
+        type: boolean
+        name: switch
+        mapping:
+          - dps_val: true
+            icon: "mdi:creation"
+          - dps_val: false
+            icon: "mdi:creation-outline"
+  - entity: switch
+    name: uv
```
make-all, line 123: "UV sterilization" is the usual name used by other devices.  Commonizing the naming will help when someone wants to make the effort to translate entity names in future.

custom_components/tuya_local/devices/AlecoAir_D25_Traditio.yaml
```diff
@@ -0,0 +1,226 @@
...
+      - id: 13
+        type: boolean
+        name: switch
+        mapping:
+          - dps_val: true
+            icon: "mdi:leaf"
+          - dps_val: false
+            icon: "mdi:leaf-outline"
+  - entity: sensor
+    name: Current humidity
```
make-all, line 136: Current humidity as a separate sensor is deprecated since HA 2023.7.  Instead, it should be added to the humidifier entity, named "current_humidity".

custom_components/tuya_local/devices/AlecoAir_D25_Traditio.yaml
```diff
@@ -0,0 +1,226 @@
...
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 19
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
```
make-all, line 155: If you know which fault code is "tank full" (usually combined with tank missing), then it is better to do this the other way around - default value is false, and only the code for "tank full" is true.  Another general problem binary_sensor can be added to catch other errors (but should set the tank full code to false as well as the 0 code)

custom_components/tuya_local/devices/AlecoAir_D25_Traditio.yaml
```diff
@@ -0,0 +1,226 @@
...
+  - entity: lock
+    name: Child lock
+    category: config
+    dps:
+      - id: 16
+        type: boolean
+        name: lock
+        mapping:
+          - dps_val: true
+            icon: "mdi:lock"
```
make-all, line 165: I generally prefer "mdi:hand-back-right-off" as an icon for child-locks.

Also, those early device configs have separate icons for all the states, I have moved away from that after discovering that it messes with themes to change things that are not generally expected to change.  So now I use this sparingly only to indicate exceptional state.

PR #1094, opened 2023-09-14
Added support for SWE SAGEWE F2s501-GB

xe5700:


make-all:
I have made a change to smartplugv2_energy instead, to avoid proliferation of smartplug configs with minor differences..

PR #1096, opened 2023-09-15
Adding support for SWS-001 weather station (sensors only)

sebastianciupinski:
Weather station supports also outside climate, but since it is taken from internet, based on location, there is no need to integrate that.

### DIFF
--- a/custom_components/tuya_local/devices/smart_weather_station.yaml
+++ b/custom_components/tuya_local/devices/smart_weather_station.yaml
@@ -1,0 +1,93 @@
+name: Smart Weather Station
+products:
+  - id: 000001bbfd
+    name: Smart Weather Station
+primary_entity:
+  entity: sensor
+  name: local temperature
+  class: temperature
+  dps:
+    - id: 131
+      name: sensor
+      type: integer
+      unit: C
+      mapping:
+        - scale: 10
+      class: measurement
+secondary_entities:
+  - entity: sensor
+    name: local humidity
+    class: humidity
+    dps:
+      - id: 132
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+        
+  - entity: sensor
+    name: sensor 1 temperature
+    class: temperature
+    dps:
+      - id: 133
+        type: integer
+        name: sensor
+        unit: C
+        mapping:
+          - scale: 10
+        class: measurement
+        
+  - entity: sensor
+    name: sensor 1 humidity
+    class: humidity
+    dps:
+      - id: 134
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+        
+  - entity: sensor
+    name: sensor 2 temperature
+    class: temperature
+    dps:
+      - id: 135
+        type: integer
+        name: sensor
+        unit: C
+        mapping:
+          - scale: 10
+        class: measurement
+        
+  - entity: sensor
+    name: sensor 2 humidity
+    class: humidity
+    dps:
+      - id: 136
+        type: integer
+        name: sensor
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/smart_weather_station.yaml
```diff
@@ -0,0 +1,93 @@
+name: Smart Weather Station
+products:
+  - id: 000001bbfd
```
make-all, line 3: Is this the product id?  It looks more like what appears in the modelId field of Query Things Data Model API call, which is something different (maybe some internal id for Tuya cloud).

custom_components/tuya_local/devices/smart_weather_station.yaml
```diff
@@ -0,0 +1,93 @@
+name: Smart Weather Station
```
make-all, line 1: HA naming convention: Only capitalise the first word of a name (unless a trademark or abbreviation that requires it)

custom_components/tuya_local/devices/smart_weather_station.yaml
```diff
@@ -0,0 +1,93 @@
...
+products:
+  - id: 000001bbfd
+    name: Smart Weather Station
+primary_entity:
+  entity: sensor
+  name: local temperature
```
make-all, line 7: HA naming convention: capitalise the first word of an entity name.
I'm not sure whether "local" is the right word, it implies maybe a web service used to get the local temperature for your area. Many other devices use "Internal" for built in sensors where they also support external probes, which is probably clearer.

custom_components/tuya_local/devices/smart_weather_station.yaml
```diff
@@ -0,0 +1,93 @@
...
+    - id: 131
+      name: sensor
+      type: integer
+      unit: C
+      mapping:
+        - scale: 10
+      class: measurement
+secondary_entities:
+  - entity: sensor
+    name: local humidity
```
make-all, line 19: Same comments abourt HA naming convention and "local" as above.

custom_components/tuya_local/devices/smart_weather_station.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - id: 133
+        type: integer
+        name: sensor
+        unit: C
+        mapping:
+          - scale: 10
+        class: measurement
+        
+  - entity: sensor
+    name: sensor 1 humidity
```
make-all, line 41: HA naming convention. Also the rest of the entities below.

custom_components/tuya_local/devices/smart_weather_station.yaml
sebastianciupinski: This is how device introduces itself in tuya app, but I am ok with changing it.

custom_components/tuya_local/devices/smart_weather_station.yaml
sebastianciupinski: This is how sensor is named in app. I also do not get naming convention. I mean I see plendy of devices having without first letter capitalized.

custom_components/tuya_local/devices/smart_weather_station.yaml
sebastianciupinski: it is going to be changed to "fpgistthfiweyaqr"

custom_components/tuya_local/devices/smart_weather_station.yaml
make-all: The convention is relatively recent, you may be looking at older integrations that are not yet updated.

PR #1099, opened 2023-09-17
Add Support for Trotec PAC-W 2600 SH

alexkn:


custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
```diff
@@ -0,0 +1,80 @@
...
+      mapping:
+        - dps_val: 1
+          value: low
+        - dps_val: 2
+          value: medium
+        - dps_val: 3
+          value: high
+    - id: 19
+      name: temperature_unit
+      type: string
```
make-all, line 51: Are the values "C" and "F" (upper case)?  If they are lower case, as is more common, then a mapping will be needed.

It would also be useful to expose this as a select entity so the units can be changed from HA.

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
```diff
@@ -0,0 +1,80 @@
...
+      mapping:
+        - dps_val: true
+          value: "vertical"
+        - dps_val: false
+          value: "off"
+    - id: 105
+      name: countdown
+      type: integer
+    - id: 106
+      name: alarm
```
make-all, line 73: I think this is better described as a fault code than alarm, based on other similar devices (also alarm2).

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - id: 3
+      type: integer
+      name: current_temperature
+    - id: 4
+      type: string
+      name: mode
+      hidden: true
+    - id: 5
+      name: fan_mode
+      type: integer
```
make-all, line 41: Is this really an integer and not a string that happens to contain integer values?

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
```diff
@@ -0,0 +1,80 @@
...
+        - dps_val: 1
+          value: low
+        - dps_val: 2
+          value: medium
+        - dps_val: 3
+          value: high
+    - id: 19
+      name: temperature_unit
+      type: string
+    - id: 101
```
make-all, line 52: It would be useful to expose Sleep as a switch entity, rather than just reporting it as an attribute on the climate entity.

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
```diff
@@ -0,0 +1,80 @@
...
+      name: sleep
+      type: boolean
+    - id: 102
```
make-all, line 55: I'm not sure whether this is an anti-frost mode that could be a switch, or an indicator that could be a binary_sensor (class: cold)

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
alexkn: Yes looks like upper case "C" and "F"

alexkn: I don't think it is really meant to be changed.
It is changeable via API, but not via tuya app (at least I can't find it) and the app does change the values but not the unit.

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - id: 104
+      name: swing_mode
+      type: boolean
+      mapping:
+        - dps_val: true
+          value: "vertical"
+        - dps_val: false
+          value: "off"
+    - id: 105
+      name: countdown
```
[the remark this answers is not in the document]
alexkn: The Timer property is an integer value in hours, which corresponds with with the Timer-Setting on the remote.
The countdown property is an integer and looks like the time left in minutes.

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
alexkn: In API Explorer it is represented as a string, but it also works as an integer.

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
alexkn: I don't know, there is no corresponding button on the remote control or in the app.

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
```diff
@@ -70,11 +70,20 @@ primary_entity:
       name: countdown
       type: integer
     - id: 106
-      name: alarm
+      name: fault_code
       type: integer
```
make-all, line 74: fault_codes are probably bitfield types. (Tuya calls them bitmap, but I used bitfield to avoid confusion with the image format).  Without a mapping there isn't any difference between integer and bitfield, but it might help future proof the config.

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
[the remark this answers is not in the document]
make-all: I just want to confirm the type of 103. Can you download the device diagnostics and check the raw dps that is in there.  Some other methods of displaying raw dps strip the quotes, making it difficult to tell.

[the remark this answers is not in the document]
alexkn: I'm not sure what you mean with "device diagnostics".
This is from "Query Properties" on the Tuya IoT Platform:
```json
{
        "code": "Timer",
        "custom_name": "",
        "dp_id": 103,
        "time": 1696182327314,
        "value": 0
},
```
And from "Query Things Data Model"
```json
{
	"abilityId": 103,
	"accessMode": "rw",
	"code": "Timer",
	"description": "",
	"name": "Timer",
	"typeSpec": {
		"max": 24,
		"min": 0,
		"scale": 0,
		"step": 1,
		"type": "value",
		"typeDefaultValue": 0,
		"unit": "Hour"
	}
},

```

custom_components/tuya_local/devices/trotec_pac-w_heatpump.yaml
alexkn: According "Query Things Data Model" it is readonly
```json
{
	"abilityId": 102,
	"accessMode": "ro",
	"code": "Defrost",
	"description": "0: abtauen\\n1: Auftauen\\n",
	"name": " Defrost",
	"typeSpec": {
		"type": "bool",
		"typeDefaultValue": false
	}
},
```

PR #1114, opened 2023-09-24
Add Support for Rotenso Roni X

cykus:


custom_components/tuya_local/devices/rotenso_ronix_heatpump.yaml
```diff
@@ -0,0 +1,309 @@
...
+      name: flags_2
+    - id: 125
+      type: string
+      name: air_quality
+      optional: true
+    - id: 132
+      type: boolean
+      name: hot_cool
+secondary_entities:
+  - entity: sensor
```
make-all, line 115: This is duplicating the climate entity's temperature reading.

custom_components/tuya_local/devices/rotenso_ronix_heatpump.yaml
```diff
@@ -0,0 +1,309 @@
...
+    class: temperature
+    dps:
+      - id: 3
+        name: sensor
+        type: integer
+        unit: C
+        class: measurement
+  - entity: sensor
```
make-all, line 123: This can be moved into the climate entity as current_humidity.

custom_components/tuya_local/devices/rotenso_ronix_heatpump.yaml
```diff
@@ -0,0 +1,309 @@
...
+        - dps_val: "low"
+          value: low
+        - dps_val: "mute"
+          value: mute
+        - dps_val: "auto"
+          value: auto
+    - id: 20
+      type: integer
+      name: fault_code
+    - id: 105
```
make-all, line 66: This duplicates the Sleep select entity below, so is not needed here.

custom_components/tuya_local/devices/rotenso_ronix_heatpump.yaml
```diff
@@ -0,0 +1,309 @@
...
+    icon: "mdi:water-off-outline"
+    dps:
+      - id: 123
+        type: hex
+        name: switch
+        mapping:
+          - scale: 1
+            mask: "0100"
+  - entity: switch
+    name: Health
```
make-all, line 258: Is this a UV purifier?  If so, "UV sterilization" is the standard naming other devices use (keeping consistent naming will help future translation efforts)

custom_components/tuya_local/devices/rotenso_ronix_heatpump.yaml
```diff
@@ -0,0 +1,309 @@
+name: Rotenso Roni X
```
make-all, line 1: Use a generic name here in case other brands match.

custom_components/tuya_local/devices/rotenso_ronix_heatpump.yaml
```diff
@@ -0,0 +1,309 @@
...
+      type: integer
+      name: current_temperature
+    - id: 4
+      type: string
+      name: mode
+      hidden: true
+    - id: 5
+      type: string
+      name: fan_mode
+      mapping:
```
make-all, line 46: values should be standard values from https://github.com/home-assistant/core/blob/dev/homeassistant/components/climate/strings.json (under fan_mode) or extended values from https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/en.json lines 67-73, in which case `translation_key: aircon_extra` needs to be added at the entity level.

custom_components/tuya_local/devices/rotenso_ronix_heatpump.yaml
cykus: i'm not 100% sure, but it's probably ionizer not uv, how should i name it in this case?

make-all: "Ionizer" is the standard naming if you are sure that is the case.

cykus: manual says Ionizer(optional), i added optional: true to entity

make-all: Optional works on the dp level, not the entity level.  There isn't really any way to make an individual bit from a bigger flags dp optional, as there would not be any way to differentiate "off" from "not supported".

PR #1134, opened 2023-10-02
Add Aroma Diffuser YX316WIFI

spuljko:
Added Tuya Smart WiFi 500ml Aromatherapy Essential Oil Diffuser Air Humidifier
    "name": "YX316WIFI",
    "product_id": "arkdxfhlkpdyof9x",
    "product_name": "YX316WIFI",
https://www.aliexpress.com/item/1005005262690569.html

### DIFF
--- a/custom_components/tuya_local/devices/yx316wifi_aroma_diffuser.yaml
+++ b/custom_components/tuya_local/devices/yx316wifi_aroma_diffuser.yaml
@@ -14,6 +14,18 @@ primary_entity:
           icon: "mdi:scent"
         - dps_val: false
           icon: "mdi:scent-off"
+    - id: 3
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: "MID"
+          value: "medium"
+        - dps_val: "2H"
+          value: "low"
+        - dps_val: "4H"
+          value: "high"
+        - dps_val: "10SEC"
+          value: "10 seconds"
     - id: 11
       name: fault_code
       type: bitfield
@@ -76,22 +88,6 @@ secondary_entities:
             range:
               min: 0
               max: 255
-  - entity: select
-    name: Work Mode
-    category: config
-    dps:
-      - id: 3
-        name: option
-        type: string
-        mapping:
-          - dps_val: "MID"
-            value: "MID"
-          - dps_val: "10SEC"
-            value: "10SEC"
-          - dps_val: "2H"
-            value: "2H"
-          - dps_val: "4H"
-            value: "4H"
   - entity: switch
     name: Sound
     category: config
@@ -99,4 +95,3 @@ secondary_entities:
       - id: 6
         name: option
         type: boolean
-
--- a/custom_components/tuya_local/devices/yx316wifi_aroma_diffuser.yaml
+++ b/custom_components/tuya_local/devices/yx316wifi_aroma_diffuser.yaml
@@ -93,5 +93,5 @@ secondary_entities:
     category: config
     dps:
       - id: 6
-        name: option
         type: boolean
+        name: switch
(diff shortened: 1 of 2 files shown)


spuljko:
@make-all why does fan needs preset? This device doesn't have preset. (there is working mode, but that is not speed - it works normally, in 10sec intervals, every x hours...)
I also have toilet ventilation fan, which sldo doesn't have preset...

make-all:
Because without a preset or speed, a fan is just a switch.  In this case, the work_mode seems to work as a preset mode for the fan, but needs a better option naming to make the options understandable.

custom_components/tuya_local/devices/yx316wifi_aroma_diffuser.yaml
```diff
@@ -0,0 +1,97 @@
...
+          value: "high"
+        - dps_val: "10SEC"
+          value: "10 seconds"
+    - id: 11
+      name: fault_code
+      type: bitfield
+      optional: true
+secondary_entities:
+  # "master" switch - turns on and off both light and diffuser
+  - entity: light
```
make-all, line 35: Maybe should be a switch rather than light, as it does not directly control the light.  And the name Power or Master would be better than "Switch" I think (which would be the default if it is a switch entity, so no name needed).

custom_components/tuya_local/devices/yx316wifi_aroma_diffuser.yaml
```diff
@@ -0,0 +1,97 @@
...
+            range:
+              min: 0
+              max: 255
+          - name: v
+            bytes: 1
+            range:
+              min: 0
+              max: 255
+  - entity: switch
+    name: Sound
```
make-all, line 92: This would be better with an icon, I think.

custom_components/tuya_local/devices/yx316wifi_aroma_diffuser.yaml
```diff
@@ -0,0 +1,97 @@
+# DPS sample:  { "1": true, "2": true, "3": "MID", "7": true, "8": 1000, "9": "colourful1", "10": "", "11": 0 }
+name: Aroma Diffuser With Lamp
+products:
+  - id: arkdxfhlkpdyof9x
+    model: YX316WIFI
+primary_entity:
+  entity: fan
```
make-all, line 7: Suggest adding `icon: "mdi:scent"` to differentiate it from an actual fan.

custom_components/tuya_local/devices/yx316wifi_aroma_diffuser.yaml
```diff
@@ -0,0 +1,97 @@
...
+secondary_entities:
+  # "master" switch - turns on and off both light and diffuser
+  - entity: light
+    name: Switch
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+  - entity: light
+    name: Lamp
```
make-all, line 42: Lamp is so similar to the default "Light", that I think it is better to not name it, as then it will benefit from translations.

syst3mboot:
How can we add this device with Tuya Local?
I tried adding it and am prompted with a list of device types to pick from, the options are:
-mellerware_citymove_vacuum
-simple_contact_sensor
-simple_dual_switch
-simple_switch
-simple_switch_timer

my device's product id seems to match too: arkdxfhlkpdyof9x

syst3mboot:
Update, figured it out! The issue is related to the device's "Sound" setting in the Smart Life app.
It appears that if you have "Sound" setting disabled (sent as ID 6), instead of sending ID 6 as `False`, it just stops sending it altogether. This makes the configuration yaml not match the device so Tuya Local doesn't recognize it. Enabling the Sound setting fixed the issue. Alternatively, I imagine creating a clone of the `yx316wifi_aroma_diffuser.yaml` file and removing the last `switch` entity (id 6) will also work.

hacshacdgacs:
Has anyone tried adding this recently? 
image
All my settings are correct and I've been unable to add this for the last few hours

PR #1145, opened 2023-10-05
Revert changes causing device  integration issues

craibo:
Revert changes from https://github.com/craibo/tuya-local/commit/370ddcfcb6aa580b4e5bff7c3c1936de5e1dc646 to Jiahong ET-72W thermostate.

These changes break the climate details reflecting  in Home Assistant.

Issues noticed:
1. If the device is manually updated these updates do not reflect on home assistant. eg. Device is on but HA shows it is off
2. Modifying the lock screen turns on the device incorrectly.

### DIFF
--- a/custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
+++ b/custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
@@ -37,23 +37,23 @@ primary_entity:
               value: auto
             - dps_val: false
               value_redirect: power
-            - dps_val: "Manual"
-              icon: "mdi:fire"
-              constraint: power
-              conditions:
-                - dps_val: true
-                  value: heat
-                - dps_val: false
-                  value_redirect: power
-                  value: "off"
-            - dps_val: "Anti_frozen"
-              icon: "mdi:snowflake"
-              constraint: power
-              conditions:
-                - dps_val: true
-                  value: cool
-                - dps_val: false
-                  value_redirect: power
+        - dps_val: "Manual"
+          icon: "mdi:fire"
+          constraint: power
+          conditions:
+            - dps_val: true
+              value: heat
+            - dps_val: false
+              value_redirect: power
+              value: "off"
+        - dps_val: "Anti_frozen"
+          icon: "mdi:snowflake"
+          constraint: power
+          conditions:
+            - dps_val: true
+              value: cool
+            - dps_val: false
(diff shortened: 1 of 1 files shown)


craibo:
Hi @make-all 
How can I ignore these 2 generic tests for this device?
This device does not function in the way that the PR which added these assumes.

Thank you

```
async def async_turn_on(self):
        """Turn on the climate device."""
        # Bypass the usual dps mapping to switch the power dp directly
        # this way the hvac_mode will be kept when toggling off and on.
        if self._hvac_mode_dps and self._hvac_mode_dps.type is bool:
            await self.device.async_set_property(self._hvac_mode_dps, True)
        else:
            await super.async_turn_on()

    async def async_turn_off(self):
        """Turn off the climate device."""
        # Bypass the usual dps mapping to switch the power dp directly
        # this way the hvac_mode will be kept when toggling off and on.
        if self._hvac_mode_dps and self._hvac_mode_dps.type is bool:
            await self.device.async_set_property(self._hvac_mode_dps, False)
        else:
            await super.async_turn_off()
```

make-all:
I can't tell what problem you are trying to fix by this change, and the failed test cases indicate that your change is breaking some functionality.

craibo:
I have 2 of these devices. The changes made by your commit here https://github.com/craibo/tuya-local/commit/370ddcfcb6aa580b4e5bff7c3c1936de5e1dc646 making HA remember the state when the device is powered off. However what is actually happening is that when you turn on the device form the devices touch screen, this doesn't reflect as turned on in Home Assistant. Changing the hvac_mode from the device also does not reflect in the Home Assistant.

I'm trying to revert the change just for this device, so that Home Assistant correctly reflects the current hvac state of the device.

Additionally I noticed another issue, disabling the lock screen turns on the device if its in an off state. This is not how the device functions when using it from the touch screen. Locking the device should leave it in its current state. This stops my kids messing with it.

craibo:
Hi @make-all 
Looks like it was my formatting breaking the tests. Apologies.
Resolved and the device is now workign as expected

make-all:
I'm going to reject this PR, because the tests show that correct info is being returned for the current hvac_mode, so the description of the problem is not explained by the changes you have undone.

If the problem is still occuring after reverting back to the released config, then look in the log for clues as to why.  I suspect that the cause is something else which caused the communication to stop working, not the specific mappings you have undone here.

PR #1136, opened 2023-10-03
device: add zigbee smart bulb and new temphumid sensor

felipejfc:
add a new zigbee lightbulb and a new temphumid sensor

light: https://pt.aliexpress.com/item/1005005834314406.html?spm=a2g0o.order_list.order_list_main.63.365bcaa4bWJTwl&gatewayAdapt=glo2bra

sensor: https://pt.aliexpress.com/item/1005004944375599.html?spm=a2g0o.order_list.order_list_main.27.365bcaa4bWJTwl&gatewayAdapt=glo2bra

Added another one, gu10 moes light:

https://pt.aliexpress.com/item/1005004871251106.html?spm=a2g0o.order_list.order_list_main.47.21efcaa41oZlWF&gatewayAdapt=glo2bra

PR #1146, opened 2023-10-05
Feature/add suport for zigbee tenky temphumid sensor

romicaiarca:
Is the same with zigbee_zb10wt_temphumid_sensor, only that the scale for battery is 1 instead of 10.


make-all:
I think this is the same as the sensor in PR #1136

PR #1147, opened 2023-10-06
Add Support for Parkside Inverter Mikrowechselrichter PMW 300

gtawelt:


custom_components/tuya_local/devices/solar_inverter2.yaml
```diff
@@ -0,0 +1,96 @@
+name: LIDL Inverter
```
make-all, line 1: Device and entity naming should follow the HA naming convention of only capitalising the first word (abbreviations and proper names or trademarks are exceptions).

make-all:
The file was renamed, so you will need to remove and readd the device after upgrading.

PR #1151, opened 2023-10-07
expand ZG205W configuration

blakadder:
I have a ZG/205W/A model of the sensor which uses the same model and dp's as the existing configuration despite not having any alarm/siren capabilities.

 

```
2023-10-07 15:59:28.905 DEBUG (MainThread) [custom_components.tuya_local.device] ZG205 received {"1": "none", "2": 8, "3": 1000, "4": 600, "101": 989, "102": 8, "103": true, "104": "disarm", "105": 600, "106": 2, "107": "high", "108": 7, "109": 2, "110": 0, "111": 0, "112": 0, "113": 0, "114": false, "115": true, "116": "none", "117": false, "118": false, "119": 1, "120": false, "full_poll": true}
```

I discovered a few new dp's so I included all of them for completeness and changes the "Detection state" sensor name and values to better reflect their function.

The dp's marked diagnostic are all undocumented by the manufacturer, my assumption is that it does some sort of self test to adapt to room conditions but its all just speculation.

### DIFF
--- a/custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
+++ b/custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
@@ -2,6 +2,9 @@ name: mmWave presence sensor
 products:
   - id: qbapy7r4deyp6zbx
     name: ZG-205W
+  - id: uykwakpl6rbxvjdd
+    name: ZG-205W/A
+
 primary_entity:
   entity: binary_sensor
   class: occupancy
@@ -14,42 +17,15 @@ primary_entity:
           value: true
         - dps_val: none
           value: false
+    # - id: 111
+    #   name: breathingselftest
 secondary_entities:
-  - entity: light
-    name: Led indicator
-    category: config
-    dps:
-      - id: 103
-        type: boolean
-        name: switch
   - entity: number
     name: Motion sensitivity
     category: config
     icon: mdi:motion-sensor
     dps:
       - id: 2
-        type: integer
-        name: value
-        range:
-          min: 0
-          max: 10
-  - entity: number
-    name: Stationary sensitivity
-    category: config
-    icon: mdi:motion-sensor
-    dps:
-      - id: 102
-        type: integer
-        name: value
-        range:
-          min: 0
-          max: 10
-  - entity: number
-    name: Small motion sensitivity
-    category: config
-    icon: mdi:motion-sensor
-    dps:
-      - id: 106
         type: integer
         name: value
         range:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
```diff
@@ -177,11 +171,133 @@ secondary_entities:
...
+      - id: 113
         type: integer
+        name: value
+        unit: s
+        range:
+          min: 0
+          max: 300
+  - entity: switch
+    name: Adaptive test duration
+    category: diagnostic
```
make-all, line 237: switch should be config, not diagnostic

custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
```diff
@@ -177,11 +171,133 @@ secondary_entities:
...
+        unit: m
+        range:
+          min: 0
+          max: 1000
+        mapping:
+          - scale: 100
+            step: 10
+  - entity: number
+    name: Adaptive testing time
+    category: diagnostic
```
make-all, line 224: number should be config, not diagnostic

custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
```diff
@@ -177,11 +171,133 @@ secondary_entities:
...
+    name: Adaptive test duration
+    category: diagnostic
+    icon: mdi:test-tube
+    dps:
+      - id: 114
+        name: switch
+        type: boolean
+  - entity: button
+    name: Factory reset
+    category: diagnostic
```
make-all, line 245: Button should be config, not diagnostic

custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
```diff
@@ -177,11 +171,133 @@ secondary_entities:
...
+            value: Large
+          - dps_val: small_move
+            value: Small
+          - dps_val: breathe
+            value: Breathe
+          - dps_val: none
+            value: None
+  - entity: switch
+    name: Small motion self-test
+    category: diagnostic
```
make-all, line 271: Switch should be config, not diagnostic. Would this be better as a button? (ie, when the test is finished, does this automatically return to false?)

custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
```diff
@@ -177,11 +171,133 @@ secondary_entities:
...
+    icon: "mdi:test-tube"
+    dps:
+      - id: 117
+        name: switch
+        type: boolean
+  - entity: switch
+    name: Breathing self-test
+    category: diagnostic
```
make-all, line 279: Switch should be config, not diagnostic. Would this be better as a button? (ie, when the test is finished, does this automatically return to false?)

make-all (commented):
None of the added dps seem to be optional.  Will this still match the earlier device?

blakadder:
> None of the added dps seem to be optional. Will this still match the earlier device?

They don't when you look at the dp's reported but those features don't exist in the Smart Life app options, but neither do the alarm options for my device. 
I'm not at all sure they will match the existing device but they use the same model number with different product id. Knowing Tuya it could go either way: be identical or two different devices.

I've marked the features diagnostic since they're called self-tests in Chinese which would be diagnostics but in the end its not quite clear what exactly they do because there's no documentation and i can't observe any visible changes when using them

custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
blakadder: the switch is active during the test and the adaptive testing time sensor counts down during it

make-all:
Home Assistant only uses "diagnostic" for read-only sensors. https://developers.home-assistant.io/blog/2021/10/26/config-entity/#entity-categories

PR #1163, opened 2023-10-10
Added 'ZC34T-03-3A_swing_arm_opener.yaml'

Levanterman:
Added 'ZC34T-03-3A_swing_arm_opener.yaml'
For the Ali Express Tuya Window Opener (Kit 4 in URL Below) :

https://www.aliexpress.com/item/1005002548271633.html?spm=a2g0o.order_list.order_list_main.76.24631802UmaT1q

### DIFF
--- a/custom_components/tuya_local/devices/ZC34T-03-3A_swing_arm_opener.yaml
+++ b/custom_components/tuya_local/devices/ZC34T-03-3A_swing_arm_opener.yaml
@@ -1,12 +1,10 @@
-name: ZC34T-03-3A Swing Arm Opener
+name: Window opener
 products:
   - id: ozttm5n6okadq7tq
-    name: ZC34T-03-3A Swing Arm Opener
-    model: ZC34T-03-3A
+    name: ZC34T-03-3A swing arm opener
 primary_entity:
   entity: cover
   class: window
-  name: Control
   dps:
     - id: 101
       type: string
@@ -18,26 +16,23 @@ primary_entity:
           value: close
         - dps_val: stop
           value: stop
+    - id: 105
+      type: string
+      name: fault
 secondary_entities:
-  - entity: switch
-    name: Lock
-    class: enum
+  - entity: lock
     dps:
       - id: 102
         type: string
-        name: switch
+        name: lock
         optional: true
         mapping:
           - dps_val: lock
-            value: lock
-            default: true
-            icon: "mdi:lock"
+            value: true
           - dps_val: unlock
-            value: unlock
-            icon: "mdi:lock-open-variant"
-  - entity: sensor
-    name: Window Current Status
-    class: enum
+            value: false
+  - entity: binary_sensor
+    class: window
     dps:
       - id: 1
         type: string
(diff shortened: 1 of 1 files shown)


Levanterman:
I am not sure why its failing with the enum error in the Python Test above, in the dps_mappings they are all listed as enum. Any ideas please? thanks

make-all:
Since a few months ago, HA requires all sensors that report a fixed list of values to be of device class "enum".  You need to either add `class: enum` to the entity definition, or remove the mapping so HA does not see a list of values.

Levanterman:
Thanks @make-all that worked :-) 
This PULL is still not finished though. 
The DPS says the window status (i.e. if its Open or Closed) says it comes through on DPS 1, but it never seems to send "1", although the SmartLife App knows the status. I have it set to optional cos it doesnt seem to send it. 
So I need to find out why, although its all working now, it doesnt display the status of the window.

Levanterman:
image

At the top of this log the window is Open, and I click Close in HA.
You can see the Close being processed, but still no DP of 1 (status) arrives , but it toggles 107 from Off to False.

Then I click OPEN, still no status on DP1, but it does toggle 107 from On to True
image

107 is defined as 'Window switch message Push Messages'
but for example if i click STOP half way through a Window Close, then 107 doesnt know the window never closed. 

Not sure how to tackle this one.

make-all:
So 107 seems to be your open/close status.  Maybe it can be cross checked against 101 to determine opening/closing as well.
HA doesn't really have any way to handle stopping in the middle for cover devices that do not report current position.  The only states it recognizes are open, closed, opening and closing, though you can also report None if it is indeterminate.

make-all:
There is also a question of whether the Rain and wind detection is a switch or a binary sensor.  Sometimes manufacturers configure their device with rw for sensors that should actually be ro, so it isn't always obvious from the developer portal docs for the device.

Either this is a switch to turn the feature on and off, in which case since there isn't a separate sensor to feedback the status, I presume the window automatically closes when rain or wind is detected. Or it is actually a sensor that returns whether wind or rain has been detected, and alerts you to close it manually (or through an automation).

Levanterman:
> There is also a question of whether the Rain and wind detection is a switch or a binary sensor. Sometimes manufacturers configure their device with rw for sensors that should actually be ro, so it isn't always obvious from the developer portal docs for the device.
> 
> Either this is a switch to turn the feature on and off, in which case since there isn't a separate sensor to feedback the status, I presume the window automatically closes when rain or wind is detected. Or it is actually a sensor that returns whether wind or rain has been detected, and alerts you to close it manually (or through an automation).

There is the option on the window controller switch to add a rain or wind sensor but i dont need them so I didnt buy them, but i noticed the status of the wind/rain is reported as unknown.  So I wired the sensor ground to ground, so it wasnt a floating logic . Now the switch reports the weather as 'fine' (i.e. not raining) perfect. 

image

For the window current position, thats on going.

Levanterman:
Got enums on all the entities except the primary, but its back to erroring on the validation check, not sure why.

PR #1167, opened 2023-10-14
Add support for Reiga 52 Ceiling fan with Light

kbirger:
Tested each setting myself. 

DP Status from tinytuya:
{'dps': {'1': True, '2': 'normal', '3': 2, '8': 'reverse', '15': False, '16': 60, '17': 32, '22': 'off'}}

DP Configuration:
I had to piece this together manually because my api trial ended. I got these from debugging the normal Tuya integration and from using tinytuya. 

I manually tested each of the values for each of the settings

```
code='mode',
  type='Enum',
  values='{"range":["normal", "sleep", "nature"]}'

# This is actually wrong. They are numeric values, not strings.
code='fan_speed',
 type='Enum',
 values='{"range":["1", "2", "3", "4", "5", "6"]}

code='fan_direction',
  type='Enum',
   values='{"range":["forward", "reverse"]}'

code='light',
 type='Boolean',
 values='{}'

# also wrong. the minimum it goes to is 2
code='bright_value',
 type='Integer',
 values='{"min":0, "max":100, "scale":0, "step":1}

code='temp_value',
 type='Integer',
 values='{"min":0, "max":100, "scale":0, "step":1}')

# wrong again. these values don't work, but values like '1hour' work. 'cancel' is wrong. the actual value I get from tinytuya is 'off'.
code='countdown_set',
 type='Enum',
 values='{"range":["cancel", "1h", "2h", "4h", "8h"]

code='mode',
 type='Enum',
 values='{"range":["normal", "sleep", "nature"
```

### DIFF
--- a/custom_components/tuya_local/devices/reiga_52_fan_light.yaml
+++ b/custom_components/tuya_local/devices/reiga_52_fan_light.yaml
@@ -1,7 +1,7 @@
-name: Reiga Ceiling Fan
+name: Ceiling fan with light
 products:
   - id: g0ewlb1vmwqljzji
-    name: Reiga 52 Ceiling Fan with Light
+    name: Reiga 52 ceiling fan with light
 primary_entity:
   entity: fan
   translation_key: fan_with_presets
@@ -61,14 +61,15 @@ secondary_entities:
       - id: 16
         type: integer
         name: brightness
-
+        range:
+          min: 0
+          max: 255
         mapping:
           - scale: 0.392156862745098
       - id: 17
         type: integer
         name: color_temp
         mapping:
-          - step: 1
           - target_range:
               min: 2700
               max: 6500
--- a/custom_components/tuya_local/devices/reiga_52_fan_light.yaml
+++ b/custom_components/tuya_local/devices/reiga_52_fan_light.yaml
@@ -40,12 +40,14 @@ secondary_entities:
         type: string
         name: option
         mapping:
-          - dps_val: cancel
+          - dps_val: "off"
             value: "Off"
           - dps_val: "1hour"
             value: "1 hour"
-          - dps_val: "3hour"
-            value: "3 hours"
+          - dps_val: "2hour"
+            value: "2 hours"
+          - dps_val: "4hour"
+            value: "4 hours"
           - dps_val: "8hour"
             value: "8 hours"
   - entity: light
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/reiga_52_fan_light.yaml
```diff
@@ -0,0 +1,77 @@
...
+        type: integer
+        name: brightness
+
+        mapping:
+          - scale: 0.392156862745098
+      - id: 17
+        type: integer
+        name: color_temp
+        mapping:
+          - step: 1
```
make-all, line 71: Remove the - step: 1 line, as it will interfere with the following line as  it is a separate entry in the list and neither specifies a dps_val, so both are competing to be the default mapping, but anyway 1 is the default step so it is not needed.

custom_components/tuya_local/devices/reiga_52_fan_light.yaml
```diff
@@ -0,0 +1,77 @@
+name: Reiga Ceiling Fan
```
make-all, line 1: HA naming convention: Only capitalise the first word of a device or entity name, unless it is an abbreviation or proper name that needs capitalising even in the middle of sentences.  I also prefer to leave brand names off the top level name if there is specific product info listed, as there are often other brands sold in different markets, or different retailers in the same market that are the same product rebranded.

custom_components/tuya_local/devices/reiga_52_fan_light.yaml
```diff
@@ -0,0 +1,77 @@
...
+    category: config
+    dps:
+      - id: 15
+        type: boolean
+        name: switch
+      - id: 16
+        type: integer
+        name: brightness
+
+        mapping:
```
make-all, line 65: I'm not sure if not having a range here will affect anything.  I have always included it in the past for any numeric value that can be set, but since light brightness has a fixed range of 0-255 in HA, maybe it is OK to leave off.  Keep watching for errors in the log I guess.

kbirger: I added it to be safe. Seems not to have broken anything. Thanks for the tip.

custom_components/tuya_local/devices/reiga_52_fan_light.yaml
kbirger: Thanks. removed

PR #1171, opened 2023-10-15
Add popotan fairy lights device configuration

Fredzxda:


Fredzxda:
The différence are in the scene values.

make-all:
It shouldn't be an issue to add additional scenes to the existing config.  The format of the scene definitions seems to be the same, and many are common between the two devices.

PR #1172, opened 2023-10-15
Fixed fan levels for wilfa_haze_hu400bc_humidifier

hsand:
The humidifier supports 5 levels, but the current config only supports up to level 4. I added "level_5" and adjusted the percentages accordingly.

I tested this with my local installation and it seems to work well.

### DIFF
--- a/custom_components/tuya_local/devices/wilfa_haze_hu400bc_humidifier.yaml
+++ b/custom_components/tuya_local/devices/wilfa_haze_hu400bc_humidifier.yaml
@@ -58,14 +58,16 @@ secondary_entities:
         name: speed
         mapping:
           - dps_val: level_0
-            value: 20
+            value: 0
           - dps_val: level_1
-            value: 40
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/wilfa_haze_hu400bc_humidifier.yaml
```diff
@@ -58,14 +58,16 @@ secondary_entities:
         name: speed
         mapping:
           - dps_val: level_0
-            value: 20
+            value: 0
```
make-all, line 61: A speed of zero results in the fan being turned off by the switch.  I'm not sure that is correct behaviour for a setting of level_0.

make-all:
As there are actually 5 levels in the original config, your change makes 6.  Since there is no reply to my previous comment about mapping level_0 to 0 (effectively off), I cannot proceed with this PR.

PR #1178, opened 2023-10-16
GiotoHun Pet Feeder

lnx85:
Add Support for GiotoHun Pet Feeder PE-A03

### DIFF
--- a/custom_components/tuya_local/devices/giotohun_pet_feeder.yaml
+++ b/custom_components/tuya_local/devices/giotohun_pet_feeder.yaml
@@ -13,7 +13,6 @@ primary_entity:
       range:
         min: 1
         max: 12
-      optional: false
     - id: 1
       name: meal_plan
       type: string
@@ -27,7 +26,6 @@ secondary_entities:
       - id: 15
         name: sensor
         type: integer
-        optional: false
   - entity: button
     icon: "mdi:reload-alert"
     class: restart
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/giotohun_pet_feeder.yaml
```diff
@@ -0,0 +1,81 @@
...
+  name: Feed
+  icon: "mdi:bowl"
+  dps:
+    - id: 3
+      name: value
+      type: integer
+      range:
+        min: 1
+        max: 12
+      optional: false
```
make-all, line 16: Leave out the "optional" when it is false.

make-all (commented):
Is this significantly different than the Catit Pixi feeder, which has basically the same dps with an additional "Quick feed" button?

lnx85:
> Is this significantly different than the Catit Pixi feeder, which has basically the same dps with an additional "Quick feed" button?

In my pet feeder there is a DPS 4 for feeder status (feeding / standby); in HA that DPS is converted into a binary_sensor that become ON when the feeder is releasing the food. It is useful to check if the food has been succesfull released.

lnx85:
> > Is this significantly different than the Catit Pixi feeder, which has basically the same dps with an additional "Quick feed" button?
> 
> In my pet feeder there is a DPS 4 for feeder status (feeding / standby); in HA that DPS is converted into a binary_sensor that becomes ON when the feeder is releasing the food. It is useful to check if the food has been succesfull released.

You can probably add the same DPS in the Catit Pixi feeder with optional: true instead of using another config file.

make-all:
Merged with Catit Pixi Smart feeder, as mentioned in the comment above - added optional Feeding sensor.

PR #1174, opened 2023-10-16
Create configuration for Novostella RGB String Light

scottfridwin:
Configuration taken from the following tinytuya device result:

```{
        "name": "String Light",
        "id": "{removed}",
        "key": "{removed}",
        "mac": "{removed}",
        "uuid": "{removed}",
        "category": "dd",
        "product_name": "String Light",
        "product_id": "q5dhxp3mjrbmr5pi",
        "biz_type": 18,
        "model": "magic-no-ir-StringLight-fc",
        "sub": false,
        "icon": "https://images.tuyaus.com/smart/icon/bay1585732422040GgH0/add9a7a9ed5ab116d430c78b040a3000.jpg",
        "mapping": {
            "20": {
                "code": "switch_led",
                "type": "Boolean",
                "values": {}
            },
            "21": {
                "code": "work_mode",
                "type": "Enum",
                "values": {
                    "range": [
                        "white",
                        "colour",
                        "scene",
                        "music"
                    ]
                }
            },
            "24": {
                "code": "colour_data",
                "type": "Json",
                "raw_values": "{\"h\":{\"min\":0,\"scale\":0,\"unit\":\"\",\"max\":360,\"step\":1},\"s\":{\"min\":0,\"scale\":0,\"unit\":\"\",\"max\":1000,\"step\":1},\"v\":{\"min\":0,\"scale\":0,\"unit\":\"\",\"max\":1000,\"step\":1}}",
                "values": {
                    "h": {
                        "min": 0,
                        "scale": 0,
                        "unit": "",
                        "max": 360,
                        "step": 1
                    },
                    "s": {
                        "min": 0,
                        "scale": 0,
                        "unit": "",
                        "max": 1000,
                        "step": 1
                    },
                    "v": {
                        "min": 0,
                        "scale": 0,
                        "unit": "",
                        "max": 1000,
                        "step": 1
                    }
                }
            },
            "26": {
                "code": "countdown",
                "type": "Integer",
                "values": {
                    "unit": "s",
                    "min": 0,
                    "max": 86400,
                    "scale": 0,
                    "step": 1
                }
            },
            "27": {
                "code": "music_data",
                "type": "Json",
[...]

PR #1180, opened 2023-10-17
Add Novostella RGB string light to existing configuration file.

scottfridwin:


scottfridwin:
Replacement for #1174

PR #1186, opened 2023-10-17
Add Support for Ecostrad iQ Heating Element

jacobpennington821:
Support for the following: https://ecostrad.com/ecostrad-iq-wifi-heating-element/

### DIFF
--- a/custom_components/tuya_local/devices/ecostrad_iq_heating_element.yaml
+++ b/custom_components/tuya_local/devices/ecostrad_iq_heating_element.yaml
@@ -1,6 +1,7 @@
 name: Ecostrad iQ Heating Element
 primary_entity:
   entity: climate
+  translation_key: thermostat
   icon: "mdi:radiator"
   dps:
     - id: 1
@@ -16,17 +17,13 @@ primary_entity:
       name: preset_mode
       mapping:
         - dps_val: "auto"
-          value: Program
-          icon: "mdi:clock-outline"
+          value: program
         - dps_val: "eco"
-          value: Eco
-          icon: "mdi:moon-waxing-crescent"
+          value: eco
         - dps_val: "hot"
-          value: Comfort
-          icon: "mdi:weather-sunny"
+          value: comfort
         - dps_val: "only_inside"
-          value: Radiator
-          icon: "mdi:radiator"
+          value: boost
     - id: 16
       type: integer
       name: temperature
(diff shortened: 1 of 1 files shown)


make-all:
Schedule is not easily decoded and reencoded, so just extract it into string attributes of the main entity, so it is available for further processing if you want.  Making it into something meaningful is a whole project in itself, involving both decoding/encoding work on the integration side and UI work in a custom card.

custom_components/tuya_local/devices/ecostrad_iq_heating_element.yaml
```diff
@@ -0,0 +1,116 @@
...
+        - dps_val: true
+          value: "heat"
+        - dps_val: false
+          value: "off"
+    - id: 2
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: "auto"
+          value: Program
```
make-all, line 19: Where possible use standard preset names (none, eco, away, boost, comfort, home, sleep, activity) and extended ones defined in one of the climate classes in translations/en.json (manual, program, temp_override, perm_override), (cool, fan_only, heat, floor_cool, floor_heat) or (smart_heat, quick_heat, quiet_heat, smart_cool. quick_cool, quiet_cool, auto).

Note these are case sensitive, so should be in all caps, and if you use any extended ones, you need to set the class of the entity appropriately.

custom_components/tuya_local/devices/ecostrad_iq_heating_element.yaml
```diff
@@ -0,0 +1,116 @@
...
+        - scale: 10
+          step: 5
+    - id: 24
+      type: integer
+      name: current_temperature
+      mapping:
+        - scale: 10
+secondary_entities:
+  - entity: number
+    name: Radiator Mode Temperature
```
make-all, line 46: If this is only valid in Radiator mode, then it may be better inside the climate entity, and linked to the temperature dp by:
```
      mapping:
         - scale: 10
           step: 5
           constraint: preset_mode
           conditions:
             - dps_val: only_inside
               value_redirect: radiator_temperature

File: custom_components/tuya_local/devices/ecostrad_iq_heating_element.yaml [hunk 3]
```diff
@@ -0,0 +1,116 @@
...
+      - id: 57
+        type: integer
+        name: value
+        unit: C
+        range:
+          min: 300
+          max: 700
+        mapping:
+          - scale: 10
+            step: 50
```
make-all, line 58: Is step 50 correct? Not 5?

custom_components/tuya_local/devices/ecostrad_iq_heating_element.yaml
```diff
@@ -0,0 +1,116 @@
...
+            value: "90 mins"
+            icon: "mdi:window-open"
+  - entity: number
+    name: Boost time
+    category: config
+    dps:
+      - id: 111
+        type: integer
+        name: value
+        unit: hours
```
make-all, line 95: unit: h

https://github.com/home-assistant/core/blob/dev/homeassistant/const.py#L612

custom_components/tuya_local/devices/ecostrad_iq_heating_element.yaml
```diff
@@ -0,0 +1,116 @@
...
+          min: 0
+          max: 7
+  - entity: number
+    name: Calibration offset
+    category: config
+    dps:
+      - id: 27
+        type: integer
+        name: value
+        unit: C
```
make-all, line 106: Leave the units off these relative temperature settings, or make them general degrees (using the Unicode character for degree sign) because they do unexpected things if Home Assistant decides to apply unit conversion.

custom_components/tuya_local/devices/ecostrad_iq_heating_element.yaml
```diff
@@ -0,0 +1,116 @@
...
+        range:
+          min: -5
+          max: 5
+  - entity: sensor
```
make-all, line 110: This is not interesting enough to make it a sensor, instead list the dp under the main climate entity with a name like `schedule`.

jacobpennington821:
I've made some changes to the config, hopefully they address your comments 👍

PR #1188, opened 2023-10-18
new devices

erelke:
A couple of new devices. There are already similar ones, but the devices don't work with them.


make-all:
One of the commits says "hungarian translation".  Did you make a modification to the Hungarian translations which you forgot to stage when committing?

make-all:
I also suggest breaking these into separate pull requests.  Some can probably be merged quite quickly, others may have some work to do, or even be better dealt with by modifying existing configs. Submitting all together like this delays them all until the most time consuming one has finally been dealt with.

custom_components/tuya_local/devices/earu_smart_wifi_circuit_breaker_with_energy_monitor.yaml
```diff
@@ -0,0 +1,364 @@
+products:
+  - id: ine4hhapkgxonij1
+    name: EARU smart breaker
+    # https://www.aliexpress.com/item/1005004894566311.html
+
```
make-all, line 5: The config still needs a top level name, I was suggesting a generic name like "Smart breaker" for the top level name, and move the brand and model number to the products name.

But I also noted that this config does not appear to be any different than `smartmcb_smt006_energymeterv2.yaml` so probably a better solution is to add the product id and name to that config, and delete this one.

custom_components/tuya_local/devices/switch_module_rdcbc_t_j_z_j.yaml
```diff
@@ -0,0 +1,150 @@
+name: Energy monitoring RDCBC
```
make-all, line 1: This config also seems like it should match one of the generic smartplug configs already.

custom_components/tuya_local/devices/earu_smart_wifi_circuit_breaker_with_energy_monitor.yaml
```diff
@@ -0,0 +1,366 @@
...
+        type: integer
+        readonly: true
+        persist: true
+        unit: kWh
+        class: total_increasing
+        name: sensor
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: 'Breaker ID'
```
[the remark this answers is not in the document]
erelke: How can add as attribute?

custom_components/tuya_local/devices/earu_smart_wifi_circuit_breaker_with_energy_monitor.yaml
erelke: the device does not work with the smartmcb_smt006_energymeterv2.yaml config

custom_components/tuya_local/devices/earu_smart_wifi_circuit_breaker_with_energy_monitor.yaml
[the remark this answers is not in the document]
make-all: Actually it already is `breaker_number` above

custom_components/tuya_local/devices/earu_smart_wifi_circuit_breaker_with_energy_monitor.yaml
make-all: Based on the dps mappings, it should, so you'll have to be more specific what you mean by "does not work".

custom_components/tuya_local/devices/simple_contact_sensor.yaml
```diff
@@ -0,0 +1,18 @@
+name: Simple (door/window) sensor
+products:
+  - id: 0mjj67aaztcfmc3i
+    name: AW201_XR3
+primary_entity:
+  entity: cover
```
make-all, line 6: If it is just a sensor, then it should be implemented as a binary_sensor.
cover is for control of curtains/doors etc.

make-all:
Manually merged to resolve conflicts and exclude duplicate config with generic smartplugv2_energyv2

PR #1193, opened 2023-10-19
Add support for ZY-M201 aka ZY-M100-24G

blakadder:
Adds support for new version of ZY-M100 with 24Ghz radar. Box labelled as M100, Tuya shows device as "ZY-M201-WIFI24G存在感应器"

LMC_12 8 4 pq20231019_112850_PQ F3-LMCR13-5 F

### DIFF
--- a/custom_components/tuya_local/devices/zym201_presence_sensor.yaml
+++ b/custom_components/tuya_local/devices/zym201_presence_sensor.yaml
@@ -4,7 +4,7 @@ products:
     name: ZY-M201-WIFI24G
 primary_entity:
   entity: binary_sensor
-  class: occupancy
+  class: presence
   dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/zym201_presence_sensor.yaml
```diff
@@ -4,7 +4,7 @@ products:
     name: ZY-M201-WIFI24G
 primary_entity:
   entity: binary_sensor
-  class: occupancy
+  class: presence
```
make-all, line 7: The way I understand the difference between these two classes is that occupancy is tied to rooms, and presence is tied to people.  I think occupancy is correct in that case, as we cannot identify an individual who is present.

PR #1201, opened 2023-10-20
Add support for Pro Breeze 30L Dehumidifier

dmitriy5181:
This one: https://probreeze.com/products/30l-high-capacity-dehumidifier-with-smart-app-control

### DIFF
--- a/custom_components/tuya_local/devices/probreeze_pb20.yaml
+++ b/custom_components/tuya_local/devices/probreeze_pb20.yaml
@@ -12,24 +12,25 @@ primary_entity:
       mapping:
         - dps_val: false
           icon: "mdi:air-humidifier-off"
-          icon_priority: 1
         - dps_val: true
           icon: "mdi:air-humidifier"
-          icon_priority: 4
     - id: 2
-      name: mode
+      name: work_mode
       type: string
+      hidden: true
+    - id: 101
+      type: boolean
+      name: mode
       mapping:
-        - constraint: inside_drying
+        - dps_val: false
+          constraint: work_mode
           conditions:
-            - dps_val: false
-              mapping:
-                - dps_val: "1"
-                  value: boost
-                - dps_val: "0"
-                  value: auto
-            - dps_val: true
-              value: laundry
+            - dps_val: "1"
+              value: boost
+            - dps_val: "0"
+              value: auto
+        - dps_val: true
+          value: Internal drying
     - id: 3
       name: current_humidity
       type: integer
@@ -49,10 +50,6 @@ primary_entity:
           value: OK
         - dps_val: 8
           value: tank
-    - id: 101
-      type: boolean
-      name: inside_drying
-      hidden: true
 secondary_entities:
   - entity: fan
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/probreeze_pb20.yaml
```diff
@@ -0,0 +1,165 @@
+name: Pro Breeze 30L Dehumidifier
```
make-all, line 1: Just use a generic name here, in case other brands match.  The products section has the branding details (and one day will be used for matching products)

custom_components/tuya_local/devices/probreeze_pb20.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - id: 7
+        type: boolean
+        name: lock
+        mapping:
+          - dps_val: false
+            icon: "mdi:hand-back-right"
+          - dps_val: true
+            icon: "mdi:hand-back-right-off"
+  - entity: select
+    name: Countdown
```
make-all, line 77: Other devices use "Timer", it is better to be consistent with them.

custom_components/tuya_local/devices/probreeze_pb20.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - id: 11
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 8
+            value: true
+            icon: "mdi:cup-water"
+            icon-priority: 1
+          - value: false
+  - entity: switch
```
make-all, line 159: How does this interact with the presets? Does it override and always use continuous?  Maybe it is better to merge this with the presets if it does using `constraint` and `conditions`.

dmitriy5181: Then it makes more sense to merge it with "on - off". So that it will be "off - on - int. drying" Because physical button "internal drying" turns the device on and modes "continuous" and "auto" are not working when drying is active.

make-all: You can't add additional options to off/on.  You can add presets.

  The following also takes advantage of some translations that Home Assistant provides.

```
    - id: 2
      name: mode
      type: string
      mapping:
        - constraint: inside_drying
          conditions:
            - dps_val: false
              mapping:
                - dps_val: "1"
                  value: boost
                - dps_val: "0"
                  value: auto
            - dps_val: true
              value: laundry
```

dmitriy5181: Ok, understood

make-all: It should show as a preset.

dmitriy5181: I see only "boost" and "auto":

image

make-all: I need to take a look at how the code handles these multiply nested mappings.  Maybe it will work better to make 101 the mode, and 2 something else like work_mode, to bring things down a level.
```
    - id: 2
      name: work_mode
      type: string
      hidden: true
    - id: 101
      type: boolean
      name: mode
      mapping:
        - dps_val: false
          constraint: work_mode
          conditions:
            - dps_val: "1"
              value: boost
            - dps_val: "0"
              value: auto
        - dps_val: true
          value: laundry
```

dmitriy5181: This works, but "laundry" or "Dry clothes" is not applicable for it. This mode is to make the internal parts of the device dry.

image

PR #1207, opened 2023-10-22
add Greenmigo ALPHA-Q25 Pro WiFi Dehumidifier

Privatecoder:


custom_components/tuya_local/translations/en.json
```diff
@@ -156,7 +156,10 @@
                     "mode": {
                         "state": {
                             "laundry": "Dry clothes",
-                            "purify": "Purify"
+                            "purify": "Purify",
```
make-all, line 159: You should not need these changes.  The Home Assistant humidifier platform already defines auto, manual and sleep, and overriding them here will prevent the translations provided by Home Assistant from being found.

Privatecoder: I cannot find "manual" [here](https://developers.home-assistant.io/docs/core/entity/humidifier/#modes) – should I leave it?

make-all: I think I used "normal" in other configs with the same options available.

Privatecoder: Ok, I'll change that then

custom_components/tuya_local/devices/greenmigo_alpha_q25_pro_wifi.yaml
```diff
@@ -0,0 +1,207 @@
+name: Greenmigo Alpha Q25 Pro WiFi Dehumidifier
```
make-all, line 1: If possible to get the product id, it would be better to move this to the name under `products:` and use a generic name (ie just Dehumidifier).  But also, Home Assistant naming convention calls for non-name related parts to be lower case if not the first word (I think it only affects "Dehumidifier" here, as the rest is part of the branding).

custom_components/tuya_local/translations/en.json
```diff
@@ -195,4 +195,4 @@
             }
         }
     }
-}
+}
```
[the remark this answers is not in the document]
Privatecoder: my vscode instance removes it on saving by default and [It seems as there is no new line in your en.json as well](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/en.json)

custom_components/tuya_local/devices/greenmigo_alpha_q25_pro_wifi.yaml
```diff
@@ -0,0 +1,207 @@
...
+          icon: "mdi:alert-circle-outline"
+          icon_priority: 1
+        - dps_val: 4
+          icon: "mdi:cup-water"
+          icon_priority: 1
+        - dps_val: 8
+          icon: "mdi:snowflake-melt"
+          icon_priority: 1
+secondary_entities:
+  - entity: sensor
```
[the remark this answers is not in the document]
Privatecoder: will replace it

custom_components/tuya_local/devices/greenmigo_alpha_q25_pro_wifi.yaml
Privatecoder: alright

custom_components/tuya_local/translations/en.json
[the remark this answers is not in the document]
make-all: I suggest to revert this file. There is a new line in the version in git, otherwise it would not detect this as a change, and the location of the error indicator on the right hand side indicates that it is the new version that is incorrect.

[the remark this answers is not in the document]
Privatecoder: alright. off-topic: I don't understand icon-priority, can you point me to how it works?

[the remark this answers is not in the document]
make-all: A lower number is higher priority (ie priority number 1 overrides all others).  Where the icon is set in more than one location, and they might indicate different icons at the same time, the priority is needed to decide which one will apply.

But setting icons should be used sparingly, because it interferes with any icon changes that the user's chosen theme makes.  The icon_priority was introduced early in the development because it was a feature in the Goldair integration that this was originally based on, and predated Home Assistant themes.  Now I mostly only set icons at the entity level for switches and sensors which would benefit from a more descriptive icon for their function, and since it is only one location setting the icon, there is no need for a priority.

make-all:
I already rebased and merged this PR.

PR #1220, opened 2023-10-27
Update blitzwolf_bwsh5_humidifier.yaml

spuljko:
Fixed recognition problems.
Added problem and spray level

### DIFF
--- a/custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
+++ b/custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
@@ -1,6 +1,7 @@
-name: BlitzWolf BWSH5 humidifier
+name: Humidifier
 products: 
   - id: lqzme7o0kzhumdxr
+    name: BlitzWolf BWSH5
 primary_entity:
   entity: humidifier
   class: humidifier
@@ -82,30 +83,33 @@ secondary_entities:
             icon: "mdi:led-on"
           - dps_val: false
             icon: "mdi:led-off"
-  - entity: select
+  - entity: fan
     name: Spray level
     icon: "mdi:heat-wave"
     category: config
     dps:
+      - id: 1
+        name: switch
+        type: boolean
       - id: 23  
-        name: option  
+        name: speed 
         type: string
         optional: true
         mapping:
           - dps_val: "level_1"
-            value: "Level 1 "
+            value: 15
           - dps_val: "level_2"
-            value: "Level 2"
+            value: 30
           - dps_val: "level_3"
-            value: "Level 3"
+            value: 45
           - dps_val: "level_4"
-            value: "Level 4"
+            value: 55
           - dps_val: "level_5"    
-            value: "Level 5"           
+            value: 70
           - dps_val: "level_6"
-            value: "Level 6"
+            value: 85
           - dps_val: "level_7" 
-            value: "Level 7"
+            value: 100
   - entity: sensor
     class: temperature
     dps:
--- a/custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
+++ b/custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
@@ -29,7 +29,7 @@ primary_entity:
     - id: 13             
       name: humidity
       type: integer
-      optional: true
+      #optional: true
       range:
         min: 35
         max: 75
@@ -38,14 +38,6 @@ primary_entity:
     - id: 14
       name: current_humidity
       type: integer
-    - id: 22
-      name: unknown_22
-      type: integer
-      optional: true
-    - id: 23
-      name: unknown_23
-      type: string
-      optional: true
 secondary_entities:
   - entity: switch
     name: UV sterilization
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
```diff
@@ -29,7 +29,7 @@ primary_entity:
     - id: 13             
       name: humidity
       type: integer
-      optional: true
+      #optional: true
```
make-all, line 32: I think this was made optional based on previous logs that showed it missing.
Removing that now will affect compatibility, as the device appears to be inconsistent about which dps it reports (probably based on the mode it is in).

custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
```diff
@@ -153,10 +163,24 @@ secondary_entities:
...
+  - entity: binary_sensor
+    name: Low water
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 22
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
```
make-all, line 183: Leave the `dps_val` off here, so that any other value is caught by this as a default.
You should leave dp 22 on the main entity above, but rename it to fault_code, so the actual value can be read.  Probably another binary_sensor called Fault, which is false on 0 and 1, and otherwise true should be added to report other Faults (unless you know there aren't any other faults reported by the device - but usually there are at least 2 fault codes documented for dehumidifiers).

spuljko: I am not sure how this field is conveted, this is from specification:
```
{
	"abilityId": 22,
	"accessMode": "ro",
	"code": "fault",
	"description": "",
	"extensions": {
		"iconName": "icon-Panic",
		"attribute": "128"
	},
	"name": "故障告警",
	"typeSpec": {
		"label": [
			"lack_water"
		],
		"maxlen": 1,
		"type": "bitmap",
		"typeDefaultValue": 0
	}
},
```
Can "bitmap" type have any other value than 0 and 1?

make-all: "Bitmap" can have any number of bytes. The documentation says there is only one, so adding another catch-all sensor is not needed, but in case the documentation is wrong, or a later firmware upgrade adds additional fault detection and reporting, it is still better to just catch 1 for "Tank full/missing" and map everything else (whether no bits set or bits 2 and above) to false.

custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
```diff
@@ -84,17 +82,30 @@ secondary_entities:
             icon: "mdi:led-on"
           - dps_val: false
             icon: "mdi:led-off"
-  - entity: sensor
-    deprecated: humidifier
-    name: Current humidity
-    class: humidity
-    category: diagnostic
+  - entity: select
```
make-all, line 85: Normally I would represent spray level as a fan entity, as I think the UI is a bit better, and it also comes with richer voice control.  It means duplicating the switch between both fan and humidifier, as both require a switch, and converting the levels to percentages rather than strings.

custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
```diff
@@ -1,4 +1,3 @@
-# sample DPS {'1': True, '4': 'health', '5': True, '10': 24, '12': 75, '13': 55, '14': 77, '18': 'c', '19': 'cancel', '20': 0, '21': True, '22': 0, '23': 'level_1', '26': True, '35': True} 
 name: BlitzWolf BWSH5 humidifier
```
make-all, line 1: Move this name under products, and add a more generic name here (just Humidifier for example)

spuljko: How will people know the difference if there is few products that are recognized as humidifier if all humidifiers has same name?

custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
spuljko: I think it looks very ugly and not user friendly. Thera are 7 levels, so step between levels is 14.28571428571429, and converted to fan is 14.28%, 28.57% etc. How will user set level to 3 via voice control that way? I know it will step to nearest value, but I doubt that anyone will even try to set it that way...

make-all: The levels do not need to be perfectly evenly spaced.  You can set them at 15%, 30%, 40%, ...

custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
make-all: If someone has multiple humidifiers, they will by default receive a name with a number suffix.  It is up to the user to then name the devices in a way that they will recognise them.

custom_components/tuya_local/devices/blitzwolf_bwsh5_humidifier.yaml
spuljko: No matter levels, this still looks ugly

PR #1229, opened 2023-10-30
Add support for Vacplus Dehumidifiers

bdf0506:
Used to be sold on Amazon, though not any longer. This is the product: https://supplyleader.com/product/Vacplus-50-Pints-3-000-Sq-Ft-Dehumidifier-with-WiFi-Remote-for-Medium-Spaces-and-Basements-Nickname-VA-D1901/B0816SXSG8

### DIFF
--- a/custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
@@ -1,4 +1,4 @@
-name: Vacplus Dehumidifier
+name: Dehumidifier
 products:
   - id: yttlroi7nr18vqbo
     name: Vacplus Dehumidifier
@@ -9,13 +9,6 @@ primary_entity:
     - id: 1
       name: switch
       type: boolean
-      mapping:
-        - dps_val: false
-          icon: "mdi:air-humidifier-off"
-          icon_priority: 1
-        - dps_val: true
-          icon: "mdi:air-humidifier"
-          icon_priority: 4
     - id: 2
       name: humidity
       type: integer
--- a/custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
@@ -1,0 +1,145 @@
+name: Vacplus Dehumidifier
+products:
+  - id: yttlroi7nr18vqbo
+    name: Vacplus Dehumidifier
+primary_entity:
+  entity: humidifier
+  class: dehumidifier
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+      mapping:
+        - dps_val: false
+          icon: "mdi:air-humidifier-off"
+          icon_priority: 1
+        - dps_val: true
+          icon: "mdi:air-humidifier"
+          icon_priority: 4
+    - id: 2
+      name: humidity
+      type: integer
+      range:
+        min: 35
+        max: 80
+      mapping:
+        - step: 5
+    - id: 6
+      name: current_humidity
+      type: integer
+    - id: 5
+      name: mode
+      type: string
+      mapping:
+        - dps_val: dehumidify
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
```diff
@@ -0,0 +1,145 @@
+name: Vacplus Dehumidifier
```
make-all, line 1: Leave out the brand name above, as other brands may match the same dps layout.  The brand in the products section ensures it is only applied to the correct product.

custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
```diff
@@ -0,0 +1,145 @@
...
+    - id: 1
+      name: switch
+      type: boolean
+      mapping:
+        - dps_val: false
+          icon: "mdi:air-humidifier-off"
+          icon_priority: 1
+        - dps_val: true
+          icon: "mdi:air-humidifier"
+          icon_priority: 4
```
make-all, line 18: Remove these priorities, as the icon is not set elsewhere.  Actually you can remove the whole mapping, because these icons are basically the default, but setting them explicitly overrides Home Assistant theming.

custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
```diff
@@ -0,0 +1,145 @@
...
+          value: Dehumidify
+        - dps_val: rainy
+          value: Rainy
+        - dps_val: dry
+          value: Dry
+        - dps_val: sleep
+          value: Sleep
+secondary_entities:
+  - entity: switch
+    name: Anion
```
make-all, line 44: Call it Ionizer to be consistent with other similar devices.  "Anion" is a technical term that isn't familiar to most end users.

custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
```diff
@@ -0,0 +1,145 @@
...
+    dps:
+      - id: 12
+        type: boolean
+        name: switch
+        optional: true
+  - entity: sensor
+    name: Humidity
```
make-all, line 51: Remove this sensor, as it is just a duplicate of the humidifier current_humidity attribute, which will already be available in the ui, recorder and elsewhere.

custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
```diff
@@ -0,0 +1,145 @@
...
+          - dps_val: 1
+            value: Tank Full
+            icon: "mdi:cup-water"
+            icon_priority: 2
+          - dps_val: 2
+            value: Fault
+          - dps_val: 3
+            value: Fault
+  - entity: select
+    name: Countdown
```
make-all, line 88: Use the name "Timer" to be consistent with other similar devices.

custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
```diff
@@ -0,0 +1,145 @@
...
+    name: Humidity
+    class: humidity
+    dps:
+      - id: 6
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    name: Temperature
```
make-all, line 60: Remove the name, as the `class: temperature` already selects an appropriate translated name.

custom_components/tuya_local/devices/vacplus_dehumidifier.yaml
bdf0506: What do you recommend here? Dehumidifier seems too generic for something like this, as there are already plenty of different Dehumidifier models in this project.

bdf0506:
All changes are complete - I believe it should be good to go now

PR #1233, opened 2023-10-31
Aleco Air fault sensors

mihsu81:
https://github.com/make-all/tuya-local/discussions/1199#discussioncomment-7377882

### DIFF
--- a/custom_components/tuya_local/devices/alecoair_d16_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/alecoair_d16_dehumidifier.yaml
@@ -12,7 +12,7 @@ primary_entity:
       mapping:
         - dps_val: false
           icon: "mdi:air-humidifier-off"
-          icon_priority: 1
+          icon_priority: 2
         - dps_val: true
           icon: "mdi:air-humidifier"
           icon_priority: 4
@@ -52,7 +52,6 @@ primary_entity:
       mapping:
         - dps_val: 0
           value: OK
-          icon: "mdi:cup-outline"
         - dps_val: 1
           value: "Water Tank Full or Removed"
           icon: "mdi:cup-water"
@@ -111,7 +110,6 @@ secondary_entities:
     class: problem
     name: Tank full
     icon: "mdi:cup-water"
-    optional: true
     dps:
       - id: 19
         type: bitfield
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/alecoair_d16_dehumidifier.yaml
```diff
@@ -48,19 +48,19 @@ primary_entity:
     - id: 19
       name: error
       type: bitfield
-      # Docs mention E1 and E2, it isn't clear which is tank full, so for now
-      # treat both as that.
+      # E2 might be a compressor fault.
       mapping:
         - dps_val: 0
           value: OK
+          icon: "mdi:cup-outline"
```
make-all, line 55: Don't set the icon here for the OK case, you want the preset icons to show through when the status is normal.

custom_components/tuya_local/devices/alecoair_d16_dehumidifier.yaml
```diff
@@ -48,19 +48,19 @@ primary_entity:
...
         - dps_val: 1
-          value: E1
+          value: "Water Tank Full or Removed"
           icon: "mdi:cup-water"
-          icon_priority: 2
+          icon_priority: 1
```
make-all, line 59: I think it is better to leave these as 2, so they do not clash with the power off icon.  If you feel strongly that the abnormal conditions should be indicated even when the switch is off, then change that to 2 so the priorities are clearly separate.

custom_components/tuya_local/devices/alecoair_d16_dehumidifier.yaml
```diff
@@ -108,16 +108,34 @@ secondary_entities:
...
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 1
+            value: true
+          - value: false
+  - entity: binary_sensor
+    class: problem
+    name: Fault
+    icon: "mdi:engine-off-outline"
```
make-all, line 126: I don't think this icon is a substantial improvement over the generic problem icon for a fault indicator.  I think it is better to leave the icon at the default so that user themes can customise it when there isn't a strong case that it is an improvement.

mihsu81:
Sorry, I was away for a month and didn't have time to make the suggested changes. Thank you for merging the PR and adding your requested changes as well.

PR #1236, opened 2023-11-01
Create qoto_05_water_valve.yaml

illuzn:
Implements support for the Tuya Zigbee QOTO QT-05 Water Valve Controller

### DIFF
--- a/custom_components/tuya_local/devices/qoto_05_water_valve.yaml
+++ b/custom_components/tuya_local/devices/qoto_05_water_valve.yaml
@@ -1,7 +1,7 @@
-name: Water Valve
+name: Water valve
 products:
   - id: arge1ptm
-    name: 
+    name: QOTO 05 water valve
 primary_entity:
   entity: number
   icon: "mdi:pipe-valve"
@@ -19,7 +19,7 @@ secondary_entities:
   - entity: sensor
     category: diagnostic
     icon: "mdi:valve"
-    name: Current Valve Position
+    name: Current valve position
     dps:
       - id: 3
         name: sensor
@@ -27,7 +27,7 @@ secondary_entities:
         unit: "%"
   - entity: sensor
     category: diagnostic
-    name: Auto Shutoff Time Remaining
+    name: Auto shutoff time remaining
     class: duration
     dps:
       - id: 101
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/qoto_05_water_valve.yaml
```diff
@@ -0,0 +1,66 @@
...
+        unit: s
+  - entity: number
+    category: config
+    name: Auto Shutoff Timer
+    icon: "mdi:timer"
+    dps:
+      - id: 11
+        type: integer
+        name: value
+        unit: s
```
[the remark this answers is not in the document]
illuzn: Copied all of these from the other QOTO device (QOTO 03 Sprinkler) which looks like it was quite old and not following these conventions.

Having a little issue with this one though. I too think that having it in seconds is a bit bizarre. Will break it down into the two timers.

**Shutoff Timer**
The device itself allows setting it with second accuracy (not sure when/ why that would be important but I figured if someone wants to water their plant for 1 minute and 1 second why should I stop them. I hate that it displays in HA as pure seconds rather than HH:MM:SS format but not sure if there is a way to do this.

**Time Remaining**
The device has a special way of doing this. For time > 60s it only updates every minute (presumably to save battery power). For time < 60s it updates every 10 seconds to provide more granularity.
Is there a way with scaling to preserve this granularity when less than 60s (was thinking of using float but that seems potentially fraught with disaster).

custom_components/tuya_local/devices/qoto_05_water_valve.yaml
```diff
@@ -34,26 +37,25 @@ secondary_entities:
...
     category: config
     name: Auto Shutoff Timer
-    icon: "mdi:timer"
+    class: duration
     dps:
       - id: 11
         type: integer
         name: value
-        unit: s
+        unit: min
```
make-all, line 45: To convert from s to min, this also needs
```
     mapping:
        - scale: 60
          step: 60
```

custom_components/tuya_local/devices/qoto_05_water_valve.yaml
```diff
@@ -1,4 +1,7 @@
-name: QOTO 05 Water Valve
+name: Water Valve
```
make-all, line 1: This and many of the entity names are still not following HA naming convention completely, as they have capitalised every word.

PR #1243, opened 2023-11-02
Update 3 devices to conventions

illuzn:
Updates 3 tuya devices which I have to current HA/ tuya-local conventions.


make-all (commented):
It looks like only the products and name changes are correct.  We need to be careful changing names of entities that are already released, as it will change their id in the entity registry.  This is why I have not done any bulk changes to commonise names before now, it really needs to come with a migration to prevent zombie entities being left behind and potentially breaking user automations and scripts.

make-all:
Merged the changes that do not cause backward compatibility issues.

PR #1260, opened 2023-11-06
Add EKF Connect Pro powerstrip

ahydrax:
Hi there!
There are two models from EKF which don't exactly match with current Tuya powerstrip analogs.
List of added devices:
- RCE-1-WF (removed, use quad_powestrip instead)
- RCE-2-WF

The differences between them in datapoints is power consumption monitoring and usb hub switch.
Both tested manually.

Any comments are appreciated.


custom_components/tuya_local/devices/ekf_connect_pro_wifi_powerstrip.yaml
```diff
@@ -0,0 +1,153 @@
+name: EKF Connect Pro WiFi 4 outlet + USB powerstrip
```
make-all, line 1: This one seems to be the same as the generic "quad_powerstrip.yaml" config, with the only addition being the non-useful test bit (which is always 1 in shipped devices)

custom_components/tuya_local/devices/ekf_connect_pro_wifi_type_c_powerstrip.yaml
```diff
@@ -0,0 +1,241 @@
...
+        mapping:
+          - scale: 10
+      - id: 22
+        type: integer
+        name: calibration
+        readonly: true
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    name: Test Bit
```
make-all, line 179: I don't think this needs to be a sensor.  In shipped devices it is always 1.  In any case, the binary_sensor needs to have a dp with name "sensor", the name "factory_test" is suitable if it is an extra attribute on some other entity.

custom_components/tuya_local/devices/ekf_connect_pro_wifi_type_c_powerstrip.yaml
```diff
@@ -0,0 +1,241 @@
...
+        name: factory_test
+        readonly: true
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    name: Error
```
make-all, line 192: For this, I plan to standardise the naming in future to "Fault", so it would be better to use that from the start.

custom_components/tuya_local/devices/ekf_connect_pro_wifi_powerstrip.yaml
ahydrax: Should I remove the config for it then?

custom_components/tuya_local/devices/ekf_connect_pro_wifi_type_c_powerstrip.yaml
ahydrax: Then I would remove it if it's not usable. I tried to make this as close as possible to other powerstrip configuration files.

ahydrax:
I removed explicit config for RCE-1-WF and adjusted properties for RCE-2-WF

ahydrax:
hi @make-all! Could you merge PR as all checks have passed?

make-all:
I merged this as a generic "quad_powerstrip_with_usb", as I suspect that there are many powerstrips that would match this (especially with the "extra" features made optional)

PR #1264, opened 2023-11-07
Adds support for my Rojeco PTM-001 Cat feeder

ciaranj:
The provided PTM-001 YAML did not match my DPs. This one does, it's quite unclear to me whether or not this will cause conflicts with the existing one, but I don't *think* it does!


ciaranj:
Thanks for sorting the naming conventions out @make-all. I /believe/ I just blindly copied them across from the original PTM-001 file so that may also require adjusting.

make-all:
There's a lot of historical cruft that needs cleaning up, but once it is in there it is harder to change entity names, as it can mess up automations etc that users have taken time to set up.

 The HA naming conventions were only introduced last year, so some configs predate that.

PR #1274, opened 2023-11-08
Create momcozy_white_noise_machine.yaml

bdf0506:
Add support for Mom Cozy White Noise machine, v1.
https://www.amazon.com/Machine-Momcozy-Sleeping-Soothing-Personal/dp/B099RSXLGH

This is v1 which has support for Alexa and Google Home. It requires the timer to be done at 30/60/90 minute intervals only. They seem to have discontinued the initial version, though it likely is probably still sold across different names. If you were to buy this device on Amazon today, you'll likely get the v2 which requires a different config #1275 

DPS mappings as such:
{"1":"Switch","2":"Work Mode","3":"Light Switch","4":"Light Mode","5":"Light Color","6":"Brightness","8":"Music Switch","9":"Volume","10":"Music","11":"Song Choose","12":"Scene","13":"Customize Scene","14":"Customize Scene Set","20":"Timer"}
image
IMG_1099


custom_components/tuya_local/devices/momcozy_white_noise_machine.yaml
```diff
@@ -0,0 +1,278 @@
...
+            value: Colourful
+          - dps_val: play
+            value: Play
+          - dps_val: reading
+            value: Reading
+          - dps_val: night
+            value: Nightlight
+          - dps_val: null
+            value: None
+  - entity: number
```
make-all, line 175: I realise siren entities do not have a good UI in Home Assistant, but I don't think it is the place of this integration to work around that.  Instead of duplicating everything here, you can use input helpers to build a UI for the parameters to pass the siren.turn_on service call (and push the HA frontend developers to make a better UI for sirens).

custom_components/tuya_local/devices/momcozy_white_noise_machine.yaml
```diff
@@ -0,0 +1,278 @@
+name: White noise machine
+products:
+  - id: woqiobw0jfafgmar
+    name: Mom Cozy White Noise Machine, v1 - no childlock, timer 30,60,90
```
make-all, line 4: One day this will be used for the default device name for matching product ids, so it shouldn't be too long.  I suggest "Mom Cozy white noise machine" is enough - it doesn't need to be unique, as only one of the two configs will match the device.

custom_components/tuya_local/devices/momcozy_white_noise_machine.yaml
bdf0506: I see both sides here. While there should be a way to control the various settings from the device entry without needing to add extra helpers and build a dashboard, I see your point. Once changes are merged I still may wind up forking and use my own fork that way I could have these settings in the device page of my fork.
I think what we really need to do is build in functionality to support media_player component instead of using the siren component. I'll open a separate issue to address that, and if/when that is ever built, we could revisit this logic.

PR #1275, opened 2023-11-08
Create momcozy_white_noise_machine_v2.yaml

bdf0506:
Add support for Mom Cozy White Noise machine, v2.
https://www.amazon.com/Machine-Momcozy-Sleeping-Soothing-Personal/dp/B099RSXLGH

This is v2 which does NOT support for Alexa and Google Home. It has a child lock available, as well as a timer that can be configured on a per minute basis up to 1440 minutes. Favorite settings (aka Scenes) are done on DPS 13, compared to v1 where Scenes are done on DPS 12. It is presumed that if you buy this product today from the amazon link, this is the likely version that is shipping.

DPS mappings as such:
{"1":"Switch","2":"Work Mode","3":"Light Switch","4":"Light Mode","5":"Light Color","6":"Brightness","8":"Music Switch","9":"Volume","10":"Music","11":"Music","13":"Scene","14":"Customize Scene Set","15":"Wake Up Set","16":"Sleep Aid Set","18":"Alarm Stop","19":"Status","20":"Timer","101":"Child Lock","102":"行程启动","103":"预览模式","104":"倒计时时间上报","105":"启用预设"}

Translations:
102: Trip Start
103: Preview Mode
104: Countdown Time Report
105: Enable Preset

IMG_1096
IMG_1097

### DIFF
--- a/custom_components/tuya_local/devices/momcozy_white_noise_machine_v2.yaml
+++ b/custom_components/tuya_local/devices/momcozy_white_noise_machine_v2.yaml
@@ -185,19 +185,6 @@ secondary_entities:
             icon: "mdi:hand-back-right-off"
           - dps_val: false
             icon: "mdi:hand-back-right"
-  - entity: number
-    name: Timer
-    category: config
-    icon: "mdi:timer"
-    mode: box
-    dps:
-      - id: 20
-        name: value
-        type: integer
-        unit: min
-        range:
-          min: 0
-          max: 1440
   - entity: sensor
     name: Timer Remaining
     class: duration
--- a/custom_components/tuya_local/devices/momcozy_white_noise_machine_v2.yaml
+++ b/custom_components/tuya_local/devices/momcozy_white_noise_machine_v2.yaml
@@ -1,0 +1,300 @@
+name: White noise machine
+products:
+  - id: dymvl7axplliptxx
+    name: Mom Cozy White Noise, v2 with child lock and full timer
+primary_entity:
+  entity: light
+  dps:
+    - id: 3
+      type: boolean
+      name: switch
+    - id: 4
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/momcozy_white_noise_machine_v2.yaml
```diff
@@ -0,0 +1,300 @@
+name: White noise machine
+products:
+  - id: dymvl7axplliptxx
+    name: Mom Cozy White Noise, v2 with child lock and full timer
```
make-all, line 4: As with the v1, device, this should be shortened, and following HA naming guidelines (only first word, brand names, abbreviations capitalized).

custom_components/tuya_local/devices/momcozy_white_noise_machine_v2.yaml
```diff
@@ -0,0 +1,300 @@
...
+            value: Scene 5
+          - dps_val: mode_6
+            value: Scene 6
+          - dps_val: mode_7
+            value: Scene 7
+          - dps_val: mode_8
+            value: Scene 8
+          - dps_val: null
+            value: None
+  - entity: number
```
make-all, line 175: As with the v1 config, volume, tone and duration do not need to be duplicated here.

custom_components/tuya_local/devices/momcozy_white_noise_machine_v2.yaml
make-all: A few of the entities also need minor changes to follow the naming convention

make-all:
Merged manually to avoid the conflict in ACKNOWLEDGEMENTS.md

PR #1281, opened 2023-11-10
new device hyd weair plus fan

xzeertacat:
Add support for hyd weair plus fan (D-68)
https://www.hyd.com.tw/products/weair-plus-iot%E9%81%A0%E7%AB%AF%E6%99%BA%E8%83%BD%E6%B6%BC%E6%9A%96%E9%A2%A8%E7%A9%BA%E6%B0%A3%E6%B8%85%E6%B7%A8%E6%A9%9F-d-68

Unfortunately, this product does not come with an English user manual.
In simple terms, this is a bladeless electric fan with both cooling and heating functions, as well as UV disinfection capability.
78cc61a6bc-Gd-9819477

### DIFF
--- a/custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
+++ b/custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
@@ -6,9 +6,18 @@ primary_entity:
   entity: climate
   dps:
     - id: 1
+      name: hvac_mode
       type: boolean
-      name: mode
-      hidden: true
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: true
+              value: "heat"
+            - dps_val: false
+              value: "fan_only"
     - id: 2
       name: fan_mode
       type: string
@@ -65,21 +74,9 @@ primary_entity:
       name: fault
       optional: true
     - id: 103
-      name: hvac_mode
       type: boolean
-      mapping:
-        - dps_val: true
-          value: "heat"
-          icon: "mdi:radiator"
-        - dps_val: false
-          value: "off"
-          icon: "mdi:radiator-disabled"
-          constraint: mode
-          conditions:
-            - dps_val: true
-              value: "cool"
-            - dps_val: false
-              value: "off"
+      name: mode
+      hidden: true
 secondary_entities:
   - entity: select
     name: Timer
@@ -88,26 +85,25 @@ secondary_entities:
       - id: 11
         type: string
         name: option
-        unit: "hour"
         mapping:
           - dps_val: "0"
             value: "Off"
           - dps_val: "1"
-            value: "1"
+            value: "1 hour"
           - dps_val: "2"
-            value: "2"
+            value: "2 hours"
           - dps_val: "3"
-            value: "3"
+            value: "3 hours"
           - dps_val: "4"
-            value: "4"
+            value: "4 hours"
           - dps_val: "5"
-            value: "5"
+            value: "5 hours"
           - dps_val: "6"
-            value: "6"
+            value: "6 hours"
           - dps_val: "7"
-            value: "7"
+            value: "7 hours"
           - dps_val: "8"
-            value: "8"
+            value: "8 hours"
   - entity: sensor
     name: Time remaining
     class: "duration"
--- a/custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
+++ b/custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
@@ -1,0 +1,149 @@
+name: HYD WeAir Plus Fan
+products:
+  - id: 6p7jm8rrqimnlusw
+    name: D-68
+primary_entity:
+  entity: fan
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+    - id: 13
+      type: integer
+      name: fault
+      optional: true
+secondary_entities:
+  - entity: select
+    name: Wind Speed
+    category: config
+    dps:
+      - id: 2
+        type: string
+        name: option
+        mapping:
+          - dps_val: "1"
+            value: "1"
+          - dps_val: "2"
+            value: "2"
+          - dps_val: "3"
+            value: "3"
+          - dps_val: "4"
+            value: "4"
+          - dps_val: "5"
+            value: "5"
+          - dps_val: "6"
+            value: "6"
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -0,0 +1,149 @@
...
+    - id: 1
+      type: boolean
+      name: switch
+    - id: 13
+      type: integer
+      name: fault
+      optional: true
+secondary_entities:
+  - entity: select
+    name: Wind Speed
```
make-all, line 17: This should be speed (mapped to percentages) in the fan entity.

custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -0,0 +1,149 @@
...
+          - dps_val: "6"
+            value: "6"
+          - dps_val: "7"
+            value: "7"
+          - dps_val: "8"
+            value: "8"
+          - dps_val: "9"
+            value: "9"
+  - entity: select
+    name: Wind Mode
```
make-all, line 43: This should be preset_mode in the fan entity, with mappings chosen from the predefined preset modes: https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/en.json#L169
(eg sleep, strong, fresh, off) then the fan should have a class of "fan_with_presets".

custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -0,0 +1,149 @@
...
+          - dps_val: "sleep"
+            value: "sleep"
+          - dps_val: "heavy"
+            value: "heavy"
+          - dps_val: "fresh"
+            value: "fresh"
+          - dps_val: "close"
+            value: "close"
+  - entity: switch
+    name: Windleftright
```
make-all, line 60: This should be oscillate in the fan entity.

custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -0,0 +1,149 @@
...
+    icon: "mdi:thermometer-lines"
+    dps:
+      - id: 8
+        type: boolean
+        name: switch
+  - entity: number
+    name: TempSet
```
make-all, line 67: If there is a settable temperature, it may be better to make the main entity a climate entity rather than a fan entity, and make this the temperature.

custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -0,0 +1,149 @@
...
+  - entity: select
+    name: Timer
+    category: config
+    dps:
+      - id: 11
+        type: string
+        name: option
+        mapping:
+          - dps_val: "0"
+            value: "0"
```
make-all, line 92: Generally "0" should be "Off" for timers, and other numbers should be marked with the units (presumably hour), to make it clear to end users.

custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -0,0 +1,149 @@
...
+          - dps_val: "5"
+            value: "5"
+          - dps_val: "6"
+            value: "6"
+          - dps_val: "7"
+            value: "7"
+          - dps_val: "8"
+            value: "8"
+  - entity: sensor
+    name: Countdown
```
make-all, line 110: Call this "Time remaining" to be consistent with other configs, and give it a class of duration.

custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -0,0 +1,149 @@
...
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Warm Switch
+    dps:
+      - id: 103
+        type: boolean
+        name: switch
+  - entity: select
+    name: Warm Speed
```
make-all, line 129: Does this override the fan speed?  If it is a strength of the heating element, then "level" would be better than "speed".

custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -0,0 +1,149 @@
...
+          - dps_val: "1"
+            value: 1"
+          - dps_val: "2"
+            value: "2"
+          - dps_val: "3"
+            value: "3"
+          - dps_val: "4"
+            value: "4"
+  - entity: switch
+    name: UVC
```
make-all, line 145: Use "UV sterilization" to be consistent with other configs.

custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -0,0 +1,149 @@
...
+    dps:
+      - id: 9
+        type: integer
+        name: value
+        unit: "C"
+        range:
+          min: 1
+          max: 30
+  - entity: sensor
+    name: NowTemp
```
make-all, line 77: If a climate entity is used, this should be the current_temperature of that.

xzeertacat:
I have already modify the main entity to climate entity and modify string to follow other configs.
thanks.

custom_components/tuya_local/devices/hyd_weair_plus_fan.yaml
```diff
@@ -3,93 +3,95 @@ products:
   - id: 6p7jm8rrqimnlusw
     name: D-68
 primary_entity:
-  entity: fan
+  entity: climate
   dps:
     - id: 1
       type: boolean
-      name: switch
+      name: mode
```
make-all, line 10: For the "toggle" service call to work correctly, I think this one needs to be hvac_mode, and 103 the hidden mode.

I think the logic is easier that way too, as the current logic is probably wrong for the case dps: {"1": false, "103": true}

```
  - id: 1
    type: boolean
    name: hvac_mode
    mapping:
      - dps_val: false
        value: "off"
      - dps_val: true
        constraint: mode
        conditions:
           - dps_val: true
             value: heat
           - dps_val: false
             value: fan_only
  - id: 103
    name: mode
    type: boolean
    hidden: true
```

xzeertacat:
You're right, this product only “fan” mode , not really cool the air
and I fix the hvac_mode logic too.
thank you for your help.

PR #1286, opened 2023-11-11
add support for EUROM Hot-Shot 2000 WiFi

dmitriy5181:


custom_components/tuya_local/devices/hotshot2000_heater.yaml
```diff
@@ -0,0 +1,72 @@
...
+      name: work_mode
+      type: string
+      hidden: true
+    - id: 6
+      name: preset_mode
+      type: boolean
+      mapping:
+        - dps_val: false
+          constraint: work_mode
+          conditions:
```
make-all, line 42: This is missing the "level1" condition.  If you don't want it showing up as an option, you can mark it hidden, but it would help to have a value mapped to it to show in the UI.

custom_components/tuya_local/devices/hotshot2000_heater.yaml
```diff
@@ -0,0 +1,72 @@
...
+      name: swing_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: "on"
+secondary_entities:
+  - entity: switch
+    name: Light
+    category: diagnostic
```
make-all, line 60: Should be config, not diagnostic, as switch is a modifiable entity (you should see warnings in the log about this).

custom_components/tuya_local/devices/hotshot2000_heater.yaml
```diff
@@ -0,0 +1,72 @@
...
+    dps:
+      - id: 10
+        type: boolean
+        name: switch
+  - entity: sensor
+    name: Remaining Time
```
make-all, line 66: Remaining time, but no Timer to go with it?

If the timer is set by setting this same dp, then it should be a number entity rather than a sensor.

dmitriy5181: Unfortunately it is not possible to set this, it is read only. I tried "number" instead of "sensor" but when I set a value in the UI the device resets it back. It is a 9 hours timer and it starts when you turn on the device.

custom_components/tuya_local/devices/hotshot2000_heater.yaml
dmitriy5181: As I understood from the docs for the Climate Entity, the "preset_mode" does not have "fan_only" option. The "level1" is present under the "hvac_mode" which supports "heat" and "fan_only", see line 17. And it is selectable in the UI:

image

custom_components/tuya_local/devices/hotshot2000_heater.yaml
dmitriy5181: How would you call the red ring below (it can be blue, green and red, indicating level1, level2 and level3 values correspondingly)?

custom_components/tuya_local/devices/hotshot2000_heater.yaml
make-all: Unlike HVAC mode, preset mode does not have to be from the predefined modes, so it won't hurt to use "Fan" here. But the predefined mode "none" might also work, and be better than whatever happens now if you select fan_only from HVAC mode then try to change the preset mode.

custom_components/tuya_local/devices/hotshot2000_heater.yaml
make-all: That is an indicator, maybe "Mode indicator"

custom_components/tuya_local/devices/hotshot2000_heater.yaml
make-all: If it is a fixed timer, that is always active when the device is on, then it makes sense without the corresponding Timer select or number  entity.

custom_components/tuya_local/devices/hotshot2000_heater.yaml
dmitriy5181: Yes it is indeed more straightforward to have fan only under the preset_mode. But I'm missing a nice fan icon. Why is it not visible when I changed this dp to:
```
    - id: 6
      name: preset_mode
      type: boolean
      mapping:
        - dps_val: false
          constraint: work_mode
          conditions:
            - dps_val: "level1"
              value: Fan only
              icon: "mdi:fan"
            - dps_val: "level2"
              value: comfort
            - dps_val: "level3"
              value: boost
        - dps_val: true
          value: eco
```

image

make-all: I can't see any way for an integration to define new preset icons.  The `icon` directive in the config files is for the entity icon.

dmitriy5181: That's a pity.

custom_components/tuya_local/devices/hotshot2000_heater.yaml
```diff
@@ -12,14 +12,7 @@ primary_entity:
...
         - dps_val: true
-          constraint: work_mode
-          conditions:
-            - dps_val: level1
-              value: fan_only
-            - dps_val: level2
-              value: heat
-            - dps_val: level3
-              value: heat
+          value: heat
```
make-all, line 15: I think this was OK to leave in, unless you observed problems with it.

dmitriy5181: Ok, then level1 will be fan_only under hvac_mode and none + hidden under preset_mode. This looks logical in the UI with built-in icons for all modes and presets.

PR #1296, opened 2023-11-13
support for CALDO UP T WIFI

sebastianciupinski:


custom_components/tuya_local/devices/caldo_up_t_wifi.yaml
```diff
@@ -0,0 +1,56 @@
+name: CALDO UP T WIFI
```
make-all, line 1: Move this name under products, and put a name here that describes the device's purpose.

custom_components/tuya_local/devices/caldo_up_t_wifi.yaml
```diff
@@ -0,0 +1,56 @@
...
+      type: boolean
+      name: swing_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: "on"
+secondary_entities:
+   - entity: select
+     name: Heating mode
```
make-all, line 34: This can be a preset_mode on the climate entity

PR #1314, opened 2023-11-17
Add new device Blitzwolf BW-LT31 LED Strip

MackoMici:


custom_components/tuya_local/devices/blitzwolf_bw_lt31.yaml
```diff
@@ -0,0 +1,155 @@
+name: Blitzwolf BW-LT31 LED Strip
+products:
+  - id: 66072501f4cfa2fe2887
```
make-all, line 3: This does not look like the usual format for a product id.  Is it really the product id, and not the device id?

custom_components/tuya_local/devices/blitzwolf_bw_lt31.yaml
```diff
@@ -0,0 +1,155 @@
+name: Blitzwolf BW-LT31 LED Strip
```
make-all, line 1: Top level name should be just LED strip, without branding.  The branding belongs under products.

custom_components/tuya_local/devices/blitzwolf_bw_lt31.yaml
```diff
@@ -0,0 +1,155 @@
...
+            min: 0
+            max: 1000
+    - id: 102
+      name: music_data
+      type: integer
+    - id: 103
+      name: line_sequence_adjustment
+      # 线序调整
+      type: integer
+      hidden: true
```
make-all, line 60: It doesn't make sense for these to be hidden, and they are not surfaced any other way.

custom_components/tuya_local/devices/blitzwolf_bw_lt31.yaml
```diff
@@ -0,0 +1,155 @@
...
+            # color: red, static
+          - dps_val: "CJ_XK"
+            value: "Starry"
+            # color: blue, flash
+          - dps_val: "CJ_JH"
+            value: "Party"
+            # color* All, Flash, Mid speed
+          - dps_val: "CJ_YS"
+            value: "Film"
+            # color: lightblue, static
```
make-all, line 155: Last line should be terminated with a newline.

PR #1328, opened 2023-11-20
add support for VALEHAUS SYZN119 towel warmer

latel:
name: VALEHAUS Towel Warmer
products:
  - id: 1dxb14emvqzxzc5q
    name: VALEHAUS SYZN119

## VALEHAUS SYZN119 DPS
```
"mapping": {
                "1": {
                    "code": "switch",
                    "type": "Boolean",
                    "values": {}
                },
                "2": {
                    "code": "temp_set",
                    "type": "Integer",
                    "values": {
                        "unit": "\u00b0C",
                        "min": 35,
                        "max": 60,
                        "scale": 0,
                        "step": 5
                    }
                },
                "3": {
                    "code": "temp_current",
                    "type": "Integer",
                    "values": {
                        "unit": "\u00b0C",
                        "min": 0,
                        "max": 99,
                        "scale": 0,
                        "step": 1
                    }
                },
                "12": {
                    "code": "countdown_set",
                    "type": "Enum",
                    "values": {
                        "range": [
                            "cancel",
                            "1h",
                            "2h",
                            "3h",
                            "4h",
                            "5h",
                            "6h"
                        ]
                    }
                },
                "13": {
                    "code": "countdown_left",
                    "type": "Integer",
                    "values": {
                        "unit": "min",
                        "min": 0,
                        "max": 1440,
                        "scale": 0,
                        "step": 1
                    }
                },
                "14": {
                    "code": "work_state",
                    "type": "Enum",
                    "values": {
                        "range": [
                            "heating",
                            "standby"
                        ]
                    }
                }
            }
```

## self test
image

### DIFF
--- a/custom_components/tuya_local/devices/valehaus_syzn119_tow_w.yaml
+++ b/custom_components/tuya_local/devices/valehaus_syzn119_tow_w.yaml
@@ -1,0 +1,84 @@
+name: VALEHAUS Towel Warmer
+products:
+  - id: 1dxb14emvqzxzc5q
+    name: VALEHAUS SYZN119
+
+primary_entity:
+  entity: switch
+  name: power
+  icon: "mdi:power"
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+
+secondary_entities:
+  - entity: number
+    name: temp_set
+    category: config
+    dps:
+      - id: 2
+        type: integer
+        name: value
+        range:
+          min: 35
+          max: 60
+        unit: "℃"
+        mapping:
+          - step: 5
+  - entity: sensor
+    name: temp_current
+    dps:
+      - id: 3
+        type: integer
+        name: sensor
+        range:
+          min: 0
+          max: 99
+        unit: "℃"
+  - entity: select
+    name: countdown_set
+    category: config
+    dps:
+      - id: 12
+        type: string
+        name: option
+        optional: true
+        mapping:
+          - dps_val: cancel
+            value: cancel
+          - dps_val: "1h"
+            value: "1h"
+          - dps_val: "2h"
+            value: "2h"
+          - dps_val: "3h"
+            value: "3h"
+          - dps_val: "4h"
+            value: "4h"
+          - dps_val: "5h"
+            value: "5h"
+          - dps_val: "6h"
+            value: "6h"
+  - entity: sensor
+    name: countdown_left
+    class: duration
+    dps:
+      - id: 13
+        type: integer
+        name: sensor
+        unit: "min"
+        range:
+          min: 0
+          max: 1440
+  - entity: sensor
+    name: work_state
+    class: enum
+    dps:
+      - id: 14
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: "heating"
+            value: "standby"
+          - dps_val: "standby"
+            value: "heating"


latel:
There is one thing I have no idea how to fix yet, this device has a count_down dp 12, it's an enum, I map it into homeassistant's select entity, however when I change count_down time to eg 6h, it still reporting dp 12 status with "cancel"
```
 {
        "code": "countdown_set",
        "custom_name": "",
        "dp_id": 12,
        "time": 1699687833000,
        "value": "cancel"
      },
```

this result in that, when I try to set countdown from homeassistant ui, it will result in unkown, after a while.

make-all:
You may need to observe the dp when you make changes from the app to see if the values match the dps mapping info you have above. If this information is not from the QueryThingsDataModel cloud API call, you may also get more accurate information from that.

latel:
> You may need to observe the dp when you make changes from the app to see if the values match the dps mapping info you have above. If this information is not from the QueryThingsDataModel cloud API call, you may also get more accurate information from that.

I select different values with the enum, and immediate query the latest staus data from IOT webpage, the latest value is not the one I selected, it's just always "cancel"

custom_components/tuya_local/devices/valehaus_syzn119_tow_w.yaml
```diff
@@ -0,0 +1,84 @@
...
+          max: 1440
+  - entity: sensor
+    name: work_state
+    class: enum
+    dps:
+      - id: 14
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: "heating"
```
make-all, line 81: Is it correct that these are reversed?

make-all (commented):
This device seems like it should be a climate entity, with two timer related secondary entities.

power = hvac_mode (true = heat, false = "off")
temp_set = temperature
temp_current = current_temperature
work_mode = hvac_action (standby = idle, heating = heating / swapped if that is really a bug in the device)

For the Timer issue, as long as "Time remaining" updates when the timer is set, it should be enough feedback.

custom_components/tuya_local/devices/valehaus_syzn119_tow_w.yaml
```diff
@@ -0,0 +1,84 @@
...
+    dps:
+      - id: 3
+        type: integer
+        name: sensor
+        range:
+          min: 0
+          max: 99
+        unit: "℃"
+  - entity: select
+    name: countdown_set
```
make-all, line 40: Suggested name is "Timer", to be consistent with other devices.

custom_components/tuya_local/devices/valehaus_syzn119_tow_w.yaml
```diff
@@ -0,0 +1,84 @@
...
+          - dps_val: "3h"
+            value: "3h"
+          - dps_val: "4h"
+            value: "4h"
+          - dps_val: "5h"
+            value: "5h"
+          - dps_val: "6h"
+            value: "6h"
+  - entity: sensor
+    name: countdown_left
```
make-all, line 63: Suggested name is "Time remaining" to be consistent with other devices.

make-all:
Need feedback on the swapped standby and heating, other issues are minor, which I would fix myself, but I am not sure whether the swapped mapping is intentional, or a mistake.

latel:
Yes, the swap is intentional, it's a device bug. I have this device myself.

PR #1344, opened 2023-11-24
Adding integration for Fisher Summer Air Conditioners

urgelaszlo:
Adding integration for Fisher Summer Air Conditioners.
Tested on FSAI-SU-125FE3 model (3.4kW)
Did not find English description: https://www.fisherklima.hu/termekek/split-klima/oldalfali/summer-sorozat/summer-2500w-inverteres-split-klima-520-669-adatlap?limitstart=0

Adding the JSON retrieved from Tuya IOT. 
I created a separate secondary entity for all the DPs available. DPs 115, 116, 122, 128 and 136 were breaking the integration (could not add the device with them in the YAML) so they are removed here. Some other DPs from the JSON were accepted, but not applicable to the current device (no sensor for humidity, so it returns just 0), I kept them to allow other models use the integration. Chinese descriptions and names were translated using Google Translate, tried to use more descriptive English names in some cases.

In the current state I can confirm the integration is working, however can't test extensively (e.g. to find out some DP functions which is not clear from the translation).
[Fisher_Summer_tuya_iot.json](https://github.com/make-all/tuya-local/files/13462422/Fisher_Summer_tuya_iot.json)

### DIFF
--- a/custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
+++ b/custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
@@ -1,0 +1,463 @@
+name: Air conditioner
+products:
+  - id: bf8705744de2dd5ab4ogqj
+    name: Fisher Summer Air Conditioner
+primary_entity:
+  entity: climate
+  icon: "mdi:air-conditioner"
+  dps:
+    - id: 1
+      name: hvac_mode
+      type: boolean
+      mapping:
+        - dps_val: false
+          value: "off"
+          icon: "mdi:hvac-off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: auto
+              icon: mdi:thermostat-auto
+              value: heat_cool
+            - dps_val: cold
+              icon: "mdi:snowflake"
+              value: cool
+            - dps_val: wind
+              icon: "mdi:fan"
+              value: fan_only
+            - dps_val: wet
+              icon: "mdi:water"
+              value: dry
+            - dps_val: hot
+              value: heat
+              icon: "mdi:heat-wave"
+    - id: 2
+      name: temperature
+      type: integer
+      range:
+        min: 160
+        max: 310
+      mapping:
+        - scale: 10  
+          step: 5 
+      unit: C
+    - id: 3
+      name: current_temperature
+      type: integer
+      readonly: true
+    - id: 4
+      name: mode
+      type: string
+    - id: 5
+      name: fan_mode
+      type: string
+      mapping:
+        - dps_val: auto
+          value: Auto
+        - dps_val: mute
+          value: Quiet
+        - dps_val: low
+          value: Low
+        - dps_val: mid_low
+          value: Medium Low
+        - dps_val: mid
+          value: Medium
+        - dps_val: mid_high
+          value: Medium High
+        - dps_val: high
+          value: High
+        - dps_val: strong
+          value: Strong
+secondary_entities:
+  - entity: switch
+    name: Power
+    category: config
+    icon: "mdi:power"
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+  - entity: number
+    name: Target Temperature
+    category: config
+    class: temperature
+    dps:
+      - id: 2
+        name: value
+        type: integer
+        unit: C
+        range:
+          min: 160
+          max: 310
+        mapping:
+          - scale: 10
+            step: 5
+  - entity: sensor
+    name: Current Temperature
+    class: temperature
+    dps:
+      - id: 3
+        name: sensor
+        type: integer
+        unit: C
+        readonly: true
+  - entity: select
+    name: Mode
+    category: config
+    dps:
+      - id: 4
+        name: option
+        type: string
+        mapping:
+          - dps_val: auto
+            icon: mdi:thermostat-auto
+            value: heat_cool
+          - dps_val: cold
+            icon: "mdi:snowflake"
+            value: cool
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
```diff
@@ -0,0 +1,463 @@
...
+      readonly: true
+    - id: 4
+      name: mode
+      type: string
+    - id: 5
+      name: fan_mode
+      type: string
+      mapping:
+        - dps_val: auto
+          value: Auto
```
make-all, line 56: These should use the translations from https://github.com/home-assistant/core/blob/dev/homeassistant/components/climate/strings.json#L40 if necessary supplemented by `translation_key: aircon_extra` values from https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/en.json#L65 - the value to use here is the key (left hand side) in the strings files, and is case-sensitive so should be all lower case (lower case is used to easily differentiate where translations are used in the config values)

custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
```diff
@@ -0,0 +1,463 @@
...
+        - dps_val: mid
+          value: Medium
+        - dps_val: mid_high
+          value: Medium High
+        - dps_val: high
+          value: High
+        - dps_val: strong
+          value: Strong
+secondary_entities:
+  - entity: switch
```
make-all, line 72: Controls and sensors that are part of the climate entity should not be duplicated here.

custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
```diff
@@ -0,0 +1,463 @@
...
+          value: Medium Low
+        - dps_val: mid
+          value: Medium
+        - dps_val: mid_high
+          value: Medium High
+        - dps_val: high
+          value: High
+        - dps_val: strong
+          value: Strong
+  - entity: sensor  # no sensor in Fisher Summer AC, returns 0
```
make-all, line 151: This can also go under the climate entity as current_humidity.

custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
```diff
@@ -0,0 +1,463 @@
...
+    class: humidity
+    dps:
+      - id: 18
+        name: sensor
+        type: integer
+        unit: "%"
+        readonly: true
+        optional: true
+  - entity: binary_sensor
+    name: Fault
```
make-all, line 162: It is useful to also list this as a fault_code extra attribute under the climate entity, so that the raw value is exposed.

custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
```diff
@@ -0,0 +1,463 @@
...
+    dps:
+      - id: 20
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+  - entity: sensor  # no sensor in Fisher Summer AC, returns 0
+    name: PM 2.5
```
make-all, line 174: sensors with class do not require a name, unless there are multiple of them or the class default name does not fit the sensor.

custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
```diff
@@ -0,0 +1,463 @@
...
+        name: option
+        type: string
+        mapping:
+          - dps_val: off
+            value: Off
+            default: true
+          - dps_val: normal
+            value: Normal
+          - dps_val: old
+            value: Old
```
make-all, line 199: maybe Elderly is better than Old here.

custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
```diff
@@ -0,0 +1,463 @@
...
+          - dps_val: child
+            value: Child
+        optional: true
+  - entity: sensor
+    name: Function Markbit
+    dps:
+      - id: 110
+        name: sensor
```
make-all, line 207: This does not need to be a sensor, the value will be fixed depending on the model, and is not interesting as an end-user sensor.  Just make it an extra attribute on the climate entity.

custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
```diff
@@ -0,0 +1,463 @@
...
+            default: true
+          - dps_val: L1
+            value: L1
+          - dps_val: L2
+            value: L2
+          - dps_val: L3
+            value: L3
+        optional: true
+  - entity: sensor 
+    name: Boolcode Options
```
make-all, line 314: This one is probably also not useful as a separate sensor and can just be an extra attribute.

custom_components/tuya_local/devices/fisher_summer_air_conditioner.yaml
```diff
@@ -0,0 +1,463 @@
...
+          - dps_val: 5
+            value: Right  
+          - dps_val: 6
+            value: Wide Angle Left
+          - dps_val: 7
+            value: Wide Angle Right
+          - dps_val: 8
+            value: Wide Angle
+  - entity: select
+    name: Power
```
make-all, line 375: Should this be configurable by the user? Is it not a fixed property of the particular model (which could just be an extra attribute) ?

PR #1345, opened 2023-11-26
docs: update Tuya's "Devices Management" path

eyarz:
the "device management" is located under the "API Explorer" which is not maintained in the guide so I fixed it.
image

image


make-all:
The previous line already guides you to select Cloud -> API Explorer

PR #1347, opened 2023-11-27
Add support for Breville Smart Dry Connect Dehumidifier (LAD208)

patoh:
Add support for Breville Smart Dry Connect Dehumidifier (LAD208)[0]

This may also support the the Smart Dry Plus Connect Dehumidifier (LAD308)[1] as it appears to just be a larger version of the LAD208, but I don't have one onhand to test with.

Attached the JSON retrieved from the Tuya query things API console - [breville_lad208_dehumidifier.json](https://github.com/make-all/tuya-local/files/13470742/breville_lad208_dehumidifier.json)

DP 11 (FAULT) according to the API should be 1 to 3, with 3 as the water full / tank removed alarm. However in testing water full / tank removed was found to be 8 not 3. I am unable to test bitmap's 1 and 2 (damaged sensors) to confirm if they are also incorrect, so have left them out and treat anything not 0, 3 or null as an Fault

dehumidifier_overview
dehumidifier

0 - https://www.breville.com/au/en/products/air-dehumidifier/bad208.html?sku=LAD208WHT2IAN1
1 - https://www.breville.com/au/en/products/air-dehumidifier/bad308.html?sku=LAD308WHT2IAN1


custom_components/tuya_local/devices/breville_lad208_dehumidifier.yaml
```diff
@@ -0,0 +1,208 @@
...
+        - dps_val: false
+          icon: "mdi:air-humidifier-off"
+          icon_priority: 2
+        - dps_val: true
+          icon: "mdi:air-humidifier"
+          icon_priority: 4
+    - id: 2
+      name: mode
+      type: string
+      mapping:
```
make-all, line 22: Preset values should be chosen from the list of translations for [HA humidifier](https://github.com/home-assistant/core/blob/dev/homeassistant/components/humidifier/strings.json#L57) or the supplemental ones for [tuya-local dehumidifiers](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/en.json#L157).  Generally "ventilation" would only be appropriate if the device is providing fresh air from outside, usually dehumidifiers are stand alone, and the only purpose of the fan only mode is to filter the recirculated air, so "purify" is usually used for this mode, as it is effectively doing the same job as an air purifier.  For continuous, either normal or boost (probably the latter, as "auto" is the normal mode really) and laundry and auto are defined already (use lower case to match the translation keys).

custom_components/tuya_local/devices/breville_lad208_dehumidifier.yaml
```diff
@@ -0,0 +1,208 @@
...
+    dps:
+      - id: 11
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - dps_val: 8
+            value: false
+          - dps_val: null
```
make-all, line 206: If null needs to be handled, then the dp 11 should probably be marked as optional (everywhere it appears).

PR #1355, opened 2023-11-28
ZY-M201 fix service call error for detection distance slider

blakadder:
fixes service call error that pops up when slider is set to 5.5

image

### DIFF
--- a/custom_components/tuya_local/devices/zym201_presence_sensor.yaml
+++ b/custom_components/tuya_local/devices/zym201_presence_sensor.yaml
@@ -55,10 +55,10 @@ secondary_entities:
         unit: m
         range:
           min: 150
-          max: 600
+          max: 550
         mapping:
           - scale: 100
-            step: 100
+#            step: 100
   - entity: sensor
     name: Target distance
     class: distance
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/zym201_presence_sensor.yaml
```diff
@@ -58,7 +58,7 @@ secondary_entities:
           max: 550
         mapping:
           - scale: 100
-            step: 100
+#            step: 100
```
make-all, line 61: Where did the 100 come from in the first place?  Does the device accept being set to other values than 1.5, 2.5, 3.5, 4.5, 5.5?  If not maybe a select is better for this setting.

make-all:
There also seems to be a bug based on that error message (a < where there should be a <= perhaps).  If the step was the issue, then i would expect "(6.0) must be between 1.5 and 5.5" when trying to set to 5.5, as it would be rounded up.

blakadder:
Step was the initial issue but while playing around with it I realised it accepts steps of 1 and reports them back to the app despite the tuya iot data having step: 100
```
  - abilityId: 4
    accessMode: rw
    code: far_detection
    description: ''
    extensions:
      iconName: icon-gongneng
      attribute: '1152'
    name: 最远检测距离
    typeSpec:
      max: 550
      min: 150
      scale: 2
      step: 100
      type: value
      typeDefaultValue: 150
      unit: M
```

PR #1358, opened 2023-11-29
Create morris_premier_plus_dehumidifier.yaml

ChriZathens:
Hello! This is just a clone of the klarstein_dryfy_pro_connect_dehumidifier.yaml file with minor changes for the Morris MDP-24400IW dehumidifier. From Tuya API there are no more DPs returned, so they are basically identical. The only difference is that while the Klarstein one has the following modes: Auto/Boost/Eco/Purify, the Morris one has Auto/High/Low/Fan respectively. So I just changed the relative values to the ones the Morris one uses. What I cannot find is where the Boost/Eco icons were coming from. So I simply added icons to the relevant section but I doubt they display. This pull request is not actually created with the intention to be commited (unless you believe it is OK ) but rather in order to let you know that the Morris dehumidifier is working fine with the klarstein yaml file, except from those minor differences, so that you can perhaps add a "proper" version yourself if you like.

P.S.: Thanks for your great work !
P.P.S.: The Boost icon (mdi: rocket-launch) is nowhere to be found in the component. How is it displayed?

### DIFF
--- a/custom_components/tuya_local/devices/morris_premier_plus_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/morris_premier_plus_dehumidifier.yaml
@@ -1,0 +1,133 @@
+name: Morris MDP-xx400IW series
+# {'updated_at': 1674605185.2896934, '1': True, '2': 'auto', '3': 57, '4': 55, '10': True, '101': True, '102': '0_90', '103': 24}
+products:
+  - id: 6cguug02fnygou5c
+primary_entity:
+  entity: humidifier
+  class: dehumidifier
+  dps:
+    - id: 1
+      name: switch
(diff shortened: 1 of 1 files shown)


make-all:
There is essentially no difference between there - even the product id is the same, so there is no need for a new config.

ChriZathens:
I missed your comment, sorry about that..
This was a mistake on my part.. I just managed to get the necessary details from tuya Api Explorer, in case you still want to add this device..
      "model": "MDP-xx400IW series",
      "name": "MDP-xx400IW series",
      "product_id": "m0wtqltrepknlpdj",
      "product_name": "MORRIS Premier+ Dehumidifier",

PR #1365, opened 2023-11-30
Support for Klatra Stor2 Air Purifier

kamilosxd678:
Adding support for Klatra Stor2 Air Purifier
image

image

1 - https://klarta.pl/oczyszczacze-powietrza/klarta-stor-2/

### DIFF
--- a/custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
+++ b/custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
@@ -12,7 +12,7 @@ primary_entity:
       name: switch
     - id: 3
       type: string
-      name: preeset_mode
+      name: preset_mode
       mapping:
         - dps_val: auto
           value: smart
--- a/custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
+++ b/custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
@@ -28,15 +28,15 @@ primary_entity:
       mapping:
         - dps_val: speedmode1
           value: 17
-        - dps_val: speedmode1
+        - dps_val: speedmode2
           value: 33
-        - dps_val: speedmode1
+        - dps_val: speedmode3
           value: 50
-        - dps_val: speedmode1
+        - dps_val: speedmode4
           value: 67
-        - dps_val: speedmode1
+        - dps_val: speedmode5
           value: 83
-        - dps_val: speedmode1
+        - dps_val: speedmode6
           value: 100
 secondary_entities:
   - entity: sensor
--- a/custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
+++ b/custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
@@ -1,14 +1,43 @@
-name: Klarta Stor 2 air purifier
+name: Air purifier
 products:
   - id: ku6ae3y8ad6tbx3b
-    name: "KLARTA STOR 2 "
+    name: "Klarta Stor 2 "
 primary_entity:
-  entity: switch
+  entity: fan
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
```diff
@@ -0,0 +1,149 @@
...
+        unit: "%"
+  - entity: select
+    name: AutoMode
+    dps:
+      - id: 101
+        name: option
+        type: string
+        mapping:
+          - dps_val: AutoSleep
+            value: "AutoSleep"
```
make-all, line 73: value should be the human readable value that will appear in the UI (unless a matching translation is available, which in the case of select entities generally isn't the case).  So adding spaces (or perhaps even removing the redundant "Auto" prefix) is better.

custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
```diff
@@ -0,0 +1,149 @@
...
+    name: Active filter life
+    icon: "mdi:air-filter"
+    category: diagnostic
+    dps:
+      - id: 5
+        type: integer
+        name: sensor
+        unit: "%"
+  - entity: select
+    name: AutoMode
```
make-all, line 66: Should be the human readable name, following [HA naming conventions](https://developers.home-assistant.io/blog/2022/07/10/entity_naming/), so "Auto mode".

custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
```diff
@@ -0,0 +1,149 @@
...
+  - entity: button
+    name: Filter reset
+    class: restart
+    category: config
+    dps:
+      - id: 11
+        type: boolean
+        name: button
+  - entity: select
+    name: Countdown
```
make-all, line 109: To be consistent with other devices, this should be called "Timer"

custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
```diff
@@ -0,0 +1,149 @@
+name: Klarta Stor 2 air purifier
+products:
+  - id: ku6ae3y8ad6tbx3b
+    name: "KLARTA STOR 2 "
+primary_entity:
+  entity: switch
```
make-all, line 6: This would be better represented as a fan, along with speed and preset_mode dps.

custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
```diff
@@ -0,0 +1,149 @@
...
+  - entity: select
+    name: Speed
+    dps:
+      - id: 4
+        name: option
+        type: string
+        icon: "mdi:fan"
+        mapping:
+          - dps_val: speedmode1
+            value: "Silent"
```
make-all, line 45: When converted to a fan entity, these should be mapped to percentages. (17, 33, 50, 67, 83, 100)

custom_components/tuya_local/devices/klarta_stor2_purifier.yaml
```diff
@@ -0,0 +1,149 @@
...
+        class: measurement
+        unit: ugm3
+  - entity: select
+    name: Mode
+    dps:
+      - id: 3
+        name: option
+        type: string
+        mapping:
+          - dps_val: auto
```
make-all, line 28: When converted to fan, the fan should be given a `translation_key: fan_with_presets`, and mapped to values chosen from [available translations](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/en.json#L169) (eg: auto => smart, manual => normal, sleep => sleep, screenoff => displayoff)

PR #1369, opened 2023-12-01
Added support for Beok TGM50-WIFI-WP

cart3a:
Added support for Beok TGM50-WIFI-WP.

I tested all the entities and they work correctly. 

The Tuya IoT Platform json response is [Query Things Data Model.json](https://github.com/make-all/tuya-local/files/13532139/Query.Things.Data.Model.json)

Some screenshots of the Home Assistant entities:

image
image
image
image
image
image


cart3a:
Maybe a dumb comment, as this is my first pull request ever.

I tried to run the python tests on the forked repo, before adding the yaml file for my thermostat, and they fail the same.
Is there something I should do?

make-all:
I think something changed in one of the underlying libraries. The last two PRs have failed but it does not seem to be caused by any error in the configs.

PR #1385, opened 2023-12-08
Create wifi_smart_garage_door

illuzn:
This looks like a clone of the loratap (or perhaps the loratap is a clone of this). I've fleshed out the dps with descriptions of what they do for mine - not sure if loratap is the same.

The open/ close timer has been hidden - you can use an automation in home assistant to do this anyway. I've only left it in to help with device recognition (not sure if this is the correct approach).

Also, not sure if this is possible to implement but it would be good with devices like this that are supposed to attach to "dumb" garages to use the travel time. Here's the issue with the current implementation:
- The switch is actually just a dumb switch that starts/ stops the garage motor i.e. pressing on will open/ close/ stop the motor depending on the current state.
- With the way it's currently implemented, say the garage detects something is in the way while closing (using its maximum force sensor or an IR sensor), it will automatically reverse direction and open fully.
- If this occurs HA will be stuck in a permanently "closing" state and the controls will not work.

This would also be good so that a stop feature can be implemented (just in case there is an obstruction and you do need to stop it quickly).


custom_components/tuya_local/devices/wifi_smart_garage_door.yaml
```diff
@@ -0,0 +1,65 @@
...
+        unit: s
+        range:
+            min: 10
+            max: 120
+        mapping:
+          - step: 1
+  - entity: number
+    name: Timer
+    hidden: true
+    icon: mdi:timer
```
make-all, line 39: When there is a colon in the value, it needs to be quoted to avoid yaml parsing errors.

custom_components/tuya_local/devices/wifi_smart_garage_door.yaml
```diff
@@ -0,0 +1,65 @@
...
+    dps:
+      - id: 7
+        name: value
+        type: integer
+        unit: s
+        range:
+          min: 0
+          max: 86400
+        mapping:
+          - step: 1
```
make-all, line 50: Is it really useful to have second precision on this timer?  The app probably has hourly or half hourly, and generally in the configs I scale them to minutes.

custom_components/tuya_local/devices/wifi_smart_garage_door.yaml
```diff
@@ -0,0 +1,65 @@
...
+    dps:
+      - id: 103
+        name: value
+        type: integer
+        unit: min
+        range:
+          min: 0
+          max: 1440
+        mapping:
+          - scale: 60
```
make-all, line 64: The representation here is of a timer that goes up to 24 minutes, which you are scaling to allow setting with minute resolution.  Is that correct, or you intended to scale it to hour resolution, or the actual timer is in seconds up to 24 hours?

The range should be the range of the raw dps values. The unit should be the unit after scaling is applied (for the value that is presented to HA).

make-all:
I will modify the loratap config rather than add a new one, but first there are a few queries and tips for future reference.

custom_components/tuya_local/devices/wifi_smart_garage_door.yaml
illuzn: Oops dumb typo

custom_components/tuya_local/devices/wifi_smart_garage_door.yaml
illuzn: The app has it by the minute... Should be scaled as you say

custom_components/tuya_local/devices/wifi_smart_garage_door.yaml
illuzn: Oops silly typo

illuzn:
> I will modify the loratap config rather than add a new one, but first there are a few queries and tips for future reference.

Shall I close pull request?

Have made commits to fix those silly typos/ logic. Sorry, I'm very new to this.

make-all:
I'll close it once the loratap modifications are committed. For now, leave it open for tracking.

make-all:
You don't seem to have pushed the commit, so I can't see what the final resolution is for that last timer, which had me confused as to what actually was meant.

make-all:
For the logic changes, it is quite difficult as it involves tracking previous state and timings, while this integration tries to stick to a simple mapping of device values to HA values.  It is a shame so many Tuya cover devices are implemented so simply without proper feedback, and expect the app to use timers to keep track of the closing and opening times - at least this one has a dp that tells the opening/closing time though.

illuzn:
Oops, I stepped out for coffee before hitting commit... done now.

I'll raise an issue regarding my two suggested fixes for the logic.

PR #1397, opened 2023-12-10
Add support for Cleverio AR10 / AR20 panel heaters

InfectedSoap:
Add support for Cleverio AR10 / AR20 panel heaters from the Swedish retailer Kjell & Co. 

Both models identify as “Cleverio AR10 / AR20” in Tuya. The only difference between them is the wattage (1000W vs 2000W) and the physical size. 

IMG_4321

AR10: https://www.kjell.com/se/produkter/smarta-hem/temperaturstyrning/smarta-element/cleverio-smart-element-1000-w-p40720

IMG_4319

AR20: https://www.kjell.com/se/produkter/smarta-hem/temperaturstyrning/smarta-element/cleverio-smart-element-2000-w-p40721

IMG_4320

### DIFF
--- a/custom_components/tuya_local/devices/cleverio_ar10_ar20.yaml
+++ b/custom_components/tuya_local/devices/cleverio_ar10_ar20.yaml
@@ -1,0 +1,166 @@
+name: Cleverio AR10 / AR20
+products:
+  - id: urxauhwqufusih5h
+    name: Cleverio AR10 / AR20
+primary_entity:
+  entity: climate
+  dps:
+    - id: 1
+      name: hvac_mode
+      type: boolean
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: heat
+    - id: 2
+      name: temperature
+      type: integer
+      range:
+        min: 5
+        max: 45
+      mapping:
+        - constraint: temperature_unit
+          conditions:
+            - dps_val: F
+              range:
+                min: 41
+                max: 113
+    - id: 3
+      name: current_temperature
+      type: integer
+    - id: 4
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: low
+          value: eco
+        - dps_val: high
+          value: boost
+    - id: 12
+      name: fault_code
+      type: bitfield
+    - id: 101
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/cleverio_ar10_ar20.yaml
```diff
@@ -0,0 +1,166 @@
...
+                - dps_val: false
+                  value: idle
+                - dps_val: true
+                  value: heating
+    - id: 103
+      name: temperature_unit
+      type: string
+      mapping:
+        - dps_val: C
+          value: TEMP_CELSIUS
```
make-all, line 62: These should just be C and F, so no mapping is needed (there is no recognition of the Python constant names from the source code, only the final values)

custom_components/tuya_local/devices/cleverio_ar10_ar20.yaml
```diff
@@ -0,0 +1,166 @@
+name: Cleverio AR10 / AR20
```
make-all, line 1: Ue a generic name like "Panel heater" at the top level in case other devices match (many Tuya devices are whitelabel products that are rebranded as local brands, so the same device can have different names in different markets).  Leave the brand under products, as normally the product id will be changed even if other firmware is not (though there are cases where different brands use the same product id too, but if we find that we will adjust to list both).

custom_components/tuya_local/devices/cleverio_ar10_ar20.yaml
InfectedSoap: Ah gotcha. I got the values from here: https://www.home-assistant.io/docs/configuration/basic/

For some reason the when I just had C and F the temperature would always display the current temperature in Celsius regardless of what I had selected. I realize now that Home Assistant will convert the temperature automatically based on my geolocation:
https://www.home-assistant.io/docs/configuration/basic/

custom_components/tuya_local/devices/cleverio_ar10_ar20.yaml
InfectedSoap: Will do!

PR #1400, opened 2023-12-11
 Added support for Prosto PTC 2000 Heater

kinkamez:


custom_components/tuya_local/devices/prosto_ptc2000_heater.yaml
```diff
@@ -0,0 +1,91 @@
...
+        max: 45
+    - id: 3
+      type: integer
+      name: current_temperature
+    - id: 5
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: "level_0"
+          value: fan
```
make-all, line 34: You could use the predefined "none" preset for fan, otherwise capitalise it, as fan is not a key for a translation string.

custom_components/tuya_local/devices/prosto_ptc2000_heater.yaml
```diff
@@ -0,0 +1,91 @@
...
+        - dps_val: "level_1"
+          value: comfort
+        - dps_val: "level_2"
+          value: boost
+    - id: 4
```
make-all, line 39: level is not part of the climate API, so this probably needs to be a separate select entity.

PR #1408, opened 2023-12-12
Add configuration for Argo DRY PURY EVO WF dehumidifier

alirow:
Support for Argo DRY PURY EVO WF 21 https://argoclima.com/en/prodotti/argo-dry-pury-evo-wf-21-21lt-gg-32ac80-u-r-2/

image
image
image

Works well but:
- no action on device when I press `Filter reset` in HA
- no action on device when I change `Temperature unit` in HA
- no action on device when I change `Timer` in HA
- I don't know how to check diagnostic values
- I don't know what values should I use to localise the labels in HA

### DIFF
--- a/custom_components/tuya_local/devices/dry_pury_evo_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/dry_pury_evo_dehumidifier.yaml
@@ -1,0 +1,160 @@
+name: Argo DRY PURY EVO WF dehumidifier
+products:
+  - id: oya6k3fqen43zyi7
+    name: Argo DRY PURY EVO WF 21
+primary_entity:
+  entity: humidifier
+  class: dehumidifier
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+    - id: 2
+      name: humidity
+      type: integer
+      range:
+        min: 30
+        max: 80
+      mapping:
+        - step: 5
+    - id: 5
+      name: mode
+      type: string
+      mapping:
+        - dps_val: dehumidify
+          value: comfort
+        - dps_val: drying
+          value: laundry
+          icon: "mdi:tshirt-crew-outline"
+          icon_priority: 3
+        - dps_val: sleep
+          value: sleep
+          icon: "mdi:power-sleep"
+          icon_priority: 3
+    - id: 6
+      name: current_humidity
+      type: integer
+secondary_entities:
+  - entity: fan
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+      - id: 5
+        name: dehumidifier_mode
+        type: string
+        hidden: true
+      - id: 4
+        name: speed
+        type: string
+        mapping:
+          - dps_val: low
+            value: 50
+            constraint: dehumidifier_mode
+            conditions:
+              - dps_val: laundry
+                invalid: true
+          - dps_val: high
+            value: 100
+            constraint: dehumidifier_mode
+            conditions:
+              - dps_val: sleep
+                invalid: true
+  - entity: sensor
+    name: Current humidity
+    # deprecated: humidifier      # without that I see humidity values and it's history
+    # category: diagnostic        # without that I see humidity values and it's history
+    class: humidity
+    dps:
+      - id: 6
+        name: sensor
+        type: integer
+        precision: 0
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 7
+        name: sensor
+        type: integer
+        unit: C
+        class: measurement
+  - entity: switch
+    name: Ionizer
+    icon: "mdi:creation"
+    optional: true
+    dps:
+      - id: 10
+        type: boolean
+        name: switch
+  - entity: lock
+    name: Child lock
+    category: config
+    dps:
+      - id: 16
+        type: boolean
+        name: lock
+        mapping:
+          - dps_val: false
+            icon: "mdi:hand-back-right"
+          - dps_val: true
+            icon: "mdi:hand-back-right-off"
+  - entity: select
+    name: Timer
+    category: config
+    dps:
+      - id: 17
+        type: string
+        name: option
+        mapping:
+          - dps_val: "cancel"
+            value: Continuous
+          - dps_val: 1h
+            value: 1 hour
+          - dps_val: 2h
+            value: 2 hours
+          - dps_val: 3h
+            value: 3 hours
+  - entity: binary_sensor # based on API doc. I don't know how to test
+    name: Fault
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 19
+        name: fault
+        type: bitfield
+        mapping:
+          - dps_val: 0
+            value: E1
+          - dps_val: 1
+            value: CL
+          - dps_val: 2
+            value: CH
+          - dps_val: 3
+            value: LO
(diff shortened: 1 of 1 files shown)


alirow:
**Standard Instruction Set:**
```
switch | Boolean | "{true,false}"
dehumidify_set_value | Integer | {   "unit": "%",   "min": 30,   "max": 80,   "scale": 0,   "step": 5 }
fan_speed_enum | Enum | {   "range": [     "low",     "high"   ] }
mode | Enum | {   "range": [     "dehumidify",     "drying"   ] }
anion | Boolean | "{true,false}"
child_lock | Boolean | "{true,false}"
countdown_set | Enum | {   "range": [     "cancel",     "1h",     "2h",     "3h"   ] }
filter_reset | Boolean | "{true,false}"
temp_unit_convert | Enum | {   "range": [     "c",     "f"   ] }
runtime_total_reset | Boolean | "{true,false}"

```

**Standard Status Set**
```
switch | Boolean | "{true,false}"
dehumidify_set_value | Integer | {   "unit": "%",   "min": 30,   "max": 80,   "scale": 0,   "step": 5 }
fan_speed_enum | Enum | {   "range": [     "low",     "high"   ] }
mode | Enum | {   "range": [     "dehumidify",     "drying"   ] }
humidity_indoor | Integer | {   "unit": "%",   "min": 20,   "max": 90,   "scale": 0,   "step": 1 }
temp_indoor | Integer | {   "unit": "℃",   "min": 5,   "max": 38,   "scale": 0,   "step": 1 }
anion | Boolean | "{true,false}"
child_lock | Boolean | "{true,false}"
countdown_set | Enum | {   "range": [     "cancel",     "1h",     "2h",     "3h"   ] }
fault | Bitmap | {   "label": [     "E1",     "CL",     "CH",     "LO",     "FL",     "Cleaning"   ] }
filter_reset | Boolean | "{true,false}"
filter_life | Integer | {   "unit": "%",   "min": 0,   "max": 100,   "scale": 0,   "step": 1 }
temp_unit_convert | Enum | {   "range": [     "c",     "f"   ] }
runtime_total_reset | Boolean | "{true,false}"
```

**DPS**
```
- name: "power"
  code: 1  
- name: "Set Dehumidify"
  code: 2
- name: "Speed"
  code: 4
- name: "Mode"
  code: 5
- name: "Indoor Humidity"
  code: 6
- name: 68  #Typ sprzętu
  code: 102 # 68
- name: "Indoor Temp"
  code: 7
- name: "Purify"
  code: 10
- name: "Child Lock"
  code: 16
- name: "Countdown"
  code: 17
- name: "Fault"
  code: 19
- name: "Filter Reset"
  code: 20
- name: "Filter Life"
  code: 23
- name: "Reset"
  code: 28
```

make-all:
You can update pull requests by pushing changes to the same branch.  There is no need to create a new PR for each change.

alirow:
> You can update pull requests by pushing changes to the same branch. There is no need to create a new PR for each change.

Thanks. I'll leave it as is for now and apply this advice in the future :)

custom_components/tuya_local/devices/dry_pury_evo_dehumidifier.yaml
```diff
@@ -0,0 +1,160 @@
+name: Argo DRY PURY EVO WF dehumidifier
```
make-all, line 1: Just use "Dehumidifer" for this top level name, in case the same device is sold under different brands elsewhere.

custom_components/tuya_local/devices/dry_pury_evo_dehumidifier.yaml
```diff
@@ -0,0 +1,160 @@
...
+        - step: 5
+    - id: 5
+      name: mode
+      type: string
+      mapping:
+        - dps_val: dehumidify
+          value: comfort
+        - dps_val: drying
+          value: laundry
+          icon: "mdi:tshirt-crew-outline"
```
make-all, line 28: If you define value specific icons here, you will need to define a default icon as well (most likely a higher priority (lower number) icon for off, and a lower priority (higher number) for on)

custom_components/tuya_local/devices/dry_pury_evo_dehumidifier.yaml
```diff
@@ -0,0 +1,160 @@
...
+            conditions:
+              - dps_val: laundry
+                invalid: true
+          - dps_val: high
+            value: 100
+            constraint: dehumidifier_mode
+            conditions:
+              - dps_val: sleep
+                invalid: true
+  - entity: sensor
```
make-all, line 63: Don't add deprecated sensors to new configs.
You can see the humidity values and history in the humidifier entity.

custom_components/tuya_local/devices/dry_pury_evo_dehumidifier.yaml
```diff
@@ -0,0 +1,160 @@
...
+    name: Fault
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 19
+        name: sensor
+        type: bitfield
+        mapping:
+          - dps_val: 0
+            value: E1
```
make-all, line 129: binary_sensor values should map to either true or false.
Normally 0 is false, and the default value for all others is true.  If the dp is optional (not always reported), you may need to add a mapping from null to false as well.

To keep access to the raw error values, add dp 19 under the humidifier entity with name fault_code as well.

make-all: Usually I would map any normal operation values to false in the Fault sensor, and make a separate binary_sensor for those (commonly for dehumidifiers there is a "tank full" indication that this would apply to).

custom_components/tuya_local/devices/dry_pury_evo_dehumidifier.yaml
```diff
@@ -0,0 +1,160 @@
...
+            value: CL
+          - dps_val: 2
+            value: CH
+          - dps_val: 3
+            value: LO
+          - dps_val: 4
+            value: FL
+          - dps_val: 5
+            value: Cleaning
+  - entity: button # based on API doc. Nothing happened when I run on my device
```
make-all, line 140: This is maybe related to one of the fault indicators.

custom_components/tuya_local/devices/dry_pury_evo_dehumidifier.yaml
```diff
@@ -0,0 +1,160 @@
...
+    name: Filter reset
+    category: config
+    optional: true
+    dps:
+      - id: 20
+        type: boolean
+        name: button
+  - entity: select # based on API doc. Nothing happened when I run on my device
```
make-all, line 148: If it does not do anything, and the device is not reporting it, then just leave it out of the config.

custom_components/tuya_local/devices/dry_pury_evo_dehumidifier.yaml
```diff
@@ -0,0 +1,160 @@
...
+          - dps_val: 3
+            value: LO
+          - dps_val: 4
+            value: FL
+          - dps_val: 5
+            value: Cleaning
+  - entity: button # based on API doc. Nothing happened when I run on my device
+    name: Filter reset
+    category: config
+    optional: true
```
make-all, line 143: `optional: true` is at the dps level, as it is the dp that is optional.

PR #1414, opened 2023-12-13
Update klarstein_dryfy_pro_connect_dehumidifier.yaml

ChriZathens:
Thanks for adding the Morris dehumidifier after all. 
The ionizer changes are inverted though.. 
I don't know if in the klarstein one they are supposed to be like this, or if it is a bug. 
With the Morris dehumidifier and the original values, though, in HA, when Ionizer is off, it shows as on and vise versa

### DIFF
--- a/custom_components/tuya_local/devices/klarstein_dryfy_pro_connect_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/klarstein_dryfy_pro_connect_dehumidifier.yaml
@@ -119,9 +119,9 @@ secondary_entities:
         optional: true
         mapping:
           - dps_val: false
-            value: true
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/klarstein_dryfy_pro_connect_dehumidifier.yaml
```diff
@@ -119,9 +119,9 @@ secondary_entities:
         optional: true
         mapping:
```
make-all, line 120: You can remove the mapping entirely if the values do not need to be mapped to different ones.  I'm not sure why the original author of the config switched them like this, it does seem strange.

Issue #1410, opened 2023-12-12
Request support for Hama Thermostat


andrei-cerbulescu:
# Log Message

```
Device matches None with quality of 0%. DPS: {"updated_at": 1702411739.984545, "103": 160}
Device matches None with quality of 0%. DPS: {"updated_at": 1702411798.5560288, "103": 160}
Device matches None with quality of 0%. DPS: {"updated_at": 1702412186.4454596, "103": 160}

The mapping I get from tinytuya is as follows:
 "mapping": {
            "101": {
                "code": "switch",
                "type": "Boolean",
                "values": {}
            },
            "102": {
                "code": "temp_current",
                "type": "Integer",
                "values": {
                    "unit": "\u00b0C",
                    "min": 0,
                    "max": 700,
                    "scale": 1,
                    "step": 1
                }
            },
            "103": {
                "code": "temp_set",
                "type": "Integer",
                "values": {
                    "unit": "\u00b0C",
                    "min": 50,
                    "max": 300,
                    "scale": 1,
                    "step": 1
                }
            }
        }
```

The actual gateway has the following fields:
```
{
  "result": {
    "properties": [
      {
        "code": "upward",
        "custom_name": "",
        "dp_id": 101,
        "time": 1691421889625
      },
      {
        "code": "down",
        "custom_name": "",
        "dp_id": 102,
        "time": 1691421889625
      }
    ]
  },
  "success": true,
  "t": 1702418139744,
  "tid": "30405d35993911ee9c9e428ec26fb7a3"
}
```

# Information about DPS mappings

```
{
  "result": {
[...]

PR #1415, opened 2023-12-13
Added support for Hama Heater Controller

andrei-cerbulescu:
Fixed issue: #1410 

image

### DIFF
--- a/custom_components/tuya_local/devices/hama_radiator_controller.yaml
+++ b/custom_components/tuya_local/devices/hama_radiator_controller.yaml
@@ -1,6 +1,6 @@
 name: Hama Radiator Controller
 products:
-  - id: 176592/176593
+  - id: yw7cahqs
     name: Hama Radiator Controller
 primary_entity:
   entity: climate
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/hama_radiator_controller.yaml
```diff
@@ -0,0 +1,32 @@
+name: Hama Radiator Controller
+products:
+  - id: 176592/176593
```
make-all, line 3: This is not the normal format for a Tuya product_id.  If you took this from iot.tuya.com, the value `product_id` under "Device Management" / "Query Device Details" is what should go here.

custom_components/tuya_local/devices/hama_radiator_controller.yaml
```diff
@@ -0,0 +1,32 @@
...
+          icon: "mdi:radiator"
+        - dps_val: false
+          value: "off"
+          icon: "mdi:radiator-disabled"
+    - id: 102
+      name: current_temperature
+      type: integer
+      mapping:
+        - scale: 10
+          step: 1
```
make-all, line 23: Step is not needed for read-only values, whatever comes from the device will always be used (and anyway 1 is the default for step)

PR #1418, opened 2023-12-15
add GratKit filament dryer

blakadder:
Adds GratKit Firefly smart filament dryer

dp101 with time remaining is repeated as sensor and number input so its easier to read time remaining since there's no datetime input option for number entities

### DIFF
--- a/custom_components/tuya_local/devices/gratkit_filament_dryer.yaml
+++ b/custom_components/tuya_local/devices/gratkit_filament_dryer.yaml
@@ -1,4 +1,4 @@
-name: GratKit filament dryer
+name: Filament dryer
 products:
   - id: mxct11k2jqpak13a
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/gratkit_filament_dryer.yaml
```diff
@@ -0,0 +1,199 @@
...
+primary_entity:
+  entity: switch
+  icon: "mdi:heat-pump"
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+secondary_entities:
+  - entity: number
+    name: Drying temperature
```
make-all, line 14: I wonder if the switch, drying temperature and temperature sensor should be combined into a climate entity, to allow the use of the climate card for control.

PR #1421, opened 2023-12-17
Add Thermoval TVT 40 WiFi Thermostat

madpeteguy:
Home: https://thermoval.pl/en/contact-en/
Catalog: https://thermoval.pl/wp-content/uploads/2023/09/THERMOVAL-Cennik-ENG.pdf

Screenshots:
obraz
obraz
obraz
obraz
obraz


custom_components/tuya_local/devices/thermoval_tvt40_thermostat.yaml
```diff
@@ -0,0 +1,315 @@
...
+          value: program
+          icon: "mdi:home-thermometer"
+        - dps_val: Anti_frozen
+          value: away
+          icon: "mdi:snowflake-melt"
+        - dps_val: Manual
+          value: manual
+          icon: "mdi:cursor-pointer"
+    - id: 4
+      name: fault
```
make-all, line 32: I would call this fault_code, and add a binary_sensor for it as well, that maps 0 to false, and any other value to true.  The binary sensor should be more convenient to use in automations and for an indication in the UI that there is a fault, and the attribute here gives access to the full fault code for reference.

custom_components/tuya_local/devices/thermoval_tvt40_thermostat.yaml
```diff
@@ -0,0 +1,315 @@
...
+      type: integer
+    - id: 5
+      name: room_temperature
```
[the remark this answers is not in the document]
make-all: You may want to make this conditionally redirect to floor_temperature when sensor_select is "0", the config will be more complex, but the UI might make more sense in terms of offset between the set temperature and current temperature.  If you do this, then keep the room temperature separate sensor, so that it is always available regardless of the sensor select setting.

custom_components/tuya_local/devices/thermoval_tvt40_thermostat.yaml
```diff
@@ -0,0 +1,315 @@
...
+      type: integer
+      mapping:
+        - scale: 10
+    - id: 6
+      name: floor_temp_limit
+      type: integer
+      mapping:
+        - scale: 10
+    - id: 7
```
make-all, line 44: Attributes that are not used in the climate platform, and have separate entities under secondary_entities should not be listed here.  Basically the attributes listed under climate should fall under one of the following:
1. standard attributes of the climate platform
2. attributes that are used in conditions to create advanced behavior
3. read-only attributes that do not make sense as separate entities on their own

custom_components/tuya_local/devices/thermoval_tvt40_thermostat.yaml
```diff
@@ -0,0 +1,315 @@
...
+          value: Home
+    - id: 8
+      name: floor_temperature
+      type: integer
+      mapping:
+        - scale: 10
+    - id: 9
+      name: hvac_mode
+      type: boolean
+      readonly: true
```
make-all, line 60: hvac_mode should not be read-only.  If this is really read-only, rename this attribute (I guess the original tuya name is work_mode, which would be appropriate) to avoid buttons that cannot be used in the UI (the mode will then always show as auto).

madpeteguy:
@make-all Thanks for review!

It's been six months since I created this configuration, I've forgotten a lot. 🤯 
But I got tired of catching up with own configuration file after each plugin update, although I could mount it via docker.
I'll check with the suggestions when I find some free time, probably after the new year.

Anyway, it's based on another device's configuration that someone provided somewhere as a workaround, but I can't find that entry. The replacement worked fine with the basic functionality of the thermostat.
I upgraded it a bit to make most of the functionality of the TVT 40 device available in HA.
Mainly displaying cooling mode (without hacks like overriding hvac_mode it always displayed heating, if i remember correctly it had problems with writing to id:9 so read only seemed right).
Also temperature reading for sensor_select = "2" with limits for both sensors, where one has a common weekly schedule in auto mode or a fixed value in manual mode, and a separate value for the other sensor.

Thermostate cards do not support such a combo, I'm fine with apexchart:
obraz
It is controlled with entities card (i think it might be the reason for sensor duplication):
obraz

Thermostate card hide so much info...
obraz

make-all (commented):
Fixed most review comments

PR #1423, opened 2023-12-17
Create ustellar_ut99911_rgb.yaml

neilmartin83:
Add support for Ustellar UT99911-RGB Smart Monitor Light Bar

Fixes issue: #1383 

Tested with physical device.

### DIFF
--- a/custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
+++ b/custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
@@ -1,10 +1,10 @@
-name: Smart Monitor Light Bar
+name: Monitor light bar
 products:
   - id: oqtk22eg6jxorl78
     name: Ustellar UT99911-RGB
 primary_entity:
   entity: light
-  name: Front LED (White)
+  name: Front White LED
   icon: "mdi:wall-sconce-flat-outline"
   dps:
     - id: 101
@@ -30,7 +30,7 @@ primary_entity:
         max: 1000
 secondary_entities:
   - entity: light
-    name: Rear LED (RGB)
+    name: Rear RGB LED
     icon: "mdi:wall-sconce-flat-variant-outline"
     dps:
       - id: 20
@@ -70,9 +70,6 @@ secondary_entities:
       - id: 103
         name: line_sequence_adjustment
         type: integer
-      - id: 104
-        name: strip_points
-        type: integer
   - entity: select
     name: Music
     icon: "mdi:playlist-music"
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
```diff
@@ -0,0 +1,174 @@
+name: Smart Monitor Light Bar
+products:
+  - id: oqtk22eg6jxorl78
+    name: Ustellar UT99911-RGB
+primary_entity:
+  entity: light
+  name: Front LED (White)
```
make-all, line 7: I don't recommend brackets in names, they may complicate the entity ids.  If you need the info, then "Front white LED" and "Rear RGB LED" would be better.

custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
```diff
@@ -0,0 +1,174 @@
+name: Smart Monitor Light Bar
```
make-all, line 1: HA naming convention is to only capitalise the first word of device and entity names.  Also, this will be prepended to every entity id, so it is best to keep it short by dropping redundant words like "Smart", "WiFi" etc from the actual product name.

custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
```diff
@@ -0,0 +1,174 @@
...
+              max: 1000
+          - name: v
+            bytes: 2
+            range:
+              min: 0
+              max: 1000
+      - id: 103
+        name: line_sequence_adjustment
+        type: integer
+      - id: 104
```
make-all, line 73: This can be removed, as it is listed below as a separate entity, and is not part of the light definition. (line_sequence_adjustment on the other hand does not appear elsewhere, so if it doesn't make sense as a separate entity can stay as an extra attribute here for reference).

custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
neilmartin83: Renamed, thank you.

custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
neilmartin83: Renamed, thank you.

custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
neilmartin83: Removed 104 - left 103 in there. This was based on the blitzwolf config but no need for that attribute to exist.

neilmartin83:
Also made a few attributes optional as noticed they weren't always returned by the device when polled, meaning it wasn't detected reliably.

PR #1424, opened 2023-12-18
Create hombli_thermostatic_radiator_valve.yaml

MarkPaxton:
Adding device profile for Hombli TRV

### DIFF
--- a/custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
+++ b/custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
@@ -5,25 +5,6 @@ products:
 primary_entity:
   entity: climate
   dps:
-    - id: 8
-      type: boolean
-      name: window_check
-    - id: 10
-      type: boolean
-      name: frost
-      mapping:
-        - dps_val: true
-          value: "On"
-        - dps_val: false
-          value: "Off"
-    - id: 27
-      type: integer
-      name: temp_correction
-      mapping:
-        - scale: 10
-    - id: 40
-      type: boolean
-      name: child_lock
     - id: 101
       type: boolean
       name: hvac_mode
@@ -31,7 +12,7 @@ primary_entity:
         - dps_val: false
           value: "off"
         - dps_val: true
-          value: "auto"
+          value: auto
     - id: 102
       type: integer
       name: current_temperature
@@ -43,15 +24,95 @@ primary_entity:
       name: temperature
       mapping:
         - scale: 10
-    - id: 105
-      type: integer
-      name: battery
-    - id: 106
-      type: boolean
-      name: away
-    - id: 108
-      type: boolean
-      name: program
-    - id: 130
-      type: boolean
-      name: antiscale
+          max: 70
+          min: 0
+          step: 10
+secondary_entities:
+  - entity: switch
+    name: Window check
+    category: config
+    dps:
+      - id: 8
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Anti frost
+    category: config
+    dps:
+      - id: 10
+        type: boolean
+        name: switch
+        mapping:
+          - dps_val: true
+            value: "On"
+          - dps_val: false
+            value: "Off"
+  - entity: select
+    name: Temperature unit
+    category: config
+    dps:
+      - id: 23
+        type: string
+        name: option
+        mapping:
+          - dps_val: "c"
+            value: Celsius
+            icon: "mdi:temperature-celsius"
+          - dps_val: "f"
+            value: Fahrenheit
+            icon: "mdi:temperature-fahrenheit"
+  - entity: number
+    name: Temperature calibration
+    category: config
+    dps:
+      - id: 27
+        type: integer
+        name: temp_correction
+        mapping:
+          - scale: 10
+            min: -60
+            max: 60
+            step: 5
+  - entity: lock
+    name: Child lock
+    icon: "mdi:hand-back-right-off"
+    category: config
+    dps:
+      - id: 40
+        type: boolean
+        name: lock
+  - entity: binary_sensor
+    name: Battery
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 105
+        type: bitfield
+        name: sensor
+        class: measurement
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+  - entity: switch
+    name: Away
+    category: config
+    dps:
+      - id: 106
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Schedule
+    icon: "mdi:calendar-clock"
+    category: config
+    dps:
+      - id: 108
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Antiscale
+    category: config
+    dps:
+      - id: 130
+        type: boolean
+        name: switch
--- a/custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
+++ b/custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
@@ -76,6 +76,7 @@ secondary_entities:
         name: value
         mapping:
           - scale: 10
+        range:
             min: -60
             max: 60
             step: 5
--- a/custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
+++ b/custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
@@ -1,0 +1,58 @@
+name: Hombli Thermostatic Radiator Valve
+products:
+  - id: ngxyxyc8
+    name: Hombli HBRT-0109
+primary_entity:
+  entity: climate
+  dps:
+    - id: 8
+      type: boolean
+      name: window_check
+      icon: "mdi:window-open"
+    - id: 10
+      type: boolean
+      name: frost
+      mapping:
+        - dps_val: true
+          value: "On"
+        - dps_val: false
+          value: "Off"
+    - id: 27
+      type: integer
+      name: temp_correction
+      mapping:
+        - scale: 10
+    - id: 40
+      type: boolean
+      name: child_lock
+    - id: 101
+      type: boolean
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: "auto"
+    - id: 102
+      type: integer
+      name: current_temperature
+      readonly: true
+      mapping:
+        - scale: 10
+    - id: 103
+      type: integer
+      name: temperature
+      mapping:
+        - scale: 10
+    - id: 105
+      type: integer
+      name: battery
+    - id: 106
+      type: boolean
+      name: away
+    - id: 108
+      type: boolean
+      name: program
+    - id: 130
+      type: boolean
+      name: antiscale
--- a/custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
+++ b/custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
@@ -1,4 +1,4 @@
-name: Hombli Thermostatic Radiator Valve
+name: Radiator Thermostat
 products:
   - id: ngxyxyc8
     name: Hombli HBRT-0109
@@ -12,11 +12,10 @@ primary_entity:
         - dps_val: false
           value: "off"
         - dps_val: true
-          value: auto
+          value: heat
     - id: 102
       type: integer
       name: current_temperature
-      readonly: true
       mapping:
         - scale: 10
     - id: 103
@@ -24,6 +23,7 @@ primary_entity:
       name: temperature
       mapping:
         - scale: 10
+      range:
           max: 70
           min: 0
     - id: 106
@@ -37,6 +37,7 @@ primary_entity:
 secondary_entities:
   - entity: switch
     name: Window check
+    icon: "mdi:window-open-variant"
     category: config
     dps:
       - id: 8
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+products:
+  - id: ngxyxyc8
+    name: Hombli HBRT-0109
+primary_entity:
+  entity: climate
+  dps:
+    - id: 8
+      type: boolean
+      name: window_check
+      icon: "mdi:window-open"
```
make-all, line 11: Don't set the icon at dp level without a condition, it is no different than setting it at entity level, but causes confusion because another unfamiliar user will think it is related to the dp.  If window_check is not used in the climate entity, it should be its own binary_sensor or switch entity.

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - id: 10
+      type: boolean
+      name: frost
```
make-all, line 14: This should also be its own binary_sensor or switch entity.

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+      mapping:
+        - dps_val: true
+          value: "On"
+        - dps_val: false
+          value: "Off"
+    - id: 27
+      type: integer
+      name: temp_correction
```
make-all, line 22: This should be its own number entity

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+      mapping:
+        - scale: 10
+    - id: 40
+      type: boolean
+      name: child_lock
```
make-all, line 27: This should be its own lock entity

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - id: 101
+      type: boolean
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: "auto"
```
make-all, line 35: I would add a condition here on the program dp, so you can switch that through hvac_mode as well:
```
      - dps_val: true
        constraint: program
        conditions:
           - dps_val: true
             value: auto
           - dps_val: false
             value: heat
```

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - id: 102
+      type: integer
+      name: current_temperature
+      readonly: true
+      mapping:
+        - scale: 10
+    - id: 103
+      type: integer
+      name: temperature
```
make-all, line 44: Requires a range (min and max)

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+      mapping:
+        - scale: 10
+    - id: 105
+      type: integer
+      name: battery
```
make-all, line 49: Should be a separate sensor entity (class: battery, unit: "%")

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - id: 106
+      type: boolean
+      name: away
+    - id: 108
+      type: boolean
+      name: program
+    - id: 130
```
make-all, line 56: Should be a separate switch entity

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+        - scale: 10
+    - id: 103
+      type: integer
+      name: temperature
+      mapping:
+        - scale: 10
+    - id: 105
+      type: integer
+      name: battery
+    - id: 106
```
make-all, line 50: Name this "preset_mode", with true mapped to "away", and false mapped to "home", to expose it as preset_mode.

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,58 @@
...
+      type: boolean
+      name: away
+    - id: 108
+      type: boolean
+      name: program
```
make-all, line 55: If this is exposed through hvac_mode as above, you can add `hidden: true` to hide the raw value (saves a small amount of recorder memory)

MarkPaxton:
Thanks for the feedback - I'll make the changes and get back to you!

Mark

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
MarkPaxton: I've been looking into this an kind of disagree on the logic. The TRV has modes like this:

off: valve closed permanently
on: valve open always
auto: value open/closed according to set temperature threshold 
program: mode off/on/auto according to programmed schedule

I think 'heat' mode doesn't make a lot of sense as a mode, since it's actually a state of "valve open".

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
MarkPaxton: I'm a bit confused about this one, as the tuya capability report looks like below, so I don't think it's a direct integer mapping...

```
   "105": {
                    "code": "battery_power",
                    "type": "Bitmap",
                    "values": {
                        "label": [
                            "bit0"
                        ],
                        "maxlen": 1
                    }
                },
```

MarkPaxton:
This is the full DP set the device reports... just to help make things more clear... 
`"mapping": {
                "8": {
                    "code": "window_check",
                    "type": "Boolean",
                    "values": {}
                },
                "10": {
                    "code": "frost",
                    "type": "Boolean",
                    "values": {}
                },
                "23": {
                    "code": "temp_unit_convert",
                    "type": "Enum",
                    "values": {
                        "range": [
                            "c",
                            "f"
                        ]
                    }
                },
                "27": {
                    "code": "temp_correction",
                    "type": "Integer",
                    "values": {
                        "unit": "",
                        "min": -6,
                        "max": 6,
                        "scale": 0,
                        "step": 1
                    }
                },
                "40": {
                    "code": "child_lock",
                    "type": "Boolean",
                    "values": {}
                },
                "101": {
                    "code": "switch",
                    "type": "Boolean",
                    "values": {}
                },
                "102": {
                    "code": "current_temp",
                    "type": "Integer",
                    "values": {
                        "unit": "\u00b0C",
                        "min": 0,
                        "max": 700,
                        "scale": 1,
                        "step": 1
                    }
                },
                "103": {
                    "code": "set_temp",
                    "type": "Integer",
                    "values": {
                        "unit": "\u00b0C",
                        "min": 50,
                        "max": 300,
                        "scale": 1,
                        "step": 5
                    }
                },
                "104": {
                    "code": "motor_opening",
                    "type": "Integer",
                    "values": {
                        "unit": "%",
                        "min": 0,
                        "max": 100,
                        "scale": 1,
                        "step": 1
                    }
                },
                "105": {
                    "code": "battery_power",
                    "type": "Bitmap",
                    "values": {
                        "label": [
                            "bit0"
                        ],
                        "maxlen": 1
                    }
                },
                "106": {
                    "code": "a_key_home",
                    "type": "Boolean",
                    "values": {}
                },
                "107": {
                    "code": "prog_mode",
                    "type": "Raw",
                    "values": {}
                },
                "108": {
                    "code": "prog_switch",
                    "type": "Boolean",
                    "values": {}
                },
                "109": {
                    "code": "prog_data",
                    "type": "Raw",
                    "values": {}
                },
                "110": {
                    "code": "historical_day_set",
                    "type": "Raw",
                    "values": {}
                },
                "111": {
                    "code": "system_time",
                    "type": "Raw",
                    "values": {}
                },
                "112": {
                    "code": "historical_week_set",
                    "type": "Raw",
                    "values": {}
                },
                "113": {
                    "code": "historical_month_set",
                    "type": "Raw",
                    "values": {}
                },
                "114": {
                    "code": "historical_year_set",
                    "type": "Raw",
                    "values": {}
                },
                "115": {
                    "code": "historical_day_now",
                    "type": "Raw",
                    "values": {}
                },
                "116": {
                    "code": "historical_week_now",
                    "type": "Raw",
                    "values": {}
                },
                "117": {
                    "code": "historical_month_now",
                    "type": "Raw",
                    "values": {}
                },
                "118": {
                    "code": "historical_year_now",
                    "type": "Raw",
                    "values": {}
                },
                "119": {
                    "code": "historical_day_pow",
                    "type": "Raw",
                    "values": {}
                },
                "120": {
                    "code": "historical_week_pow",
                    "type": "Raw",
                    "values": {}
                },
                "121": {
                    "code": "historical_month_pow",
                    "type": "Raw",
                    "values": {}
                },
                "122": {
                    "code": "historical_year_pow",
                    "type": "Raw",
                    "values": {}
                },
                "123": {
                    "code": "prog_data_1",
                    "type": "Raw",
                    "values": {}
                },
                "124": {
                    "code": "prog_data_2",
                    "type": "Raw",
                    "values": {}
                },
                "125": {
                    "code": "prog_data_3",
                    "type": "Raw",
                    "values": {}
                },
                "126": {
                    "code": "prog_data_4",
                    "type": "Raw",
                    "values": {}
                },
                "127": {
                    "code": "prog_data_5",
                    "type": "Raw",
                    "values": {}
                },
                "128": {
                    "code": "prog_data_6",
                    "type": "Raw",
                    "values": {}
                },
                "129": {
                    "code": "prog_data_7",
                    "type": "Raw",
                    "values": {}
                },
                "130": {
                    "code": "water_sca",
                    "type": "Boolean",
                    "values": {}
                }
            }`

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
MarkPaxton: Away mode is program=off + set temp = 16 as well

MarkPaxton:
How it looks so far - have not yet got calibration to appear. Temperature unit is not reported. Battery is boolean: normal (+low?)

image

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
make-all: Seems more like a binary_sensor then.  Maybe leaving it as an attribute is better if the indication it is giving is not useful.

make-all:
"have not yet got calibration to appear" = probably related to the following:

`FAILED tests/test_device_config.py::TestDeviceConfig::test_config_files_parse - AssertionError: False is not true : hombli_thermostatic_radiator_valve.yaml missing required  value in number_temperature_calibration`

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,118 @@
...
+  - entity: number
+    name: Temperature calibration
+    category: config
+    dps:
+      - id: 27
+        type: integer
+        name: temp_correction
+        mapping:
+          - scale: 10
+            min: -60
```
make-all, line 80: Again, the min and max need to be inside a `range:` block, not `mapping:`

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
```diff
@@ -0,0 +1,118 @@
...
+  - entity: lock
+    name: Child lock
+    icon: "mdi:hand-back-right-off"
+    category: config
+    dps:
+      - id: 40
+        type: boolean
+        name: lock
+  - entity: binary_sensor
+    name: Battery
```
make-all, line 92: Does not need a name, as there is a matching class.

custom_components/tuya_local/devices/hombli_thermostatic_radiator_valve.yaml
make-all: - off: valve closed permanently **= HVACMode.OFF**
- on: valve open always *HA Climate platform has no equivalent*
- auto: value open/closed according to set temperature threshold **= HVACMode.HEAT**
- program: mode off/on/auto according to programmed schedule **= HVACMode.AUTO**

make-all: I guess the options then are to implement dp 101 as switching between "off" and "heat", and implement "auto" and "program" by separate switches (I only actually see the switch for program though).  But I'm not sure of the behaviour of dp 101 in the case that program is enabled. Does it become a sensor that indicates the actual state of the valve? Or is it required to be on (dp 101 overrides dp 108), or does the state not matter (dp 108 overrides dp 101)?

MarkPaxton:
Thanks for making the final few changes… didn’t have time over the holidays! I’ll test it out over the next week or so… though I need to get the BLE gateway to work more reliably :(

MarkPaxton:
Hello, to report back on this, it all seems to be working reliably if I tick the "poll only" option on the config.

I do want to ask for assistance as there's adp 104 which represents the valve state 0=closed, 100=fully open. I'd like to get this and use it trigger my heating boiler on and off when open/closed.

I added the following as a secondary entity but not getting much happening:
  - entity: number
    name: Valve opening
    category: diagnostic
    dps:
      - id: 104
        type: integer
        name: value
        readonly: true
        range:
            max: 100
            min: 0

PR #1450, opened 2023-12-25
Add support for Eeese Anna dehumidifier

sevimuelli:
Based on the existing Eeese dehumidifiers, but needed a separate one to be recognised.

For some reason, my VS Code formatted the Acknowledgements.md, but only corrected some typos.

### DIFF
--- a/custom_components/tuya_local/devices/eeese_anna_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/eeese_anna_dehumidifier.yaml
@@ -1,0 +1,170 @@
+name: Eeese Anna dehumidifier
+primary_entity:
+  entity: humidifier
+  class: dehumidifier
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+      mapping:
+        - dps_val: false
+          icon: "mdi:air-humidifier-off"
+          icon_priority: 2
+        - dps_val: true
+          icon: "mdi:air-humidifier"
+          icon_priority: 4
+    - id: 2
+      name: humidity
+      type: integer
+      range:
+        min: 30
+        max: 80
+      mapping:
+        - step: 5
+    - id: 5
+      name: mode
+      type: string
+      mapping:
+        - dps_val: Manual
+          value: normal
+        - dps_val: Auto
+          value: auto
+        - dps_val: Laundrydrying
+          value: laundry
+          icon: "mdi:tshirt-crew"
+          icon_priority: 3
+        - dps_val: Sleeping
+          value: sleep
+          icon: "mdi:sleep"
+          icon_priority: 3
+        - dps_val: Purifying
+          value: purify
+          icon: "mdi:air-purifier"
+          icon_priority: 3
+    - id: 6
+      type: integer
+      name: current_humidity
+    - id: 19
+      type: bitfield
+      name: error
+      optional: true
+      mapping:
+        - dps_val: 4
+          icon: "mdi:cup-water"
+          icon_priority: 1
+secondary_entities:
+  - entity: sensor
+    name: Current humidity
+    deprecated: humidifier
+    category: diagnostic
+    class: humidity
+    dps:
+      - id: 6
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: binary_sensor
+    name: Tank
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 19
+        type: bitfield
+        name: sensor
+        optional: true
+        mapping:
+          - dps_val: 0
+            value: false
+          - dps_val: null
+            value: false
+          - value: true
+  - entity: switch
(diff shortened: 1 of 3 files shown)


custom_components/tuya_local/devices/eeese_anna_dehumidifier.yaml
```diff
@@ -0,0 +1,170 @@
...
+    - id: 19
+      type: bitfield
+      name: error
+      optional: true
+      mapping:
+        - dps_val: 4
+          icon: "mdi:cup-water"
+          icon_priority: 1
+secondary_entities:
+  - entity: sensor
```
make-all, line 56: Remove this deprecated sensor, it is not needed for compatibility in new configs.

custom_components/tuya_local/devices/eeese_anna_dehumidifier.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - id: 19
+        type: bitfield
+        name: sensor
+        optional: true
+        mapping:
+          - dps_val: 0
+            value: false
+          - dps_val: null
+            value: false
+          - value: true
```
make-all, line 81: Tank should be one fault code (4 based on the above error attribute, but if that was just copied from another config, confirm it first).
I would split this into two binary_sensors, one for the routine Tank full indication (true for dps_val: 4, otherwise false), and one for actual faults (false for 0, 4, null, true for other values)

custom_components/tuya_local/devices/eeese_anna_dehumidifier.yaml
```diff
@@ -0,0 +1,170 @@
...
+        - dps_val: Purifying
+          value: purify
+          icon: "mdi:air-purifier"
+          icon_priority: 3
+    - id: 6
+      type: integer
+      name: current_humidity
+    - id: 19
+      type: bitfield
+      name: error
```
make-all, line 49: Recently I try to standardise this naming to "fault_code" (and corresponding binary sensor called "Fault")

custom_components/tuya_local/devices/eeese_anna_dehumidifier.yaml
sevimuelli: I split it into Fault, Tank, Defrost, in accordance with the bitmap i got from the Tuya's Developer Portal.
`{
  "label": [
    "E1",
    "E2",
    "tankfull",
    "defrost"
  ]
}`, for dps_value 1+2, 4, 8.

custom_components/tuya_local/devices/eeese_anna_dehumidifier.yaml
sevimuelli: I'm not really sure which sensor do you mean here? Do you mean the current humidity under secondary entities?
`secondary_entities:
  - entity: sensor
    name: Current humidity
    deprecated: humidifier
    category: diagnostic
    class: humidity
    dps:
      - id: 6
        type: integer
        name: sensor
        unit: "%"
        class: measurement`

make-all: Yes

sevimuelli:
There is one error: `FAILED tests/test_device_config.py::TestDeviceConfig::test_config_files_parse - AssertionError: False is not true : eeese_anna_dehumidifier.yaml missing required  sensor in binary_sensor_tank`,

I expect the problem to be that I changed the name from sensor to fault_code also in the secondary entities. Does it need to be "sensor" or does it only have to be unique?

```
secondary_entities:
  - entity: binary_sensor
    name: Tank
    class: problem
    category: diagnostic
    dps:
      - id: 19
        type: bitfield
        name: fault_code
        optional: true
        mapping:
          - dps_val: 4
            value: true
          - value: false
  - entity: binary_sensor
    name: Defrost
    class: problem
    category: diagnostic
    dps:
      - id: 19
        type: bitfield
        name: fault_code
        optional: true
        mapping:
          - dps_val: 8
            value: true
          - value: false
  - entity: binary_sensor
    name: Fault
    class: problem
    category: diagnostic
    dps:
      - id: 19
        type: bitfield
        name: fault_code
        optional: true
        mapping:
          - dps_val: 0
            value: false
          - dps_val: null
            value: false
          - dps_val: 4
            value: false
          - dps_val: 8
            value: false
          - value: true
```

Can you please tell me what I need to do here?

make-all:
binary_sensor requires a dp named sensor, which will be the value of the sensor.  The name does not need to be the same when the dp is listed in multiple entities, so the same id can be listed as fault_code in the humidifier entity to make the raw value available as an extra attribute there, and also listed as the sensor dp of the binary_sensor entity.

sevimuelli:
Ok I understand. Thank you for your help

make-all:
Merged manually to workaround the conflict in doc file

PR #1454, opened 2023-12-26
Add support for 2 more Zigbee devices working through Tuya hub (Cover switch and Door/Window sensor)

wisemanny:
1. Added Tuya Smart Zigbee Door Sensor support via hub
2. Added support for Loratap QCSC400ZB-V2 covers switch

### DIFF
--- a/custom_components/tuya_local/devices/tuya_smart_zigbee_door_sensor.yaml
+++ b/custom_components/tuya_local/devices/tuya_smart_zigbee_door_sensor.yaml
@@ -1,0 +1,26 @@
+name: Tuya Smart Zigbee Door Sensor
+products:
+  - id: 7jIGJAymiH8OsFFb
+    name: Tuya Smart Zigbee Door Sensor
+primary_entity:
+  entity: binary_sensor
+  class: door
+  dps:
+    - id: 101
+      type: boolean
+      name: sensor
+secondary_entities:
+  - entity: sensor
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        optional: true
+        unit: ""
+        mapping:
+          - dps_val: low
+            value: 0
+          - dps_val: high
+            value: 500
--- a/custom_components/tuya_local/devices/tuya_smart_zigbee_door_sensor.yaml
+++ b/custom_components/tuya_local/devices/tuya_smart_zigbee_door_sensor.yaml
@@ -18,9 +18,10 @@ secondary_entities:
         type: integer
         name: sensor
         optional: true
-        unit: ""
+        unit: "%"
+        class: measurement
         mapping:
           - dps_val: low
             value: 0
           - dps_val: high
-            value: 500
+            value: 100
--- a/custom_components/tuya_local/devices/tuya_smart_zigbee_door_sensor.yaml
+++ b/custom_components/tuya_local/devices/tuya_smart_zigbee_door_sensor.yaml
@@ -20,8 +20,3 @@ secondary_entities:
         optional: true
         unit: "%"
         class: measurement
-        mapping:
-          - dps_val: low
-            value: 0
-          - dps_val: high
-            value: 100
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -432,6 +432,7 @@ of device.
 - LoraTap GDC100W garage door opener
 - LoraTap QCSC420W double curtain switch
 - LoraTap SC500W-V1 curtain switch (supports many other simple curtain/blind controllers)
+- LoraTap QCSC400ZB-V2 curtain switch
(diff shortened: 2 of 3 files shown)


custom_components/tuya_local/devices/tuya_smart_zigbee_door_sensor.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - id: 103
+        type: integer
+        name: sensor
+        optional: true
+        unit: ""
+        mapping:
+          - dps_val: low
+            value: 0
+          - dps_val: high
+            value: 500
```
make-all, line 26: 500 is incorrect, battery range is 0 - 100%.  Unit should also be "%", not "".

But given the granularity, this would be better as a binary_sensor with class "battery",  low mapped to true, and "high" mapped to false.

wisemanny: This is how it was reported by tinytuya:

```
    {
        "name": "Window 1",
        "id": "******",
        "key": "*******",
        "mac": "",
        "category": "mcs",
        "product_name": "Door Sensor",
        "product_id": "7jIGJAymiH8OsFFb",
        "biz_type": 18,
        "model": "",
        "sub": true,
        "icon": "https://images.tuyaeu.com/smart/icon/1524817291ohyty7jfc6kyokuqeyvnp14i_0.png",
        "uuid": "****",
        "node_id": "****",
        "mapping": {
            "101": {
                "code": "switch",
                "type": "Boolean",
                "values": {}
            },
            "103": {
                "code": "battery",
                "type": "Integer",
                "values": {
                    "unit": "",
                    "min": 0,
                    "max": 500,
                    "scale": 0,
                    "step": 1
                }
            }
        },
        "parent": "",
        "ip": "",
        "version": ""
    },

```

Let me know if I stilll should adjust it. Will do.

make-all: You need to look at the QueryThingsDataModel result really, don't trust any other source of info, as most sources are for the cloud protocol, which can differ from the local protocol.

>

wisemanny: @make-all I see. Thank you for the support. Here is what Things Data Model shows (queried via Tuya API Explorer):

```
{
  "modelId": "0000000d1z",
  "services": [
    {
      "actions": [],
      "code": "",
      "description": "",
      "events": [],
      "name": "默认服务",
      "properties": [
        {
          "abilityId": 101,
          "accessMode": "ro",
          "code": "prm_switch",
          "description": "上报门窗开关状态",
          "name": "开关",
          "typeSpec": {
            "type": "bool",
            "typeDefaultValue": false
          }
        },
        {
          "abilityId": 103,
          "accessMode": "ro",
          "code": "Battery",
          "description": "备注数据范围：0-30 低电量状态； 31-60中等电量；61-100正常；101-500 感叹号（非正常状态）；初始配网给105；另外电量小于10 ，告警推送",
          "name": "电池电量",
          "typeSpec": {
            "max": 500,
            "min": 0,
            "scale": 0,
            "step": 1,
            "type": "value",
            "typeDefaultValue": 0,
            "unit": ""
          }
        }
      ]
    }
  ]
}
```

And here is the translation for convenience:

"Remarks data range: 0-30 low power state; 31-60 medium power; 61-100 normal; 101-500 exclamation mark (abnormal state); initial distribution network is 105; in addition, if the power is less than 10, an alarm will be pushed",
           "name": "battery level",

Please let me know how do you recommand I adjust this?

make-all: It looks like it is an integer then, with unit "%", and values above 100% aree abnormal.

wisemanny: @make-all, I updated the file. Let me know if you see more issues.

custom_components/tuya_local/devices/tuya_smart_zigbee_door_sensor.yaml
```diff
@@ -0,0 +1,27 @@
...
+    category: diagnostic
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        optional: true
+        unit: "%"
+        class: measurement
+        mapping:
+          - dps_val: low
```
make-all, line 24: If this is an integer, it should not geet values "low" and "high".  I don't understand why this mapping is here.

wisemanny: @make-all learn as I go, so bear with me please. Updated.
If there is a doc that I am missing - please let me know. I checked with other files, seems like this is how it must be done.

PR #1457, opened 2023-12-27
Add support for Proscenic A9 Air Purifier

weltall:
Adds support for the prescenic Air Purifier A9.

Most sensors are normal aside:
* the display "light" which is inverted in what actually happens (true is off)
* the filter sensor is in hours and not days differently than shown in tuya IOT platform
* There is an sensor which has no meaning and just shows numbers like 100 000 001, seems a sort of bitfield in a string but couldn't figure out what it tries to expose, the A8 device description (already present) seems to have the same thing so i kept it the same way
* Finally technically the fan is a choice which has some speed settings and Auto and Sleep mode - so presets. I tried to map them to speed % in order to allow to use it more easily. only thing is a bit quirky is that i set 0 to auto making it only possible to set it as a preset as 0 would mean off usually. It works in home assistant but i didn't know how to map it to a fan speed aside that option.

immagine

immagine

### DIFF
--- a/custom_components/tuya_local/devices/proscenic_a9_airpurifier.yaml
+++ b/custom_components/tuya_local/devices/proscenic_a9_airpurifier.yaml
@@ -1,0 +1,149 @@
+name: Air Purifier
+products:
+  - id: giwqfeiezesbkkbe
+    name: Proscenic A9
+primary_entity:
+  entity: fan
+  icon: "mdi:air-purifier"
+  translation_key: fan_with_presets
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+    - id: 4
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: "Sleep"
+          value: sleep
+        - dps_val: "1"
+          value: low
+        - dps_val: "2"
+          value: medium
+        - dps_val: "3"
+          value: high
+        - dps_val: "4"
+          value: turbo
+        - dps_val: "Auto"
+          value: auto
+    - id: 4
+      name: speed
+      type: string
+      mapping:
+        - dps_val: "Sleep"
+          value: 20
+        - dps_val: "1"
+          value: 40
+        - dps_val: "2"
+          value: 60
+        - dps_val: "3"
+          value: 80
+        - dps_val: "4"
+          value: 100
+        - dps_val: "Auto"
+          value: 0
+    - id: 102
+      name: unknown_102
+      type: string
+secondary_entities:
+  - entity: sensor 
+    class: pm25
+    dps:
+      - id: 2
+        name: sensor
+        class: measurement
+        type: integer
+        unit: ugm3
+  - entity: light # this is inverted to what actually happens (false = light on)
+    category: config
+    name: Display
+    icon: "mdi:lightbulb-night"
+    dps:
+      - id: 8
+        name: switch
+        type: boolean
+        mapping:
+          - dps_val: false
+            value: true
+          - dps_val: true
+            value: false
+  - entity: sensor
+    name: Filter usage
+    category: diagnostic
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/proscenic_a9_airpurifier.yaml
```diff
@@ -0,0 +1,149 @@
...
+          value: low
+        - dps_val: "2"
+          value: medium
+        - dps_val: "3"
+          value: high
+        - dps_val: "4"
+          value: turbo
+        - dps_val: "Auto"
+          value: auto
+    - id: 4
```
make-all, line 29: Probably better to just use preset_mode, since all the options do not fit naturally into "speed"

custom_components/tuya_local/devices/proscenic_a9_airpurifier.yaml
```diff
@@ -0,0 +1,149 @@
...
+            value: "Medium"
+          - dps_val: "severe"
+            value: "Severe"
+  - entity: select
+    name: Timer
+    icon: "mdi:fan-clock"
+    category: config
+    dps:
+      - id: 101
+        type: integer
```
make-all, line 137: Is it really an integer, or a string?

custom_components/tuya_local/devices/proscenic_a9_airpurifier.yaml
weltall: Well, the negative part of that is that, aside not being able to do stuff like increase fan speed commands, is that it makes impossible to directly start with a preset mode. 

In order to power on the device, instead of directly selecting a speed or a preset, you need to to turn on the fan and then select a preset mode from the home assistant interface from what i can see.

Else it will still report last preset mode but it actually goes in auto, which updates only by reopening the interface. It does a similar thing with the speed when just turning on the device with the power button in the same interface but at least it updates the interface saying auto and you can directly select a speed and it will turn at that specific mode directly.

If that's preferable i can surely change it but seems to reduce functionality. maybe mapping auto to 1% or something like that instead of 0%?

make-all: fan.turn_on can take a preset_mode as an argument, and as of 2023.12.2, this integration will use that.

weltall: I tried it by commenting out the speed part:

```
primary_entity:
  entity: fan
  icon: "mdi:air-purifier"
  translation_key: fan_with_presets
  dps:
    - id: 1
      name: switch
      type: boolean
    - id: 4
      name: preset_mode
      type: string
      mapping:
        - dps_val: "Sleep"
          value: sleep
        - dps_val: "1"
          value: low
        - dps_val: "2"
          value: medium
        - dps_val: "3"
          value: high
        - dps_val: "4"
          value: turbo
        - dps_val: "Auto"
          value: auto
#    - id: 4
#      name: speed
#      type: string
#      mapping:
#        - dps_val: "Sleep"
#          value: 20
#        - dps_val: "1"
#          value: 40
#        - dps_val: "2"
#          value: 60
#        - dps_val: "3"
#          value: 80
#        - dps_val: "4"
#          value: 100
#        - dps_val: "Auto"
#          value: 1
    - id: 102
      name: unknown_102
      type: string
```

but it has that behaviour:
if i choose a preset from HA nothing happens, if i turn on then the fan manually it keeps showing the preset i choose, but didn't actually set it and the device wakes up in auto mode. after several seconds it shows auto again. If I'm not doing anything wrong this pretty much doesn't make possible to directly turn on the fan on a specific preset.


https://github.com/make-all/tuya-local/assets/1127663/57ff8f19-0c38-4911-a98c-b3a5619055a3

PR #1460, opened 2023-12-27
Add support for Arlec PA1123BKHA portable air conditioner

tataihono:
portable ac
[Add support for Arlec PA1123BKHA portable air conditioner](https://www.bunnings.co.nz/arlec-3-3kw-smart-portable-aircon_p0453765)

- Mobile Wifi control Heating & Cooling reverse switch
- Touch control panel with LED display
- 3 fan speed with Remote control
- 3.3kW cooling & 2.6kW Heating Power

### DIFF
--- a/custom_components/tuya_local/devices/arlec_pa1123bkha_portable_air_conditioner.yaml
+++ b/custom_components/tuya_local/devices/arlec_pa1123bkha_portable_air_conditioner.yaml
@@ -9,10 +9,10 @@ primary_entity:
       name: hvac_mode
       type: boolean
       mapping:
-        - dps_val: false
+        - dps_val: False
           value: "off"
           icon: "mdi:hvac-off"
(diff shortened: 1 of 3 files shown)


custom_components/tuya_local/devices/arlec_pa1123bkha_portable_air_conditioner.yaml
```diff
@@ -0,0 +1,134 @@
+name: Arlec PA1123BKHA portable air conditioner
```
make-all, line 1: Leave the branding off this top level name, in case other products match (due to the same device being sold in other markets with different branding). So just "Portable air conditioner" here, and "Arlec PA1123BKHA" in the products section below.

make-all:
Manually merged with conflict resolution in ACKNOWLEDGEMENTS.md

PR #1461, opened 2023-12-27
Air quality sensor - Air housekeeper (6 in 1)

kkot86:
Added temperature and air quality sensor - Air housekeeper (6 in 1). 
Tuya product_id: ugjbnqj2mffaexp5

Measurement support:
1) Temperatures
2) Humidity
3) PM2.5
4) VOC
5) CH2O
6) CO2

### DIFF
--- a/custom_components/tuya_local/devices/airquality6in1.yaml
+++ b/custom_components/tuya_local/devices/airquality6in1.yaml
@@ -1,20 +1,16 @@
-name: Air quality monitor 6 in 1
+name: Air quality monitor
 products:
   - id: ugjbnqj2mffaexp5
-    name: Air Housekeeper
+    name: Air Housekeeper 6-in-1
 primary_entity:
   entity: sensor
-  name: PM2.5
-  class: volatile_organic_compounds
-  icon: "mdi:factory"
+  class: pm25
   dps:
     - id: 2
       type: integer
       name: sensor
       unit: µg/m³
       class: measurement
-      mapping:
-        - scale: 1
 secondary_entities:
   - entity: sensor
     class: temperature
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/airquality6in1.yaml
```diff
@@ -0,0 +1,70 @@
+name: Air quality monitor 6 in 1
+products:
+  - id: ugjbnqj2mffaexp5
+    name: Air Housekeeper
+primary_entity:
+  entity: sensor
+  name: PM2.5
+  class: volatile_organic_compounds
```
make-all, line 8: class for a PM2.5 sensor should be pm25, and then it will not need a name, as the class defines a translated name already.

custom_components/tuya_local/devices/airquality6in1.yaml
```diff
@@ -0,0 +1,70 @@
...
+  icon: "mdi:factory"
```
make-all, line 9: Also should not need an icon, as the class should come with an appropriate icon as well.

custom_components/tuya_local/devices/airquality6in1.yaml
```diff
@@ -0,0 +1,70 @@
...
+  dps:
+    - id: 2
+      type: integer
+      name: sensor
+      unit: µg/m³
+      class: measurement
+      mapping:
```
make-all, line 16: Mapping is not needed here, as a scale of 1 is default. 

However, if that scale is copied directly from the iot.tuya.com docs, then the actual scale is 10, as the Tuya definition of scale seems to be how many decimal places to shift.

custom_components/tuya_local/devices/airquality6in1.yaml
```diff
@@ -0,0 +1,70 @@
...
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: Formaldehyde
+    class: volatile_organic_compounds
+    dps:
+      - id: 20
+        type: integer
+        name: sensor
+        unit: mg/m³
```
make-all, line 46: According to https://www.home-assistant.io/integrations/sensor/, units should be µg/m³ (ugm3 can be used as a substitute in the config file), so I would remove the scaling to make it so.

custom_components/tuya_local/devices/airquality6in1.yaml
```diff
@@ -0,0 +1,70 @@
...
+        class: measurement
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    class: volatile_organic_compounds
+    dps:
+      - id: 21
+        type: integer
+        name: sensor
+        unit: mg/m³
```
make-all, line 56: Likewise here

PR #1477, opened 2024-01-02
Add support for Smartmi Air Purifier E1

Balya:


custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
```diff
@@ -0,0 +1,61 @@
+name: Smartmi Air Purifier E1
```
make-all, line 1: Just make this "Air purifier", in case other brands match the same config.

custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
```diff
@@ -0,0 +1,61 @@
...
+products:
+  - id: pwrmggdhwkz1lnvh
+    name: Smartmi Air Purifier E1
```
make-all, line 4: Don't repeat the "Air purifier here", just the brand and model is enough (I haven't decided exactly how it will be used in future, but most likely displayed during device detection, and used for the model and manufacturer info for the device.

custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
```diff
@@ -0,0 +1,61 @@
...
+primary_entity:
+  entity: fan
+  translation_key: fan_with_presets
+  name: Air Purifier
```
make-all, line 8: This will be appended to the device name, so repeating air purifier here is redundant. No name is OK for the primary entity.

custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
```diff
@@ -0,0 +1,61 @@
...
+        - dps_val: true
+          icon: "mdi:air-purifier"
+        - dps_val: false
+          icon: "mdi:air-purifier-off"
+    - id: 3
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: sleep
+          value: Silent
```
make-all, line 24: I strongly suggest adding `translation_key: fan_with_presets` to the entity definition above, and using preset names from https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/en.json#L166 (the keys in the list under fan_wth_presets). ie. sleep, smart, normal, or if the original tuya dps_val are misleading sleep, normal, strong.

custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
```diff
@@ -0,0 +1,61 @@
...
+          icon: "mdi:air-purifier"
+        - dps_val: manual
+          value: Turbo
+          icon: "mdi:wind-power-outline"
+    - id: 22
+      type: bitfield
+      name: fault_code
+secondary_entities:
+  - entity: light
+    name: LED lights
```
make-all, line 37: Are these indicator leds? If so "Indicators" is probably a clearer name.  If it is a light "feature" of the device, then probably no name and removing the category: config.

custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
```diff
@@ -0,0 +1,61 @@
...
+    dps:
+      - id: 101
+        name: sensor
+        type: integer
+        range:
+          min: 0
+          max: 4320
+        mapping:
+          - step: 1
+        unit: hours
```
make-all, line 61: 'h' is the correct unit for hours.  In that case the sensor should have `class: duration` added.

custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
```diff
@@ -0,0 +1,61 @@
...
+            icon: "mdi:led-off"
+  - entity: sensor
+    name: Filter usage
+    category: diagnostic
+    icon: "mdi:shield-bug"
+    dps:
+      - id: 101
+        name: sensor
+        type: integer
+        range:
```
make-all, line 56: No need for a range on a sensor, values from the device will always be used as-is whether or not they are in range (unless there is a target_range mapping, in which case the range is needed to calculate scale and shift amount)

custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
```diff
@@ -0,0 +1,61 @@
...
+          min: 0
+          max: 4320
+        mapping:
```
make-all, line 59: A step of 1 is also not required to be declared, and step is only relevant for numbers the user can set.

custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
Balya: I looked in other configuration files and the brand is indicated everywhere in them. For example, "Breville Smart Air Connect Purifier" or "Himox H06 air purifier". Is it really worth removing the brand name here?

make-all: Yes, there are a lot of old configs that have not been updated, but that does not mean new ones should not follow new naming rules that have evolved through experience.

custom_components/tuya_local/devices/smartmi_e1_purifier.yaml
```diff
@@ -0,0 +1,56 @@
...
+        - dps_val: true
+          icon: "mdi:air-purifier"
+        - dps_val: false
+          icon: "mdi:air-purifier-off"
+    - id: 3
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: sleep
+          value: Sleep
```
make-all, line 23: values that are using translations should be all lower case to match the keys.  

Even though "sleep" will map to itself, you still need to list the mapping here, since this list is used to tell HA which preset modes are available.

make-all: Also, icons listed here are for the entity, not for the specific option.  Since there is no icon_priority, it will be random whether the preset_mode icon or the switch icon above is applied.

Changing icons dynamically interferes with themes and user customisation, so should be used sparingly, only when there is useful information conveyed. (again, this is another thing that you will see overused in older configs)

PR #1493, opened 2024-01-08
Update beca_bht002_thermostat_c.yaml

thomos81:
ID:104 is the temperature switch/heater switch for MOES thermostats.
Just debugged by TUYA IOT platform tools.

### DIFF
--- a/custom_components/tuya_local/devices/beca_bht002_thermostat_c.yaml
+++ b/custom_components/tuya_local/devices/beca_bht002_thermostat_c.yaml
@@ -47,7 +47,12 @@ primary_entity:
         - scale: 2
     - id: 104
       type: boolean
-      name: unknown_104
+      name: temperature_switch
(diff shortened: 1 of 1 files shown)


make-all:
I'm not sure I understand what "temperature switch" is and how it differs from the main switch on dp 1.

thomos81:
Main switch is the input switch to power on/off the device however temperature switch is an output switch to control the heater output, like gas boiler dry_contact switch (GC) or 10A electrical heating (GB) or the 3rd type is the water floor heating (GA)
https://moeshouse.com/products/zigbee-002-thermostat-for-heating-boiler?variant=39802433011793

thomos81:
Sorry was wrong about the BHT002 Type of device, i found there is a separate MOES_BHT002 yaml file as well. I can use the MOES config instead.

PR #1499, opened 2024-01-08
Fix smart_watering_device.yaml

illuzn:
Apologies, I didn't see your code reviews on my PR (holidays probably).

This implements those revisions and I'll provide some context here where necessary:
1. The device is a small water pump for your plants. The runtime is in seconds because this is ample time to supply water to your house plants.
2. Implemented DPS 10 as an attribute of DPS 102 (which was missing). Again DPS 10 seems to do nothing (but perhaps the data will be useful to someone else in the future).

### DIFF
--- a/custom_components/tuya_local/devices/smart_watering_device.yaml
+++ b/custom_components/tuya_local/devices/smart_watering_device.yaml
@@ -11,10 +11,8 @@ primary_entity:
       mapping:
         - dps_val: true
           icon: mdi:water-pump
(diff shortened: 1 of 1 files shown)


make-all:
Ignore the test failure, it seems to be related to https://developers.home-assistant.io/blog/2023/12/18/config-entry-minor-version/ which I guess was pushed into a 2023.1 point release which the unit tests updated to today.

PR #1500, opened 2024-01-09
Add dimmer without color (0-10v S10 Wireless Dimmer)

antibill51:
[https://fr.aliexpress.com/item/1005003286827653.html](url)


make-all:
This is basically dimmable_lightbulb.yaml with the timer added. Since both supported lightbulbs listed in PR #557 are listed online as having timers too, I have added the timer to the existing config, making it identical to this.

PR #1502, opened 2024-01-09
added support for Alutherm 2000 v2

blancqua:
Last week I bought an Eurom Alutherm 2000. The existing integration doesn't seem to work with my version. I also noticed in the shop they were selling the same product in two different boxes. I took the "newest" model but the spec seemed to have changed. I took a copy of the existing eurom integration and tweaked it to work with the model I have.


blancqua:
I see the same test fails in other Pull Requests. Any clue @make-all ?

make-all:
It is a change in Home Assistant yesterday to make the newly introduced in HA 2024.1 minor version attribute on integrations configs mandatory.

PR #1503, opened 2024-01-09
Touchstone Sideline Fireplace Fixed Ember Brightness

lundyfpv:
Added code to ID 109 and made 105 optional. I believe this will fix the issue with brightness control of the embers.


make-all:
I made a fix for this yesterday but forgot to push it.

PR #1558, opened 2024-01-24
Add support for generic gas CO meter

tomd-tc:


custom_components/tuya_local/devices/tuya_gas_sensor.yaml
```diff
@@ -0,0 +1,12 @@
+name: TUYA Gas Leak Detector
+products:
+  - name: TUYA Gas Leak Detector
+primary_entity:
+  entity: sensor
+  class: carbon_dioxide
+  name: Sensor Reading
```
make-all, line 7: Remove the name, so that translations for carbon dioxide sensor are used.

custom_components/tuya_local/devices/tuya_gas_sensor.yaml
```diff
@@ -0,0 +1,12 @@
...
+  dps:
+    - id: 5
+      type: float
```
make-all, line 10: Is it really float?  Most Tuya devices are using integer, and then apply a scale factor to get fixed point.

custom_components/tuya_local/devices/tuya_gas_sensor.yaml
```diff
@@ -0,0 +1,12 @@
+name: TUYA Gas Leak Detector
```
make-all, line 1: Top level name should be generic, like "Gas leak detector"

custom_components/tuya_local/devices/tuya_gas_sensor.yaml
```diff
@@ -0,0 +1,12 @@
...
+products:
+  - name: TUYA Gas Leak Detector
```
make-all, line 3: Products list must have id for each product, otherwise it will not be able to detect which is which.
Branding in name should be actual brand and model, I think Tuya is not selling any end products themselves.

custom_components/tuya_local/devices/tuya_gas_sensor.yaml
```diff
@@ -0,0 +1,12 @@
...
+primary_entity:
+  entity: sensor
+  class: carbon_dioxide
```
make-all, line 6: Title says CO.  Is it carbon_monoxide, or carbon_dioxide?

custom_components/tuya_local/devices/tuya_gas_sensor.yaml
make-all: Need confirmation on this to proceed to merge

custom_components/tuya_local/devices/tuya_gas_sensor.yaml
make-all: Product ids can be obtained by scanning your network with tinytuya if you do not have access to the iot.tuya.com portal to get the info from the cloud.

custom_components/tuya_local/devices/tuya_gas_sensor.yaml
make-all: Need confirmation on this to proceed to merge.

make-all:
As there are too many unanswered questions about this config that have been open for 9 months now, I will close this.

PR #1566, opened 2024-01-25
New Device Holman WX8 Irrigation Controller

TwoWheelPro:


custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -0,0 +1,208 @@
+name: Holman WX8
+products:
+  - id: 3c8bxnseywh7radc
+    model: WX8
+    manufacturer: Holman
+primary_entity:
+  entity: sensor
```
make-all, line 7: Should be a `binary_sensor`, suggest `class: running`.

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -0,0 +1,208 @@
...
+  name: Watering
+  dps:
+    - id: 123
+      type: boolean
```
make-all, line 11: Missing `name: sensor`

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -0,0 +1,208 @@
...
+secondary_entities:
+  - entity: switch
+    name: Zone 1
```
make-all, line 14: Suggest `icon: "mdi:fountain"` or `"mdi:sprinkler"` or `"mdi:sprinkler-variant"`

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -0,0 +1,208 @@
...
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Zone 8
+    dps:
+      - id: 112
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Start A
```
make-all, line 62: Based on the name, this may be better as a button

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -0,0 +1,208 @@
...
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Start C
+    dps:
+      - id: 115
+        type: boolean
+        name: switch 
+  - entity: sensor
+    name: Firmware Version
```
make-all, line 80: I don't think this needs to be a sensor - it will never change and is informational only.  Instead, add it to the primary_entity as
```
    - id: 116
      name: firmware_version
      type: integer
```

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -0,0 +1,208 @@
...
+    category: diagnostic
+    dps:
+      - id: 116
+        name: sensor
+        type: integer     
+  - entity: sensor
+    name: Power Supply Voltage
```
make-all, line 87: Add `class: voltage`

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -0,0 +1,208 @@
...
+    dps:
+      - id: 117
+        name: sensor
+        type: integer
+        unit: "V"
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: Battery Voltage
+    category: diagnostic
```
make-all, line 98: Add `class: voltage`

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -0,0 +1,208 @@
...
+    name: Device Flag
+    category: diagnostic
+    # dps:
+      # - id: 120
+        # type: bitfield
+        # name: sensor
+    dps:
+      - id: 120
+        optional: true
+        name: 9V Battery Good
```
make-all, line 126: If you want to split this out to a separate sensor, this can be a binary_sensor with class: battery, but the logic is reversed (true means low battery, false means good).
If you need more debugging to figure out the masking and values, just add it as an attribute on the primary_entity without any masking.

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
TwoWheelPro: class: voltage did not work on my system

TwoWheelPro:
Updated as much as I could. I think it's working well for basic functionality. Next version will need a more experienced developer.

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -0,0 +1,240 @@
+name: Holman WX8
```
make-all, line 1: The top level name should be generic and describe the purpose of the device (like "Irrigation controller").  The reason is that many of these Chinese made devices are often sold under different brand names in different countries.

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
make-all: "Didn't work" meaning what?  The only difference in the UI I would expect is the icon.

stas1000:
Hi, has anybody got this working natively through the integration? I still get the error: Sorry, there is no support for this device.

TwoWheelPro:
Works on my device fine. It's based off the product id. If you can get yours from the Tuya development portal you can check it. in the device definition YAML.

name: Irrigation Controller
products:
  - id: 3c8bxnseywh7radc
    model: WX8
    manufacturer: Holman

stas1000:
> custom_components/tuya_local/devices/holma

below is the error I keep getting when attempting to load the WX8 via the integration with the device id and local key.
I also tried to update the tuya local configuration.yaml as below to my device Id but still have the same error :

name: Irrigation Controller
products:

id: 3c8bxnseywh7radc - replaced with the device id of my WX8
model: WX8
manufacturer: Holman

Is the product id the same as the product id device ID?

image

make-all:
The error message in the log will have more details.

stas1000:
> The error message in the log will have more details.

Thanks I have 2 logs:
Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:97
Integration: Tuya Local ([documentation](https://github.com/make-all/tuya-local), [issues](https://github.com/make-all/tuya-local/issues))
First occurred: 21:30:26 (1 occurrences)
Last logged: 21:30:26

Device matches None with quality of 0%. DPS: {"updated_at": 1707908425.5580792, "102": false, "103": false, "104": false, "105": false, "106": false, "107": false, "108": false, "112": false, "113": false, "114": false, "115": false, "116": 6, "117": 115, "118": 87, "119": 0, "120": 99, "121": 1, "122": 11, "123": false}

stas1000:
And this one:
Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:103
Integration: Tuya Local (documentation, issues)
First occurred: 21:30:26 (1 occurrences)
Last logged: 21:30:26

Report this to https://github.com/make-all/tuya-local/issues/

make-all:
The problem seems to be this one: `120": 99`

In the config dp 120 is defined as base64, but your device is returning an integer.

stas1000:
I have no idea how you figured that out from my logs but it worked! 
I updated from base64 to integer in the device config on lines 131,142,154,167,179 & 191.
Everything looks to be picked up now and working.
Thank you

make-all:
Now we have to figure out if there are two different firmware revisions here, or the device uses int up to some value (maybe 255) then switches to base64. Worst case we will have to remove those flags from the config.

stas1000:
If it helps at all my units firmware version is 6

TwoWheelPro:
I'm also on firmware V6. Could it be the platform? I'm running on Rasberry PI with home assistant yellow.

I will change mine to integers as well and see if it creates any issues.

make-all:
I notice that you originally had these as bitfields, which is an integer type. I wonder if you originally got your devices added when they were defined like that, and you are not experiencing the issue since you have not added any more devices since you changed the type to base64.  If they are bitfields, they don't need the mask attribute to function, just map the relevant bit to true, and anything else to false (or vice-versa if they are inverted)

TwoWheelPro:
Possibly, essentially the data is coded as 8bits with each representing a state, e.g. Bit 4 is the Wet/Dry indicator for a wired rain sensor. How would you "just map" that value without using a mask? The below works to show the correct value if I short/open the terminal with a jumper.

 - entity: binary_sensor
    name: Rain sensor
    class: moisture
    dps:
      - id: 120
        name: sensor
        type: base64
        mapping:
          - mask: "000F0000"
            dps_val: 0
            value: true
          - value: false

make-all:
If the data is in integer format, that should fail on the base64 decode step, but maybe it is getting through after the failure with the value intact which it can still use for masking.  However that mask is pulling out 4 bits from the middle of a 32 bit value, so I'm not sure how it is working if it really is bit 4 of an 8 bit value you want to extract.

For bit 4 in a bitfield, the mapping for an inverted value would be:
```
  - entity: binary_sensor
    name: Rain sensor
    class: moisture
    dps:
      - id: 120
        name: sensor
        type: bitfield
        mapping:
          - dps_val: 16
            value: false
          - value: true
```

TwoWheelPro:
OK, I'll do some testing tomorrow. 

@stas1000 I'll send you my code once I've tested. Hopefully you can test as well. I believe an interpreted value of "99" corresponds to having an Evie sensor connected, which I don't have one, so my value is "3" when viewed as an integer.

stas1000:
Can do. Cheers

TwoWheelPro:
@stas1000  I've updated the file and tested on my system by fully removing and re-adding the device. This should work.

https://github.com/TwoWheelPro/tuya-local/blob/main/custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml

stas1000:
> @stas1000 I've updated the file and tested on my system by fully removing and re-adding the device. This should work.
> 
> https://github.com/TwoWheelPro/tuya-local/blob/main/custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml

Hi mate,

I removed deleted my configuration and my edited yaml file and holman_wx8_irrigation_controller.yaml
uploaded the new file holman_wx8_irrigation_controller.yaml
reset home assistant
connected WX8 to home assistant - it worked
the only thing that has not updated is the evie sensor will still show as wet - although it is dry in the app

image

make-all:
How is the battery sensor?  Is it perhaps swapped with the Evie moisture sensor? (ie the bits are reversed)

TwoWheelPro:
Based on the JSON definition from the development portal it seems correct. Is it the mapping? I started at 128 for the DPS value, should it be 256? 128, 64, 32, 16, (8), 4, 2, 1
image

make-all:
128 is correct for bit 7. 256 would be bit 8, which requires a second byte (and the integer value would be >256). You seem to have interpreted the bit positions from left to right as written on paper, which is unusual, but unusual conventions are not unusual for Tuya devices so I wouldn't jump to conclusions about that.

TwoWheelPro:
@stas1000 coudl you chuck some water in the Evie and confirm that is changes value?

Hopefully it's just inverted polarity.

make-all:
I guess most of these dp 120 bits are flags for abnormal states, so difficult to test.  Is there anything else that is easy to trigger?
If they are mirrored, then the following should be swapped:

Evie sensor : Battery voltage
Evie battery : Power supply
Evie : Watering
(reserved) : Rain sensor

I don't know if removing the power supply will make it drop Wifi to conserve battery, so I don't know how feasible that is to test, but maybe the Watering sensor is easy to control, and check whether it affects the correct sensor, or the Evie connection sensor?

stas1000:
i have tried to add water but the evie sensor stays in wet status. i double checked to see if maybe the battery was flat but the tuya app is definitely still updating its status between wet and dry

make-all:
I meant to check the "Battery voltage" sensor under diagnostics that reads Normal or Low, and see if it changes when you expect the Evie sensor to change (to test the hypothesis that the sensors are mirrored)

make-all:
Another test could be to unplug the Evie sensor and see if the "Watering" sensor changes.

TwoWheelPro:
The Mapping is not working. I'll look at it more later. If i unplug the battery then the moister sensor goes to "wet" but also the battery status goes to "low" so it is mixing the values together. It's not correctly reading a single bit but looking at the whole byte.

stas1000:
Here are some screen grabs of the sensors with the battery connected:
Watering Off
Screenshot_20240217-143649
Watering On
Screenshot_20240217-143703

make-all:
The screen grabs seem to confirm that the Watering and Evie connection are swapped. The battery removal may be causing some knock on effects in the device.

TwoWheelPro:
I have reversed the order of the bitfield values and that seems to have fixed the issue.

@stas1000 if you download the latest it should work, i've tested the wired sensor, watering, battery ok, and the all work ok. If you can test the Evie flags it should be ok now?

stas1000:
Thank you, it looks to function correctly now with the evie sensor updating between wet & dry

image

PR #1569, opened 2024-01-25
Add support for Goodcell 3-phase 22kW EV charger

philippedeswert:
Pre-eliminary support for the Goodcell 3-phase 22kW EV charger.

A few issues remain:
* the two other phases current and voltage are not visible.
* the energy reset does not work (might also not be implemented correctly for all I know on the charger)

### DIFF
--- a/custom_components/tuya_local/devices/goodcell_ev_charger.yaml
+++ b/custom_components/tuya_local/devices/goodcell_ev_charger.yaml
@@ -1,7 +1,7 @@
 name: EV charger
 products:
-  - id: hytatexziwaifxj1
-    name: Feyree 32A 85-264V EV charger
+  - id: j6bzjwhiv2cljjcy
+    name: Goodcell 32A 85-264V 3 phase EV charger
 primary_entity:
   entity: sensor
   class: enum
@@ -12,7 +12,7 @@ primary_entity:
       type: string
       name: sensor
       mapping:
-        - dps_val: no_connect
+        - dps_val: no_connet
           value: Disconnected
         - dps_val: connect
           value: Connected
@@ -30,6 +30,23 @@ primary_entity:
       type: string
       optional: true
       name: charger_status
+      mapping:
+        - dps_val: charger_free
+          value: Standby
+        - dps_val: charger_insert
+          value: Cable connected
+        - dps_val: charger_free_fault
+          value: Charger fault
+        - dps_val: charger_wait
+          value: Charger ready
+        - dps_val: charger_charging
+          value: Charging
+        - dps_val: charger_pause
+          value: Charging paused
+        - dps_val: charger_end
+          value: Charging ended
+        - dps_val: charger_fault
+          value: Charger fault
     - id: 10
       type: bitfield
       name: fault_code
@@ -72,15 +89,12 @@ primary_entity:
       type: string
       optional: true
       name: online_state
-    - id: 105
-      type: boolean
-      optional: true
-      name: require_authorization
 secondary_entities:
   - entity: button
     name: Clear energy
     class: restart
     category: config
+    hidden: true
     dps:
       - id: 16
         type: boolean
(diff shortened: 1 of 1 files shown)


make-all:
You have multiple unnamed sensors with the same class (for at least voltage and current).  These need names to distinguish them from each other (max is one unnamed sensor of each class, and one unnamed sensor without a class).  Basically each entity needs to be uniquely identified by (entity type)_(name or class)  ie. name is used if available, or class if unnamed.

philippedeswert:
> You have multiple unnamed sensors with the same class (for at least voltage and current). These need names to distinguish them from each other (max is one unnamed sensor of each class, and one unnamed sensor without a class). Basically each entity needs to be uniquely identified by (entity type)_(name or class) ie. name is used if available, or class if unnamed.

Ofc, this is so obvious now. Well it was getting late last night. Anyway updated my PR with names as I can now see all different phases. I do not know if it is needed to mark them optional, as that *might* make this also work with the 1-phase models.  It seems there might be something I can do with the constraints too, so current can be adjusted to max 16A for the models not going to 32A.

philippedeswert:
Rebased and pushed some small tweaks. 

Could not figure out how to use dps 113 to select dps 114 or 115 for the current setting. (As in dps 114 if dps 113 says Max16A, and dps 115 if it is Max32A), If you can guide me on how to do it I can implement and fix that. Are there other changes needed still otherwise?

make-all:
Something like:
```
  - entity: number
    name: Set charge current
    category: config
    icon: "mdi:ev-plug-type2"
    dps:
      - id: 115
        type: integer
        name: value
        unit: A
        range:
          min: 6
          max: 32
        mapping:
          - constraint: max_current
            conditions:
              - dps_val: Max16A
                value_redirect: value_alt
                range:
                  min: 6
                  max: 16
      - id: 114
        type: integer
        name: value_alt
        unit: A
        range:
          min: 6
          max: 16
      - id: 113
        type: string
        name: max_current
 
```

philippedeswert:
Updated as per your suggestion, but I am not sure it works (did not seem to switch to pds 114 when I changed the Max16A condition to Max32A to trigger it....

Otherwise it pretty much works only exception is that energy reset button does not seem to work, but it is not supported in the app. So I guess it is not supposed to work. Should I remove that?

Otherwise anything I should do before you consider this done?

custom_components/tuya_local/devices/goodcell_ev_charger.yaml
```diff
@@ -0,0 +1,275 @@
...
+                    min: 6
+                    max: 16
+      - id: 114
+        type: integer
+        name: value_alt
+        unit: A
+        range:
+          min: 6
+          max: 16
+      - id: 113
```
make-all, line 129: This may need to be a switch or select entity as well if the intention is for the user to make this change.  I am not sure though whether this is an install time option, a user option, or automatic based on vehicle capabilities.

custom_components/tuya_local/devices/goodcell_ev_charger.yaml
```diff
@@ -0,0 +1,275 @@
...
+      name: fault_code
+    - id: 11
+      type: string
+      optional: true
+      name: alarm_set_1
+    - id: 12
+      type: string
+      optional: true
+      name: alarm_set_2
+    - id: 14
```
make-all, line 61: I think this should be a select entity to select the mode of charging.  Otherwise the settings linked to inactive charging modes will be redundant if you cannot switch to that mode.

make-all:
Energy reset probably only affects balance_energy, which is probably only relevant in `charge_energy` charging mode.  I don't think it affects the total_forward_energy, or the charge_energy_once (which I think is the last/current charge only)

custom_components/tuya_local/devices/goodcell_ev_charger.yaml
```diff
@@ -0,0 +1,275 @@
...
+    icon: "mdi:ev-plug-type2"
+    dps:
+      - id: 115
+        type: integer
+        name: value
+        unit: A
+        range:
+          min: 6
+          max: 32
+          mapping:
```
make-all, line 114: This `mapping` should be at the same level of indentation as `range`.  Currently it is under `range`, which prevents this from working.

PR #1582, opened 2024-01-31
Create poolex_arctic_action.yaml

a-r-j-a-n:
Initial commit for the Poolex Arctic Action

### DIFF
--- a/custom_components/tuya_local/devices/poolex_arctic_action.yaml
+++ b/custom_components/tuya_local/devices/poolex_arctic_action.yaml
@@ -1,0 +1,269 @@
+name: Pool heatpump
+products:
+  - id: ea8w1w
+    name: Poolex Arctic Action
+primary_entity:
+  entity: climate
+  icon: "mdi:hot-tub"
+  translation_key: pool_heatpump
+  dps:
+    - id: 1
+      type: boolean
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: preset_mode
+          conditions:
+            - dps_val: Heat_Intelligent
+              value: heat
+            - dps_val: Heat_Powerful
+              value: h_powerful
+            - dps_val: Heat_Energy_Saving
+              value: h_silent
+            - dps_val: Cool_Intelligent
+              value: cool
+            - dps_val: Cool_Powerful
+              value: c_powerful
+            - dps_val: Cool_Silent
+              value: c_silent
+    - id: 2
+      type: integer
+      name: temperature
+      range:
+        min: 5
+        max: 40
+      mapping:
+        - constraint: temperature_unit
+          conditions:
+            - dps_val: f
+              range:
+                min: 41
+                max: 104
+              value_redirect: temp_set_f
+    - id: 3
+      type: integer
+      name: current_temperature
+      mapping:
+        - constraint: temperature_unit
+          conditions:
+            - dps_val: f
+              value_redirect: temp_current_f
+    - id: 4
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: Heat_Intelligent
+          icon: "mdi:hot-tub"
+          value: heat
+        - dps_val: Heat_Powerful
+          icon: "mdi:hot-tub"
+          value: h_powerful
+        - dps_val: Heat_Energy_Saving
+          icon: "mdi:hot-tub"
+          value: h_silent
+        - dps_val: Cool_Intelligent
+          value: cool
+          icon: "mdi:snowflake"
+        - dps_val: Cool_Powerful
+          icon: "mdi:snowflake"
+          value: c_powerful
+          icon: "mdi:snowflake"
+        - dps_val: Cool_Silent
+          value: c_silent
+          icon: "mdi:snowflake"
+    - id: 13
+      type: string
+      name: temperature_unit
+      mapping:
+        - dps_val: c
+          value: C
+        - dps_val: f
+          value: F
+    - id: 14
+      type: integer
+      name: temp_set_f
+      range:
+        min: 41
+        max: 104
+      hidden: true
+      optional: true
+    - id: 15
+      type: integer
+      name: temp_current_f
+      hidden: true
+      optional: true
+    - id: 21
+      type: integer
+      name: error
+      mapping:
+        - dps_val: 0
+          value: "OK"
+        - dps_val: 4
+          value: "Water Flow Protection"
+          icon: "mdi:water-boiler-alert"
+          icon_priority: 2
+        - dps_val: 3
+          value: "Water Flow Fault"
+          icon: "mdi:water-boiler-alert"
+          icon_priority: 2
+        - dps_val: 4
+          value: "Frost Protection"
+          icon: "mdi:water-boiler-alert"
+          icon_priority: 2
+        - dps_val: 5
+          value: "High Pressure Fault"
+          icon: "mdi:water-boiler-alert"
+          icon_priority: 2
+        - dps_val: 6
+          value: "Low Pressure Fault"
+          icon: "mdi:water-boiler-alert"
+          icon_priority: 2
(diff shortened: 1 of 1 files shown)


a-r-j-a-n:
If I can assist any way in testing, please let me know (how)! :)

make-all:
You should be able to test by copying your added and modified files into the custom-components/tuya-local directory on your Home Assistant installation.

a-r-j-a-n:
I'll test it with my local version first then

custom_components/tuya_local/devices/poolex_arctic_action.yaml
```diff
@@ -0,0 +1,268 @@
...
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: preset_mode
+          conditions:
+            - dps_val: Heat_Intelligent
+              value: heat
+            - dps_val: Heat_Powerful
+              value: h_powerful
```
make-all, line 22: hvac_mode values must be from the predefined list https://developers.home-assistant.io/docs/core/entity/climate/#hvac-modes (the constants listed there map to all lower case strings matching the name of the constant).

In the case of multiple presets matching an hvac_mode, the full set will be accessible through preset_mode, and here all the heating values should map to heat (mark others hidden so that the  preset_mode that is selected when switching to heat mode is deterministic) and the cooling related ones to cool.

custom_components/tuya_local/devices/poolex_arctic_action.yaml
```diff
@@ -0,0 +1,268 @@
...
+          conditions:
+            - dps_val: f
+              value_redirect: temp_current_f
+    - id: 4
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: Heat_Intelligent
+          icon: "mdi:hot-tub"
+          value: heat
```
make-all, line 59: preset_mode can be anything, but it is recommended to take from existing translations, in this cas the translation_key pool_heatpump was already defined, and the presets defined for other pool heatpumps at https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/en.json#L128 closely match (smart_heat for Heat_Intelligent, quick_heat for Heat_Powerful, and quiet_heat for Heat_Energy_Saving, and likewise for cool.  The Energy_Saving does not quite match quiet, but it seems like there is a common controller, and manufacturers choose the naming based on how they want to market it).

custom_components/tuya_local/devices/poolex_arctic_action.yaml
```diff
@@ -0,0 +1,268 @@
...
+          icon: "mdi:water-boiler-alert"
+          icon_priority: 2
+        - dps_val: 42
+          value: "Internal Coil Temperature Fault"
+          icon: "mdi:water-boiler-alert"
+          icon_priority: 2
+
+
+secondary_entities:
+  - entity: water_heater
```
make-all, line 188: New configs should not have deprecated entities.  I think you copied from the one pool heatpump that was originally created with only heating modes, and only after converting to water_heater did a user point out that it also supported cooling modes.

custom_components/tuya_local/devices/poolex_arctic_action.yaml
```diff
@@ -0,0 +1,268 @@
...
+      - id: 13
+        type: string
+        name: option
+        mapping:
+          - dps_val: c
+            value: Celsius
+          - dps_val: f
+            value: Fahrenheit
+  - entity: binary_sensor
+    name: Status
```
make-all, line 258: This should be an hvac_action attribute of the climate entity, with mapping to either heating or cooling  depending on the preset mode.

make-all: But actually this is dp 1, which is just the power switch. There is no need for a separate running sensor, as you can tell from the state of the climate entity whether it is switched off, or to cooling or heating. And this shouldn't be hvac_action, as it isn't giving any more info than hvac_mode does.

a-r-j-a-n: Changed the configuration, based on the feedback provided. 

Also fixed the product_id, but it is still not discovered after placing it in my local custom_components/tuya_local/devices/ folder. Only matches for 83% with hydrotherm_dynamic_x8_water_heater and it doesnt come up in the type selection dropdown. Don't know how to test properly I'm afraid.

```
{
  "result": [
    {
      "active_time": 1698766405,
      "bind_space_id": "170682830",
      "category": "qn",
      "create_time": 1698766405,
      "custom_name": "Poolex",
      "icon": "smart/icon/ay1513418289516FAXSV/cfca410c1da90b02660fc721b48a951b.png",
      "id": "_redacted_",
      "ip": "_redacted_",
      "is_online": true,
      "lat": "21.7300",
      "local_key": "_redacted_",
      "lon": "15.3100",
      "model": "NC11440004",
      "name": "Heat Pumps",
      "product_id": "pzp8owd7lhtpssgj",
      "product_name": "南晟变频泳池机 ℃ ℉",
      "sub": false,
      "time_zone": "+01:00",
      "update_time": 1698766527,
      "uuid": "5734565c743afbbf"
    }
  ],
  "success": true,
  "t": 1706715008831,
  "tid": "9dcec602c04d11ee84c772fd0bbfcb23"
}
```

a-r-j-a-n:
Changed the configuration, based on the feedback provided.

Also fixed the product_id, but it is still not discovered after placing it in my local custom_components/tuya_local/devices/ folder. Only matches for 83% with hydrotherm_dynamic_x8_water_heater and it doesnt come up in the type selection dropdown. Don't know how to test properly I'm afraid.

```
{
  "result": [
    {
      "active_time": 1698766405,
      "bind_space_id": "170682830",
      "category": "qn",
      "create_time": 1698766405,
      "custom_name": "Poolex",
      "icon": "smart/icon/ay1513418289516FAXSV/cfca410c1da90b02660fc721b48a951b.png",
      "id": "_redacted_",
      "ip": "_redacted_",
      "is_online": true,
      "lat": "21.7300",
      "local_key": "_redacted_",
      "lon": "15.3100",
      "model": "NC11440004",
      "name": "Heat Pumps",
      "product_id": "pzp8owd7lhtpssgj",
      "product_name": "南晟变频泳池机 ℃ ℉",
      "sub": false,
      "time_zone": "+01:00",
      "update_time": 1698766527,
      "uuid": "5734565c743afbbf"
    }
  ],
  "success": true,
  "t": 1706715008831,
  "tid": "9dcec602c04d11ee84c772fd0bbfcb23"
}
```

custom_components/tuya_local/devices/poolex_arctic_action.yaml
```diff
@@ -0,0 +1,267 @@
...
+          icon_priority: 2
+        - dps_val: 42
+          value: "Internal Coil Temperature Fault"
+          icon: "mdi:water-boiler-alert"
+          icon_priority: 2
+
+secondary_entities:
+  - entity: climate
+    icon: "mdi:hot-tub"
+    deprecated: climate
```
make-all, line 189: In my previous comment, the intention was to suggest that you remove this deprecated entity, not change it to a second climate entity.

custom_components/tuya_local/devices/poolex_arctic_action.yaml
```diff
@@ -0,0 +1,267 @@
+name: Poolex Arctic Action
```
make-all, line 1: I'm not sure why you moved the branded name here in your latest revision, this should be under products as it was before, with a generic name at top level. It is very common for pool heatpumps to be sold under an importer's brand, so essentially the same device is sold in different markets with different brands.

make-all:
The log message that goes along with that "83% match" should contain a list of the dps received from the device.  This should be compared with the config, and any missing dps marked as `optional: true` in the config.

a-r-j-a-n:
Got it working. It matches now automatically. Also added two additional entities.

Issue #1601, opened 2024-02-04
Nedis air quality monitor wrong unit


JacobsenKim:
![Skærmbillede 2024-02-04 135140](https://github.com/make-all/tuya-local/assets/157890151/052e8dc4-fc2e-4645-9065-389986581bd3)

Line 46 need to change from ppm to "µg/m³" then the error stop in HA.

i had this fixed before but update the complete local tuya pack and got the problem once more.

Please update the file
https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/nedis_airquality.yaml

Thx

make-all:
Based on the original report, the device is reporting ppm.  So the correct fix is not to change the unit, but to change the class to the more recently added volatile_organic_compounds_parts class (I'm not sure exactly when this appeared, but it was within the last few months, so it is surprising that HA is throwing errors so soon after introducing the split between these).

JacobsenKim:
The change fixed the problem at my end :-D

Thx

Issue #384, opened 2023-02-10
Request support for Nedis WIFISA10CWT Air Quality Monitor


janchrillesen:
# Log Message

2023-02-10 14:49:30.314 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches None with quality of 0%. DPS: {'updated_at': 1676036969.9560797, '2': 2, '18': 196, '19': 429, '21': 1, '22': 368}

# Information about DPS mappings

{
  "result": {
    "category": "jqbj",
    "functions": [],
    "status": [
      {
        "code": "ch2o_value",
        "type": "Integer",
        "values": "{\"unit\":\"mg/m³\",\"min\":0,\"max\":1000,\"scale\":2,\"step\":1}"
      },
      {
        "code": "va_temperature",
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":-400,\"max\":2000,\"scale\":1,\"step\":1}"
      },
      {
        "code": "va_humidity",
        "type": "Integer",
        "values": "{\"unit\":\"%\",\"min\":0,\"max\":1000,\"scale\":1,\"step\":1}"
      },
      {
        "code": "voc_value",
        "type": "Integer",
        "values": "{\"unit\":\"ppm\",\"min\":0,\"max\":999,\"scale\":1,\"step\":1}"
      },
      {
        "code": "co2_value",
        "type": "Integer",
        "values": "{\"unit\":\"ppm\",\"min\":0,\"max\":1000,\"scale\":0,\"step\":1}"
      }
    ]
  },
  "success": true,
  "t": 1676036798646,
  "tid": "57987ba3a94911edbe05a63cfb33a057"
}

# Product ID

    "product_id": "ipbqjatfsxghe3dc",
    "product_name": "Air Quality Monitor",
    "category": "jqbj",
    "category_name": "Formaldehyde Detector",
    "model": "WIFISA10CWT",

# Information about how the device functions

https://nedis.dk/en-us/product/home-living/climate/thermometers/550768043/smartlife-air-quality-monitor-wi-fi-humidity-temperature-volatile-organic-compounds-voc-android-ios-white
https://www.manua.ls/nedis/wifisa10cwt/specifications

janchrillesen:
Thank you for adding this device. I added the YAML file manually to my installation and it works perfectly

PR #1602, opened 2024-02-04
Update nedis_airquality.yaml

JacobsenKim:
https://github.com/make-all/tuya-local/issues/1601

change line 46 from ppm to "µg/m³"


make-all:
This is not a correct fix, as according to the device information provided in #384, the device is providing a ppm reading, not a µg/m³ reading.

PR #1612, opened 2024-02-05
Add support for CleanAir 604B humidifier

FSJHecker:


custom_components/tuya_local/devices/cleanair_604b_humidifier.yaml
```diff
@@ -0,0 +1,165 @@
...
+        - dps_val: false
+          icon: mdi:air-humidifier-off
+          icon_priority: 2
+    - id: 13
+      name: humidity
+      type: integer
+      range:
+        min: 30
+        max: 75
+        step: 5
```
make-all, line 21: `step` belongs in a `mapping` list, `range` is only for `min` and `max`.

custom_components/tuya_local/devices/cleanair_604b_humidifier.yaml
```diff
@@ -0,0 +1,165 @@
+name: Humidifier
+products:
+  - id: cqoisddoci9i3yer
+    name: Clean Air Optima CA-604B Smart Top Filling
```
make-all, line 4: This name seems a bit long, maybe leave off the "Smart Top Filling", which is more descriptive than part of the model name.

custom_components/tuya_local/devices/cleanair_604b_humidifier.yaml
```diff
@@ -0,0 +1,165 @@
...
+primary_entity:
+  entity: humidifier
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+      mapping:
+        - dps_val: false
+          icon: mdi:air-humidifier-off
```
make-all, line 13: It looks as though HA is going to deprecate setting dynamic icons programmatically in favour of the new icon translations feature coming in HA 2024.2, so I prefer to keep the icons simple for now (just a static icon at the entity level where the default icon is not specific enough - usually for the generic entities like number, select, switch etc).

custom_components/tuya_local/devices/cleanair_604b_humidifier.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - id: 8
+        name: switch
+        type: boolean
+        mapping:
+          - dps_val: true
+            icon: mdi:volume-high
+          - dps_val: false
+            icon: mdi:volume-off
+  - entity: sensor
+    name: Current Temperature
```
make-all, line 62: It would be better to leave off the name, so the default class translated name is used.

custom_components/tuya_local/devices/cleanair_604b_humidifier.yaml
```diff
@@ -0,0 +1,165 @@
...
+            value: 8 hours
+          - dps_val: 9h
+            value: 9 hours
+          - dps_val: 10h
+            value: 10 hours
+          - dps_val: 11h
+            value: 11 hours
+          - dps_val: 12h
+            value: 12 hours
+      - id: 20
```
make-all, line 103: Time remaining can be made a `class: duration` sensor, it is generally interesting enough to have as a sensor rather than an extra attribute where it is more difficult to get at.

custom_components/tuya_local/devices/cleanair_604b_humidifier.yaml
```diff
@@ -0,0 +1,165 @@
...
+        name: remaining
+        type: integer
+        optional: true
+  - entity: binary_sensor
+    name: Watertank Empty
```
make-all, line 108: HA naming convention is to only capitalise the first word (unless an abbreviation or brand name that should be capitalised).  I think just `Tank empty` is clear enough and more concise.

custom_components/tuya_local/devices/cleanair_604b_humidifier.yaml
```diff
@@ -0,0 +1,165 @@
...
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 1
+            value: true
+            icon: mdi:water-remove
+          - value: false
+            icon: mdi:water
+  - entity: binary_sensor
+    name: Watertank Overfilled
```
make-all, line 122: `Tank overflow` would be more correct in English.

PR #1633, opened 2024-02-08
Fix returning supported color modes

hyperb1iss:
* Return a set including ColorMode.ONOFF instead of a list


make-all:
Thanks for the effort of submitting a fix.
I've installed a slightly simpler version of the change which only changes the lists to sets without altering the logic.
ONOFF is actually handled already for the devices that support on/off and nothing else.  The fallback empty set should not actually be used by any real devices, as it represents a light with no controls.

Issue #1645, opened 2024-02-11
Request support for Dynasty BTX Fireplace


MatthewBCooke:
# Log Message

```
This error originated from a custom integration.

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:97
Integration: Tuya Local (documentation, issues)
First occurred: 13:49:19 (1 occurrences)
Last logged: 13:49:19

Device matches qoto_03_sprinkler with quality of 38%. DPS: {"updated_at": 1707688158.178351, "1": false, "3": 12, "5": "0", "13": "c", "101": false, "102": "0", "103": "0", "104": "0", "105": "0", "106": "0", "107": "0", "108": "0", "109": 0}
```

# Information about DPS mappings

```
  "result": {
[...]

PR #1646, opened 2024-02-11
Added support for dynasty_btx_fireplace.yaml

MatthewBCooke:
Config for dynasty electric fireplaces as per https://github.com/make-all/tuya-local/issues/1645

### DIFF
--- a/custom_components/tuya_local/devices/dynasty_btx_fireplace.yaml
+++ b/custom_components/tuya_local/devices/dynasty_btx_fireplace.yaml
@@ -1,0 +1,193 @@
+name: Fireplace
+products:
+  - id: 5081205010521cfac95f
+    name: Dynasty BTX Fireplace
+primary_entity:
+  entity: climate
+  icon: "mdi:fireplace"
+  dps:
+    - id: 1
+      name: hvac_mode
+      type: boolean
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: heat
+    - id: 3
+      name: current_temperature
+      type: integer
+      range:
+        min: 0
+        max: 40
+    - id: 5
+      name: temperature
+      type: string
+      mapping:
+        - dps_val: "0"
+          value: 16
+        - dps_val: "1"
+          value: 19
+        - dps_val: "2"
+          value: 21
+        - dps_val: "3"
+          value: 23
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/dynasty_btx_fireplace.yaml
```diff
@@ -0,0 +1,193 @@
...
+      type: boolean
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: heat
+    - id: 3
+      name: current_temperature
+      type: integer
+      range:
```
make-all, line 20: No range needed on read-only attributes, what is received from the device will be displayed, regardless of whether it is in range or not.

make-all:
I'm not sure if temperature will work in general without a range (even when using mapped values, as the code does not infer the range from the list of possible values). Possibly it works in this case because 16 - 30 is the Home Assistant default range, but still would be better explicitly included in the config to isolate from future changes of the defaults.

custom_components/tuya_local/devices/dynasty_btx_fireplace.yaml
```diff
@@ -0,0 +1,193 @@
...
+        - dps_val: "3"
+          value: 23
+        - dps_val: "4"
+          value: 25
+        - dps_val: "5"
+          value: 27
+        - dps_val: "6"
+          value: 30
+    - id: 13
+      name: temperature_unit
```
make-all, line 42: Including this here will require more complex handling of temperature dp, as you need the values and range to be different depending on this setting.
It may be better to leave this out, and mark this config as only being valid for C.

custom_components/tuya_local/devices/dynasty_btx_fireplace.yaml
```diff
@@ -0,0 +1,193 @@
+name: Fireplace
+products:
+  - id: 5081205010521cfac95f
```
make-all, line 3: Is this the product id, or the device id? (it should be the former, which is alphanumeric, not normally limited to hex characters as device ids often are)

MatthewBCooke: Updated to the correct product ID and removed the range for the temperature

PR #1650, opened 2024-02-13
Inverted bit values to show correct status

TwoWheelPro:
Bit mask values where inverted. Updated to show correct values.

### DIFF
--- a/custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
+++ b/custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
@@ -16,7 +16,7 @@ primary_entity:
       type: integer
     - id: 101
       name: programming
-      type: string
+      type: raw
       optional: true
 secondary_entities:
   - entity: switch
@@ -107,9 +107,15 @@ secondary_entities:
         unit: "V"
         mapping:
           - scale: 10
+  - entity: sensor
+    name: Evie battery voltage
+    class: voltage
+    category: diagnostic
+    dps:
       - id: 119
-        name: catchup_voltage  # not sure what this is, so made an attribute
+        name: sensor
         type: integer
+        unit: "V"
         mapping:
           - scale: 10
   - entity: sensor
@@ -124,13 +130,28 @@ secondary_entities:
         mapping:
           - scale: 10
   - entity: binary_sensor
+    name: Battery voltage
     class: battery
+    category: diagnostic
     dps:
       - id: 120
         name: sensor
-        type: base64
+        type: bitfield
         mapping:
           - mask: "10000000"
+            dps_val: 0
+            value: true
+          - value: false
+  - entity: binary_sensor
+    name: Power supply
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 120
+        name: sensor
+        type: bitfield
+        mapping:
+          - mask: "01000000"
             dps_val: 0
             value: true
           - value: false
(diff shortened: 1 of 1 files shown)


make-all:
I'll let you test the changes to bitfield before merging (or cancelling) this PR.

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -16,7 +16,7 @@ primary_entity:
       type: integer
     - id: 101
       name: programming
-      type: string
+      type: raw
```
make-all, line 19: raw is not a type, us string as originally

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -124,22 +130,37 @@ secondary_entities:
         mapping:
           - scale: 10
   - entity: binary_sensor
+    name: Battery voltage
```
make-all, line 133: If this is a voltage it should be extracting more than one bit and be a sensor rather than a binary_sensor. If it is a low battery warning flag, then leave the name out so that translated names from the battery class are used. If it is anything else, the class is incorrect, so remove it.

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -124,22 +130,37 @@ secondary_entities:
...
     class: battery
+    category: diagnostic
     dps:
       - id: 120
         name: sensor
-        type: base64
+        type: bitfield
         mapping:
           - mask: "10000000"
```
make-all, line 141: The mask is hex, so this is bit 32.  As discussed in the other conversation, just using the dps_val when it is set will auto mask the value when mapping bitfield values, so assuming you meant bit 8:
```
mapping:
  - dps_val: 128
    value: false
  - value: true
```

make-all (changes requested):
At least the raw type needs to be changed back to string

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
TwoWheelPro: This is a flag based on the voltage, i.e. "Voltage is ok"

make-all: So inverted it would meet the spec of `class: battery` for [binary_sensors](https://www.home-assistant.io/integrations/binary_sensor/).

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
TwoWheelPro: Reverted. I was testing. DPS 101 doesn't actualy how up in Tuya_local but it is present in the Tuya Development portal.

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
TwoWheelPro: image

TwoWheelPro: look's good to me?

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
TwoWheelPro: OK made the chages to show dps values. Seems like a strange way to do it to me. But it does seem to work.

make-all (approved):

PR #1656, opened 2024-02-14
Add support for Graywind shades/blinds

dannygreg:
https://www.amazon.com/gp/product/B07Z95FTVW/?ie=UTF8

Mine have the opening/closing status flipped, but I believe that's because I have reverse enabled - probably not worth trying to fix that.

When I deployed to my local Hass instance the shades were correctly recognised and function otherwise as expected.


dannygreg:
You beat me to it! Thanks!

PR #1663, opened 2024-02-17
Updated binary value polarity for Holman Bit Flags for battery entities.

TwoWheelPro:
reversed bitfield flag order. Although I see you made that change on the main branch. Agree this is correct based on my latest testing.

### DIFF
--- a/custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
+++ b/custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
@@ -139,8 +139,8 @@ secondary_entities:
         type: bitfield
         mapping:
           - dps_val: 1
-            value: true
-          - value: false
+            value: false
+          - value: true
   - entity: binary_sensor
     name: Power supply
     class: problem
@@ -151,8 +151,8 @@ secondary_entities:
         type: bitfield
         mapping:
           - dps_val: 2
-            value: true
-          - value: false
+            value: false
+          - value: true
   - entity: binary_sensor
     name: Watering
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -174,6 +174,17 @@ secondary_entities:
           - dps_val: 8
             value: true
           - value: false
+  - entity: binary_sensor
```
make-all, line 177: I would not include this, or if you want it for observation, make it an attribute of one of the other entities rather than a standalone sensor

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
```diff
@@ -196,8 +207,8 @@ secondary_entities:
         type: bitfield
         mapping:
           - dps_val: 64
-            value: true
-          - value: false
+            value: false
```
make-all, line 210: Do all of the sensors need inverting?  Some of them may be obvious, such as Tuya flags batteries as OK=1, Low=0, while HA expects the flag to indicate the exceptional case Normal=0, Low=1, but the Evie sensor may have been correct according to what the other user reported.

TwoWheelPro: I think they are correct now. the Battery values I assume are all the same, so the Evie batter should be the same as the 9V battery state. The moisture and watering bits' show correct state changes when I trigger the system or manually trigger the rain sensor.

custom_components/tuya_local/devices/holman_wx8_irrigation_controller.yaml
TwoWheelPro: I put this in when I was testing the mapping. I will remove it. I don't believe it means anything.

PR #1686, opened 2024-02-27
Update hydrotherm_dynamic_x8_water_heaterv6.yaml

knowc:
Use inlet temperature sensor as current_temperature state attribute for the water heater, for easier reference/visibility in HA. 

(Currently tested to be working well in my install.)


make-all:
Shouldn't the temperature of interest be the outlet temperature?

knowc:
> Shouldn't the temperature of interest be the outlet temperature?

I would have thought so too, but I think the default naming of these sensors are mistranslated (or something to that effect).

The 'outlet' sensor has remained at -50C at all times - nonsensical and uninformative.
The 'inlet' sensor matches the current temperature reported in the Tuya app, and sensibly increases when the heater is running, decreases over time and with hot water use, etc. So I am quite confident this is the correct information to use for current temperature.

Issue #1431, opened 2023-12-20
MADIMACK ECO 16KW INVERTER HEAT PUMP


wholly1971:
# Log Message

```
This error originated from a custom integration.

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:97
Integration: Tuya Local (documentation, issues)
First occurred: December 17, 2023 at 12:15:02 PM (2 occurrences)
Last logged: 8:05:56 PM

Device matches None with quality of 0%. DPS: {"updated_at": 1702779299.2215989, "101": false, "102": "heating", "104": 32, "105": "boost", "106": "c", "107": 0, "108": 29, "109": 0, "110": 40, "111": 18, "112": 29, "113": 32, "114": 32, "115": 34, "116": 0, "117": 29, "118": 33, "119": 37, "120": 39, "121": 480, "122": false, "123": false, "124": "oc_0"}
Device matches None with quality of 0%. DPS: {"updated_at": 1703066752.7406576, "101": false, "102": "heating", "104": 32, "105": "boost", "106": "c", "107": 0, "108": 30, "109": 0, "110": 40, "111": 18, "112": 27, "113": 28, "114": 30, "115": 26, "116": 0, "117": 30, "118": 29, "119": 29, "120": 32, "121": 480, "122": false, "123": false, "124": "oc_0"}
```

# Information about DPS mappings

```
{
  "result": {
    "model": "{\"modelId\":\"000004kb7r\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"默认服务\",\"properties\":[
{\"abilityId\":101,\"accessMode\":\"rw\",\"code\":\"switch1\",\"description\":\"\",\"extensions\":{\"attribute\":\"1\"},\"name\":\"开关\",\"typeSpec\":{\"type\":\"bool\",\"typeDefaultValue\":false}},
{\"abilityId\":102,\"accessMode\":\"rw\",\"code\":\"mode1\",\"description\":\"\",\"name\":\"模式\",\"typeSpec\":{\"range\":[\"auto\",\"heating\",\"cold\"],\"type\":\"enum\",\"typeDefaultValue\":\"auto\"}},
{\"abilityId\":103,\"accessMode\":\"rw\",\"code\":\"child_lock1\",\"description\":\"\",\"name\":\"童锁开关\",\"typeSpec\":{\"type\":\"bool\",\"typeDefaultValue\":false}},
[...]

PR #1691, opened 2024-03-02
Madimack Elite V3 Pool heatpump with updated dps mappings

jameskoziol:
Hello, I've been using your custom integration for a while to interface with my pool heat pump. I have recently had the whole thing replaced due to an electrical failure under warranty.

The new heat pump is identically named, but the dps mappings have changed. The heatpump was not auto-recognized by the existing integration.

I have rewritten the device file with the newer correct mappings and have renamed the file so it does not conflict with the original file for the older version of the unit.

I have also renamed a few of the diagnostic entities to correspond with the original control app. I also attempted to change the preset modes on the heater to correspond with the naming in the control app. Whilst this works, I am unable to specify appropriate icons in the latest version of HA (I don't understand yet how translations and the icons.yaml file work). So, I have left them as in the original file.

The newer heat pump also has dps maps for power (kW) functionality. I tried mapping it as well, but the data from it is weird. I think it only measures power for the control board, and not the whole unit. This would have been useful if it was properly implemented.


custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
```diff
@@ -0,0 +1,200 @@
+name: Madimack Elite V3 pool heatpump
```
make-all, line 1: Ideally this config should have a `products` section listing the `id` and branding part of the `name`, and this top level name should just be the generic part: "Pool heatpump".

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
```diff
@@ -0,0 +1,200 @@
...
+      range:
+        min: 18
+        max: 40
+    # ^ in reality, the range is different for cool and auto
+    - id: 105
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: silence
+          icon: "mdi:weather-night"
```
make-all, line 47: You shouldn't try to set icons from both hvac_mode and preset_mode.  Just from one place is less complex.

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
```diff
@@ -0,0 +1,200 @@
...
+        - constraint: temperature_unit
+          conditions:
+            - dps_val: f
+              range:
+                min: 60
+                max: 104
+      range:
+        min: 18
+        max: 40
+    # ^ in reality, the range is different for cool and auto
```
make-all, line 41: Since there are min_temperature and max_temperature dps, those will be used for the range in the UI.  The range here should be the min of the min, and max of the max, as this will be enforced on input at the integration level.

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
```diff
@@ -0,0 +1,200 @@
...
+          icon: "mdi:weather-night"
+          value: sleep
+        - dps_val: perfect
+          icon: "mdi:weather-sunny"
+          value: comfort
+        - dps_val: power
+          icon: "mdi:sun-thermometer"
+          value: boost
+    - id: 106
+      name: temperature_unit
```
make-all, line 56: Is this also settable?  There should be a select entity for setting it, if so.

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
jameskoziol: Yes, I can find the product ID through the tuya platform and change. However, I have noticed the product ID code found is identical to another device by the same manufacturer already in your system (madimack_eco_heatpump.yaml). Will using the same product ID code for a different unit be a problem?

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
jameskoziol: Understood - I'd like to try figure out how translation_key works to specify icons properly, but I'm having trouble finding documentation (I'm a HA newbie)

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
jameskoziol: This is how I found it in the original version. The dp is writeable - I'll add a select primary entity for this

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
make-all: There isn't much in the way of documentation, just the announcement here https://developers.home-assistant.io/blog/2024/01/19/icon-translations/ and official documentation here https://developers.home-assistant.io/docs/core/entity#icons

They document that state_attributes can be used to define icons for attributes like preset_mode, but don't go into detail about priorities when both state_attributes and state define icons.  But the default climate component defines icons for preset modes, so it must be handled.

I haven't looked at defining icons for climate entities yet, as the defaults are fairly comprehensive, but it is probably worth it for pool heatpumps.

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
make-all: I imagine when I eventually use this, it will be capable of showing a list, the same as the dps matching that is currently used (and will continue to be used as a backup).  But if the product ids are the same, maybe the config is also compatible (maybe only partially specified, as the knowledge of where to get good info from to make a complete config has evolved over time).

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
jameskoziol: I have added (as a secondary entity) a select switch for the units. It does successfully change the units from celsius to fahrenheit in the backend of tuya, changing the set temps and temperature limits accordingly.

This is difficult to test in home assistant though - the resulting climate card looks like it is being translated by HA from F to C, as the temperatures listed don't change. I suspect the system settings over-ride the units being used.

It may be best if I left this select switch undefined - all the remaining temperature sensors pull this DP to ensure the correct units, and HA converts to the system settings? What would you prefer?

make-all: You are correct. Home Assistant translates the temperatures in the frontend to the user's preferred units.  So if the backend is working correctly, the displayed value should not change.  If the backend is not doing things correctly, then you would see the values change but not match the units that are displayed in the frontend.

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
```diff
@@ -0,0 +1,203 @@
...
+              range:
+                min: 54
+                max: 104
+      range:
+        min: 12
+        max: 40
+    - id: 105
+      name: preset_mode
+      type: string
+      translation_key: madimack_presets
```
make-all, line 43: I don't think device specific translation keys are the way to go.  There are currently 627 configs which is a lot of individual icon customisation.

It is better to group into common functionality.  Pool heatpumps fairly commonly support "quiet" "quick" and "smart" presets, though many of them annoyingly have separate presets for cooling and heating modes, and I think we can group these together for both translations and icons.

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
jameskoziol: I have added a product ID line and the specific name of the unit I am using to my fork. I'm also a github newbie so let me know if you cannot see this.

As mentioned before - there may be an issue in that the product ID appears to be duplicated with other completely separate products made by the same manufacturer

custom_components/tuya_local/devices/madimack_elite_v3_heatpump_updated.yaml
jameskoziol: Yep - I gave it a crack to see if it would work on my end and no go. I'll revert the changes

make-all:
Can you check the info you are using to make this config against the info here: https://github.com/make-all/tuya-local/issues/1431

It looks like the dps assignments are the same, but probably some need  to be made optional, as they are not showing up on your unit.  They may have taken the Madimack Eco firmware and disabled a few features without changing the product id, I guess, or they use the same Tuya module firmware, but it has some conditions to enable/disable features based on some input to the module.

jameskoziol:
Hi yes - the DPS mappings do appear identical, even down to the product ID number. It's interesting that when I re-added the new unit, this particular device file was unable to match (I was presented only with "tuya_switch" as an option).

Scanning through the DPS values for the Eco unit - the main difference is in the preset modes. The Eco unit appears to have a "silence," "perfect," "power," and "boost" presets. The unit I have (Elite V3) does not have a "boost" mode option, just the other presets. Without knowing how your DPS mapping algorithm works, I suspect this may have made the difference.

The remaining DPS's are accurate - even the "power" field. On my unit however, the power field appears to be supplying garbage data that does not correspond to actual power outputs. I don't know if the Eco unit does this correctly.

Happy to be guided by you on next steps. It sounds like the manufacturer is just cobbling things together. I'm happy to run a custom file in my own instance and not add to the integration repository.

PR #1693, opened 2024-03-02
add support Smartlife Plant Sensor SGS01

almirus:


custom_components/tuya_local/devices/ble_smart_plant_moisture.yaml
```diff
@@ -0,0 +1,48 @@
+name: Moisture sensor
+products:
+  - id: bfbf6c7bb11573ed23j4ya
+    model: SGS01
```
make-all, line 4: This should be `name`, not `model`

custom_components/tuya_local/devices/ble_smart_plant_moisture.yaml
```diff
@@ -0,0 +1,48 @@
...
+        mapping:
+          - dps_val: "c"
+            value: "C"
+          - dps_val: "f"
+            value: "F"
+  -  entity: sensor
+     class: battery
+     category: diagnostic
+     dps:
+      - id: 14
```
make-all, line 42: I think the error on this line becomes OK if you fix the above one

custom_components/tuya_local/devices/ble_smart_plant_moisture.yaml
```diff
@@ -0,0 +1,48 @@
...
+    class: temperature
+    dps:
+      - id: 5
+        name: sensor
+        type: integer
+        mapping:
+          - scale: 10
+        unit: C
+  - entity: select
+    name: Temperature unit
```
make-all, line 27: In current head, there should be no name here, instead use `translation_key: temperature_unit`

make-all:
It is probably better to add additional devices as separate pull requests (using branches), as the first one is ready to merge, but now there are other devices with issues blocking the merge

PR #1696, opened 2024-03-03
add support Smartlife Plant Sensor SGS01

almirus:


custom_components/tuya_local/devices/ble_smart_plant_moisture.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - entity: sensor
+    name: Soil Temperature
+    class: temperature
+    dps:
+      - id: 5
+        name: sensor
+        type: integer
+        mapping:
+          - scale: 10
+        unit: C
```
make-all, line 25: What happens if temperature unit is changed? Does the device start reporting in F? If so, then the unit dp needs to be added to this entity rather than specifying a fixed unit here.

custom_components/tuya_local/devices/ble_smart_plant_moisture.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - entity: select
+    translation_key: temperature_unit
+    category: config
+    dps:
+      - id: 9
+        name: option
+        type: string
+        mapping:
+          - dps_val: "c"
+            value: "C"
```
make-all, line 35: values in the select entity should be `celsius` and `fahrenheit` for the translations and icons to work correctly (the `C` and `F` are correct when it is used as a unit dp to specify units on the actual temperature sensor though)

custom_components/tuya_local/devices/ble_smart_plant_moisture.yaml
almirus: Okay, I will delete it. Despite what is stated in the document, it doesn't work.

```
{
                    "abilityId": 9,
                    "accessMode": "rw",
                    "code": "temp_unit_convert",
                    "description": "",
                    "extensions": {
                        "iconName": "icon-dp_mode",
                        "attribute": "1152"
                    },
                    "name": "温标切换",
                    "typeSpec": {
                        "range": ["c", "f"],
                        "type": "enum",
                        "typeDefaultValue": "c"
                    }
                }
```

almirus: @make-all Without this parameter (DP 9), the device cannot be added, but as mentioned earlier, changing the unit of measurement does not lead to anything. What if the parameter is needed but needs to be hidden?

make-all: How did you determine that the change did not work?

The UI converts temperatures according to the user preferences, so if the device reports 20C and units are set to C, it will show 20C, but if the device reports 68F, it will also show 20C.

But a missing DP should not prevent a device from being added. It might make a different config look like a better match  than the actual config you want, but this config should still be listed, even if it is not selected by default.

almirus: I based it on the Tuya documentation. There is a DP for changing the temperature unit of measurement. However, it is not possible to change the unit of measurement within the Tuya Smart application as there is no such option available. I make the change in tuya-local, but after a minute, the parameter returns to "Celsius".

There was nothing better than my converter, so it simply refused to be added without that DP.

custom_components/tuya_local/devices/ble_smart_plant_moisture.yaml
```diff
@@ -0,0 +1,42 @@
+name: Moisture sensor
+products:
+  - id: bfbf6c7bb11573ed23j4ya
```
make-all, line 3: Usually BLE devices will have a shorter product id.  Actually the format (valid hex digits except the last 4) looks more like a device id (which is specific to only your device)

almirus: ooops 😣
gvygg3m8

pilonull:
Hey, um, potentially stupid question... I've been trying to get SGS01 plant sensor info in to home assistant. 

Does this integration require a tuya bluetooth hub to work? Or would it work with the HA bluetooth integration (Bluetooth dongle on HA PC, or esphome bluetooth proxies)? 

I have tried several forks of the ha_tuya_ble integration but none of them seem to work either, I can get the SGS01 sensor added in to HA but it does not give me any data, either the temp and moisture sensors show up as unavailable or they do not show up at all...

almirus:
@pilonull yes, you need hub

Joao-Sousa-71:
I’ve integrated this sensor and it doesn’t update regularly, but if I open the Smart Life mobile app it does update the sensor data right away. Anybody else with the same issue?

almirus:
> t doesn’t update regularly

I have exactly this behavior of this sensor, you need to write to the developers of the sensor.

Joao-Sousa-71:
> > t doesn’t update regularly
> 
> I have exactly this behavior of this sensor, you need to write to the developers of the sensor.

It is a fw issue then… I’ll try to find who are the developers of this sensor. Didn’t knew if it was something that the integration could manage it.

make-all:
If you can find out how the app manages to get the sensor to update more frequently, it is possible that some way to use that same method could be added.  But most likely this is a battery conservation measure, so doing it all the time would run through batteries too quickly.

Joao-Sousa-71:
> If you can find out how the app manages to get the sensor to update more frequently, it is possible that some way to use that same method could be added. But most likely this is a battery conservation measure, so doing it all the time would run through batteries too quickly.

That seems a great idea, I guess that every 60 minutes it is more than enough. This sensor uses two AAA batteries which for a BLE is quite powerful (thinking on a PC mouse that batteries last for years nowadays).
I can enable logs or anything that can help, l’m new on Tuya wold and apologies for my ignorance but glad to help the community.

almirus:
I think is that a defect resides within the firmware's code (or perhaps a universal firmware intended for low-capacity battery devices is employed), whereby the sensor perpetually amasses data, yet only transmits said data upon receiving a specialized request from the application or hub. Regrettably, this renders the sensor devoid of utility for the purpose of creating automatizations

Joao-Sousa-71:
I can confirm that json definitely works! If sent within a short period of time will not update the hub, but at least with 40 minutes it works. @make-all can you implement this?

Joao-Sousa-71:
> No access to the C/F setting has been provided in the pull request, so that cannot be done without more info on the dps mappings.

Understand. looking to the code dps mappings seems to have a relation with the teh json commands:

image

@almirus can you add something on this?

make-all:
DPS mappings are in Cloud / API Explorer / Device Control / Query Things Data Model.  You need to set your data center are the top of the page, and fill in the device id for your SGS01.

make-all:
Apologies, the temperature unit is exposed as read-only within the temperature sensor. So it just needs to expose the same dp via a select entity.

Joao-Sousa-71:
> Apologies, the temperature unit is exposed as read-only within the temperature sensor. So it just needs to expose the same dp via a select entity.

No need to apologize. I've applied your change and _voilá_ it works as charm 👍 
With the new select, the service to call is very straight:

```
service: select.select_option
target:
  entity_id: select.sensor_humidade_vasos_temperature_unit
data:
  option: celsius
```
I'm now checking how often can be called starting with an automation every 30 minutes for test purposes, my goal is to update the  sensorwhile the irrigation is running and before its stops. 
Thank you @make-all

Joao-Sousa-71:
A new update: Like in the Smart Life app the "update" retrieves not the actual/last value, but recorded data stored in the device, this behavior is confirmed in this discussion  #1910 by Adrian.
See here the humidity:

image

and the same for the temperature:

image

Now it will be just a matter of findind the right interval to retrieve the last value only and profit 👍

almirus:
@Joao-Sousa-71 as I had previously stated, the sensor stores all statistics over a defined period of time (temperature and humidity), but transmits the entire data set simultaneously upon receiving a request from the hub or application.
It is plausible that in conjunction with the aforementioned data transmission, a timestamp is also reported

Joao-Sousa-71:
> @Joao-Sousa-71 as I had previously stated, the sensor stores all statistics over a defined period of time (temperature and humidity), but transmits the entire data set simultaneously upon receiving a request from the hub or application. It is plausible that in conjunction with the aforementioned data transmission, a timestamp is also reported

Make sense. So far 30 minutes seems a good interval and yet didn't receive more that one value of humidity and temperature. Need to change the location of the sensor and force temperature changes within the period.

Joao-Sousa-71:
i've been making adjustments and with 30 and 20 minutes cadence, still have a just a few syncs where more than one value is sent to the Gateway. I'm now testing 15 minutes (which is 4 updates per hour) to see how it goes.
Keep you posted.

image

almirus:
@Joao-Sousa-71 I have a version that periodically switches to Celsius (only). I am using the following automation:
@jmotor 

```
alias: >-
  Для обновления датчиков влажности почты нужно дернуть установку единицы
  измерения
description: ""
trigger:
  - platform: time_pattern
    minutes: "01"
    hours: "13"
    seconds: "00"
condition: []
action:
  - device_id: dd122e973bd3631ebf95e6c23b2099ea
    domain: select
    entity_id: 1b60addd90ea0d4b1b01798a178431b0
    type: select_last
    enabled: false
  - device_id: e9bd7c36bda01143ee01902642c4fb01
    domain: select
    entity_id: 3b9254e2fdc12e862aeec6cc2cc4afc5
    type: select_last
    enabled: false
  - delay:
      hours: 0
      minutes: 0
      seconds: 2
      milliseconds: 0
    enabled: false
  - device_id: dd122e973bd3631ebf95e6c23b2099ea
    domain: select
    entity_id: 1b60addd90ea0d4b1b01798a178431b0
    type: select_first
  - device_id: e9bd7c36bda01143ee01902642c4fb01
    domain: select
    entity_id: 3b9254e2fdc12e862aeec6cc2cc4afc5
    type: select_first
mode: single

```

jmotor:
> @Joao-Sousa-71 I have a version that periodically switches to Celsius (only). I am using the following automation: @jmotor
> 
> ```
> alias: >-
>   Для обновления датчиков влажности почты нужно дернуть установку единицы
>   измерения
> description: ""
> trigger:
>   - platform: time_pattern
>     minutes: "01"
>     hours: "13"
>     seconds: "00"
Хорошо быть умным...
Я туповат в этом вопросе...
Это надо вписать в...? Куда?
Или ещё ID поменять?

Joao-Sousa-71:
> How did you do it? Do you have detailed instructions?

just this:
´´´
# Update soil moisture sensor every X minutes
- id: update_ble_soil_moisture_sensor
  alias: update_ble_soil_moisture_sensor
  description: 'Atualiza o Sensor de Humidade Horta'
  initial_state: true
  trigger:
    - platform: time_pattern
      minutes: "/15"  
  action:     
    - service: select.select_option
      target:
        entity_id: select.sensor_humidade_vasos_temperature_unit
      data:
        option: celsius
´´´

DanilinSA:
```
    device_id: dd122e973bd3631ebf95e6c23b2099ea
    entity_id: 1b60addd90ea0d4b1b01798a178431b0
    device_id: e9bd7c36bda01143ee01902642c4fb01
    entity_id: 3b9254e2fdc12e862aeec6cc2cc4afc5
```

What do these numbers mean? Where did you get them?

make-all:
These are internal ids that HA uses when you set up an automation from the UI. Unlike `select.sensor_humidade_vasos_temperature_unit`, the hex id cannot be changed by the user.

make-all:
The refresh button is using a hack to try to get the hub to communicate immediately with the devices so they send updates. What I have observed is that when you use the phone app, the app communicates directly with the devices via Bluetooth, so it may be using a method which is not exposed via the hubs to get them to send the data. When the devices do send data, they seem to send a batch of updates in succession, in the cloud there seems to be a timestamp which allows these to be displayed properly in a graph, but the timestamp is not exposed via the local protocol, so they all appear as a spike within a few seconds. Often the last one sent is identical to the previous reading (maybe a bug in the hub firmware) so the sensors do not appear to be updating even though the history shows updates received.

PR #1698, opened 2024-03-03
add support Smart curtain robot Dongguan LY1678-2

almirus:


custom_components/tuya_local/devices/dongguan_electric_curtain.yaml
```diff
@@ -0,0 +1,89 @@
+name: Smart curtain robot
+products:
+  - id: kcy0x4pi
+    name: Dongguan LY1678-2 Electric Curtain Motor
+primary_entity:
+  entity: cover
+  name: Curtain
```
make-all, line 7: No need to include a name when there is a class of the same name already.
A translated name is already defined here: https://github.com/home-assistant/core/blob/dev/homeassistant/components/cover/strings.json#L56

PR #1700, opened 2024-03-03
add LightStar track light

almirus:


make-all (commented):
I'm not sure what is materially different about this config compared to other cct lights, as the dp mapping seems standard and only dp 106 is additional, but it isn't really anything useful.

PR #1703, opened 2024-03-03
add Royal fresh full inverter

almirus:


custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+      type: boolean
+      name: hot_cool
+    - id: 133
+      type: string
+      name: swing_action
+    - id: 134
+      type: json
+      name: statistics
+secondary_entities:
+  - entity: fan
```
make-all, line 151: 102 is already defined as preset_mode of the climate entity, it does not need to be duplicated as a partially functional fan, and again as a select entity,

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+            value: "Off"
+          - dps_val: low
+            value: Fresh Air Low
+          - dps_val: mid
+            value: Fresh Air Medium
+          - dps_val: strong
+            value: Fresh Air High
+          - dps_val: auto
+            value: Fresh Air Auto
+  - entity: sensor
```
make-all, line 181: This is already available in the climate entity, so does not need to be duplicated as a sensor

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+    class: temperature
+    dps:
+      - id: 3
+        name: sensor
+        type: integer
+        unit: C
+        class: measurement
+  - entity: sensor
```
make-all, line 189: climate supports current_humidity also, so this can go in there.

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+    name: Beep
+    category: config
+    dps:
+      - id: 123
+        type: hex
+        name: switch
+        mapping:
+          - mask: "0010"
+  - entity: binary_sensor
+    name: Fault
```
make-all, line 372: Should be unnamed, to make use of the default HA translations for the "problem" class.

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+    dps:
+      - id: 20
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+  - entity: sensor
+    name:
```
make-all, line 384: Remove this empty name

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+        min: 160
+        max: 880
+      mapping:
+        - scale: 10
+          step: 5
+      unit: C
+    - id: 3
+      name: current_temperature
+      type: integer
+      unit: C
```
make-all, line 41: climate entities only have a single temperature unit, so this does not need to be specified here as well as on the temperature dp

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+        - dps_val: mid_high
+          value: medhigh
+        - dps_val: high
+          value: high
+        - dps_val: strong
+          value: strong
+    - id: 102
+      name: preset_mode
+      type: string
+      translation_key: Recirculate
```
make-all, line 69: translation_key can only be specified at entity level, and also needs to be lower case.
Perhaps some preset_mode mappings can be added to aircon_extra to cover these, or separate preset_mode into a fan entity with class ventilation, mapping these to "circulate", 2 of "eco"/"sleep"/"anti-condensation", "fresh" and "auto" (use dp 1 as the switch for the fan as well) since this isn't really a preset of the climate control, but a separate control for air circulation.

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+        - dps_val: mid
+          value: Fresh Air Medium
+        - dps_val: strong
+          value: Fresh Air High
+        - dps_val: auto
+          value: Fresh Air Auto
+    - id: 20
+      type: bitfield
+      name: fault_code
+    - id: 105
```
make-all, line 84: Since this is exposed as a switch below, it does not need to be included here.

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+      type: string
+      name: horizontal_swing
+      hidden: true
+    - id: 119
+      type: string
+      name: electricity_management
+    - id: 120
+      type: string
+      name: gen_mode
+    - id: 123
```
make-all, line 125: This should perhaps be duplicated with the mask for eco_mode, so the value can be used in a condition for the temperature dp, to use eco_temp instead.

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -0,0 +1,409 @@
...
+    category: config
+    icon: "mdi:leaf"
+    dps:
+      - id: 123
+        type: hex
+        name: switch
+        mapping:
+          mask: "0001"
+  - entity: switch
+    name: Anti-freeze
```
make-all, line 333: name here should be replaced with `translation_key: anti_frost`, to use standard translations and icons. icon below should also be removed.

PR #1724, opened 2024-03-09
Added new light devices

zeosson:
added two new light devices  
used tiny tuya to get the values


custom_components/tuya_local/devices/peteme_smart_recessed_lighting.yaml
```diff
@@ -0,0 +1,71 @@
+name: Peteme Smart Recessed Lighting
+products:
+    - id: sjaun9de
+    name: SIG_DL_RGBCW
+primary_entity:
+  entity: light
```
make-all, line 6: The indentation here is OK, it is because the first indented line has 4 leading spaces, the lint checker has configured itself for 4 rather than 2 spaces.

PR #1731, opened 2024-03-10
Added support for Silvercrest coffee maker

popeen:
Added support for Silvercrest Coffee maker.

### DIFF
--- a/custom_components/tuya_local/devices/silvercrest_coffee_maker.yaml
+++ b/custom_components/tuya_local/devices/silvercrest_coffee_maker.yaml
@@ -1,7 +1,7 @@
 name: Coffee maker
 products:
   - id: iegq4tjqeuiew5cl
-    name: Coffee Maker
+    name: Silvercrest SKMS 900 A1
 primary_entity:
   entity: switch
   icon: "mdi:coffee-maker"
@@ -34,19 +34,22 @@ secondary_entities:
         type: string
         name: sensor
   - entity: sensor
-    name: Warm left
-    icon: "mdi:clock"
+    name: Warm time remaining
+    class: duration
     category: diagnostic
     dps:
       - id: 12
         type: integer
         name: sensor
         unit: min
-  - entity: sensor
-    name: Lack water
-    icon: "mdi:water"
+  - entity: binary_sensor
+    translation_key: tank_empty
     category: diagnostic
     dps:
       - id: 105
         type: bitfield
         name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/silvercrest_coffee_maker.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - id: 1
+      type: boolean
+      name: operation_mode
+      mapping:
+        - dps_val: false
+          value: "Off"
+        - dps_val: true
+          value: "On"
+    # Dummy dps, water_heater requires a temperature.
+    # The coffee maker has no sensor for this so we put in a fake dps for it.
```
make-all, line 18: water_heater does not require temperature, do not add a dummy.

custom_components/tuya_local/devices/silvercrest_coffee_maker.yaml
```diff
@@ -0,0 +1,51 @@
...
+    name: Coffee Maker
+primary_entity:
+  entity: water_heater
+  icon: "mdi:coffee-maker"
+  dps:
+    - id: 1
+      type: boolean
+      name: operation_mode
+      mapping:
+        - dps_val: false
```
make-all, line 13: It is better to use predefined operation modes from https://github.com/home-assistant/core/blob/dev/homeassistant/components/water_heater/strings.json#L11
"off" and "electric" are probably best (note all lower case)

custom_components/tuya_local/devices/silvercrest_coffee_maker.yaml
```diff
@@ -0,0 +1,51 @@
...
+      range:
+        min: 0
+        max: 0
+    - id: 4
+      type: string
+      name: current_state
+    - id: 12
+      type: string
+      name: keep_warm_min
+    - id: 101
```
make-all, line 33: Does not need to be here if it is unused by the water_heater, and has a dedicated entity below.

custom_components/tuya_local/devices/silvercrest_coffee_maker.yaml
```diff
@@ -0,0 +1,51 @@
+name: Silvercrest Coffee Maker
```
make-all, line 1: This name should be swapped with the product specific name below.  Also, Home Assistant naming convention is to only capitalise the first word of names, unless trademarks or abbreviations require it.

popeen:
Thanks for great feedback!

Made me go back and try switch again as specified by the data model and this time could make it work, I must have done something wrong the first time. Meaning it is no longer a water_heater. I am still not sure why I couldn't add the water meter without temperature but I'm sure it was something on my end.

I have changed the name and moved all the properties to separate entities. I think it should be ok now.

PR #1736, opened 2024-03-11
New device: Madimack Inverflow Pro variable speed pool water pump

jameskoziol:
This device is a variable speed pool water pump. The specific model I have is Madimack Inverflow Pro P300i. There are another two pumps in the range with more capacity. The manufacturer also makes a Inverflow Ultra pump which may be compatible. The Inverflow Eco does not have wifi/Tuya control. These water pumps are often used with Madimack heatpumps.

There is an obscure control mechanism which may need review:
The primary entity is set as a fan (closest matching entity available). The only controllable aspect is on/off. For the speed setting, I have added a read-only DP from Tuya that reports the pump actual speed, on a range of 30-120. This DP is always updated, regardless of the pump mode.

I have added several secondary config entities enabling full control of the pump. Firstly, there are three control modes:
1. "Boost" mode - which sets the motor at 100% and runs on a settable timer. This is the default mode on startup, and the default timer is 180s.
2. Manual control - this allows the pump to be set to a % operating capacity, with a range of 30-120%
3. AI flow mode - this allows the pump to be set to a flow rate with 4 different options of flow units.
Of note - the controls for all 3 of the above modes work independently. Mapping the control DP's of these modes to the primary entity "speed" control can result in odd behaviour. (e.g. if primary entity "speed" reads off the Manual control DP, but a flow rate is set in "AI" mode, the primary entity "speed" is not updated - it can only be updated by using the read-only "real-speed" DP.

There is also a selectable flow rate units option in the pump. This interestingly does not work on my unit, despite it being "R/W" in Tuya - I cannot change from the L/min default.

There are also several useful diagnostic DP's. These include a measurement of real flow rates, current power, diagnostic fault codes and an hourly energy usage DP (this DP I have set as "measurement" type, not "total-increasing" as it would not be compatible with HA's energy panel - I'd recommend end users setup an Integration helper).

I'd like feedback on this setup - there are a few other possible options if I activate some constraints that may be less confusing?

### DIFF
--- a/custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
+++ b/custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
@@ -1,0 +1,177 @@
+name: Pool pump
+products:
+  - id: ircs2n82vgrozoew
+    name: Madimack Inverflow Pro P300i
+primary_entity:
+  entity: fan
+  dps:
+    - id: 105
+      type: boolean
+      name: switch
+    - id: 102
+      type: integer
+      name: speed
+      range:
+        min: 30
+        max: 120
+      readonly: true
+    - id: 116
+      name: unknown_116
+      type: string
+    - id: 117
+      name: unknown_117
+      type: string
+    - id: 115
+      name: unknown_115
+      type: boolean
+secondary_entities:
+  - entity: select
+    name: Pump mode
+    translation_key: mode
+    category: config
+    dps:
+      - id: 103
+        type: string
+        name: option
+        mapping:
+          - dps_val: "MI"
+            value: "Manual"
+          - dps_val: "AI"
+            value: "AI Flow set"
+          - dps_val: "backwash"
+            value: "Boost"
+  - entity: select
+    name: Pump flow units
+    translation_key: units
+    category: config
+    dps:
+      - id: 110
+        type: string
+        name: option
+        mapping:
+          - dps_val: "m3_h"
+            value: "m³/h"
+          - dps_val: "l_min"
+            value: "L/min"
+          - dps_val: "us_gpm"
+            value: "US gpm"
+          - dps_val: "ipm_gpm"
+            value: "International gpm"
+  - entity: number
+    name: Boost timer set
+    category: config
+    mode: box
+    dps:
+      - id: 104
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 1500
+        unit: sec
+  - entity: number
+    name: AI flow rate set
+    category: config
+    mode: slider
+    dps:
+      - id: 106
+        type: integer
+        name: value
+      - id: 101
+        type: integer
+        name: maximum
+      - id: 107
+        type: integer
+        name: minimum
+      - id: 113
+        type: integer
+        name: step
+      - id: 110
+        type: string
+        name: unit
+        mapping:
+          - dps_val: "m3_h"
+            value: "m³/h"
+          - dps_val: "l_min"
+            value: "L/min"
+          - dps_val: "us_gpm"
+            value: "gal/min"
+          - dps_val: "ipm_gpm"
+            value: "gal/min"
+  - entity: number
+    name: Manual percentage power set
+    category: config
+    mode: slider
+    dps:
+      - id: 111
+        type: integer
+        name: value
+        range:
+          min: 30
+          max: 120
+        step: 5
+        unit: "%"
+  - entity: sensor
+    name: Boost time remaining
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 108
+        type: integer
+        name: sensor
+        unit: sec
+  - entity: sensor
+    name: Current power
+    class: power
+    category: diagnostic
+    dps:
+      - id: 5
+        type: integer
+        name: sensor
+        unit: W
+  - entity: sensor
+    name: Flow rate
+    class: volume_flow_rate
+    category: diagnostic
+    dps:
+      - id: 112
+        type: integer
+        name: sensor
+      - id: 110
+        type: string
+        name: unit
+        mapping:
+          - dps_val: "m3_h"
+            value: "m³/h"
+          - dps_val: "l_min"
+            value: "L/min"
+          - dps_val: "us_gpm"
+            value: "gal/min"
+          - dps_val: "ipm_gpm"
+            value: "gal/min"
+  - entity: sensor
+    name: Hourly energy usage
+    class: energy
+    category: diagnostic
+    dps:
+      - id: 109
+        type: integer
+        name: sensor
+        class: measurement
+        unit: kWh
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Fault code
+    category: diagnostic
+    dps:
+      - id: 2
+        type: bitfield
+        name: sensor
+  - entity: sensor
+    name: Flow pressure warning
+    category: diagnostic
+    dps:
+      - id: 114
+        type: boolean
+        name: sensor


custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+        name: option
+        mapping:
+          - dps_val: "m3_h"
+            value: "m³/h"
+          - dps_val: "l_min"
+            value: "L/min"
+          - dps_val: "us_gpm"
+            value: "US gpm"
+          - dps_val: "ipm_gpm"
+            value: "International gpm"
```
make-all, line 59: I think it is Imperial rather than International (SI units are the only international units)

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+    category: config
+    mode: box
+    dps:
+      - id: 104
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 1500
+        unit: sec
```
make-all, line 71: `s` is the correct unit for seconds. (only min uses multiple letters, to distinguish from meters)

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+          - dps_val: "m3_h"
+            value: "m³/h"
+          - dps_val: "l_min"
+            value: "L/min"
+          - dps_val: "us_gpm"
+            value: "US gpm"
+          - dps_val: "ipm_gpm"
+            value: "International gpm"
+  - entity: number
+    name: Boost timer set
```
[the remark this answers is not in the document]
make-all: Although this probably benefits from keeping the name to make it clear what the timer is for, it should probably still have `translation_key: timer` for the icon customization and any future "translation" related features.

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+        unit: "%"
+  - entity: sensor
+    name: Boost time remaining
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 108
+        type: integer
+        name: sensor
+        unit: sec
```
make-all, line 122: `s` definitely needed here to avoid a warning about incompatible units for a duration sensor.

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+  - entity: sensor
+    name: Current power
```
make-all, line 124: Remove the name, it is better to inherit from the power class.

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+    class: power
+    category: diagnostic
+    dps:
+      - id: 5
+        type: integer
+        name: sensor
+        unit: W
+  - entity: sensor
+    name: Flow rate
```
make-all, line 133: Remove the name, it is better to inherit from the volume_flow_rate class

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+    category: diagnostic
+    dps:
+      - id: 109
+        type: integer
+        name: sensor
+        class: measurement
+        unit: kWh
+        mapping:
+          - scale: 100
+  - entity: sensor
```
make-all, line 164: Without any decoding of the sensor values, it is generally better to use a binary_sensor with `class: problem` with 0 mapped to false, and any other value mapped to true. Then include the dp a second time with a name like `fault_code` so the raw value is still available as an attribute.

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+    name: Fault code
+    category: diagnostic
+    dps:
+      - id: 2
+        type: bitfield
+        name: sensor
+  - entity: sensor
```
make-all, line 171: Since the dp is boolean, this should be a binary_sensor. Again the `class: problem` is probably appropriate, but this one should keep its name, since it is not a general problem indicator, but for a specific purpose.

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+    category: config
+    mode: slider
+    dps:
+      - id: 111
+        type: integer
+        name: value
+        range:
+          min: 30
+          max: 120
+        step: 5
```
make-all, line 112: `step` needs to be set inside a mapping (the logic is that a step causes set values to snap to nearest step, so it modifies the values - hence under mapping)

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
+name: Pool pump
+products:
+  - id: ircs2n82vgrozoew
+    name: Madimack Inverflow Pro P300i
+primary_entity:
+  entity: fan
```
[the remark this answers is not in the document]
jameskoziol: Done. I wasn't sure of the best logic here. Trying to decide between fan/switch/valve. I suppose valve makes sense as it is a water pump

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
```diff
@@ -0,0 +1,177 @@
...
+            value: "gal/min"
+  - entity: sensor
+    name: Hourly energy usage
+    class: energy
+    category: diagnostic
+    dps:
+      - id: 109
+        type: integer
+        name: sensor
+        class: measurement
```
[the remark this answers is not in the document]
jameskoziol: This is a potential problem for this sensor - it quite literally only reports average energy usage over an hour, and as a result is incompatible with HA's logic (or just common sense) for energy sensors. Should I perhaps unmap the sensor as an unknown DP?

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: Thanks! shows my ignorance living in a metric only country

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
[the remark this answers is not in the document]
jameskoziol: removed "set" from config entities and added translation key

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: Thanks. Noted after publishing initial commit. Was working on another project at same time that used sec as abbreviation.

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: Surprisingly it still worked without an error, but updated.

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: Done

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: Also done

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: Done. Fault code now separate attribute for fault warning sensor

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: Done

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: step request removed. Behaviour of slider is that the unit auto-corrects to a 5% increment anyway

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: As an aside - I can if you'd prefer map the error codes to their published descriptions. Just needs a fair bit of copy/paste. Probably unnecessary though with the fault code as a separate attribute

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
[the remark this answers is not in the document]
jameskoziol: In my own instance, I've created a Riemann integration for energy based on the power sensor, which is probably going to be more accurate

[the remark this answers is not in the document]
make-all: I think you should just remove the energy class if it is unsuitable for use in the Energy dashboard.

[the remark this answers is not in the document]
jameskoziol: OK. I will remove the energy class, leaving it as a reported sensor only.

I've just pushed a commit with human readable error codes as well.

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
make-all: This depends on how useful those specific problems are. I tend to split out fault codes that are common user maintenance actions such as tank full/empty, filter replacement etc into their own binary_sensor entity, and leave the rare faults grouped as a single sensor with more details in an attribute (decoded if you know all the details, or at least the raw number so the user can work it out over time if a fault is recurring)

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
make-all: This device seems to validate its input properly then, but many tuya devices don't which is why we need the option to encode that in the UI to avoid sending unexpected values to the device (also in some specific cases HA uses step, but not here).

custom_components/tuya_local/devices/madimack_inverflow_pro.yaml
jameskoziol: My latest upload to my fork has the error codes mapped (turned out the online manual had the entries in a table, which was easy to transcribe). Being a github newbie, not sure if the updates are visible on your end or require a new pull request?

make-all: It is visible, pushing to the same branch you submitted as a PR updates the PR.

PR #1742, opened 2024-03-12
Improve Avatto ME81 thermostats

erelke:
reflect hvac action on icon


make-all:
Now that icon translations are supported, I don't want to add additional icon config like this.  The default climate entity already has logic to handle hvac_action, and the "fix" is perhaps to remove any other icon customization that might be overriding it.

make-all:
Manually merged the product name changes.

PR #1748, opened 2024-03-13
Add support for Shinco 50D Dehumidifier

shiner66:


make-all:
This is substantially the same as Shinco 30D. Only the ionizer switch seems to be missing.

shiner66:
It is, but it is necessary to make the integration recognize it.

make-all:
You seem to have deleted your branch. Does that mean the commit I made above is working?

shiner66:
It’s unstable, I tried multiple times to add the dehumidifier and it doesn’t always recognize it, but most of the times it does work.

Sent with Spark
Il 14 mar 2024, 06:06 +0100, Jason Rumney ***@***.***>, ha scritto:

PR #1761, opened 2024-03-20
Update and rename zx_db11b_doorbell_alarm.yaml to zx_db11b_doorbell_a…

salvin-nl:
…larm2.yaml


make-all:
Renaming config files after they ship is a breaking change, and there does not seem to be any reason for it.
I also don't see the rationale for renaming the product to another non-brand name, and making a base64 value available as a sensor.

salvin-nl:
Hi, I'm trying to add base64 sensor, so that i can differentiate between different types of devices being used. modifying binary sensor to sensor is giving a list of sensors, that is helpful in identifying which calling button is pressed

PR #1763, opened 2024-03-21
Add support for HeatStorm HS-1500 heater

cw-cdz:
Add support for the [HeatStorm HS-1500](https://www.amazon.com/Heat-Storm-HS-1500-PHX-WIFI-Infrared-Heater/dp/B07JXRWJ8D) wall mounted heater. 

I validated DPs via Tuya developer account website with my own devices. It is mostly similar to the HS-6000 excepting some DPs, namely the hvac_mode one which is not present on the hs1500.

### DIFF
--- a/custom_components/tuya_local/devices/heatstorm_hs1500_heater.yaml
+++ b/custom_components/tuya_local/devices/heatstorm_hs1500_heater.yaml
@@ -1,0 +1,132 @@
+name: Heatstorm heater
+products:
+  - id: gifk0oubmovuz3ky
+    name: HS-1500
+primary_entity:
+  entity: climate
+  dps:
+    - id: 1
+      type: boolean
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: heat
+    - id: 2
+      type: integer
+      name: temperature
+      range:
+        min: 4
+        max: 37
+      mapping:
+        - constraint: temperature_unit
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/heatstorm_hs1500_heater.yaml
```diff
@@ -0,0 +1,132 @@
...
+      - id: 7
+        type: boolean
+        name: lock
+        mapping:
+          - dps_val: true
+            value: false
+          - dps_val: false
+            value: true
+  - entity: binary_sensor
+    name: Fault
```
make-all, line 109: name should be removed, ut use the standard translations for problem class

custom_components/tuya_local/devices/heatstorm_hs1500_heater.yaml
```diff
@@ -0,0 +1,132 @@
...
+    dps:
+      - id: 13
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+  - entity: select
+    name: Temperature unit
```
make-all, line 121: name and icon should be replaced with `translation_key: temperature_unit` and values be in all lower case to use translations.

make-all:
My condolences for your loss. I am actually in the same situation myself, which is delaying the next release and proper review of PRs which are starting to stack up.

PR #1770, opened 2024-03-23
Add support for Sol*Aire Vitra Wifi Electric Panel Heater

benzoid:
Adapted from andersson_gsh_heater.yaml which mostly worked, but led to "antifreezen" in the Preset names instead of Away.  Also adjusted min and max temperature to the values limited in the Smart Life app.
This is tested with a 2000W Sol*Aire Vitra Wifi Electric Panel Heater but should work with all Vitra models.
See https://adax-solaire.co.uk/product/vitra-wifi-electric-panel-heater-wall-mounted-or-portable/


benzoid:
Updating the config with secondary entities

benzoid:
Secondary entities Child lock, Temperature calibration and Display added.
Preset mode mapping corrected.

custom_components/tuya_local/devices/solaire_vitra_wifi_electric_panel_heater.yaml
```diff
@@ -0,0 +1,72 @@
+name: Sol*Aire Vitra WiFi Electric Panel Heater
+primary_entity:
```
make-all, line 2: It would be useful to have a `products` section here with the product_id (or productKey) of your heater and the branding parts of the name.  The top level name can then be shortened to just the generic portion without redundant words like WiFi, in case other brands also match (such as rebadged heaters sold elsewhere in the world).

custom_components/tuya_local/devices/solaire_vitra_wifi_electric_panel_heater.yaml
```diff
@@ -0,0 +1,72 @@
...
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 6
+        type: boolean
+        name: lock
+        mapping:
+          - dps_val: true
+            value: true
+            icon: "mdi:account-lock"
```
make-all, line 62: It is not necessary to override the icons here.  You are free to suggest a change in icons.json, but I don't see a strong case for the change unless there are multiple other integrations using these icons.

custom_components/tuya_local/devices/solaire_vitra_wifi_electric_panel_heater.yaml
benzoid: Where would I find the product_id (if available)?  I can't find anything in the json from Download Diagnostics.
In the eu.iot.tuya.com Devices page, there is a column "Product" (between "Device ID" and "Source") with unpromising value "New-Solaire smart heater S2".  I'm guessing that might be products.name (can I have a name without an id?)?  The manual refers to "Vitra (s2)" so I think that might be the range of heaters.
This is my first Tuya device (that I'm using through Smart Things anyway -- always trying to stay local hence this integration!) so some guidance would be very welcome, thank you.

custom_components/tuya_local/devices/solaire_vitra_wifi_electric_panel_heater.yaml
benzoid: These icons more closely match the equivalent function icons in the Smart Things app, but I can appreciate consistency for functions across devices within Home Assistant so happy to leave those as default.
I noticed that the hvac_mode icons are overridden in the andersson_gsh_heater.yaml that got me started figuring out this heater, should they be taken out and default (climate?) icon used?
Thanks again for any help.

make-all: I haven't yet defined icon translations for common climate subcategories, so they are still using the legacy method to customise.  I think at least the default hvac_off icon probably doesn't match heaters well, so have not removed those yet.

custom_components/tuya_local/devices/solaire_vitra_wifi_electric_panel_heater.yaml
make-all: The name does not have to be exactly what is in iot.tuya.com, but somewhere near that (in the same calls that give you local_key) there should be a product_id as well.  It is also visible if you do a network scan from tinytuya on the command line.

It is not in device diagnostics, because there is no known method to get this from a device that is connected locally. Only from the discovery broadcast messages and the cloud.

benzoid: I'll get round to having a look at using tinytuya soon.

benzoid: Added products.id, products.name and changed name and filename to match.

bigheadsmith:
No way, I was spending all of last night trying to get this exact heater working

benzoid:
> No way, I was spending all of last night trying to get this exact heater working

@bigheadsmith Glad the config will get some use straight away, once released.  Apologies I didn't get round to doing it sooner and saved you an evening!

PR #1790, opened 2024-04-01
Add support for Tongou TO-Q-SY1-JWT

spunky3378:
Add support for Tongou Device TO-Q-SY1-JWT (WIFI Smart Circuit Breaker Power Metering 1P DIN Rail)


make-all:
Device is already supported by smartplugv2_energyv2.

PR #1810, opened 2024-04-09
Added file for new device: Eurom Sani towel radiator

jan-gerard:
Hi,
I added the device "Eurom Sani-Bathroom-Radiator Towel Wifi heater". I tested it with the radiator. 
One thing that deviates from the manual, is that I created a non-existing dummy DP id for the current temperature, while maintaining the availability of the two sensor values: room temperature and surface temperature. The heater can control both, depending on the mode of operation, whether it controls the room or the heater itself.
Furthermore I wanted to specify an icon for the new operating mode 'anti-frost' but I did not manage to find out how to do that, using the icons.json file. If you have any suggestions? 
Jan-Gerard

### DIFF
--- a/custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
+++ b/custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
@@ -12,10 +12,10 @@ primary_entity:
         - dps_val: true
           constraint: mode
           conditions:
-            - dps_val: r
-              value: heat
-            - dps_val: t
-              value: dry
+            - dps_val: "r"
+              value: "heat"
+            - dps_val: "t"
+              value: "dry"
               icon: "mdi:hanger"
     - id: 2
       type: integer
@@ -26,7 +26,7 @@ primary_entity:
       mapping:
         - constraint: mode
           conditions:
-            - dps_val: t
+            - dps_val: "t"
               value_redirect: towel_dry_temperature
               range:
                 min: 40
@@ -42,9 +42,9 @@ primary_entity:
       mapping:
         - constraint: mode
           conditions:
-            - dps_val: t
+            - dps_val: "t"
               value_mirror: heater_temperature
-            - dps_val: r
+            - dps_val: "r"
               value_mirror: room_temperature
     - id: 3
       type: integer
@@ -54,7 +54,7 @@ primary_entity:
       type: bitfield
       mapping:
         - dps_val: 0
-          value: OK
+          value: "OK"
     - id: 101
       type: string
       name: hvac_action
@@ -62,11 +62,11 @@ primary_entity:
         - dps_val: true
           constraint: mode
           conditions:
-            - dps_val: r
-              value: heating
-            - dps_val: t
-              #value: heating
-              value: drying
+            - dps_val: "r"
+              value: "heating"
+            - dps_val: "t"
+              #value: "heating"
+              value: "drying"
         - dps_val: false
           value: idle
           constraint: hvac_mode
@@ -77,10 +77,10 @@ primary_entity:
       name: mode
       type: string
       mapping:
-        - dps_val: r
-          value: room
-        - dps_val: t
-          value: towel
+        - dps_val: "r"
+          value: "room"
+        - dps_val: "t"
+          value: "towel"
     - id: 105
       name: heater_temperature
       type: integer
@@ -93,27 +93,20 @@ primary_entity:
       range:
         min: 40
         max: 70
-    - id: 102
+    - id: 103
       name: preset_mode
       type: boolean
       mapping:
-        - dps_val: true
-          constraint: anti_frost_mode
-          conditions:
-            - dps_val: false
-              value: eco
-            - dps_val: true
-              hidden: true
-              value: invalid
         - dps_val: false
-          constraint: anti_frost_mode
-          conditions:
-            - dps_val: false
-              value: none
-            - dps_val: true
-              # todo: icon for preset
-              value: Anti-frost
-    - id: 103
-      name: anti_frost_mode
-      type: boolean
-      hidden: true
+          value: "none"
+        - dps_val: true
+          value: "Anti-frost"
+secondary_entities:
+  - entity: switch
+    name: "Eco-mode"
+    icon: "mdi:leaf"
+    category: config
+    dps:
+      - id: 102
+        name: value
+        type: boolean
--- a/custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
+++ b/custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
@@ -1,7 +1,7 @@
-name: Sani Bathroom Radiator
+name: Bathroom radiator
 products:
   - id: bc6gdgt0cpq1jmjz
-    name: LBH-63 TOWEL RAIL EUROM
+    name: Eurom Sani LBH-63
 primary_entity:
   entity: climate
   translation_key: heater
@@ -15,10 +15,10 @@ primary_entity:
         - dps_val: true
           constraint: mode
           conditions:
-            - dps_val: "r"
-              value: "heat"
-            - dps_val: "t"
-              value: "dry"
+            - dps_val: r
+              value: heat
+            - dps_val: t
+              value: dry
               icon: "mdi:hanger"
     - id: 2
       type: integer
@@ -30,7 +30,7 @@ primary_entity:
       mapping:
         - constraint: mode
           conditions:
-            - dps_val: "t"
+            - dps_val: t
               value_redirect: surface_temperature_setpoint
               range:
                 min: 40
@@ -42,7 +42,7 @@ primary_entity:
       mapping:
         - constraint: mode
           conditions:
-            - dps_val: "t"
+            - dps_val: t
               value_mirror: surface_temperature_internal
     - id: 105
       name: surface_temperature_internal
@@ -54,7 +54,7 @@ primary_entity:
       type: bitfield
       mapping:
         - dps_val: 0
-          value: "OK"
+          value: OK
     - id: 101
       type: string
       name: hvac_action
@@ -62,11 +62,10 @@ primary_entity:
         - dps_val: true
           constraint: mode
           conditions:
-            - dps_val: "r"
-              value: "heating"
-            - dps_val: "t"
-              #value: "heating"
-              value: "drying"
+            - dps_val: r
+              value: heating
+            - dps_val: t
+              value: drying
         - dps_val: false
           value: idle
           constraint: hvac_mode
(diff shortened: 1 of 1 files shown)


jan-gerard:
Ai, I notice now that setting the eco preset-mode does not work from Home Assistant. All other presets do work. And retrieving the preset-mode from the device works as well. The difficulty is that triggers are two binary switches that are mutually exclusive. How can I combine two binary entities into one (string) preset-mode, providing both getting and setting? I need nested conditions?

jan-gerard:
I fixed the two errors in the code. Now it should pass the tests.

jan-gerard:
There is one annoying thing: Switching preset modes from Anti-frost to Eco is only possible when switching first to preset None. Because the preset-mode is a combination of two boolean entities, and the condition statement is used, then for one of the states only one of the booleans is relevant, and the other is not switched when switching mode. I don't see a method to solve this. It's not blocking though.

make-all:
You can mark entries in a conditional mapping as `hidden: true` to avoid selecting that one - useful when you want selecting eco to always set boost off, but you still want to recognise it as eco if both boost and eco are on (or vice versa). In that case, marking the "both true" case as hidden will prevent that from being used when setting.

jan-gerard:
> You can mark entries in a conditional mapping as `hidden: true` to avoid selecting that one - useful when you want selecting eco to always set anti-frost off, but you still want to recognise it as eco if both anti-frost and eco are on (or vice versa). In that case, marking the "both true" case as hidden will prevent that from being used when setting.
I tried that, but it doesn't work. Probably, the values are set the right way, but first both mode should be set to false, before the other can go go true, in the heater firmware.

jan-gerard:
I think I can live with the limited preset switching flow. Who wants to switch from Eco to Anti-frost at once. So I think the code is now optimal. And the heating state does not show the drying state anymore: heating means heater is on, also when used for drying.

make-all:
If the device does not support setting two dps together, then I suggest that you simplify the config and do not try to combine the eco and anti-frost together.  Make anti_frost a separate switch, and just leave "eco" and "comfort" as presets.

"off" requires the quotes in hvac_mode and hvac_action, as it is defined in HA as the string "off", not the boolean value False, which is what yaml converts unquoted off to.

hvac_action should report drying if the hvac_mode is dry.  If that breaks the chart display, then that is a bug in the frontend, the integration should not be crippled to workaround frontend bugs.

make-all:
I just realised I have some unsent comments, so at least one of them is probably obsolete now.

custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
```diff
@@ -108,5 +108,9 @@ secondary_entities:
     category: config
     dps:
       - id: 102
-        name: value
         type: boolean
+        mapping:
```
make-all, line 112: This is incorrect, the value should be a boolean, not a string.

The correct fix would be to change the "name: value", which you removed, to "name: switch"

custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
```diff
@@ -0,0 +1,112 @@
...
+      mapping:
+        - constraint: mode
+          conditions:
+            - dps_val: "t"
+              value_redirect: towel_dry_temperature
+              range:
+                min: 40
+                max: 70
+    - id: 333
+      # dummy variable for current temperature of
```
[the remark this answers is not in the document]
jan-gerard: That's the point, I need to create an entity that doesn't exist, since the others (heater_temperature:105 and room_temperature:3) are already used. And the "current_temperature" that is used depends on the operation mode:2. I could use one of the above as the base entity (let's say heater_temperature), but then that one behaves rather strange in the logging. So the current_temperature entity is the actual temperature as used by the internal control loop, corresponding to the mode of operation. 
If there is another option to create an entity without a dps id, I'd like to use that, but for now, this seems to be the only workaround.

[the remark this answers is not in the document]
jan-gerard: Or is it possible to use the same dp id in both the primary and secondary entities? That would allow the actual values to be used (heater surface temperature and room temperature) as individual sensors (nice for making graphs) and use either of the two as the current temperature in the climate/heater device. That would be most elegant and correct I think.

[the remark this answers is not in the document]
make-all: Yes, it is possible to reuse the same dp in another entity, or even in the same entity with a different name.  Making secondary sensor entities for both temperatures would make them easier to work with than as non-standard attributes on the climate entity.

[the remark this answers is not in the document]
jan-gerard: I made a change accordingly.

custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
```diff
@@ -0,0 +1,152 @@
+name: Eurom Sani-Bathroom-Radiator Towel Wifi heater
+primary_entity:
```
make-all, line 2: Ideally a products section should be here with the product id and branding as the name. The top level name should be generic in case other devices match.

custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
```diff
@@ -0,0 +1,152 @@
...
+      unit: C
+      range:
+        min: 40
+        max: 70
+    - id: 103
+      name: preset_mode
+      type: boolean
+      mapping:
+        - dps_val: false
+          value: "none"
```
make-all, line 115: Appropriate standard presets would be `home` and `away`

make-all (commented):
The commented out portions can just be removed if it is working.  This will resolve the lint warnings about yamllint's fussy indentation preferences

custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
jan-gerard: The I remove the Towel WiFi heater. "Eurom Sani Bathroom Radiator" is the common part in the name of the different models covered.

make-all: The branding is the most important part to remove. Making it more general, such as "Bathroom radiator" is not an issue, but since it has the towel_temperature logic in there, it is probably only valid for the Towel warmer models.

make-all: But like I said, finding the product id and keeping the branding info in a products section is optimum.

jan-gerard: The Product ID is some strange number, more like a serial than a recognizable product identifier. Maybe the product name makes more sense. But I don't know which part is specific for this model, since there is are black and white models, with different capacity values.
"model": "LBH-63",
"name": "Your Eurom Heater",
"product_id": "bc6g********jmjz",
"product_name": "LBH-63 TOWEL RAIL EUROM",

jan-gerard: By the way, for some reason, the integration no longer works since 1 day. Tuya cloud explorer still reports the same local key as I originally used. https://eu.iot.tuya.com/cloud/explorer

custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
```diff
@@ -0,0 +1,112 @@
...
+        - dps_val: false
+          constraint: frost_mode
+          conditions:
+            - dps_val: false
+              value: "none"
+            - dps_val: true
+              # todo; icon
+              value: "Anti-frost"
+        - dps_val: true
+          value: "eco"
```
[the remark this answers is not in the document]
jan-gerard: > The true case should also have constraint and conditions to ensure the two values are always set together.

That appears to be a bit of a problem. When one of the modes is on, to switch to the other, the first should first be switched off. So Eco to Anti-frost switching should first switch Eco off, then in a next all switch Anti-frost on.

custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
jan-gerard: > Appropriate standard presets would be `home` and `away`

That appears to be a bit of a problem. When one of the modes is on, to switch to the other, the first should first be switched off. So Eco to Anti-frost switching should first switch Eco off, then in a next all switch Anti-frost on. We can make both the switches boolean secondary entities, but they should still be mutually exclusive, and to function properly, the switching should be a 2-step process.

custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
jan-gerard: > By the way, for some reason, the integration no longer works since 1 day. Tuya cloud explorer still reports the same local key as I originally used. https://eu.iot.tuya.com/cloud/explorer

Power-cycle the heater and it's working again. Strange that the Tuya app kept working, but no longer a HA - Tuya Local connection.

custom_components/tuya_local/devices/eurom_sani_bathroom_towel_radiator.yaml
make-all: If the device does not accept these being switched together, then we don't really have a way to enforce that. Perhaps you can make some buttons in the UI that are tied to automations that do the correct sequence of events. In that case you may want to move both out to external switches which can be hidden from direct use in the UI, to avoid someone accidentally using the climate presets while the eco switch is not in the correct state.

jan-gerard: I had hoped there would be another way, to implement all functions of the heater, elegantly in the standard Climate entity format, while using this integration. Apparently the Tuya app also sets those in succession. I guess there are more Tuya devices out there that suffer from the same behavior.
But I understand that it is not straightforward implementation, while maintaining the elegant condition scheme that implicitly sets multiple states according to the yaml file of the device. Maybe adding some statements that define an order of entities to be set, which then determines whether values/switches can be set at once, or need successive calls to the API? Or that define 'invalid combinations', such that the integration knows that two contradicting states are not valid, then queues the states that are not allowed in combinations with others, and iterates until all combinations are valid and set. So if 'invalid: eco=on & away=on' is set, then switching from 'eco=on&away=off' will first switch to the valid 'eco=off&away=off', notices that it's not ready since away should be on in the end, and makes a second call to 'eco=off&away=on'. No defined sequence needed, only defining invalid combinations.

jan-gerard:
For me the integration works fine now. I can control the things I need, and include them in the Climate card in HA.

PR #1828, opened 2024-04-14
adding support for the Arlec Smart Button

gravy6:
The device supported is: 
https://grid-connect.com.au/download/sg022ha/

It has to be connected via a smart hub.

### DIFF
--- a/custom_components/tuya_local/devices/arlec_smart_button.yaml
+++ b/custom_components/tuya_local/devices/arlec_smart_button.yaml
@@ -1,10 +1,9 @@
 name: Smart Button
 products:
   - id: kpzc6pm8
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/arlec_smart_button.yaml
```diff
@@ -0,0 +1,28 @@
+name: Smart Button
+products:
+  - id: kpzc6pm8
+    name: Arlec Smart Button
+primary_entity:
+  entity: event
+  name: Button
```
make-all, line 7: The entity should not need a name if the desired name matches the class, generally classes define a default translated name.

PR #1837, opened 2024-04-18
Add support for Stepuptech solar inverter SG800

sj29fr:
Hello,
I added support for Stepuptech solar inverter bought from Materfrance.
I removed the CO2 and plant tree indicators.
I also don't know how to handle clear energy function.
Nevertheless, it provides all required statistics.
For info, the tuya protocol version is 3.5.
image


custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,137 @@
...
+    category: diagnostic
+    dps:
+      - id: 101
+        type: integer
+        name: sensor
+        unit: W
+        force: true
+        class: measurement
+        mapping:
+          scale: 100
```
make-all: likewise

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,137 @@
...
+    class: voltage
+    category: diagnostic
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        unit: V
+        force: true
+        mapping:
+          scale: 100
```
make-all: likewise

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,137 @@
...
+    class: temperature
+    dps:
+      - id: 104
+        type: integer
+        name: sensor
+        unit: C
+        force: true
+        class: measurement
+        mapping:
+          scale: 10
```
make-all: likewise

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,137 @@
...
+    class: current
+    category: diagnostic
+    dps:
+      - id: 106
+        type: integer
+        name: sensor
+        unit: A
+        force: true
+        mapping:
+          scale: 100
```
make-all: likewise

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,137 @@
...
+    category: diagnostic
+    dps:
+      - id: 109
+        type: integer
+        name: sensor
+        unit: W
+        force: true
+        class: measurement
+        mapping:
+          scale: 100
```
make-all: likewise

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,137 @@
...
+    category: config
+    dps:
+      - id: 110
+        type: integer
+        name: sensor
+        unit: "%"
+        optional: true
+        class: measurement
+        mapping:
+          scale: 1
```
make-all: This mapping is not required, as a scale of 1 is a no-op

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,137 @@
...
+    class: current
+    category: diagnostic
+    dps:
+      - id: 111
+        type: integer
+        name: sensor
+        unit: A
+        force: true
+        mapping:
+          scale: 100
```
make-all: needs `- `

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,137 @@
...
+    category: diagnostic
+    dps:
+      - id: 113
+        type: integer
+        name: sensor
+        unit: kWh
+        force: true
+        class: total_increasing
+        mapping:
+          scale: 100
```
make-all: likewise

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,137 @@
...
+    class: voltage
+    category: diagnostic
+    dps:
+      - id: 114
+        type: integer
+        name: sensor
+        unit: V
+        force: true
+        mapping:
+          scale: 100
```
make-all: likewise

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
sj29fr: I removed these lines

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,135 @@
...
+  entity: sensor
+  name: Daily energy out
+  class: energy
+  dps:
+    - id: 107
+      type: integer
+      name: sensor
+      unit: kWh
+      class: total_increasing
+      force: true
```
make-all, line 15: Do all of these need to be forced?
Usually this is only necessary for some smartplugs, dedicated meters usually give frequent updates automatically, as this is their primary function,

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,135 @@
...
+        type: integer
+        name: sensor
+        unit: W
+        force: true
+        class: measurement
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: power adjust
+    class: power
```
make-all, line 90: class power should have a unit of W, or similar, not %.  What actually is this measuring? Power factor?

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,135 @@
...
+    dps:
+      - id: 111
+        type: integer
+        name: sensor
+        unit: A
+        force: true
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: total energy
```
make-all, line 112: Untranslated names should be capitalised.  "Total energy", or remove the name to use the default translations for energy.

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
```diff
@@ -0,0 +1,135 @@
+name: Micro Inverter Stepuptech SP800
+products:
+  - id: r73jxzrtcda4uwwb
+    name: SP800
```
make-all, line 4: Move "Stepuptech" into this product specific name, and change the top level name to a generic "Solar inverter" in case other brands also match the config.

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
sj29fr: It is to limit the power output of the inverter. 
This is a setting and I may have not choose the good syntax.
Per default, the value is 100%, but you can reduce the output power.

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
sj29fr: I will correct it

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
sj29fr: OK

custom_components/tuya_local/devices/solar_inverter_stepuptech.yaml
sj29fr: No, it's a mistake

make-all:
If force is needed, it is needed. My comment was just to make sure you have checked first that it is needed (ie those dps are not returned by default, and adding `force: true` is effective to make the device return them). Otherwise it can cause problems of its own.

PR #1865, opened 2024-04-30
Update wdyk_3phase_energymonitor.yaml

rostKo:
Need change scale from 1000 to 100 for right display as in Tuya Smart.


make-all:
Does this need more checking, or was it mistakenly marked as draft?

rostKo:
> Does this need more checking, or was it mistakenly marked as draft?

Sorry, I'm new to github, so I marked it as a draft by mistake.  The change is required for the correct display of Total energy.

PR #1874, opened 2024-05-02
HA 2024.5.0 Issue Fixes

craibo:
- Attempt to resolve coroutine issues
- Update version

Confirmed working


make-all:
This change is switching from using HA managed executor tasks to unmanaged Python threads. There will be more management required, such as cleaning up when the integration is stopped, and catching and reporting any errors that are thrown.

It may be better to use another HA method of starting executor tasks to keep the HA management of these.

Also, you have a couple of changes from October in there that are already rebased and merged upstream so are causing a conflict for this PR. It would be better to start from a branch that is a fresh copy of upstream.

craibo:
Closing and updating with feedback.
Thanks @make-all

PR #1876, opened 2024-05-03
Ha 2024.5.0 issue fixes

craibo:
- Attempt to resolve coroutine issues
- Update manifest

### DIFF
--- a/custom_components/tuya_local/device.py
+++ b/custom_components/tuya_local/device.py
@@ -144,7 +144,9 @@ class TuyaLocalDevice(object):
         self._shutdown_listener = self._hass.bus.async_listen_once(
             EVENT_HOMEASSISTANT_STOP, self.async_stop
         )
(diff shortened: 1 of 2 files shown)


make-all:
Does this not cause performance problems?

It seems you have changed from starting a task to handle receiving, to scheduling receives on the HA event loop.  Since the tinytuya library is not async, this will block the HA event loop for up to 5 seconds every time we check for updates.

The other change to use schedule_update_ha_state instead of async_write_ha_state should not be necessary if the receive loop is running on the HA event loop, but without a change to async_refresh, will result in extra polling of the device, which may be what is making the difference.

make-all:
I am not sure why the test failure is occuring, maybe coroutines need to be top level functions outside of a class.

It will also be necessary to save the result of the call that creates the task so it can be awaited when the device is stopped (after ensuring that the receive loop will be terminated).

Issue #1708, opened 2024-03-04
Roll out icons and translations to more devices


make-all:
Now that #1579 is implemented, continue to add icons and translations to commonly used entities in future releases.

This issue is so the main issue of adding the support can be closed once it is released, but still keep track of future work needing to be done on the device config level.

PR #1878, opened 2024-05-04
Add support for Plikc Neve X W thermostat

beltra:
Add support for Plikc Neve X W thermostat.
Quite similar to Plikc Neve X WRF, but some issues of that integration have been fixed. Supports setting automatic, manual and holiday mode. Heating is correctly displayed when active.

### DIFF
--- a/custom_components/tuya_local/devices/plikc_neve_x_thermostat.yaml
+++ b/custom_components/tuya_local/devices/plikc_neve_x_thermostat.yaml
@@ -27,9 +27,9 @@ primary_entity:
           hidden: true
         - dps_val: holiday
           value: away
-          constraints: Durata vacanza
+          constraint: Durata vacanza
           conditions:
-          - dps_val: 1
+            - dps_val: 1
     - id: 3
       type: string
       name: hvac_action
--- a/custom_components/tuya_local/devices/plikc_neve_x_thermostat.yaml
+++ b/custom_components/tuya_local/devices/plikc_neve_x_thermostat.yaml
@@ -1,4 +1,4 @@
-name: Termostato
+name: Thermostat
 products:
   - id: 5kwgvcaqzzkzrftu
     name: Plikc Neve X W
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/plikc_neve_x_thermostat.yaml
```diff
@@ -0,0 +1,147 @@
...
+        - dps_val: manual
+          value: boost
+        - dps_val: temporary
+          value: boost
+          hidden: true
+        - dps_val: holiday
+          value: away
+          constraint: Durata vacanza
+          conditions:
+            - dps_val: 1
```
make-all, line 32: This seems incomplete. It also references a non-existing dp

custom_components/tuya_local/devices/plikc_neve_x_thermostat.yaml
```diff
@@ -0,0 +1,147 @@
...
+        range:
+          min: -30
+          max: 30
+        mapping:
+          - scale: 10
+            step: 1
+  - entity: number
+    name: Durata vacanza
+    mode: box
+    force: true
```
make-all, line 93: `force` should be at dps level.  But is it really effective?

custom_components/tuya_local/devices/plikc_neve_x_thermostat.yaml
```diff
@@ -0,0 +1,147 @@
...
+  - entity: number
+    name: Calibrazione
+    mode: box
+    category: config
+    icon: "mdi:thermometer-check"
+    dps:
+      - id: 27
+        type: integer
+        name: value
+        unit: °C
```
make-all, line 83: It is better to use ° for calibration adjustments, because you don't want HA doing unit conversion on these.

make-all (commented):
In general device and entity names should be in English, as there is a high chance of any config also matching devices sold in other markets, and for better or worse, English is the most widely spoken.

You are welcome though to contribute to #1708 by expanding translation_keys to cover more common entity types.

custom_components/tuya_local/devices/plikc_neve_x_thermostat.yaml
beltra: What I wanted to do was set the dp 33 (vacation duration) to 1, as the thermostat doesn't go in vacation mode if a value is not pushed to the vacation duration

beltra:
Thanks for the changes, sorry I didn't have time to fix the problems. I'd suggest changing the dps_val for id 1 to heat as the device only supports heating.

make-all:
The hvac_action values include both heating and cooling. Is that a mistake?

beltra:
The thermostat can operate in only one mode at a time and a reset is required to change mode (i.e. it has only one physical output to turn on/off only one device). I don't know if an option can be added in the integration to set the behaviour.

PR #1880, opened 2024-05-04
Add sh07s_ty smart sprinkler variant

AdrianGarside:


make-all:
I suggest to use local branches to separate your new device submission from the cloud setup changes.

AdrianGarside:
Yes, I created this PR first and then didn't want to lose the better device config while I was working on the other change. Once this PR goes in I'll be able to merge things back and it'll disappear as a delta in the other.

PR #1881, opened 2024-05-05
Add Tuya cloud assisted setup flow

AdrianGarside:
Pick up the streamlined Tuya cloud login from the official HA cloud Tuya integration. This allows you to get a code from the Smart Life app and then use the app to login to Tuya with a QR code. All available devices for that account can then be select to add locally. Only the IP address still needs to be discovered so a local scan is performed to get that. Just like the final device connection the scan for the IP can fail if other things (including the Smart Life app just used for login!) are connecting to the device when the scan runs so those should be closed before that and the final connection step.

New flow:

Starting page in the config flow allows users to choose to go down the cloud assisted configuration path or the original manual entry path:
image

Now you need to provide the 'User Code' available from the Smart Life app:
image

Then you will get presented with a QR code you scan in the Smart Life app:
image

Once you scan and approve the login in the app, a list of available devices that the cloud account knows about is shown. Now you must select the device you want to add and, if necessary, the gateway it connects through:
image

Next the local IP address of the device is found by discovery:
image

Finally the retrieved data is prepopulated into the existing register local device page:
image


AdrianGarside:
@make-all this is not yet ready but wanted to get your early thoughts on this new streamlined setup option that utilizes the Tuya cloud to get all the device data.

make-all:
This presents the local / cloud selection for each device that is added.  It might be better to detect that the user has already selected one of these, and skip that selection, using the cloud login details if available, otherwise jumping straight to local selection.

Ideally the way this should work is: local discovery to detect devices -> if cloud available, get local key from cloud, otherwise prompt. If local discovery is not available (due to complex network config, or conflict with localtuya), then the full manual setup needs to also be available, though device ids may also be available to be pre-populated from the cloud.

AdrianGarside:
Yes, I wanted to keep this first version simple and less disruptive. The cloud credentials can be saved into the domain data and then re-used on subsequent device additions (vs the current repetitive login required). In my experience doing that correctly and safely handling the possible states is a little hard to test all the edge cases.

If the flow was as-is (i.e. always giving the user the choice up front) but the QR login steps were skipped if credentials had already been saved from a previous configure, would that work for you?

AdrianGarside:
@make-all I haven't found a good way to save the authentication data persistently into home assistant. I can save data to self.hass.data[DOMAIN][DATA_STORE] but that doesn't survive a reload of the integation / home assistant restart. It would improve over what I have as adding multiple devices _in one go_ would only require a single login via the smart life app so the very first setup if you had multiple devices to add would be better. But the scenario of getting more Tuya devices over time wouldn't really change. Any ideas?

AdrianGarside:
I'm going to combine the device/hub selection into a single step.

AdrianGarside:
@make-all I'm not familiar with the automated python testing. It's complaining about the new tuya_sharing dependency:
custom_components/tuya_local/config_flow.py:22: in <module>
    from tuya_sharing import (
E   ModuleNotFoundError: No module named 'tuya_sharing'

Do you have any pointers on how I can fix this / run those tests locally?

make-all:
The tuya-sharing library is not a standard part of Python, so needs to be declared as a dependency in requirements.txt and requirements-dev.txt

AdrianGarside:
Need to update the README to describe the new optional config flow.

AdrianGarside:
TODO: Update README

AdrianGarside:
@make-all any concerns with this PR?

AdrianGarside:
> No, only finding the time to manually rebase it is blocking the merge now, as it cannot automatically be rebased.

You'll want to squash merge this anyway. I like to create lots of small commits - no point polluting the history permanently with them.

YuriiMaiboroda:
@AdrianGarside, hi.

I found that logging in via a QR code is very useful, since it works even after the trial subscription for the Tuya IoT Platform has expired.

I want to add this type of login to [tuya ble](https://github.com/PlusPlus-ua/ha_tuya_ble/).
Unfortunately, I couldn't find a fork of the repository where this has already been implemented or any other integration that allows connecting Bluetooth devices without a gateway. My server is running on a Raspberry Pi, and I was hoping that its built-in Bluetooth would be sufficient.

Could you please point me to the documentation for this mechanism? I'm particularly interested in where to get the values for TUYA_CLIENT_ID and TUYA_SCHEMA.

PR #1890, opened 2024-05-06
Add bluetooth dual water timer

AdrianGarside:
Add dual water timer bought on Amazon (and visually there's a number of brands selling the exact same one).
https://www.amazon.com/gp/product/B0D13W939Y/ref=ppx_yo_dt_b_asin_title_o03_s00?ie=UTF8&th=1

image

### DIFF
--- a/custom_components/tuya_local/devices/HCT-626_dual_water_timer.yaml
+++ b/custom_components/tuya_local/devices/HCT-626_dual_water_timer.yaml
@@ -72,7 +72,6 @@ secondary_entities:
   - entity: switch
     name: Smart weather delay
     category: config
-    class: duration
     dps:
       - id: 43
         type: boolean
@@ -88,19 +87,19 @@ secondary_entities:
         optional: true
         mapping:
           - dps_val: sunny
-            value: sunny
+            value: Sunny
             icon: "mdi:weather-sunny"
           - dps_val: cloudy
-            value: cloudy
+            value: Cloudy
             icon: "mdi:weather-cloudy"
           - dps_val: rainy
-            value: rainy
+            value: Rainy
             icon: "mdi:weather-rainy"
           - dps_val: snowy
-            value: snowy
+            value: Snowy
             icon: "mdi:weather-snowy"
           - dps_val: null
-            value: unavailable
+            value: Unavailable
             icon: "mdi:weather-sunny-off"
   - entity: select
     name: Weather delay
@@ -143,22 +142,30 @@ secondary_entities:
         type: integer
         name: value
         optional: true
-  - entity: number
-    name: DPS 101
+  - entity: sensor
+    name: Valve 1 last runtime
+    class: duration
     category: diagnostic
     dps:
       - id: 101
         type: integer
-        name: value
-        optional: true
-  - entity: number
-    name: DPS 102
+        name: sensor
+        mapping:
+          - scale: 60
+        unit: min
+        readonly: true
+  - entity: sensor
+    name: Valve 2 last runtime
+    class: duration
     category: diagnostic
     dps:
       - id: 102
         type: integer
-        name: value
-        optional: true
+        name: sensor
+        mapping:
+          - scale: 60
+        unit: min
+        readonly: true
   - entity: number
     name: DPS 103
     category: diagnostic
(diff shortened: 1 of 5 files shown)


AdrianGarside:
42 might be the countdown for the weather delay. When I set that to 1 day it changes from 0 to 24.

custom_components/tuya_local/devices/HCT-626_dual_water_timer.yaml
```diff
@@ -0,0 +1,197 @@
...
+            step: 60
+        unit: min
+  - entity: number
+    name: Weather delay remaining
+    category: diagnostic
+    dps:
+      - id: 42
+        type: integer
+        name: value
+        unit: hours
```
AdrianGarside, line 61: Should make this read only. You can set it as-is but it doesn't seem to be respected.

make-all: If read only, it should be a sensor with class duration. Unit should be h

custom_components/tuya_local/devices/HCT-626_dual_water_timer.yaml
```diff
@@ -0,0 +1,216 @@
...
+        name: sensor
+        readonly: true
+  - entity: sensor
+    name: Spray Schedule
+    category: diagnostic
+    dps:
+      - id: 108
+        type: string
+        name: sensor
+        hidden: true
```
AdrianGarside, line 215: This didn't work how I expected it to.

AdrianGarside: Also these don't reliably get set - other activity causes them to get sent, so maybe I need force: true?

AdrianGarside:
The only one I haven't worked out yet is DPS 47
DPS 53 is battery level somehow - setting it to 1 caused me to get a notification of very low battery where higher values did not. Maybe it's really a boolean - 1 = low battery level?
DPS 47 might be the battery meter level - one of my devices is showing 2 and the battery meter is showing 2 bars. The other is set to 5 and the battery meter is showing 5 bars. Seems likely. I've only had the first device for a few weeks so if it's really already low on batter that will make sad...

AdrianGarside:
Should I make the last runtime / current spray time just attributes on the valve states sensors? They probably won't be that useful but attributes are much more of a pain to get at / use in automations.

make-all:
The schedule values do not appear very useful, and should be moved to attributes, but current and last runtime appear to be more useful.

custom_components/tuya_local/devices/HCT-626_dual_water_timer.yaml
```diff
@@ -0,0 +1,229 @@
...
+    dps:
+      - id: 47
+        type: string
+        name: sensor
+        optional: true
+        range:
+          min: 0
+          max: 5
+  - entity: binary_sensor
+    name: Battery low
```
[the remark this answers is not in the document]
AdrianGarside: IIRC I had class battery here and that caused errors in the log output so I had to remove it. I'll double check.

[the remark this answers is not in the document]
AdrianGarside: Worked this time. Maybe having the name caused issues before.

AdrianGarside:
Is there a better way to present the durations? In the screenshot in the PR, valve 1 ran for 2s. After scaling to minutes it gets rescaled by home assistant back into seconds for display but with an introduced error. The spray time is scaled the same way but since it was full minutes there was no rounding but it gets presented in an inconsistent way. If I set a precision manually on either they become (for e.g) 0.03 min. Setting precision to 0 in the yaml gives:
image

make-all:
You shouldn't need to scale sensors with a class of duration. Just set the unit to `s`, and let HA take care of formating

AdrianGarside:
> You shouldn't need to scale sensors with a class of duration. Just set the unit to `s`, and let HA take care of formating

Then everything becomes seconds, which isn't great when the runtime will typically be more consumable in minutes:
image

My personal ideal would be consistency so get the MM:SS display for both of these:
image

AdrianGarside:
@make-all I think this is ready to go.

custom_components/tuya_local/devices/HCT-626_dual_water_timer.yaml
```diff
@@ -0,0 +1,220 @@
...
+        mapping:
+          - scale: 0.05
+        unit: '%'
+        precision: 0
+  - entity: binary_sensor
+    category: diagnostic
+    class: battery
+    dps:
+      - id: 53
+        type: integer
```
make-all, line 134: Binary sensor should have boolean values.  Is there a mapping needed? Is 53 a bitfield with only one bit indicating low battery, or is low battery the only indication in here (the mapping would be different for these two cases - either specifying the bit value for true, and defaulting to false, or specifying  0=false and defaulting to true).

AdrianGarside: @make-all The DPS is an integer and I've only ever seen it be 0. But when I had it writable setting it to any value other than 0 triggered the low battery warning in the smart life app. So a mapping to a bool is probably correct.

AdrianGarside:
@make-all removing the mode: box lines changes the entity config to:
image
Which is much less easy to use to set an accurate time than the way it presents with that setting (in PR description).

make-all:
This can be changed in the UI if the user prefers a different representation of the number entities.  Unless there is a good reason for it, the default should follow standard HA policy

PR #1914, opened 2024-05-13
Add support for PCIe computer power switch

BenJamesAndo:
Support for Wifi Computer Reset Switch PCIe Card
https://www.ebay.com.au/itm/225773542438

There is a pro version and the mini version. This is the mini version. This supports power and restart. Most interesting is the force restart option. It turns off the computer, waits 14 seconds then turns it on again. During the 14 seconds of force restart you can't turn on the PC via the app or via the physical button on the computer. The standard restart simply restarts it.

```
{
  "result": {
    "active_time": 1715411952,
    "bind_space_id": "70408609",
    "category": "cz",
    "create_time": 1715411952,
    "custom_name": "",
    "icon": "smart/icon/ay1552295613286EPqcq/88cf003bbc9f79fad1953ad1671daefc.png",
    "id": "bf50fd7b929b23099cxsg6",
    "ip": "",
    "is_online": false,
    "lat": "",
    "local_key": "",
    "lon": "",
    "model": "JH-PcMini",
    "name": "电脑",
    "product_id": "na90jncjlvw5t0ba",
    "product_name": "电脑",
    "sub": false,
    "time_zone": "+10:00",
    "update_time": 1715567812,
    "uuid": "94e9fc025f960573"
  },
  "success": true,
  "t": 1715568705935,
  "tid": "bc87d45d10d311ef9a8e62f092a4528b"
}
```
image

image


custom_components/tuya_local/devices/jh_pcmini.yaml
```diff
@@ -0,0 +1,26 @@
...
+    name: Mini card pcie pc power switch
+primary_entity:
+  entity: switch
+  class: outlet
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+secondary_entities:
+  - entity: select
```
make-all, line 13: I wonder if this might be better as 2 buttons?

BenJamesAndo: Yes I reckon it would.
Would it be something like this?
```
secondary_entities:
  - entity: button
    name: Reset
    dps:
      - id: 101
        name: button
        type: string
        mapping:
          - dps_val: "Reset"          
  - entity: button
    name: Force reset
    dps:
      - id: 101
        name: button
        type: string
        mapping:
          - dps_val: "forceReset"
```

However with this code the reset and force reset buttons both trigger the "Reset" dps_val. I can't figure out why forceReset isn't getting triggered from the Force reset button. With the select entity dropdown it worked properly.
I've tried 
```
  - entity: button
    name: Force reset
    dps:
      - id: 101
        name: button
        type: string
        mapping:
          - dps_val: "forceReset"
            value: "forceReset"
```
but same outcome.

make-all: `value: true` is what buttons expect

BenJamesAndo: That worked, thanks. Updated now.

PR #1952, opened 2024-05-26
ips-pro: fix and improve model

chris-garbas:
- fix dps value for the boost mode, it's "booster" here.
- handle temperature dynamically
- add cool and auto mode


custom_components/tuya_local/devices/ips_pro_heatpump.yaml
```diff
@@ -1,19 +1,54 @@
-name: IPS heatpump
+name: IPS Inverter Heatpump
 primary_entity:
   entity: climate
   dps:
     - id: 1
-      name: hvac_mode
+      name: power
```
make-all, line 6: This change will prevent the turn_on and turn_off services from working.
It is better to keep this as hvac_mode, and make the "on" case conditional on the dp 105 (under another name)

custom_components/tuya_local/devices/ips_pro_heatpump.yaml
```diff
@@ -1,19 +1,54 @@
-name: IPS heatpump
+name: IPS Inverter Heatpump
```
make-all, line 1: This should be moved under a "products" section linked to the product id of the device , and an unbranded name used at top level instead.

custom_components/tuya_local/devices/ips_pro_heatpump.yaml
```diff
@@ -49,26 +84,14 @@ primary_entity:
...
-              range:
-                min: 18
-                max: 40
-      range:
-        min: 18
-        max: 45
     - id: 107
       type: integer
       name: min_temperature
+      readonly: true
```
make-all, line 90: This is not needed here, as min_temperatures are not writable in any case.

custom_components/tuya_local/devices/ips_pro_heatpump.yaml
```diff
@@ -49,26 +84,14 @@ primary_entity:
...
     - id: 108
       type: integer
       name: max_temperature
+      readonly: true
```
make-all, line 94: Ditto for max_temperature

custom_components/tuya_local/devices/ips_pro_heatpump.yaml
```diff
@@ -97,6 +120,7 @@ secondary_entities:
         type: integer
         unit: "%"
         class: measurement
+        readonly: true
```
make-all, line 123: Also not needed, as sensors are inherently read-only

custom_components/tuya_local/devices/ips_pro_heatpump.yaml
chris-garbas: it is working this way, havac_mode does the value_redirect. 
Where would you place the modes then?

make-all: Either way will work for the set_hvac_mode service, and turn_off will work as long as "off" is one of the modes. But turn_on will not keep the previous mode if the hvac_mode dp is not boolean.
So you need to turn that on its head:
```
  - id: 1
    type: boolean
    name: hvac_mode
    mapping:
        - dps_val: false
          value: "off"
        - dps_val: true
          constraint: work_mode
          conditions:
            - dps_val: smart
              value: heat_cool
            - dps_val: cool
              value: cool
            - dps_val: warm
              value: heat
```

chris-garbas: it's not working this way :(

chris-garbas: i need to check it later again, please do not merge as it is not working yet

chris-garbas: Thank you for your review. I have applied your suggestion, it works with my heatpump 👍

custom_components/tuya_local/devices/ips_pro_heatpump.yaml
```diff
@@ -46,13 +46,27 @@ primary_entity:
       name: hvac_action
       type: string
       mapping:
+        - dps_val: smart
```
make-all, line 49: It is incorrect to map this to hvac_action, as it is the hvac_mode, and has no indication of when the device is idle (ie set to warm/cold/smart, but not actually heating/cooling at this time).

I guess the original author of this config did not seem to be able to get hvac_mode working using this, so instead mapped it to the read-only hvac_action instead, but it should be one or the other.

chris-garbas: would it be possible to map it to the power level? 
is there a way to check the dps_val for >0 ?

but i would need to map the hvac_mode and the power level to be able to determine if it's heating or cooling.

custom_components/tuya_local/devices/ips_pro_heatpump.yaml
```diff
@@ -35,34 +42,9 @@ primary_entity:
...
-        - dps_val: warm
-          constraint: hvac_mode
-          conditions:
-            - dps_val: false
-              value: "off"
-            - dps_val: true
-              value: idle
     - id: 106
       name: temperature
       type: integer
```
make-all, line 47: temperature requires a range.  I'm not sure why you removed this.  If it is because the temperature remains in C even when the setting is for F, then maybe it needs to be solved by a redirect to another dp (temp_set_f), or by hardcoding the unit to C and ignoring temperature_unit.

chris-garbas: I have removed it because the range is handled by min and max temperature. The values there change dependent on the mode (cool/auto/heat)

Let me know if you don't want me to contribute. I can use my fork as well ;)

make-all: Yes, you're right. I actually  noticed that the first time I reviewed, but forgot in the interim

make-all:
Merged

chris-garbas:
> Merged

even with hvac_action :heart_eyes:, thank you!

PR #1954, opened 2024-05-26
Added Amico 52 inch Smart Ceiling Fan

hwhesselink:
Hi,

This is for https://amicolight.com/collections/ceiling-fan-light/products/amico-6-speeds-52-inch-3cct-5-blade-ceiling-fan-with-light-and-remote-alexa-google-home.

I'd hoped to be able to use "step" in the mapping to limit the travel of the HA slider to 3 positions but I've not been able to get that to work.  

Cheers,

Henk Hesselink

### DIFF
--- a/custom_components/tuya_local/devices/amico_smart_ceiling_fan.yaml
+++ b/custom_components/tuya_local/devices/amico_smart_ceiling_fan.yaml
@@ -1,0 +1,80 @@
+name: Ceiling fan
+products:
+  - id: eb208a14315d7cb2bd6txt
+    name: Amico 52 inch ceiling fan
+primary_entity:
+  entity: fan
+  translation_only_key: fan_with_presets
+  dps:
+    - id: 60
+      type: boolean
+      name: switch
+    - id: 61
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: fresh
+          value: fresh
+        - dps_val: nature
+          value: nature
+    - id: 62
+      type: integer
+      name: speed
+      range:
+        min: 1
+        max: 6
+    - id: 63
+      type: string
+      name: direction
+      mapping:
+        - dps_val: forward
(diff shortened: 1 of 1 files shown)


hwhesselink:
To clarify the "step" remark: I meant the color_temp mapping.  With e.g. the (1-6) speed setting the slider will only settle on a multiple of 16% - if you select any other value the slider will go to the nearest 16% multiple.  With the (1-3) color_temp setting you can almost always set a value that is not a multiple of 33%.  The light itself only has 3 temps so will go to the nearest multiple, but the slider doesn't match.

Replacing the 3 dps_val/values pairs with the line "step: 1500" results in an error message about the target value not being between "3000.0 and 6000.0".  so maybe there's a int/float mismatch somewhere?  I've tried various ways of using step but none worked so I settled on what it is now.

hwhesselink:
Yes, that's what I originally had.  The problem is that the slider doesn't snap (unlike e.g. with speed, which does).

With speed as I drag the slider the value also "snaps".  I.e. the value stays the same as I drag until I reach a new range and then the value changes to that (in this case in 16% increments).  When I release the slider it snaps to that value.  This is, obviously, the expected behavior.

With color_temp I can drag the slider and the value changes continuously.  When I release the slider it stays wherever I release it.  If I drag the slider past the half-way point between 2 "snap points" (so 3750K between 3000 and 4500) then the slider briefly jumps to the next snap point and then starts following the cursor again.  The slider only occasionally snaps when I release it.

So the value of the underlying object changes to a supported value with the brief jump but the slider can be put almost anywhere.

My config uses target_range to set the top and bottom values of the slider.  Just using that and range would have the 3 temps at 3K, 4.5K and 6K so I use dps/value pairs to make that 3K, 4K and 6K per the fan documentation.

make-all (approved):

PR #1968, opened 2024-06-01
Added support for luko dehumidifier

cemilbrowne:
Added support for a Luko Dehumidifier.

### DIFF
--- a/custom_components/tuya_local/devices/luko_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/luko_dehumidifier.yaml
@@ -34,7 +34,6 @@ primary_entity:
         - step: 5
 secondary_entities:
   - entity: fan
-    name: Fan
     dps:
       - id: 1
         name: switch
@@ -60,7 +59,7 @@ secondary_entities:
         name: sensor
         type: integer
         unit: "C"
-        class: measurement
+        class: temperature
         readonly: true
         mapping:
           - scale: 10
--- a/custom_components/tuya_local/devices/luko_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/luko_dehumidifier.yaml
@@ -1,6 +1,7 @@
-name: Luko Dehumidifier
+name: Dehumidifier
 products:
   - id: 4w2fzgr62c1oyjsx
+    name: Luko dehumidifier
 primary_entity:
   entity: humidifier
   class: dehumidifier
@@ -11,12 +12,6 @@ primary_entity:
     - id: 8
       name: current_humidity
       type: integer
-      unit: "%"
-      class: measurement
-      readonly: true
-      range:
-        min: 15
-        max: 100
     - id: 2
       name: mode
       type: string
@@ -26,7 +21,6 @@ primary_entity:
     - id: 46
       name: humidity
       type: integer
-      unit: "%"
       range:
         min: 40
         max: 80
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -0,0 +1,66 @@
...
+      class: measurement
+      readonly: true
+      range:
+        min: 15
+        max: 100
+    - id: 2
+      name: mode
+      type: string
+      mapping:
+        - dps_val: "dehumidity"
```
make-all, line 24: If there is only one option, it does not seem worth adding it as mode, since the user cannot change it anyway.

custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -0,0 +1,66 @@
...
+          - dps_val: "2speed"
+            value: 40
+          - dps_val: "3speed"
+            value: 60
+          - dps_val: "4speed"
+            value: 80
+          - dps_val: "5speed"
+            value: 100
+  - entity: sensor
+    name: Temperature
```
make-all, line 57: change this to `class: temperature`

custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -0,0 +1,66 @@
...
+      type: integer
+      unit: "%"
+      range:
+        min: 40
+        max: 80
+      mapping:
+        - step: 5
+secondary_entities:
+  - entity: fan
+    name: Fan
```
make-all, line 37: Does not need a name if it is the same as the default naming for the entity.

custom_components/tuya_local/devices/luko_dehumidifier.yaml
cemilbrowne: I think it has to be there, otherwise when exposed to Homekit, it doesn't seem to work as well.  Equally happy to remove if I'm just doing it wrong.

custom_components/tuya_local/devices/luko_dehumidifier.yaml
cemilbrowne: updated; suggest README.md could be updated to reflect that class: temperature is an option (as well as other options).  I interpreted it as only _type_ of entity (measurement/increasing).

custom_components/tuya_local/devices/luko_dehumidifier.yaml
cemilbrowne: ack; I didn't see an immediate way to understand that (especially with capitalisation).

cemilbrowne:
I think it's ready; one change I haven't made.

custom_components/tuya_local/devices/luko_dehumidifier.yaml
make-all: class at the entity level is different than class at the dp level.  At the entity level it specifies device class (valid classes depend on the entity type, and can be found in HA documentation).  At dp level it specifies state class, which can be measurement, total, total_increasing.

custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -0,0 +1,65 @@
+name: Luko Dehumidifier
+products:
+  - id: 4w2fzgr62c1oyjsx
```
make-all, line 3: Product should have a name, containing branding (including model name or number if available)

Top level name above should be generic, without branding, in case of other devices matching the same config.

PR #1755, opened 2024-03-17
New device: Moes IR Thermostat (AC Controller)

fabianoarruda:
This is my attempt to add compatibility for Moes IR Thermostat - It is a IR controller specific to Air conditioners with built in temperature and humidity sensors. There is also a variant wihtout the LCD screen, which I believe should work the same, but I only have the LCD Screen variant, so not able to confirm.

Details about the device: https://moeshouse.com/products/wifi-ir-thermostat-with-ac-remote-and-sensor?variant=40016596303953

Now, I was able to add the device successfully and it read all attributes correctly, and it is able to set the AC functions - It will update the Tuya device status correctly, example if you change target temperature to 20 C, it will get reflected on the device screen and even in the app... The only problem is, changing the properties via DPS locally will not trigger the IR commands, So it will not set the actual AC temperature.  

I have figured out how to send the IR commands via tinytuya using the Cloud integration, but I don't know yet how to bind these commands to the climate device, so when you attempt to change the device on HA, it should trigger the IR command. 

Here is an example of code that triggers the IR command correcty so The AC will get the command, and the Tuya device screen will also reflect the change (and it will propagate to HA):

``` python
import tinytuya
import json

tinytuya.set_debug()

c = tinytuya.Cloud()

device_id = '...' # IR Thermostat

# Get a listing of all programmed "remotes". This is also the id of the subdevice you may see listed
# on Tuya Cloud as an "Air conditioner", after you configure your AC model in the app.
# In this result you will also find the remote_index attribute which I believe is 
# the IR codeset for a specific brand.

# Example response:
# "result": [
#         {
#             "area_id": 0,
#             "brand_id": 182,
#             "brand_name": "Midea",
#             "category_id": 5,
#             "operator_id": 0,
#             "remote_id": "eb0dc7........", # this is the same as subdevice id on Tuya Cloud
#             "remote_index": 11272,
#             "remote_name": "Ar condicionado"
#         }

# res = c.cloudrequest( '/v2.0/infrareds/' + device_id + '/remotes' )

remote_id = '...'

# list of supported standard AC codes:
# https://developer.tuya.com/en/docs/cloud/infrared-air-conditioner-apis?id=Kb3oe9ehg02fn

# Send command using AC standard code. This is the method that works with my specific AC (Midea)
post_data = {
  "category_id": 5,
  "remote_index": 11272,
  "code": "temp",
  "value": 20
}
[...]

PR #1970, opened 2024-06-02
New device: Moes IR Thermostat (AC Controller)

fabianoarruda:
This adds compatibility for Moes IR Thermostat - It is a IR controller specific to Air conditioners with built in temperature and humidity sensors. There is also a variant wihtout the LCD screen, which I believe should work the same, but I only have the LCD Screen variant, so not able to confirm.

Details about the device: https://moeshouse.com/products/wifi-ir-thermostat-with-ac-remote-and-sensor?variant=40016596303953


fabianoarruda:
@make-all from your [comment here](https://github.com/make-all/tuya-local/pull/1755#issuecomment-2002690176)

> I think this is a compatible/same device as ir_moes_heatpump.yaml. Some of the previously unknown dps you have gotten the documentation for, as well as others that did not show up. It seems strange that it does not send the IR commands automatically, but your docs also have the ir_send and ir_learn dps, so we can add a remote entity to the existing config. The previous report noted that the climate controller dps were only working with the sub device after setting up the remote model, but you appear to have listed the product id of the main IR device, this may affect the IR sending, as it is not tied to a model if you don't use the sub device. Probably some dps need to be set as optional if your device is not being detected as ir_moes_heatpump already.

After testing your updated `ir_moes_heatpump.yaml` with my device, I think they are not the same device. The ir_moes_heatpump you mentioned is not 100% compatible with my device. First because this device is not for heatpumps, I know Moes have other devices for which does different things. My device specifically controls Air conditioner units only. I created a new file named `moes_ir_thermostat` because it's more generic, following the pattern `brand_device_name` as other devices in the repo. Also, please note that model "S16Pro" is a name for my type of device only (other models may have different model names). That's why I also changed to "Moes IR Thermostat" which is more generic. I based on your configuration but had to make some modifications to make it work 100% with my device. So, I created this new PR. Please take a look.

make-all:
I don't see any incompatibilities between this config and the ir_moes_heatpump config that was previously contributed.  Can you explain what exactly you needed to change to make it work?

fabianoarruda:
@make-all Here is a [diff](https://github.com/fabianoarruda/tuya-local/compare/main...fabianoarruda:tuya-local:diff-test?expand=1#diff-843674e1c90d88074f54deecf71ffbe9f9ad15c11be258ab34a6f193fee0ff52) comparing the files, with the changes I needed to make it work 100%.

I considered doing a PR on the same file `ir_moes_heatpump.yaml` but I was not sure if that was the best approach, my concern was if it could break something for someone who is using that configuration. So I instead created a different file `moes_ir_thermostat.yaml`.

moes ir thermostat

make-all:
None of the changes in the diff appear to be necessary.  Some are incorrect, such as changing a binary_sensor to output string values.

fabianoarruda:
> None of the changes in the diff appear to be necessary. Some are incorrect, such as changing a binary_sensor to output string values.

@make-all I'm not sure which sensor you are referring to. Can you please confirm? I can make additional adjustments as necessary, no problem.

My main point on this PR is that the current `ir_moes_heatpump.yaml` doesn't work 100% with my device, and these changes I proposed were necessary to make it work correctly and display all the correct properties. The most important changes are:

- the `hvac_mode` mapping is incorrect. There is no `heatcool` mode. This gives an error in HA when this mode is selected. Accepted values are `auto` or `heat_cool` (with underscore). I used `auto` to be consistent with the DPS values list.

hvac_error
 
- Humidity and temperature sensors added as secondary entities so they can actually appear as sensor entities:

moes ir thermostat sensors

- minor tweaks for displaying attributes in HA: 
  - Humidity should be displayed as a percentage (%), and added min and max values.
  - Filter life is used to track when AC filter should be cleaned. Its necessary to add the max value which is default to 720, so you can create for example a gauge card which will let you now when the reading has reached "critical" value. (I also just noticed that it needs the scale to be set as 1, because you dont need to display minutes and seconds, only hours)

I'm just trying to help, I spent some time already to make this device work the best on HA, the changes I'm proposing is what I'm using currently. So I thought they can be useful to other people and will save them from having the same issues I had. Sorry if I didn't give enough details before, now I see I should have better explained what are the propsed changes and why I'm proposing. 

So please take a look again and if you still find problems I can make additional changes to make the PR acceptable.

make-all:
Of these, only the heatcool needed fixing, but should have been changed to heat_cool to match the HA meaning of the modes.

Temperature and humidity do not need duplicating, they can be accessed directly from the climate entity.  If there is some custom card that requires sensor entities, you can define a template_sensor to extract it.

The current_humidity attribute does not allow customising the units, it is always %, so that change is unnecessary and will be ignored.  Min and max will also be ignored, as that attribute is read-only, and whatever value is returned by the device will be used.

Filter life is a sensor, and therefore also read-only, so min and max will be ignored. Sensor does not have any API for the gauge card to read any value you set here, so you will anyway need to set these in the card configuration. Setting the scale to 1 will not affect anything, as that is already the default.

The other change you made was to remove the `optional: true` tag from dp 201, which will only cause detection failures.

fabianoarruda:
>Temperature and humidity do not need duplicating, they can be accessed directly from the climate entity. 

@make-all Sorry but I respectfully disagree. In some cases, you do need the sensors as secondary units. For example, you need the sensor to make it visible on Alexa because it's not exposed under the climate Entity in Alexa. 

>If there is some custom card that requires sensor entities, you can define a template_sensor to extract it.

Yes, sure you can do that, but this is an additional step needed, why not offer this already as a sensor ready to be used? I don't see any downsides on doing that, you are only making things easier for beginners who don't know yet about concepts like custom templates.

Also, I've seen other types of climate entities that do expose Temperature and humidity as secondary sensors, SmartIR for example does that. Even here tuya-local project there are other devices in the same category that expose temperature and Humidity sensors as secondary entities (I used other devices as examples when building this one): 

https://github.com/make-all/tuya-local/blob/6d46cfc7ac5ec16f3a6efd7fb58fa448e2f87f0f/custom_components/tuya_local/devices/etop_ch7100_thermostat.yaml#L166-L173

>The current_humidity attribute does not allow customising the units, it is always %, so that change is unnecessary and will be ignored

Understood, but I just want to mention I used other examples as reference, see above.

> only the heatcool needed fixing, but should have been changed to heat_cool to match the HA meaning of the modes.

Again, I started from other devices as a reference, so I thought it would make more sense. But I think you are right about using HA defaults.
https://github.com/make-all/tuya-local/blob/6d46cfc7ac5ec16f3a6efd7fb58fa448e2f87f0f/custom_components/tuya_local/devices/moes_trv_thermostat.yaml#L14-L19

> Filter life is a sensor, and therefore also read-only, so min and max will be ignored. Sensor does not have any API for the gauge card to read any value you set here, so you will anyway need to set these in the card configuration. Setting the scale to 1 will not affect anything, as that is already the default.

Makes sense to me 👍

>The other change you made was to remove the optional: true tag from dp 201, which will only cause detection failures.

The device always sends DPS 201 when querying data points, So, in my understanding, it's not optional, it will actually help detection, right? It detected my device just fine. Only DPS 202 is optional (it is not sent when querying):

query DPS

PR #1980, opened 2024-06-05
Update qlima_d820a_dehumidifier.yaml

Blcktape:
Add current humidity sensor and change a description.

### DIFF
--- a/custom_components/tuya_local/devices/qlima_d820a_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/qlima_d820a_dehumidifier.yaml
@@ -113,7 +113,7 @@ secondary_entities:
             value: Purify only
             icon: "mdi:air-filter"
           - dps_val: dehum30_air
-            value: Both
+            value: Dehumidify and Purify
             icon: "mdi:air-humidifier"
   - entity: switch
     name: Internal dry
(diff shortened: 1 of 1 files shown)


make-all:
Current humidity is already part of the humidifier entity.
Changing the naming is just a personal preference, and can be disruptive if other users have automations set up using the current name.

Blcktape:
Current humidity is indeed already part of the humidifier entity. But you are not able to use it as its own sensor entity just like temperature.
To be able to do something with the data like displaying it in a graph the current humidity should be available and in my oppinion disclosed as an sensor.

make-all:
You can display a graph of the current_humidity attribute of the humidifier entity.

Blcktape:
> You can display a graph of the current_humidity attribute of the humidifier entity.

I had an issue with that using an "Statistics graph card". It would only show "sensor" entities. As "current_humidity" isnt disclosed as an sensor it would not be able to display it as such in my case.
If there is a different way i would gladly like to know.

make-all:
Template sensors can get around limitations like this. Or you can use the standard History graph car, which supports most entity types.

Blcktape:
> Template sensors can get around limitations like this. Or you can use the standard History graph car, which supports most entity types.

Thank you, this fixed my problem as i did not now "template sensors" existed. This patch can be discarded.

Issue #1971, opened 2024-06-03
Atomberg Gorilla Fan - Not able to set max speed


aravindjaimon:
This form is for reporting bugs.  If you have encountered a bug, you should
generally have some log messages or device diagnostics to show something is
going wrong.  If your problem is more of a question or a request for help,
then [Discussions](https://github.com/make-all/tuya-local/discussions) is
probably a more suitable venue.

**Describe the bug**
Fan is successfully added, but the speed is having 3 range, Low, Medium and High with values 1,2 and 3 respectively. My fan supports 5 speed levels by default and with boost it goest to 6th speed. When looking at the percentage calculation its getting devided by 33% since the speed level is divisible by 5 it should be by 20%

**To Reproduce**
Steps to reproduce the behavior including the device the issue was observed with:
Load Gorilla Fan from Atomberg to the device, try changing speed using controls

**Expected behavior**
User should be able to control the fan speed with its full potential.

**Additional context**

If the bug involves a device, then please include device diagnostics from
Settings / Devices & Services / Tuya Local (your device) / "1 device"

Please check if there are messages from Tuya Local in the Assistant log (Settings / System / Logs), and include those too.

```
--- Logging error ---
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/logging/handlers.py", line 1497, in emit
    self.enqueue(self.prepare(record))
                 ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/logging/handlers.py", line 1479, in prepare
    msg = self.format(record)
          ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/logging/__init__.py", line 999, in format
    return fmt.format(record)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/logging/__init__.py", line 703, in format
    record.message = record.getMessage()
                     ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/logging/__init__.py", line 392, in getMessage
    msg = msg % self.args
          ~~~~^~~~~~~~~~~
TypeError: not all arguments converted during string formatting
Call stack:
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/src/homeassistant/homeassistant/__main__.py", line 220, in <module>
    sys.exit(main())
  File "/usr/src/homeassistant/homeassistant/__main__.py", line 208, in main
[...]

PR #1988, opened 2024-06-07
feat(fan): added support for gorilla fan v2

aravindjaimon:
# Add support for Gorilla Fan V2

This PR adds support for the Gorilla Fan V2 device to the Tuya Local integration.

## Changes
- Added new device configuration file `gorilla_fan_v2.yaml` defining the DPS mappings and entities for the Gorilla Fan V2

## Testing
- Manually tested with a physical Gorilla Fan V2 device

Resolves: #1971


custom_components/tuya_local/devices/gorilla_fan_v2.yaml
```diff
@@ -0,0 +1,58 @@
...
+secondary_entities:
+  - entity: light
+    name: Night Light
+    icon: mdi:ceiling-fan-light
+    dps:
+      - id: 15
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Boost
```
make-all, line 33: What is the difference between this and the mode (dp 2)?
It seems confusing to have both named the same way.

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
```diff
@@ -0,0 +1,58 @@
...
+    name: Sleep
+    icon: mdi:sleep
+    dps:
+      - id: 113
+        type: boolean
+        name: switch
+  - entity: number
+    name: Timer
+    icon: mdi:timer-edit
+    translation_key: timer
```
make-all, line 49: The translation_key should cover name and icon already, so those should be removed.

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - id: 3
+      type: integer
+      name: speed
+      range:
+        min: 1
+        max: 5
+
+secondary_entities:
+  - entity: light
+    name: Night Light
```
make-all, line 26: Consider using `translation_key: nightlight` instead of an explicit name and icon.

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
```diff
@@ -0,0 +1,58 @@
+name: Gorilla Fan V2
```
make-all, line 1: Preferably there should be a products section here, with product id and name (with branding), and the top level name should be generic (Ceiling fan, for example)

make-all: I see from #1971 that the brand is Atomberg, and the product name is Gorilla.  The file naming and the name in the products section should reflect this.  I'm not sure whether the V2 is part of the product name, or it is a misunderstanding that whatever config you originally installed this as is "V1" (but there is no atomberg or gorilla config, so there is no need to use v2 here as disambiguation if it is not part of the actual product name).

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
aravindjaimon: dp 2 changes when the switch for boost changes but the vice-versa doesn't happen, I'll remove dp2

make-all: Then I think you can move this into the fan as preset_mode (mapping: false -> normal, true -> strong) and add `translation_key: fan_with_presets` to the fan, to get those translated.  Rename the dp 2 to something not standard in the fan entity (probably the tuya naming is ok) to leave it as an extra attribute for reference (probably without any mapping is best, since it is only for reference).

tests/devices/test_gorilla_fan_v2.py
```diff
@@ -0,0 +1,71 @@
...
+        self.assertEqual(self.subject.preset_mode, "Normal")
+
+        self.dps[PRESET_DPS] = "boost"
+        self.assertEqual(self.subject.preset_mode, "Boost")
+
+    def test_preset_modes(self):
+        self.assertCountEqual(self.subject.preset_modes, ["normal", "boost"])
+
+    async def test_set_preset_mode_to_normal(self):
+        async with assert_device_properties_set(
```
make-all, line 38: The error here is because of the mapping on preset_mode.  But better would be to use lower case, with translation_key as above, and use "strong" instead of "boost" since that already has translations.

tests/devices/test_gorilla_fan_v2.py
```diff
@@ -0,0 +1,71 @@
...
+            self.subject._device,
+            {PRESET_DPS: "normal"},
+        ):
+            await self.subject.async_set_preset_mode("Normal")
+
+    async def test_set_preset_mode_to_boost(self):
+        async with assert_device_properties_set(
```
make-all, line 45: Same error cause as above

tests/devices/test_gorilla_fan_v2.py
```diff
@@ -0,0 +1,71 @@
...
+            await self.subject.async_set_preset_mode("Boost")
+
+    def test_speed(self):
+        self.dps[PRESET_DPS] = "normal"
+        self.dps[FANMODE_DPS] = 3
+        self.assertEqual(self.subject.percentage, 60)
+
+    async def test_set_speed_in_normal_mode(self):
+        self.dps[PRESET_DPS] = "normal"
+        async with assert_device_properties_set(self.subject._device, {FANMODE_DPS: 3}):
```
make-all, line 58: This error is due to FAN_PAYLOAD having `"1": False`, which means that the fan is off, so it will automatically be switched on when setting a speed > 0.  You can avoid it by adding `self.dps[SWITCH_DPS] = True` above to start the test with the fan on.

tests/devices/test_gorilla_fan_v2.py
```diff
@@ -0,0 +1,71 @@
...
+            await self.subject.async_set_percentage(60)
+
+    async def test_set_speed_in_boost_mode(self):
+        self.dps[PRESET_DPS] = "boost"
+        async with assert_device_properties_set(self.subject._device, {FANMODE_DPS: 5}):
```
make-all, line 63: Same error cause as above

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
```diff
@@ -0,0 +1,58 @@
...
+primary_entity:
+  entity: fan
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+      icon: mdi:ceiling-fan
```
make-all, line 8: This should be at entity level, not inside the dp definition.  If inside a dp definition, it is expected to be inside a mapping, so that it is conditional on the dp value.  But probably it is best to avoid explicitly setting an icon. It might be better to add a new translation_key for `ceiling_fan` and set the icon there, but then you won't be able to reuse the already existing translations from fan_with_presets for the preset modes, or you'll have to duplicate them in the translation files.

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
aravindjaimon: Removed name and icon

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
aravindjaimon: I attempted the suggested changes, and while they were beneficial, I encountered an issue. When changing the speed, the boost mode should revert to normal, but this isn't happening. Additionally, I believe it's more convenient to maintain it as a separate entity.
Although boost functions as a 6th speed, the speed value does not change when set to boost.

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
aravindjaimon: Removed explicit icon

tests/devices/test_gorilla_fan_v2.py
aravindjaimon: removed test cases as they are not necessary.

tests/devices/test_gorilla_fan_v2.py
aravindjaimon: removed test cases as they are not necessary

tests/devices/test_gorilla_fan_v2.py
aravindjaimon: removed test cases as they are unnecessary

tests/devices/test_gorilla_fan_v2.py
aravindjaimon: removed testcases as they are unnecessary

tests/devices/test_gorilla_fan_v2.py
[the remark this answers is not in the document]
aravindjaimon: removed testcases

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
aravindjaimon: The fan in question is from Atomberg and is referred to as the Renesa fan. However, this name is not present in any of our systems. In both the Tuya Dashboard and the Smart Life App, it is listed as the Gorilla Fan V2. Therefore, I used this name for consistency.

aravindjaimon:
@make-all  The configurations are updated as per your recommendations. Please review my comments and changes.

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
make-all: What about the product id? If you cannot get it from the cloud, it is also possible to read from `python -m tinytuya scan`

custom_components/tuya_local/devices/gorilla_fan_v2.yaml
make-all: The issue you encountered is not really resolved by moving Boost from a preset_mode of the fan to a separate switch.  The way to solve it would be to combine it with speed:

```
    - id: 3
      type: integer
      name: speed_setting
      hidden: true
    - id: 115
      name: speed
      type: boolean
      mapping:
        - dps_val: true
          value: 100
        - dps_val: false
          constraint: speed_setting
          conditions:
            - dps_val: 1
              value: 17
            - dps_val: 2
              value: 33
            - dps_val: 3
              value: 50
            - dps_val: 4
              value: 67
            - dps_val: 5
              value: 83
```

PR #2005, opened 2024-06-12
support for diivoo dwv010 dual water timer

jannisan:
Simple config file for DWV010 with manual control of both valves, battery sensor, total water consumption, weather delay

### DIFF
--- a/custom_components/tuya_local/devices/diivoo_dwv010.yaml
+++ b/custom_components/tuya_local/devices/diivoo_dwv010.yaml
@@ -26,6 +26,7 @@ secondary_entities:
         name: operation
   - entity: sensor
     name: Water use
+    class: volume
     category: diagnostic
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/diivoo_dwv010.yaml
```diff
@@ -0,0 +1,138 @@
...
+secondary_entities:
+  - entity: valve
+    class: water
+    name: Valve 2
+    dps:
+      - id: 121
+        type: boolean
+        name: valve
+  - entity: sensor
+    name: cumulative water consumtion
```
make-all, line 60: First word of entity names should be capitalised.  p is missing from consumption.  Maybe a shorter and easier to understand name is "Water usage".  A `class: volume` would be useful here too.

Issue #1598, opened 2024-02-04
Light adds support for EFFECT_OFF in HA 2024.2


make-all:
HA 2024.2 now supports explicitly setting light effect to off.  We'll need to look at how to best handle this change.

https://developers.home-assistant.io/blog/2024/01/23/light-color-mode-changes

make-all:
When effect is a separate dp, it should have an explicit "off" value, but if it is combined with color_mode (as is the case for many tuya bulbs supporting scene and music), we need a way to switch back out of the effect mode, so implicitly include EFFECT_OFF in that case.

Until now, it was a bit awkward to shift out of those modes, moreso when only one real color mode is supported, as the HA UI doesn't let you change to the same value it thinks it is, so you are forced to use a service call.

PR #2009, opened 2024-06-12
Extra support for LSC_SMART_Connect_RGB-CCT_Bulb

malosaa:
Support for LSC_SMART_Connect_RGB-CCT_Bulb

### DIFF
--- a/custom_components/tuya_local/devices/LSC_SMART_Connect_RGB-CCT_Bulb
+++ b/custom_components/tuya_local/devices/LSC_SMART_Connect_RGB-CCT_Bulb
@@ -1,0 +1,106 @@
+name: LSC RGB-CCT
+products:
+  - id: hpaeslvkx4znwzae 
+    name: LSC SMART Connect A60 RGB-CCT
+
+primary_entity:
+  entity: light
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+    - id: 2
+      name: color_mode
+      type: string
+      mapping:
+        - dps_val: white
+          value: color_temp
+        - dps_val: colour
+          value: hs
+    - id: 3
+      name: brightness
+      type: integer
+      range:
+        min: 25
+        max: 255
+    - id: 4
+      name: color_temp
+      type: integer
+      range:
+        min: 0
+        max: 255
+      mapping:
+        - target_range:
+            min: 2700
+            max: 6500
+    - id: 5
+      name: rgbhsv
+      type: hex
+      format:
+        - name: r
+          bytes: 1
+        - name: g
+          bytes: 1
+        - name: b
+          bytes: 1
+        - name: h
+          bytes: 2
+          range:
+            min: 0
+            max: 360
+        - name: s
+          bytes: 1
+          range:
+            min: 0
+            max: 255
+        - name: v
+          bytes: 1
+          range:
+            min: 0
+            max: 255
+
+secondary_entities:
+  - entity: select
+    name: Scene
+    icon: "mdi:palette"
+    category: config
+    dps:
+      - id: 6
+        type: string
+        name: option
+        optional: true
+        mapping:
+          - dps_val: "ffff2801005766"
+            value: Scene 1
+          - dps_val: "ffff3d0266000003660000ff48000000000000000000"
+            value: Scene 2
+          - dps_val: "ffff2301096600"
+            value: Scene 3
+          - dps_val: "ffff0505ff000000ff00ffff00ff00ff0000ff000000"
+            value: Scene 4
+
+  - entity: select
+    name: Work Mode
+    icon: "mdi:lightbulb"
+    category: config
+    dps:
+      - id: 2
+        type: string
+        name: option
+        optional: true
+        mapping:
+          - dps_val: "colour"
+            value: Color
+          - dps_val: "white"
+            value: White
+          - dps_val: "scene"
+            value: Scene
+
+  - entity: sensor
+    name: Current Work Mode
+    icon: "mdi:lightbulb"
+    category: diagnostic
+    dps:
+      - id: 2
+        type: string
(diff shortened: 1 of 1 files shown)


malosaa:
This code is for 3.2 as 3.3 is problematic when adding multiple, gives unknown error on device adding while selecting 3.3 while 3.2 has no problems so i created this code for version 3.2 and works perfect.

custom_components/tuya_local/devices/LSC_SMART_Connect_RGB-CCT_Bulb
```diff
@@ -0,0 +1,106 @@
...
+      type: boolean
+      name: switch
+    - id: 2
+      name: color_mode
+      type: string
+      mapping:
+        - dps_val: white
+          value: color_temp
+        - dps_val: colour
+          value: hs
```
make-all, line 19: Add a mapping for `scene` here and the duplicate select entity below will not be needed.

custom_components/tuya_local/devices/LSC_SMART_Connect_RGB-CCT_Bulb
```diff
@@ -0,0 +1,106 @@
...
+        optional: true
+        mapping:
+          - dps_val: "colour"
+            value: Color
+          - dps_val: "white"
+            value: White
+          - dps_val: "scene"
+            value: Scene
+
+  - entity: sensor
```
make-all, line 99: Triplicating this as a sensor is totally unnecessary.

make-all (commented):
This seems to match rgbww_lightbulb apart from the colour_temp target range and the extra "Flash scene" presets.
That would probably be a good template to follow.

custom_components/tuya_local/devices/LSC_SMART_Connect_RGB-CCT_Bulb
```diff
@@ -0,0 +1,106 @@
+name: LSC RGB-CCT
```
make-all, line 1: This name should just be generic.  Quite likely other bulbs match this, as it seems to follow the older standard layout tuya bulbs use (though none exactly match it).

custom_components/tuya_local/devices/LSC_SMART_Connect_RGB-CCT_Bulb
malosaa: like the screenshot above, we have a scene selector, that works normal, and we have a work_mode selector that works also normal. If we add the scene as a mapping it won't work and fail, the scenes won't work. see screenshot below.




image
image

Also i show you the updated code according your requests, and it make it looks bad.


```
name: Generic RGB-CCT Bulb
products:
  - id: hpaeslvkx4znwzae 
    name: LSC SMART Connect A60 RGB-CCT

primary_entity:
  entity: light
  dps:
    - id: 1
      type: boolean
      name: switch
    - id: 2
      name: color_mode
      type: string
      mapping:
        - dps_val: white
          value: color_temp
        - dps_val: colour
          value: hs
        - dps_val: scene
          value: scene
    - id: 3
      name: brightness
      type: integer
      range:
        min: 25
        max: 255
    - id: 4
      name: color_temp
      type: integer
      range:
        min: 0
        max: 255
      mapping:
        - target_range:
            min: 2700
            max: 6500
    - id: 5
      name: rgbhsv
      type: hex
      format:
        - name: r
          bytes: 1
        - name: g
          bytes: 1
        - name: b
          bytes: 1
        - name: h
          bytes: 2
          range:
            min: 0
            max: 360
        - name: s
          bytes: 1
          range:
            min: 0
            max: 255
        - name: v
          bytes: 1
          range:
            min: 0
            max: 255

secondary_entities:
  - entity: select
    name: Scene
    icon: "mdi:palette"
    category: config
    dps:
      - id: 6
        type: string
        name: option
        optional: true
        mapping:
          - dps_val: "ffff2801005766"
            value: Scene 1
          - dps_val: "ffff3d0266000003660000ff48000000000000000000"
            value: Scene 2
          - dps_val: "ffff2301096600"
            value: Scene 3
          - dps_val: "ffff0505ff000000ff00ffff00ff00ff0000ff000000"
            value: Scene 4
```

custom_components/tuya_local/devices/LSC_SMART_Connect_RGB-CCT_Bulb
malosaa: Agree, i will update that

custom_components/tuya_local/devices/LSC_SMART_Connect_RGB-CCT_Bulb
make-all: Normally I would capitalise the non-standard entries in color_mode, as they do not have translations.
But it seems the HA change referenced in #1598 also came with a behaviour change that effects no longer auto cancel when you change modes, so implementation of that may need to be prioritised.

custom_components/tuya_local/devices/LSC_SMART_Connect_RGB-CCT_Bulb
malosaa: Agree, i change it to LSC RGB-CCT

malosaa:
> This seems to match rgbww_lightbulb apart from the colour_temp target range and the extra "Flash scene" presets. That would probably be a good template to follow.

Well that template does not work as it should be, as these bulbs do work different.
There are a lot of problems the way the scenes and the selectors work, as that template works well except for working mode and the scenes.  
The way i created the selectors solve all problems for the LSC bulbs and the scenes work correct.

malosaa:
@make-all 
i managed to map the requested scenes and work now.

image

i can add a diagnostic to this device template like this screenshot:
image

if its not needed i leave it out and i update the code if you want.

PR #2011, opened 2024-06-13
Adding Poolsana Heat Pump

Cyrelion:
Adding Poolsana Heat Pump to supported devices: https://www.poolsana.de/pool-full-inverter-waermepumpe-poolsana-inverpower-next-11

### DIFF
--- a/custom_components/tuya_local/devices/poolsana_heatpump.yaml
+++ b/custom_components/tuya_local/devices/poolsana_heatpump.yaml
@@ -52,7 +52,7 @@ primary_entity:
 secondary_entities:
   - entity: sensor
     class: temperature
-    name: Effluent Temperature
+    name: Outlet Temperature
     category: diagnostic
     dps:
       - id: 25
--- a/custom_components/tuya_local/devices/poolsana_heatpump.yaml
+++ b/custom_components/tuya_local/devices/poolsana_heatpump.yaml
@@ -20,7 +20,7 @@ primary_entity:
             - dps_val: Heating
               value: heat
             - dps_val: Auto
-              value: auto
+              value: heat_cool
     - id: 2
       type: string
       name: mode
@@ -37,12 +37,8 @@ primary_entity:
     - id: 16
       type: integer
       name: current_temperature
-      range:
-        min: -300
-        max: 1000
       mapping:
         - scale: 10
-          step: 1
     - id: 5
       type: string
       name: preset_mode
@@ -62,13 +58,7 @@ secondary_entities:
       - id: 25
         type: integer
         name: sensor
-        range:
-          min: 0
-          max: 100
         unit: "°C"
-        mapping:
-          - scale: 1
-            step: 1
   - entity: binary_sensor
     category: diagnostic
     class: problem
--- a/custom_components/tuya_local/devices/poolsana_heatpump.yaml
+++ b/custom_components/tuya_local/devices/poolsana_heatpump.yaml
@@ -1,8 +1,8 @@
 name: Pool heat pump
+products:
+- id: abb8kaxcphfzkwsg
+  name: PoolSana InverPower Next
 primary_entity:
-  products:
-  - id: abb8kaxcphfzkwsg
-    name: PoolSana InverPower Next
   entity: climate
   translation_key: pool_heatpump
   dps:
--- a/custom_components/tuya_local/devices/poolsana_heatpump.yaml
+++ b/custom_components/tuya_local/devices/poolsana_heatpump.yaml
@@ -1,0 +1,113 @@
+name: PoolSana InverPower Next
+primary_entity:
+  entity: climate
+  translation_key: pool_heatpump
+  dps:
+    - id: 1
+      type: boolean
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: Cooling
+              value: cool
+            - dps_val: Heating
+              value: heat
+            - dps_val: Auto
+              value: auto
+    - id: 2
+      type: string
+      name: mode
+      hidden: true
+    - id: 4
+      type: integer
+      name: temperature
+      range:
+        min: 70
+        max: 400
+      mapping:
+        - scale: 10
+          step: 10
+    - id: 16
+      type: integer
+      name: current_temperature
+      range:
+        min: -300
+        max: 1000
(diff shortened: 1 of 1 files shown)


Cyrelion:
This adds the heat pump as a climate entity and also different dp sensors for current temperature, target temperature, effluent temperature and fault value.

custom_components/tuya_local/devices/poolsana_heatpump.yaml
```diff
@@ -0,0 +1,113 @@
...
+        type: integer
+        name: sensor
+        range:
+          min: 0
+          max: 100
+        unit: "°C"
+        mapping:
+          - scale: 1
+            step: 1
+  - entity: sensor
```
make-all, line 69: There is no need to duplicate the current_temperature and temperature from the climate entity.

custom_components/tuya_local/devices/poolsana_heatpump.yaml
```diff
@@ -0,0 +1,113 @@
...
+        type: integer
+        name: sensor
+        range:
+          min: 70
+          max: 400
+        mapping:
+          - scale: 10
+            step: 10
+        unit: "°C"
+  - entity: sensor
```
make-all, line 99: This should be a `binary_sensor`.  It does not need a name, as the `class: problem` provides one already.

custom_components/tuya_local/devices/poolsana_heatpump.yaml
```diff
@@ -0,0 +1,113 @@
+name: PoolSana InverPower Next
```
make-all, line 1: A `products` section with the product id and name (branding) would be useful, and the above name should then be changed to be generic ("Pool heat pump") in case other brands match the same device config.

Product id can be obtained from iot.tuya.com info, or `python -m tinytuya scan`

custom_components/tuya_local/devices/poolsana_heatpump.yaml
Cyrelion: I think I miss something here, and/or the line numbers are wrong. I thik you're talking about the Fault sensor, right?

custom_components/tuya_local/devices/poolsana_heatpump.yaml
Cyrelion: Sure, I've added/corrected this.

custom_components/tuya_local/devices/poolsana_heatpump.yaml
make-all: Tile card and Entity card both let you display the current temperature or temperature of a climate entity by itself easily, if that is what you want to do.  I think the entities list card you have to edit the yaml to do it, and maybe some custom cards do not support it, but in that case you can add a template_sensor helper to expose the attribute as a sensor if you really want to.

custom_components/tuya_local/devices/poolsana_heatpump.yaml
make-all: Yes, the comment is for the fault sensor.

custom_components/tuya_local/devices/poolsana_heatpump.yaml
```diff
@@ -0,0 +1,85 @@
...
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: Cooling
+              value: cool
+            - dps_val: Heating
+              value: heat
+            - dps_val: Auto
+              value: auto
```
make-all, line 23: This more likely corresponds to HA's `heat_cool` mode than `auto` based on the descriptions at https://developers.home-assistant.io/docs/core/entity/climate/#hvac-modes

custom_components/tuya_local/devices/poolsana_heatpump.yaml
```diff
@@ -0,0 +1,85 @@
...
+      range:
+        min: 70
+        max: 400
+      mapping:
+        - scale: 10
+          step: 10
+    - id: 16
+      type: integer
+      name: current_temperature
+      range:
```
make-all, line 40: Range is not needed here, as current_temperature only reports what the device is saying.  Step is also not needed (for that reason, and also that a step of 1 is the default)

custom_components/tuya_local/devices/poolsana_heatpump.yaml
```diff
@@ -0,0 +1,85 @@
...
+        - dps_val: Boost
+          value: boost
+        - dps_val: Silence
+          value: eco
+        - dps_val: Smart
+          value: comfort
+secondary_entities:
+  - entity: sensor
+    class: temperature
+    name: Effluent Temperature
```
make-all, line 59: Effluent makes it sound like sewage, maybe outlet is better.

custom_components/tuya_local/devices/poolsana_heatpump.yaml
```diff
@@ -0,0 +1,85 @@
...
+    category: diagnostic
+    dps:
+      - id: 25
+        type: integer
+        name: sensor
+        range:
```
make-all, line 65: Again, range is not needed for sensors, as they report what the device is saying regardless of whether it is in range or not. Step and scale are also both at defaults, so the whole mapping can be removed too.

custom_components/tuya_local/devices/poolsana_heatpump.yaml
Cyrelion: Yeah, you're right, it's no AI behind :-) Corrected

custom_components/tuya_local/devices/poolsana_heatpump.yaml
Cyrelion: Effluent is the original tuya naming. Because I don't know exactly what sensor this is measuring I didn't want to change naming :-) But if you want I can change

Cyrelion:
Sorry for the "problems", this is my first commit and first time I write a device yaml file :-(

custom_components/tuya_local/devices/poolsana_heatpump.yaml
make-all: Yes, tuya naming is often a bit off due to English not being their native language.  But I am pretty sure this heatpump only pumps out heated water.

Cyrelion: Yeah so true. I‘ve changed it and hope it is really only pumping water 😉

PR #2016, opened 2024-06-16
Add support for Inkbird IBT-26S - BBQ thermometer

nima-1102:
I have added support for the Inkbird IBT-26S BBQ thermometer and have tested its functionality.


custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
```diff
@@ -0,0 +1,292 @@
...
+      - id: 112
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+
+  - entity: select
+    translation_key: temperature_unit
+    category: config
+    dps:
```
make-all, line 92: Does this not affect any of the sensor readings?

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
```diff
@@ -0,0 +1,292 @@
...
+            value: celsius
+  - entity: sensor
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 101
+        type: integer
+        name: sensor
+        unit: "%"
+  - entity: number
```
make-all, line 109: May be better emulated as a light.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
```diff
@@ -0,0 +1,292 @@
...
+          max: 100
+  - entity: switch
+    name: Alarm Sound
+    category: config
+    dps:
+      - id: 104
+        type: boolean
+        name: switch
+  - entity: sensor
+    name: Status Code
```
make-all, line 129: Does this need to be a sensor?

I think it would be better as an enum class sensor with values mapped to descriptions of the actual meaning of the status codes. If the codes are not known, then maybe an attribute on another entity is enough.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
```diff
@@ -0,0 +1,292 @@
...
+    name: Alarm Probe 1
+    category: diagnostic
+    icon: "mdi:alarm-light"
+    dps:
+      - id: 120
+        type: string
+        name: option
+        optional: true
+        mapping:
+          - dps_val: "EAAAAAAKBw=="
```
make-all, line 219: I don't think the alarm is this simple. That is "On with some specific setting".  Maybe it is better to just leave the alarms as attributes on the temperature probe entities, and let the user set it in the official app, where all the functions are available.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
nima-1102: This has no influence on the sensor values, only the display on the device is changed.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
nima-1102: I have adjusted the range minimum to 5% instead of 0%, is there a light entity without an on and off switch integrated?

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
nima-1102: I looked at it again and added a Bluetooth & Battery state sensor instead, works well.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
nima-1102: I don't use the official app, just Homeassistant, so I have the option of triggering the alarm sound for a sensor via automation. It works great, otherwise you would have to use multiple apps. For this reason it is optional.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
make-all: Yes, lights can have just a brightness dp. If the brightness does not go down to 0, they will not be presented with a switch in the UI.

make-all: There is also `translation_key: backlight` or `translation_key: display` which can be used with light in place of an explicit name and icon.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
```diff
@@ -0,0 +1,330 @@
...
+    dps:
+      - id: 102
+        type: integer
+        name: value
+        unit: '%'
+        range:
+          min: 5
+          max: 100
+  - entity: switch
+    name: Alarm Sound
```
make-all, line 157: Maybe add `icon: "mdi:bullhorn"` or similar.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
make-all: By contributing to this project, you are not just meeting your immediate needs, but making this available for everyone. So I don't agree with implementing this as "On" and "Off" with your personally preferred settings.  If you just want your own custom solution, you would be better using localtuya.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
nima-1102: I've tried that several times, it's always presented as a slider with a switch in the UI. I changed it anyway, maybe I did something wrong.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
nima-1102: I'm not merely "meeting my immediate needs;" rather, I have articulated a useful purpose for this type of function. The implementation doesn't necessarily have to be exactly as I described, but I haven't found any other way to trigger the device's alarm. I've now simply changed it to a switch. If it really shouldn't be there at all, I can remove it.

I'm open to suggestions on how to implement this or a similar function while also retaining the ability to trigger the alarm.

make-all: Ideally I think if the bit that controls off/on can be determined, that can be exposed as an off / on switch.  But exposing it like now with a particular full setting of the alarm presumes that everyone will want the same settings as you, which I think is wrong.  I would rather it be a read-only attribute than to expose a switch that is geared to a single use case.

make-all: The two settings you have now decode to the following 48bit values (HEX):

- ON  100000000007
- OFF 00aa05000007

Most likely that 10 in the first byte is the on/off bit, but it isn't clear why the following two bytes are different in the off value.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
```diff
@@ -0,0 +1,342 @@
...
+          - dps_val: 2580
+            value: 1
+  - entity: binary_sensor
+    name: Bluetooth
+    icon: mdi:bluetooth
+    category: diagnostic
+    class: connectivity
+    dps:
+      - id: 128
+        type: boolean
```
make-all, line 133: The dps_vals in the mapping are integers.  I suspect this should be a bitfield, and only one bit is needed to be defined in the mapping as indicating true, rather than filtering out specific values as false or true as now.

I think the correct mapping would be:
```
- id: 128
  type: bitfield
  name: sensor
  mapping:
    - dps_val: 4
      value: true
    - value: false
```

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
```diff
@@ -0,0 +1,342 @@
...
+        name: sensor
+        mapping:
+          - dps_val: 4624
+            value: 0
+          - dps_val: 4628
+            value: 1
+          - dps_val: 2576
```
make-all, line 122: My guess is the correct mapping here would be:
```
- id: 128
  type: bitfield
  name: sensor
  mapping:
    - dps_val: 2048
      value: true
    - value: false
```

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
```diff
@@ -0,0 +1,342 @@
...
+    name: Battery level
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 101
+        type: integer
+        name: sensor
+        unit: "%"
+  - entity: binary_sensor
+    name: Battery state
```
make-all, line 110: You do not need the name, if the class is set and unique  Leaving out the name allows HA provided translations to be used.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
```diff
@@ -0,0 +1,342 @@
...
+          - dps_val: 4628
+            value: false
+          - dps_val: 2576
+            value: true
+          - dps_val: 2580
+            value: true
+  - entity: light
+    translation_key: backlight
+    category: config
+    icon: "mdi:brightness-percent"
```
make-all, line 147: Suggest to leave out icon when translation key is provided.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
nima-1102: I tested it and it doesn't work, my current config works without any problems.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
nima-1102: I tested it and it doesn't work, my current config works without any problems.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
nima-1102: Okay, I don't know enough about that, so I removed everything.

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
make-all: Did your test also include fixing the type (it is definitely not boolean)?

custom_components/tuya_local/devices/inkbird_ibt_26s_thermometer.yaml
make-all: Likewise

PR #2018, opened 2024-06-16
Add support for Zemismart roller shade - ZM25R2

rac146:
Got this new variant of one of the zemismart roller shades (wifi) - ZM25R2:
https://www.zemismart.com/products/zm25r2

Similar to the other zemismarts, but not exactly the same. Action is unreliably reported so did the same thing as one of the other zemismart configs.


make-all:
This seems to match m027_curtain the closest, except for the inversion of positions and disabling of the unreliable action.  I think you can find some improvements for this config if you study that one.

rac146:
They do look similar! The odd part here is that the zemismart seems to want ID 2, 4, and possibly 12 as optional, which is slightly different than the M027 (looks like ID 3 is optional). Are you thinking we try to combine the settings by making one file?

I also added 101 for remote pairing with the zemismart, but currently the M027 file has this marked unknown

make-all:
I think if the positions are not inverted in M027, but they are in this device, then the configs cannot be merged into one.  I was more suggesting that you look at it for ideas, as I think there were some secondary entities implemented there, which are missing from this config (left as read-only attributes)

rac146:
Got it, makes sense. I'll test out those other entities and update the file

rac146:
@make-all - updated config with the missing entities. Testing so far looks good..

rac146:
So far the shades have randomly become unavailable a few times - I'm not really an expert in this area but I'm assuming maybe some of the other DPIDs should be marked optional. I switched 10 and 20 to optional for now and haven't noticed any problems. Will keep testing.

Is there any way to tell for sure which DPIDs should be marked optional?

make-all:
Devices going unavailable is most likely related to power management on the devices themselves (are they battery powered?) or network instability. It is not affected by which dps are optional or not.  The optional dps only matter when adding new devices (when the dps received from the device are matched against the config, at which time any dps that may be missing need to be marked as optional).  To find all the dps that need to be marked optional, turn on debug logging, and watch the dps lists marked with `"full_poll": true` to see which dps are not reliably returned with every full poll.

PR #2020, opened 2024-06-17
Adding Support for BLE-YL01 Waterquality tester

Cyrelion:
BLE-YL01

### DIFF
--- a/custom_components/tuya_local/devices/ble-yl01_waterquality_tester.yaml
+++ b/custom_components/tuya_local/devices/ble-yl01_waterquality_tester.yaml
@@ -186,22 +186,11 @@ secondary_entities:
         range:
           min: -1200
           max: 1200
-  - entity: number
-    name: Minimum humidity
-    category: config
-    class: humidity
-    dps:
-      - id: 116
-        type: integer
-        name: value
-        unit: "%"
-        range:
-          min: 0
-          max: 100
   - entity: sensor
-    name: SALT TDS
+    name: Salt
     dps:
-      - id: 116
+      - id: 117
         type: integer
-        name: value
+        name: sensor
+        class: measurement
         unit: "ppm"
--- a/custom_components/tuya_local/devices/ble-yl01_waterquality_tester.yaml
+++ b/custom_components/tuya_local/devices/ble-yl01_waterquality_tester.yaml
@@ -1,0 +1,207 @@
+name: Water quality monitor
+products:
+  - id: v1jqz5cy
+    name: BLE-YL01 Waterquality tester
+primary_entity:
+  entity: sensor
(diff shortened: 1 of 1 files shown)


Cyrelion:
The only left problem is that it looks like the device isn't pushing values on a regular basis, only if you open the mobile app. Don't know if there is a switch I can enable on the device.yaml to do a regular poll

make-all:
On some other devices, users have observed that writing to any writable dp will trigger a report of all the sensor values.  This seems to be what the app is doing behind the scenes.

Cyrelion:
That doesn't work :-(

robinhood-code:
I'm using this and the TDS is not reporting (always 0) but it's showing in the Tuya app. Anyone else? 

Also can we implement the Calibration button? These are the dps below:

```
{\"abilityId\":114,\"accessMode\":\"rw\",\"code\":\"phcal\",\"description\":\"\",\"name\":\"pH Calibration\",\"typeSpec\":{\"type\":\"value\",\"max\":1400,\"min\":0,\"scale\":2,\"step\":1,\"unit\":\"\"}},
{\"abilityId\":115,\"accessMode\":\"rw\",\"code\":\"eccal\",\"description\":\"\",\"name\":\"EC Calibration\",\"typeSpec\":{\"type\":\"value\",\"max\":20000,\"min\":0,\"scale\":0,\"step\":1,\"unit\":\"us/cm\"}},
{\"abilityId\":116,\"accessMode\":\"rw\",\"code\":\"orpcal\",\"description\":\"\",\"name\":\"ORP Calibration\",\"typeSpec\":{\"type\":\"value\",\"max\":1200,\"min\":-1200,\"scale\":0,\"step\":1,\"unit\":\"mV\"}},
{\"abilityId\":118,\"accessMode\":\"rw\",\"code\":\"phreset\",\"description\":\"\",\"name\":\"pH Reset\",\"typeSpec\":{\"type\":\"bool\"}}]}]}"
```

Cyrelion:
I can‘t test anymore because I’ve send back the device as it only send updates when I’ve opened the Tuya app. This doesn’t work for my use-case

make-all:
105 is also missing from the current config, the name in tuya docs is "backightValue", and it has a range of 0-16M. It might be brightness for the backlight, but your log is showing 0, so maybe it is not useful.

robinhood-code:
@Cyrelion just curious what alternative are you switching to ? the device is really unreliable today I've only added salt to the pool and now the probe is report 0 for pH.

Cyrelion:
@robinhood-code No, didn't found any alternatives yet. Yes it's absolutely not useable, I'm still searching and open for suggestions.

PR #2032, opened 2024-06-21
Add kogan lx-8 support

ewandank:
Adds support for the Kogan LX-8 Vacuum. 
Very similar to the Kogan LX-10, so i am guessing that the error messages are the same. I can remove those if that is preferred.

### DIFF
--- a/custom_components/tuya_local/devices/kogan_lx8_vacuum.yaml
+++ b/custom_components/tuya_local/devices/kogan_lx8_vacuum.yaml
@@ -1,0 +1,323 @@
+name: Kogan LX8 Robot Vacuum
+products:
+  - id: 1bgkj0zbhbw7fppy
+    name: Kogan LX8 Robot Vacuum
+primary_entity:
+  entity: vacuum
+  dps:
+    - id: 101
+      name: activate # clean_switch
+      type: boolean
+    - id: 102
+      name: pause # pause_switch
+      type: boolean
+    - id: 104
+      name: command # clean_mode
+      type: string
+      optional: true
+      mapping:
+        - value: pause
+        - value_redirect: pause
+        - dps_val: smart
+          value: smart
+        - dps_val: pose
+          value: clean_spot
+        - dps_val: zone
+          value: zone
+        - dps_val: backcharge
+          value: return_to_base
+    - id: 105
+      name: status # robot_state
+      type: string
+      mapping:
+        - dps_val: idle
+          value: standby
+        - dps_val: pointing\areaing
+          value: pointing_areaing
+        - dps_val: totaling
+          value: cleaning
+        - dps_val: sweep
+          value: sweep
+        - dps_val: mop
+          value: mop
+        - dps_val: fault
+          value: error
+        - dps_val: pause
+          value: paused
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/kogan_lx8_vacuum.yaml
```diff
@@ -0,0 +1,323 @@
...
+      type: boolean
+    - id: 102
+      name: pause # pause_switch
+      type: boolean
+    - id: 104
+      name: command # clean_mode
+      type: string
+      optional: true
+      mapping:
+        - value: pause
```
make-all, line 19: This does not seem right.  What are you trying to do here, and why did you not just copy the config from LX10 for this dp?

custom_components/tuya_local/devices/kogan_lx8_vacuum.yaml
```diff
@@ -0,0 +1,323 @@
...
+        - dps_val: chargring
+          value: charging
+        - dps_val: tocharge
+          value: returning
+        - dps_val: fullcharge
+          value: charged
+        - dps_val: remotectl
+          value: remotectl
+    - id: 106
+      name: battery
```
make-all, line 56: battery as an attribute on vacuums is deprecated since last July, and should be replaced by a battery sensor.
https://developers.home-assistant.io/blog/2023/07/10/vacuum-updates/

make-all (commented):
The only difference see here is in the fan speeds.  I wonder if that is in fact a problem in the current config rather than a difference between the two models, as it seems strange that they would go to the trouble of keeping everything else the same, but use a different value for just one of the three fan speeds.

ewandank:
> The only difference I see here is in the fan speeds. I wonder if that is in fact a problem in the current config rather than a difference between the two models, as it seems strange that they would go to the trouble of keeping everything else the same, but use a different value for just one of the three fan speeds.
Not sure, unfortunately i don't have an lx-10 to verify, however this is the output from tinytuya:
```
{
    "109": {
        "code": "fan_mode",
        "type": "Enum",
        "values": {
            "range": [
                "quiet",
                "auto",
                "strong"
            ]
        }
    }
}
```

ewandank:
I've had a closer look and it looks like i had an older version checked out, my apologies. 
It's now the same as lx-10 with a few fields marked as optional and the fanmode difference. I have pasted the output of tinytuya below.  If there's a neater way to combine the lx-8 and lx-10 please let me know. 
```json
            "109": {
                "code": "funmode",
                "type": "Enum",
                "values": {
                    "range": [
                        "quiet",
                        "standard",
                        "strong"
                    ]
                }
                
```

custom_components/tuya_local/devices/kogan_lx8_vacuum.yaml
ewandank: Have copied the lx10 now

custom_components/tuya_local/devices/kogan_lx8_vacuum.yaml
ewandank: copied the lx10 now

PR #2036, opened 2024-06-22
Update kogan_bladeless_fan.yaml

sambarlick:
Translation only key was preventing the fans from being added to the integration.


make-all:
This doesn't seem to be the correct way to fix a bug, it is just a workaround.

What is the error that is preventing the fan from being added when this is set?

sambarlick:
The devices aren't picked up by the integration. Removing that line allowed
the integration to identify the devices.

I have no idea what that line does or why it prevented the integration from
identifying the devices.
Regards,
Sam Myers

make-all:
It seems likely to be a case of correlation, not causation if there are no log messages available.

PR #2037, opened 2024-06-22
Added a missing DP

almirus:


custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -370,4 +370,20 @@ secondary_entities:
     dps:
       - id: 103
         type: boolean
+  - entity: sensor
+    name: Power
```
make-all, line 374: Remove the name, as the class already defines a default name with translations

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -370,4 +370,20 @@ secondary_entities:
...
+    class: power
+    category: diagnostic
+    dps:
+      - id: 116
+        type: integer
+        name: sensor
+        precision: 0
```
make-all, line 381: precision should not be needed, or at least allow 1 decimal place, as a divisor of 30 should give an accurate first decimal place.

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -370,4 +370,20 @@ secondary_entities:
...
+        unit: W
+        class: measurement    
+        mapping:
+          - scale: 30
+        range:
```
make-all, line 386: Range is not needed for sensors (unless they are inverted)

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -370,4 +370,20 @@ secondary_entities:
...
+          min: 0
+          max: 1000000     
         name: sensor
```
make-all, line 389: This name belongs to the sensor above, and needs to move back up there, otherwise the sensor above is missing a sensor dp, and this one has a duplicate name.

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
almirus: No, it results in a number with many digits after the decimal point. (The sensor itself is not very accurate, so I decided to show only int)

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -371,19 +371,14 @@ secondary_entities:
       - id: 103
         type: boolean
```
make-all, line 372: The `name: sensor` which you removed from below needs to be inserted back here, as otherwise this dp is missing a name.

almirus: don't understand

Issue #2027, opened 2024-06-20
Request support for Pet Feeder


pergolafabio:
Hi, requesting device support for a Cat Pet Feeder, as discussed before, i wasnt able to add the device

![image](https://github.com/make-all/tuya-local/assets/43251136/288ae709-496c-4d4b-b573-587737b0e83d)

-->

# Log Message

```
2024-06-20 11:53:12.822 WARNING (MainThread) [custom_components.tuya_local.config_flow] Could not find device: XXXX
2024-06-20 11:53:40.547 ERROR (MainThread) [custom_components.tuya_local.device] Failed to refresh device state for Test.
```

# Information about DPS mappings

```
{
  "result": {
    "model": "{\"modelId\":\"000004ajdj\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"默认服务\",\"properties\":[
{\"abilityId\":101,\"accessMode\":\"rw\",\"code\":\"basic_indicator\",\"description\":\"设备指示灯是否打开，true打开，false关闭\",\"name\":\"指示灯\",\"typeSpec\":{\"type\":\"bool\"}},
{\"abilityId\":103,\"accessMode\":\"rw\",\"code\":\"basic_flip\",\"description\":\"true反转，false正常\",\"name\":\"录制画面翻转\",\"typeSpec\":{\"type\":\"bool\"}},
{\"abilityId\":104,\"accessMode\":\"rw\",\"code\":\"basic_osd\",\"description\":\"true打开水印，false关闭水印\",\"name\":\"视频osd功能\",\"typeSpec\":{\"type\":\"bool\"}},
{\"abilityId\":106,\"accessMode\":\"rw\",\"code\":\"motion_sensitivity\",\"description\":\"0-2，灵敏度依次增加；仅为灵敏度，0并不是关闭移动侦测报警；规定0为低灵敏度，1为中灵敏度，2为高灵敏度。\",\"name\":\"移动侦测报警灵敏度\",\"typeSpec\":{\"type\":\"enum\",\"range\":[\"0\",\"1\",\"2\"]}},
{\"abilityId\":108,\"accessMode\":\"rw\",\"code\":\"basic_nightvision\",\"description\":\"0:自动\\n1:关\\n2:开\",\"name\":\"红外夜视\",\"typeSpec\":{\"type\":\"enum\",\"range\":[\"0\",\"1\",\"2\"]}},
[...]

PR #2038, opened 2024-06-22
Add support for V330 pet feeder

pergolafabio:


make-all:
As mentioned on #2027, it is currently not possible to match a device that does not output any data, so this will need to wait until that restriction has been overcome by matching product id based on local or cloud discovery.

Issue #2029, opened 2024-06-20
Request support for Karst-400 Solar Inverter


Moutmout31:
# Log Message

```

2024-06-20 15:31:47.067 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches hama_radiator_controller with quality of 30%. DPS: {"updated_at": 1718890297.2822084, "2": 3843, "10": 277734, "15": "Karst-400", "16": "A8.2023.0705", "17": "ES20230705E01A", "18": 55, "101": true, "102": 3018, "103": 2777, "105": 100}

```

# Information about DPS mappings

```
{
	"2":"Total",
	"3":"DC Input",
	"7":"AC Grid",
	"10":"Total power",
	"11":"Alarm",
	"12":"Work mode",
	"13":"Inverter alarm set1",
	"14":"Inverter alarm set2",
	"15":"Inverter type",
	"16":"Inverter ID",
	"17":"imei imsi",
	"18":"Inverter temperature",
	"101":"开关机",
	"102":"光伏功率",
	"103":"交流功率",
	"104":"电量清零",
	"105":"逆变器功率调整"
}


model:
{
    modelId:f4da3c,
    services:
    [{
        actions:[],
	code:,
        description:,
	events:[],
        name:默认服务,
        properties: [
            {abilityId:2,accessMode:rw,code:reverse_energy_total,description:,extensions:{iconName:icon-dp_lightning,attribute:1152},name:反向总有功电量,typeSpec:{type:value,max:99999999,min:0,scale:2,step:1,unit:kW·h}},
            {abilityId:3,accessMode:ro,code:pv1_dc_data,description:1，PV1直流侧电压、电流、功率\\n2，大端模式，HEX格式，共6个字节\\n3，单位精度：电压，2字节，单位0.1V;电流，2字节，单位 0.1A ;功率，2字节，单位0.01kW\\n4，报文格式\\n举例：08 80 00 0B 03 F3 表示PV1侧电压217.6V，电流1.1A，功率10.11KW\\n5，通信逻辑：\\nA情形：配网成功时，MCU上报该数据。\\nB情形：MCU周期定时上报该值。建议：在WIFI模式时，1分钟上报一次。NB模式时，1个小时上报一次该数据。,extensions:{iconName:icon-dp_mode,attribute:1152},name:直流输入,typeSpec:{type:raw,maxlen:128}},
[...]

PR #2043, opened 2024-06-24
Add support for Avidsen Soria 400W Solar Inverter

Moutmout31:
This PR adds support for Avidsen Soria 400W solar inverter (plug & play kit).
This address issue https://github.com/make-all/tuya-local/issues/2029
PS: this is my first ever PR, so feel free to give feedback.

### DIFF
--- a/custom_components/tuya_local/devices/avidsen_soria_solarinverter.yaml
+++ b/custom_components/tuya_local/devices/avidsen_soria_solarinverter.yaml
@@ -30,7 +30,6 @@ secondary_entities:
       - id: 101
         type: boolean
         name: switch
-        # force: true
   - entity: number
     name: Power Adjustement
     category: config
@@ -63,7 +62,7 @@ secondary_entities:
         type: integer
         name: sensor
         unit: W
-        force: true
+        #force: true
         class: measurement
         mapping:
           - scale: 10
@@ -76,7 +75,7 @@ secondary_entities:
         type: integer
         name: sensor
         unit: W
-        force: true
+        #force: true
         class: measurement
         mapping:
           - scale: 1000
@@ -89,7 +88,7 @@ secondary_entities:
         type: integer
         name: sensor
         unit: W
-        force: true
+        #force: true
         class: measurement
         mapping:
           - scale: 10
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/avidsen_soria_solarinverter.yaml
```diff
@@ -0,0 +1,183 @@
...
+            value: true
+  - entity: sensor
+    name: AC Power
+    class: power
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        unit: W
+        force: true
```
make-all, line 66: DP is not optional, does it really require force? Normally DP that require force will not show up by default, so need to also be optional.

custom_components/tuya_local/devices/avidsen_soria_solarinverter.yaml
```diff
@@ -0,0 +1,183 @@
...
+      - id: 7
+        type: base64
+        name: sensor
+        unit: W
+        optional: true
+        mapping:
+          - mask: "0000000000FFFFFF0000"
+            scale: 1000
+  - entity: sensor
+    name: Temperature
```
make-all, line 175: Remove the name here, to use the class default translations

custom_components/tuya_local/devices/avidsen_soria_solarinverter.yaml
Moutmout31: OK, I wasn't sure about that. I've changed it in my HA instance, but device has just gone offline because of night coming. I'll check tomorrow it is OK.

custom_components/tuya_local/devices/avidsen_soria_solarinverter.yaml
Moutmout31: OK, thanks for the explanation.

custom_components/tuya_local/devices/avidsen_soria_solarinverter.yaml
Moutmout31: Tested and OK, I also checked again device recognicion.

Issue #1103, opened 2023-09-18
Request support for Atorch (S1WP) device Smart Plug Energy meter


jnasserifar:
Could you please add Atorch (S1WP) to the support device list please.

I tried to add it myself but when I add yaml file in adding device I receive Unknown error. when I remove the file the suggested devices to choose are not energy meters!

Thank you in advance!

From Tuya iot API explorer:

    "model": "S1WP",
    "name": "Smart Energy Meter",
    "product_id": "sqrf2g1amfutn4co",
    "product_name": "炬为电能宝-计控型(S1WP)",

{
  "result": {
    "properties": [
      {
        "code": "switch_1",
        "custom_name": "",
        "dp_id": 1,
        "time": 1691582860910,
        "value": true
      },
      {
        "code": "countdown_1",
        "custom_name": "",
        "dp_id": 9,
        "time": 1695000549682,
        "value": 0
      },
      {
        "code": "add_ele",
        "custom_name": "",
        "dp_id": 17,
        "time": 1695003267591,
        "value": 5
      },
      {
        "code": "cur_current",
        "custom_name": "",
        "dp_id": 18,
        "time": 1695005887553,
        "value": 830
      },
      {
        "code": "cur_power",
        "custom_name": "",
        "dp_id": 19,
        "time": 1695005887568,
        "value": 6613
      },
      {
        "code": "cur_voltage",
        "custom_name": "",
        "dp_id": 20,
        "time": 1695005887514,
        "value": 12385
      },
      {
        "code": "price",
        "custom_name": "",
        "dp_id": 101,
        "time": 1695000549682,
        "value": 73
      },
      {
        "code": "cost",
        "custom_name": "",
        "dp_id": 102,
        "time": 1695005872416,
        "value": 51192
      },
      {
        "code": "add_cost",
        "custom_name": "",
        "dp_id": 103,
        "time": 1695004926491,
        "value": 5
      },
      {
        "code": "ovp",
        "custom_name": "",
        "dp_id": 104,
        "time": 1695000549682,
        "value": 2650
      },
      {
        "code": "ocp",
        "custom_name": "",
        "dp_id": 105,
        "time": 1695000549682,
        "value": 1000
      },
      {
        "code": "opp",
        "custom_name": "",
        "dp_id": 106,
        "time": 1695000549682,
        "value": 3000
      },
      {
        "code": "language",
        "custom_name": "",
        "dp_id": 107,
        "time": 1695000549682,
        "value": "english"
      },
[...]

PR #2049, opened 2024-06-25
device atorch_s1wp add update interval entity

Lxeon:
this entity can set the update interval of the device, default is 60 seconds

but if set it less than 60 seconds, it will be reset to 60 seconds after 5 minutes.  it can use automate script for keep 1 second

### DIFF
--- a/custom_components/tuya_local/devices/atorch_s1wp.yaml
+++ b/custom_components/tuya_local/devices/atorch_s1wp.yaml
@@ -531,3 +531,14 @@ secondary_entities:
             value: "Off"
           - dps_val: memory
             value: Memory
+  - entity: number
+    category: config
+    name: Update interval
+    icon: "mdi:update"
+    dps:
+      - id: 139
+        type: integer
+        name: value
+        range:
+          min: 1
+          max: 90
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/atorch_s1wp.yaml
```diff
@@ -531,3 +531,15 @@ secondary_entities:
             value: "Off"
           - dps_val: memory
             value: Memory
+  - entity: number
+    category: config
+    name: Update interval
+    icon: "mdi:update"
+    dps:
+      - id: 139
+        type: integer
```
make-all, line 540: I think this requires `optional: true`. DP 139 does not appear in the log posted to #1103

Lxeon: added

pivanet: is it already available on the release?

Lxeon: > oh... so i have to use automation to prevent it going back to original 60 seconds, right?

yes, you can make trigger like this:
```
platform: state
entity_id:
  - number.ju_wei_dian_neng_bao_ji_kong_xing_s1wp_update_interval
to: "60"
from: null
```

PR #2062, opened 2024-06-29
Create bauhn_5way_powerstrip_usb_AP5W-0624.yaml

lukegackle:
Adding support for the BAUHN 5-way powerboard AP5W-0624, this is a tested working config with this device.


custom_components/tuya_local/devices/bauhn_5way_powerstrip_usb_AP5W-0624.yaml
```diff
@@ -0,0 +1,214 @@
...
+        name: value
+        unit: min
+        range:
+          min: 0
+          max: 86400
+        mapping:
+          - scale: 60
+            step: 60
+  - entity: number
+    name: Timer 4
```
make-all, line 108: Should this be Timer 5?

custom_components/tuya_local/devices/bauhn_5way_powerstrip_usb_AP5W-0624.yaml
```diff
@@ -0,0 +1,214 @@
...
+        name: value
+        unit: min
+        range:
+          min: 0
+          max: 86400
+        mapping:
+          - scale: 60
+            step: 60
+  - entity: sensor
+    name: Current
```
make-all, line 123: Name is not required when it is the same as class

custom_components/tuya_local/devices/bauhn_5way_powerstrip_usb_AP5W-0624.yaml
```diff
@@ -0,0 +1,214 @@
...
+        type: integer
+        class: measurement
+        force: true
+        unit: mA
+      - id: 23
+        type: integer
+        name: calibration
+        optional: true
+  - entity: sensor
+    name: Power
```
make-all, line 138: name not required

custom_components/tuya_local/devices/bauhn_5way_powerstrip_usb_AP5W-0624.yaml
```diff
@@ -0,0 +1,214 @@
...
+        force: true
+        unit: W
+        mapping:
+          - scale: 10
+      - id: 24
+        type: integer
+        name: calibration
+        optional: true
+  - entity: sensor
+    name: Voltage
```
make-all, line 155: name not required

custom_components/tuya_local/devices/bauhn_5way_powerstrip_usb_AP5W-0624.yaml
```diff
@@ -0,0 +1,214 @@
...
+            value: Last state
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 41
+        type: boolean
+        name: lock
+        optional: true
+  - entity: sensor
```
make-all, line 194: I don't think these last three should be sensors. Attributes on another entity (Perhaps Timer 1, since they are timer related) is enough to make them available without having them cluttering the default UI and entity list.

custom_components/tuya_local/devices/bauhn_5way_powerstrip_usb_AP5W-0624.yaml
```diff
@@ -0,0 +1,214 @@
+name: BAUHN 5-way powerstrip with USB
```
make-all, line 1: This name should be without branding.  Also, I only see 5 switches and no timers, so is it 4 way powerstrip with USB, or 5 way without a controlled USB?

custom_components/tuya_local/devices/bauhn_5way_powerstrip_usb_AP5W-0624.yaml
lukegackle: @make-all Thanks very much for the feedback, I've updated the other changes but wasn't sure what the correct syntax would be to make these ones as attributes of timer 1, is there any guidance you can provide on this one?

custom_components/tuya_local/devices/bauhn_5way_powerstrip_usb_AP5W-0624.yaml
lukegackle: Just to clarify this is a 5 way power board with 5 power outlets that are individually controllable, the USB ports are not controllable. I have removed the brand, and also the USB part from the name as it's irrelevant for the purpose of this config.

custom_components/tuya_local/devices/bauhn_5way_powerstrip_usb_AP5W-0624.yaml
make-all: Something like the way these three are handled in [aubess_2gang_switch](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/aubess_2gang_switch.yaml#L12) for example (there they are on switch 1, mainly because the primary entity usually ends up with left overs, but since these are all timing related, putting them on a timer also makes sense).  As long as you avoid names with special meaning for the entity, extra attributes just end up as read-only values in the state attributes for the entity.

lukegackle: @make-all Thanks very much, I have updated the YAML to have those three sensors as attributes of timer 1 and all seems to be working 😊

Issue #1856, opened 2024-04-29
All my switch entity names have None appended to name.


RenierM26:
**Describe the bug**
All my switch entities have None Appended to their names.

Example: Combo bedroom outlet 1 None

**To Reproduce**
All switch entities have None appended to their names.

**Expected behavior**
Switch entities should not have none appened to their name. I'm guessing the None is set a string somewhere by accident.

**Additional context**
Started about 2 - 3 releases back. 

If the bug involves a device, then please include device diagnostics from
Settings / Devices & Services / Tuya Local (your device) / "1 device":

  "integration_manifest": {
    "domain": "tuya_local",
    "name": "Tuya Local",
    "codeowners": [
      "@make-all"
    ],
    "config_flow": true,
    "dependencies": [],
    "documentation": "https://github.com/make-all/tuya-local",
    "integration_type": "device",
    "iot_class": "local_push",
    "issue_tracker": "https://github.com/make-all/tuya-local/issues",
    "requirements": [
      "tinytuya==1.13.2"
    ],
    "version": "2024.4.0",
    "is_built_in": false
  },
  "data": {
    "name": "Combo bedroom outlet 1",
    "type": "cbi_astute_outdoor_smartswitch",
    "device_id": "**REDACTED**",
    "device_cid": "",
    "local_key": "**REDACTED**",
    "host": "**REDACTED**",
    "protocol_version": 3.3,
    "tinytuya_version": "1.13.2",
    "api_version_set": 3.3,
    "api_version_used": 3.3,
    "api_working": true,
    "status": {},
    "cached_state": {
      "updated_at": 1714368864.6771305,
      "1": true,
      "7": 0,
      "14": "on",
      "20": 6,
      "21": 109,
      "22": 2386,
      "23": 139
    },
    "pending_state": {},
    "connected": true,
    "force_dps": [
      20,
      21,
      23,
      22
    ],
    "home_assistant": {
      "name": "Combo bedroom outlet 1",
      "name_by_user": null,
      "disabled": false,
      "disabled_by": null,
      "entities": [
        {
          "disabled": false,
          "disabled_by": null,
          "entity_category": null,
          "device_class": "outlet",
          "original_device_class": "switch",
          "icon": null,
          "unit_of_measurement": null,
          "state": {
            "entity_id": "switch.combo_bedroom_outlet_1",
            "state": "on",
            "attributes": {
              "device_class": "outlet",
              "friendly_name": "Combo bedroom outlet 1 None"
            },
[...]

PR #2069, opened 2024-07-01
Add support for PlantsIO Ivy smart planter

thewade:
This PR adds the PlantsIO "Ivy" Smart Planter.

Info on the device:
https://www.indiegogo.com/projects/plantsio-ivy-smart-flowerpots-endless-fun#/
https://smartplantivy.com/

Available on Amazon and AliExpress. Lots of drop shippers using various brand names.

Datapoints documented at https://gist.github.com/thewade/ef9f6014f13932bd2e77d43331e2027d

This device has a ton of sensors and config options. I haven't got them all the config done, and some enum data needs to be reverse engineered to be mapped.

Some settings like brightness should be restricted when auto brightness is on, but I wasn't able to figure out if it was possible using constraints and conditions.

I expect another PR in the future to fill in the missing details as well as constraints and conditions if they are possible, but I need a break for now.

### DIFF
--- a/custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
+++ b/custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
@@ -8,6 +8,7 @@ products:
 primary_entity:
   entity: sensor
   name: Water level
+  class: volume_storage
   dps:
     - id: 108
       type: integer
@@ -22,7 +23,7 @@ primary_entity:
       type: string
       name: version
 secondary_entities:
-  # Sensor entities
+  # Sensor entities raw
   - entity: sensor
     name: Ambient light
     class: illuminance
@@ -61,19 +62,22 @@ secondary_entities:
         type: integer
         optional: true
         unit: "%"
+  # Sensor entities intelligent
   - entity: sensor
     name: Water status
     icon: "mdi:watering-can"
     class: enum
     dps:
       - id: 106
-        type: bitfield
+        type: integer
         name: sensor
         mapping:
           - dps_val: 0
             value: "Initializing"
           - dps_val: 2
             value: "Drinking"
+          - dps_val: 3
+            value: "Absorbing from soil"
           - value: "Unknown state"
   - entity: sensor
     name: Light status
@@ -81,9 +85,11 @@ secondary_entities:
     class: enum
     dps:
       - id: 138
-        type: bitfield
+        type: integer
         name: sensor
         mapping:
+          - dps_val: 0
+            value: "Initializing"
           - dps_val: 1
             value: "Acceptable"
           - dps_val: 2
@@ -101,9 +107,11 @@ secondary_entities:
     class: enum
     dps:
       - id: 139
-        type: bitfield
+        type: integer
         name: sensor
         mapping:
+          - dps_val: 0
+            value: "Initializing"
           - dps_val: 1
             value: "Good"
           - dps_val: 2
@@ -115,13 +123,15 @@ secondary_entities:
           - value: "Unknown state"
   - entity: sensor
     name: Humidity status
-    class: enum
     icon: "mdi:water-percent"
+    class: enum
     dps:
       - id: 140
-        type: bitfield
+        type: integer
         name: sensor
         mapping:
+          - dps_val: 0
+            value: "Initializing"
           - dps_val: 1
             value: "Good"
           - value: "Unknown state"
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
```diff
@@ -0,0 +1,431 @@
+# Datapoints documented at:
+# https://gist.github.com/thewade/ef9f6014f13932bd2e77d43331e2027d
+
+name: Smart Planter
+products:
+  - id: 8n3q0y4cwov8ifxt
+    name: PlantsIO Ivy Smart Planter
+primary_entity:
+  entity: sensor
+  name: Water level
```
make-all, line 10: `class: volume_storage`

custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
```diff
@@ -0,0 +1,431 @@
...
+        - dps_val: 0
+          icon: "mdi:cup-outline"
+        - icon: "mdi:cup"
+    - id: 107
+      type: string
+      name: version
+secondary_entities:
+  # Sensor entities
+  - entity: sensor
+    name: Ambient light
```
make-all, line 27: Suggest to remove this name, so that pre-defined translations can be used.

custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
```diff
@@ -0,0 +1,431 @@
...
+    class: illuminance
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        unit: lx
+        class: measurement
+  - entity: sensor
+    name: Temperature
```
make-all, line 36: Definitely remove the names that are the same as class

custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
```diff
@@ -0,0 +1,431 @@
...
+          - dps_val: 2
+            value: "Hot"
+            icon: "mdi:thermometer-high"
+          - dps_val: 3
+            value: "Cold"
+            icon: "mdi:thermometer-low"
+          - value: "Unknown state"
+  - entity: sensor
+    name: Humidity status
+    class: enum
```
make-all, line 118: Sensors with only two states could maybe be made into binary_sensor instead, especially if there is a relevant class defined for binary_sensor.

custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
```diff
@@ -0,0 +1,431 @@
...
+    name: Battery
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 121
+        type: integer
+        name: sensor
+        unit: "%"
+  - entity: sensor
+    name: Charge state
```
make-all, line 139: This can definitely be a binary_sensor (`class: battery_charging`), since the original value is boolean even

custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
```diff
@@ -0,0 +1,431 @@
...
+  - entity: switch
+    name: Auto brightness
+    category: config
+    icon: "mdi:brightness-auto"
+    dps:
+      - id: 135
+        type: boolean
+        name: switch
+  - entity: number
+    name: Brightness
```
make-all, line 324: Is this a light?

custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
thewade: When I do that all my sensors have no names:

image

And I end up getting entity IDs like:

sensor.smart_planter
sensor.smart_planter_2
sensor.smart_planter_3

Am I doing something wrong?

This is my current code for the illuminance sensor:

```
  - entity: sensor
    class: illuminance
    dps:
      - id: 103
        type: integer
        name: sensor
        unit: lx
        class: measurement
```

custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
thewade: I would expect there to be multiple states from the "AI" derived sensors. 

e.g. Good, Dry, Too humid, Too dry

However I have no been able to determine all the values in my environment, and as I see with Light status enum the values are not in order so I don't want to guess.

The developers seems to had expressed interested in home assistant status, so I have emailed them to see if they are willing to share the mapping.

custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
thewade: This is the display brightness.

thewade:
I have added a commit with the following changes:

- fix most enums to use integer instead of bitfield (some states were not mapping until I fixed it)
- use volume_storage class for water level
- add absorbing state to water status
- add initializing states from missing enums
- use binary_sensor entity and battery_charging class for charging status
- convert plant in pot to binary_sensor
- add not selected enum for plant type
- add config options for weather and weather location

I haven't made the change regarding the names of known class types due the issue I am seeing and I will wait for your feedback.

thewade:
> * convert plant in pot to binary_sensor

There seems to be some issue with this change as I now have an error in the logs:

```
This error originated from a custom integration.

Logger: custom_components.tuya_local.sensor
Source: custom_components/tuya_local/sensor.py:59
integration: Tuya Local (documentation, issues)
First occurred: 10:52:57 AM (2155 occurrences)
Last logged: 5:24:13 PM

plantios_ivy_smart_planter.yaml/Plant in pot: Unrecognized sensor device class of problem ignored
```

The entry:
```
  - entity: sensor
    name: Plant in pot
    class: problem
    category: diagnostic
    dps:
      - id: 102
        type: boolean
        name: sensor
```

custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
make-all: This is maybe a recently introduced bug due to a side effect of the fix for #1856.  It may be better for me to revert that change.
Unfortunately this was difficult to detect, as existing entity names do not change once they are setup, only newly added ones are affected.

PR #2087, opened 2024-07-07
Add MOES Motion Sensor Smart Switch

levineds:
Note, this is different from MOES Motion Sensor Light


make-all:
Sorry, the pytest tests are a single test for the config, so they stop at the first error. I think changing the "option" to "value" in all your number entities will be the last one though.

levineds:
I appreciate the tests existing in the first place. Sorry for not understanding the documentation better.On Jul 7, 2024, at 10:16 PM, Jason Rumney ***@***.***> wrote:﻿
Sorry, the pytest tests are a single test for the config, so they stop at the first error. I think changing the "option" to "value" in all your number entities will be the last one though.

custom_components/tuya_local/devices/moes_motionsensor_switch.yaml
```diff
@@ -0,0 +1,127 @@
...
+    dps:
+      - id: 49
+        name: value
+        type: integer
+        unit: s
+        range:
+          min: 5
+          max: 3600
+  - entity: number
+    name: Appointment Time
```
make-all, line 107: More obvious naming is needed for these two times.  I would never guess that "Appointment" is the end of allowed hours, and "Target" is the beginning.  Maybe "Sensing start time" and "Sensing end time" would be better.
I would also add the following to make the display in hours rather than minutes, so for example 7pm shows as "19.0" rather than "1140".  7:30 will show as "19.5", which isn't ideal, but I think still better than "1170".
```
          mapping:
            - scale: 60
```

make-all (commented):
In general, entity names should follow HA naming guidelines, and only capitalise the first word.

PR #2093, opened 2024-07-09
Beca ZHT002: Added current temperature diagnostic

pandatautau:
Added datapoint for current temperature (for use with custom thermostat cards)


make-all:
There is already a climate entity with the current temperature embedded.

PR #2094, opened 2024-07-09
Added support for Arlec air purifier APR005HA

horse-pants:
image

### DIFF
--- a/custom_components/tuya_local/devices/arlec_apr005ha_air_purifier.yaml
+++ b/custom_components/tuya_local/devices/arlec_apr005ha_air_purifier.yaml
@@ -1,0 +1,100 @@
+name: Air Purifier
+products:
+  - id: dfgsetrdnabqeimr
+    name: Arlec Air Purifier CFR225HA
+primary_entity:
+  entity: fan
+  translation_only_key: fan_with_presets
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+    - id: 3
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: Auto
+          value: smart
+        - dps_val: Manual
+          value: normal
+        - dps_val: Sleep
+          value: sleep
+    - id: 4
+      type: integer
+      name: speed
+      range:
+        min: 1
+        max: 3
+secondary_entities:
+  - entity: switch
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/arlec_apr005ha_air_purifier.yaml
```diff
@@ -0,0 +1,101 @@
...
+          - dps_val: 9h
+            value: 9 hours
+          - dps_val: 10h
+            value: 10 hours
+          - dps_val: 11h
+            value: 11 hours
+          - dps_val: 12h
+            value: 12 hours
+  - entity: sensor
+    name: Air quality
```
make-all, line 85: It may be better to leave the name out so that translations can be used.

custom_components/tuya_local/devices/arlec_apr005ha_air_purifier.yaml
```diff
@@ -0,0 +1,101 @@
...
+    class: pm25
+    dps:
+      - id: 2
+        type: integer
+        name: sensor
+        unit: µg/m³
+  - entity: sensor
+    name: Total time
```
make-all, line 93: Total what time?  Maybe "Running time" is better, or just leave the name off and use the translations for "duration" class. Or if it is the time left for the timer, `translation_key: time_remaining` may be more suitable.

horse-pants: Yeah, running time, I'll take the names out and use the translations

custom_components/tuya_local/devices/arlec_apr005ha_air_purifier.yaml
horse-pants: Sounds good, cheers

horse-pants: ah that's right, without the names everything becomes air quality, should I put the name back in or is there a better way to do it?
image

make-all: I think that was a side effect of an earlier fix that is now reverted.

Issue #2097, opened 2024-07-10
Request support for Casdon TD Pro 2 device


jianyun8023:
I will create configuration files for this device.

# Log Message

```
2024-07-10 09:30:52.594 DEBUG (MainThread) [custom_components.tuya_local.device] Refreshing device state for Test
2024-07-10 09:30:52.594 INFO (MainThread) [custom_components.tuya_local.device] Setting protocol version for Test to 3.3
2024-07-10 09:30:52.619 DEBUG (SyncWorker_48) [custom_components.tuya_local.device] Test refreshed device state: {"dps": {"1": false, "2": false, "7": 0, "8": 40, "10": 0, "11": 0, "13": 0, "102": "wait", "103": false, "105": "1", "108": 0}}
2024-07-10 09:30:52.619 DEBUG (SyncWorker_48) [custom_components.tuya_local.device] new state (incl pending): {"updated_at": 1720575052.619872, "1": false, "2": false, "7": 0, "8": 40, "10": 0, "11": 0, "13": 0, "102": "wait", "103": false, "105": "1", "108": 0}

```

# Information about DPS mappings

```json
{
  "result": {
    "category": "kx",
    "functions": [
      {
        "code": "switch",
        "desc": "{}",
        "name": "开关",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "start",
        "desc": "{}",
        "name": "启动/停止",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "cook_temperature",
        "desc": "{\"unit\":\"℃\",\"min\":0,\"max\":500,\"scale\":0,\"step\":1}",
        "name": "烹饪温度",
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":0,\"max\":500,\"scale\":0,\"step\":1}"
      },
      {
        "code": "appointment_time",
        "desc": "{\"unit\":\"min\",\"min\":0,\"max\":1440,\"scale\":0,\"step\":1}",
        "name": "预约时间",
        "type": "Integer",
        "values": "{\"unit\":\"min\",\"min\":0,\"max\":1440,\"scale\":0,\"step\":1}"
      },
      {
        "code": "cook_time",
        "desc": "{\"unit\":\"min\",\"min\":0,\"max\":1440,\"scale\":0,\"step\":1}",
        "name": "烹饪时间",
        "type": "Integer",
[...]

PR #2098, opened 2024-07-10
Add support for Casdon TD Pro 2 device

jianyun8023:
Add oven type device: Casdon Oven TD Pro 2.
Cloesd #2097

Preview:
image

### DIFF
--- a/custom_components/tuya_local/devices/casdon_td_pro_2.yaml
+++ b/custom_components/tuya_local/devices/casdon_td_pro_2.yaml
@@ -168,7 +168,7 @@ secondary_entities:
           - dps_val: 4
             value: "E3: Chamber low temperature protection (below 35°C)"
           - dps_val: 8
-            value: "E4: Evaporation tray low temperature protection (below 35°C)"
+            value: "E4: Evaporation tray low temp protection (below 35°C)"
           - dps_val: 16
             value: "E5: Sensor connection error (Open circuit)"
           - dps_val: 32
--- a/custom_components/tuya_local/devices/casdon_td_pro_2.yaml
+++ b/custom_components/tuya_local/devices/casdon_td_pro_2.yaml
@@ -1,0 +1,193 @@
+name: Casdon TD Pro 2
+products:
+  - id: wab6oustdukujpkn
+    name: 蒸烤箱TD-pro二代
+primary_entity:
+  entity: switch
+  icon: "mdi:power"
+  name: Switch
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+
+secondary_entities:
+  - entity: switch
+    name: Start
+    icon: "mdi:play-pause"
+    dps:
+      - id: 2
+        name: switch
+        type: boolean
+  - entity: number
+    name: cook_temperature
+    dps:
+      - id: 7
+        name: value
+        type: integer
+        unit: C
+        range:
+          min: 0
+          max: 500
+  - entity: sensor
+    name: temp_current
+    class: temperature
+    dps:
+      - id: 8
+        name: sensor
+        type: integer
+        unit: C
+        readonly: true
+        range:
+          min: 0
+          max: 500
+  - entity: number
+    name: appointment_time
+    dps:
+      - id: 9
+        name: value
+        type: integer
+        optional: true
+        range:
+          min: 0
+          max: 1440
+        unit: min
+        class: measurement
+  - entity: number
+    icon: "mdi:timer-edit"
+    name: cook_time
+    dps:
+      - id: 10
+        name: value
+        type: integer
+        range:
+          min: 0
+          max: 1440
+        unit: min
+        class: measurement
+  - entity: sensor
+    icon: "mdi:timer"
+    name: remain_time
+    dps:
+      - id: 11
+        name: sensor
+        type: integer
+        readonly: true
+        range:
+          min: 0
+          max: 1440
+        unit: min
+        class: measurement
+  - entity: binary_sensor
+    name: fault
+    category: diagnostic
+    dps:
+      - id: 13
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: OK
+          - dps_val: 1
+            value: "er1：腔体高温保护(腔体传感器温度高于290°C)"
+          - dps_val: 2
+            value: "er2：蒸发盘高温保护(蒸发盘传感器温度高于200°C)"
+          - dps_val: 4
+            value: "er3：腔体低温保护(连续工作5分钟温度低于35°C)"
+          - dps_val: 8
+            value: "er4：蒸发盘低温保护(连续工作5分钟温度低于35°C)"
+          - dps_val: 16
+            value: "er5：传感器连接出故障(腔体或蒸发盘传感器开路)"
+          - dps_val: 32
+            value: "er6：传感器感应出故障(腔体或蒸发盘传感器短路)"
+          - dps_val: 64
+            value: "er7：显示板通信故障"
+          - dps_val: 128
+            value: "er8：水箱开关（CN7）未闭合或者蒸发盘NTC（CN3）温度高于125℃并持续时间达25s后，缺水提示，并停止工作。"
+  - entity: sensor
+    name: cook_status
+    class: enum
+    dps:
+      - id: 102
+        name: sensor
+        type: string
+        readonly: true
+        mapping:
+          - dps_val: wait
+            value: 烹饪等待
+          - dps_val: reservation
+            icon: "mdi:cup-water"
+            value: 预约中
+          - dps_val: cooking
+            icon: "mdi:cup"
+            value: 烹饪中
+          - dps_val: cancel
+            icon: "mdi:cup"
+            value: 烹饪取消
+          - dps_val: done
+            icon: "mdi:cup"
+            value: 烹饪完成
+          - dps_val: pause
+            icon: "mdi:cup"
+            value: 烹饪暂停
+  - entity: switch
+    name: Lamp
+    icon: "mdi:lightbulb-on"
+    dps:
+      - id: 103
+        name: switch
+        type: boolean
+  - entity: select
+    name: all_mode
+    category: config
+    dps:
+      - id: 105
+        name: option
+        type: string
+        mapping:
+        - dps_val: "1"
+          value: 健康蒸
+        - dps_val: "2"
+          value: 鲜嫩蒸
+        - dps_val: "3"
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
+name: Casdon TD Pro 2
```
make-all, line 1: This name should be generic (Oven).  The branding belongs below.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+products:
+  - id: wab6oustdukujpkn
+    name: 蒸烤箱TD-pro二代
```
make-all, line 4: If this is sold in China market only, then a Chinese name is OK here, but if it is sold outside China also, it may be better as "Casdon TD Pro 2" so the user does not need to read Chinese to recognise it.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+primary_entity:
+  entity: switch
+  icon: "mdi:power"
+  name: Switch
```
make-all, line 8: name is not needed

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+        name: value
+        type: integer
+        range:
+          min: 0
+          max: 1440
+        unit: min
+        class: measurement
+  - entity: sensor
+    icon: "mdi:timer"
+    name: remain_time
```
make-all, line 70: Use `translation_key: time_remaining` instead of name and icon.
And add `class: duration`, for better formatting of HA display.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+          - dps_val: cancel
+            icon: "mdi:cup"
+            value: 烹饪取消
+          - dps_val: done
+            icon: "mdi:cup"
+            value: 烹饪完成
+          - dps_val: pause
+            icon: "mdi:cup"
+            value: 烹饪暂停
+  - entity: switch
```
make-all, line 146: I think it is better to use `entity: light` and no name.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+
+secondary_entities:
+  - entity: switch
+    name: Start
+    icon: "mdi:play-pause"
+    dps:
+      - id: 2
+        name: switch
+        type: boolean
+  - entity: number
```
make-all, line 22: I think this may be better as a climate entity, as it seems most other ovens supported by HA expose themselves this way, and users expect a thermostat-like interface for their oven. The climate entity may cover dps 2 (hvac_mode), 7 (temperature) and 8 (current_temperature), and maybe also dp 105 (preset_mode).
You could also convert dp 102 values to "heating", "idle", "off" hvac_action, but since it would only give partial information, you need to also keep the sensor in that case.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+      - id: 8
+        name: sensor
+        type: integer
+        unit: C
+        readonly: true
+        range:
+          min: 0
+          max: 500
+  - entity: number
+    name: appointment_time
```
make-all, line 45: I would use `translation_key: timer` here, or if it is wall clock time rather than a "from now" delay, a name like "Start time".

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+        type: integer
+        optional: true
+        range:
+          min: 0
+          max: 1440
+        unit: min
+        class: measurement
+  - entity: number
+    icon: "mdi:timer-edit"
+    name: cook_time
```
make-all, line 58: Name should be human readable (not with underscores), such as "Cooking time"

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+    dps:
+      - id: 13
+        type: bitfield
+        name: sensor
+        mapping:
+          - value: OK
+          - dps_val: 0
+            value: OK
+          - dps_val: 1
+            value: "er1：腔体高温保护(腔体传感器温度高于290°C)"
```
make-all, line 94: Again, unless this is really only Chinese market only, these would be better translated to English so more people can read them.  Maybe this does not need to be a sensor, but can instead be an extra attribute on the problem binary_sensor below.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+          - dps_val: 16
+            value: "er5：传感器连接出故障(腔体或蒸发盘传感器开路)"
+          - dps_val: 32
+            value: "er6：传感器感应出故障(腔体或蒸发盘传感器短路)"
+          - dps_val: 64
+            value: "er7：显示板通信故障"
+          - dps_val: 128
+            value: "er8：水箱开关（CN7）未闭合或者蒸发盘NTC（CN3）温度高于125℃并持续时间达25s后，缺水提示，并停止工作。"
+  - entity: binary_sensor
+    name: fault
```
make-all, line 110: Instead of name, use `class: problem` for HA to display appropriately.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+    dps:
+      - id: 13
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+  - entity: sensor
+    name: cook_status
```
make-all, line 121: Name can be just "Status", I think.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -0,0 +1,206 @@
...
+            value: 烹饪暂停
+  - entity: switch
+    name: Lamp
+    icon: "mdi:lightbulb-on"
+    dps:
+      - id: 103
+        name: switch
+        type: boolean
+  - entity: select
+    name: all_mode
```
make-all, line 154: If this remains separate from climate entity, the name can be just "Mode".

make-all (approved):

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
jianyun8023: I searched for product information, and found that this company only sells its products in China, but I feel using English would create a more harmonious feel.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
jianyun8023: Okay

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
jianyun8023: Okay, I’ll remove the `mode`.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
jianyun8023: Okay

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
jianyun8023: I have now translated these descriptions into English, but I would like to have versions in Chinese as well. This product is sold only in China.

jianyun8023:
@make-all I have finished the modifications and have localized some of the descriptions into English. I want to do localization adaptation in Chinese, and I can do this task.

I currently don’t understand how to use the translation_key to configure the fault_message value.

- English: `er1: Chamber Overheat Protection (Sensor temperature exceeds 290°C)`
- Chinese: `er1：腔体高温保护(腔体传感器温度高于290°C)`

```yaml
      - id: 13
        type: bitfield
        name: fault_message
        mapping:
          - value: OK
          - dps_val: 0
            value: OK
          - dps_val: 1
            value: "er1: Chamber Overheat Protection (Sensor temperature exceeds 290°C)"
          - dps_val: 2
            value: "er2: Evaporation Tray Overheat Protection (Sensor temperature exceeds 200°C)"
          - dps_val: 4
            value: "er3: Chamber Low Temperature Protection (Temperature remains below 35°C for 5 minutes)"
          - dps_val: 8
            value: "er4: Evaporation Tray Low Temperature Protection (Temperature remains below 35°C for 5 minutes)"
          - dps_val: 16
            value: "er5: Sensor Connection Error (Open circuit in chamber or evaporation tray sensor)"
          - dps_val: 32
            value: "er6: Sensor Malfunction (Short circuit in chamber or evaporation tray sensor)"
          - dps_val: 64
            value: "er7: Display Board Communication Error"
          - dps_val: 128
            value: "er8: Water Tank Switch (CN7) Not Closed or Evaporation Tray NTC (CN3) Exceeds 125°C for 25 Seconds, Triggering Water Shortage Alert and Stopping Operation"

```

jianyun8023:
@make-all Thank you for your assistance with my pull request.

make-all:
If you want to translate those messages, I suggest using a more specific translation_key, such as oven_fault at the sensor level.   I am not sure that translation handles non-standard extra attributes like "fault_message", so you may need to go back to an enum class sensor just for the message.
Make the values in the config "error_1", "error_2" etc, with the full message moved to the en.json and zh_CN.json files.

jianyun8023:
> If you want to translate those messages, I suggest using a more specific translation_key, such as oven_fault at the sensor level.   I am not sure that translation handles non-standard extra attributes like "fault_message", so you may need to go back to an enum class sensor just for the message.
> 
> Make the values in the config "error_1", "error_2" etc, with the full message moved to the en.json and zh_CN.json files.

@make-all No problem, I will make this modification after completing the merge and localization.

make-all:
I think the translations can be a new PR, I will merge this one now.

PR #2101, opened 2024-07-11
Implement Arlec PC191HA Series 2 Socket

illuzn:
Despite stating the model number suggesting that it is a merely a revision of the original Arlec PC181HA, this is actually a redesign with the CB2S module replacing the WB2S module in the previous series.

The CB2S is newer but less technically capable and as such has quite a few features stripped out.

This PR implements the Series 2 which prevents various non-existent entities from appearing.

### DIFF
--- a/custom_components/tuya_local/devices/arlec_pc191ha_s2_socket.yaml
+++ b/custom_components/tuya_local/devices/arlec_pc191ha_s2_socket.yaml
@@ -1,0 +1,121 @@
+name: Advanced energy monitoring smartplug
+products:
+  - id: auojpnb4hpc13ftb
+    name: Arlec PC191HA Series 2
+primary_entity:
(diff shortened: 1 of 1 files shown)


make-all:
There doesn't seem to be any reason to add a new config for this device rather than use the existing smartplugv2_energy.yaml config and ignore the missing config options.  If we added a config for every single model of smartplug on the market globally, there would be thousands of configs.

illuzn:
> There doesn't seem to be any reason to add a new config for this device rather than use the existing smartplugv2_energy.yaml config and ignore the missing config options. If we added a config for every single model of smartplug on the market globally, there would be thousands of configs.

Understood - makes sense. I was coming at this the other way of users might be confused with all these extraneous configuration options that don't actually work.

PR #2115, opened 2024-07-15
Implement Simplified Chinese localization for the Casdon TD Pro 2.

jianyun8023:
Implement Simplified Chinese localization for the Casdon TD Pro 2.
Added English configuration, other multilingual configurations not added.


custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -158,47 +160,47 @@ secondary_entities:
...
+            value: cooking
           - dps_val: cancel
-            value: Cooking Canceled
+            value: cancel
           - dps_val: done
-            value: Cooking Completed
+            value: done
           - dps_val: pause
-            value: Cooking Paused
+            value: pause
```
jianyun8023, line 203: The retention of value here is necessary for enum usage.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
```diff
@@ -4,6 +4,7 @@ products:
     name: Casdon TD Pro 2
 primary_entity:
   entity: climate
+  translation_only_key: casdon_oven
```
make-all, line 7: I think just "oven" is better, as other ovens are likely to have similar presets.  Even if some are different, it is no problem having extra presets defined in the translations.

Unlike the error codes, these do not have the "E1" prefix tying them strongly to a specific device implementation.

make-all:
Maybe also add an icon for the "oven" climate translation key to icons.json.

mdi:microwave, mdi:toaster-oven and mdi:stove may be suitable candidates.

custom_components/tuya_local/devices/casdon_td_pro_2.yaml
jianyun8023: I think using an oven is also feasible; I’m just not sure if this list will become particularly large in the future.

PR #2116, opened 2024-07-15
Add support for Tera W01 EV Charger

tomassanjuann:
Adding yaml file for Tera W01 EV Charger, previously not supported.


custom_components/tuya_local/devices/tera_w01_evcharger.yaml
```diff
@@ -0,0 +1,270 @@
+name: Tera W01 EV charger
```
make-all, line 1: This top level name should be without product-specific branding (eg, just "EV charger")

tomassanjuann: Sure!

PR #2134, opened 2024-07-18
Add support for DH-CSK03W Dehumidifier

neilmartin83:
Also sold in the UK as BLACK+DECKER BXEH60013GB 3.2L Smart Dehumidifier & Air Purifier

image

### DIFF
--- a/custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
@@ -6,13 +6,6 @@ primary_entity:
     - id: 1
       name: switch
       type: boolean
-      mapping:
-        - dps_val: false
-          icon: "mdi:air-humidifier-off"
-          icon_priority: 2
-        - dps_val: true
-          icon: "mdi:air-humidifier"
-          icon_priority: 4
     - id: 2
       name: humidity
       type: integer
@@ -26,9 +19,9 @@ primary_entity:
       type: string
       mapping:
         - dps_val: "mode_strong"
-          value: "Strong"
+          value: "boost"
         - dps_val: "mode_sleep"
-          value: "Sleep"
+          value: "sleep"
     - id: 6
       type: integer
       name: current_humidity
@@ -77,14 +70,6 @@ secondary_entities:
             value: "Light mode 2"
           - dps_val: "light_mode3"
             value: "Light mode 3"
-  - entity: sensor
-    class: humidity
-    dps:
-      - id: 6
-        type: integer
-        name: sensor
-        unit: "%"
-        class: measurement
   - entity: sensor
     class: temperature
     dps:
--- a/custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
@@ -1,0 +1,95 @@
+name: "DH-CSK03W Dehumidifier"
+primary_entity:
+  entity: humidifier
+  class: dehumidifier
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+      mapping:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
```diff
@@ -0,0 +1,95 @@
...
+        mapping:
+          - dps_val: "ligth_off"
+            value: "Light Off"
+          - dps_val: "light_mode1"
+            value: "Light mode 1"
+          - dps_val: "light_mode2"
+            value: "Light mode 2"
+          - dps_val: "light_mode3"
+            value: "Light mode 3"
+  - entity: sensor
```
make-all, line 80: This is already in the humidifier entity, so does not need to be duplicated.

custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
```diff
@@ -0,0 +1,95 @@
+name: "DH-CSK03W Dehumidifier"
+primary_entity:
+  entity: humidifier
+  class: dehumidifier
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+      mapping:
```
make-all, line 9: Custom icons should not be needed. Some older dehumidifier configs provide them because the tank indicator is combined with the humidifier, so to get the icon priorities right they are all specified even when they match the default. but since binary_sensors have been introduced the tank indicator is generally separate.

custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
```diff
@@ -0,0 +1,95 @@
...
+        min: 40
+        max: 70
+      mapping:
+        - step: 5
+    - id: 5
+      name: mode
+      type: string
+      mapping:
+        - dps_val: "mode_strong"
+          value: "Strong"
```
make-all, line 29: It is generally better to use predefined modes with [translations](https://github.com/home-assistant/core/blob/dev/homeassistant/components/humidifier/strings.json#L62), such as `boost` and `sleep` (all lower case)

custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - entity: select
+    name: "Light mode"
+    category: config
+    dps:
+      - id: 31
+        type: string
+        name: option
+        optional: true
+        mapping:
+          - dps_val: "ligth_off"
```
make-all, line 72: This misspelling is by tuya?

custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
neilmartin83: Removed mappings

custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
neilmartin83: Using `boost` now and corrected `sleep` to lower case, displaying correctly in HA

custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
neilmartin83: It's an error on their side - if I set it to `light_off` it blanks out when I use the Smart Life app to switch Light modes. Otherwise works properly with the incorrect spelling.

custom_components/tuya_local/devices/DH-CSK03W_dehumidifier.yaml
neilmartin83: Removed

PR #2137, opened 2024-07-20
Casdon td pro 2 add local Recip

jianyun8023:
Casdon td pro 2: Add Local recipes

[【PDF】Casdon td pro 2 Local Recipes](https://www.casdon.com.cn/uploads/editor/file/20221124/1669256852863137.pdf)


custom_components/tuya_local/translations/en.json
```diff
@@ -355,6 +355,42 @@
...
+                "name": "Local Recipe",
+                "state": {
+                    "none": "None",
+                    "steamed_egg_with_okra": "Steamed Egg with Okra",
+                    "steamed_sea_bass": "Steamed Sea Bass",
+                    "steamed_prawns": "Steamed prawns",
+                    "handmade_steamed_bread": "Handmade Steamed Bread",
+                    "fan_steamed_baby_vegetables": "Fan Steamed Baby Vegetables",
+                    "braised_pork": "Braised Pork",
+                    "snow_fungus_and_bird_s_nest": "Snow fungus and bird s nest",
```
make-all, line 369: If ' cannot go into the string, then it is better to remove the space between "bird" and "s". But I think inside "double-quotes", 'apostrophe' should be OK.

custom_components/tuya_local/translations/en.json
```diff
@@ -355,6 +355,42 @@
...
+                    "crab_pot": "Crab Pot",
+                    "potato_ribs": "Potato Ribs",
+                    "coconut_chicken_soup": "Coconut Chicken Soup",
+                    "snack_platter": "Snack Platter",
+                    "chicken_skewers": "Chicken Skewers",
+                    "roasted_pork_knuckle": "Roasted Pork Knuckle",
+                    "dried_lemon": "Dried Lemon",
+                    "pork_shop": "Pork Shop",
```
make-all, line 377: I think maybe this is "Pork chop", not shop (店)

custom_components/tuya_local/translations/en.json
```diff
@@ -355,6 +355,42 @@
                     "celsius": "Celsius",
                     "fahrenheit": "Fahrenheit"
                 }
+            },
+            "oven_local_recipe": {
+                "name": "Local Recipe",
```
make-all, line 360: I think "Built-in" is a closer translation of the Chinese than "Local".
Also, Home Assistant convention is for "recipe" in lower case.

custom_components/tuya_local/translations/en.json
jianyun8023: These English and Chinese words are from the Tuya application. I checked the word, and the original is wrong. It should be `猪肉脯` (pork jerky) instead.

custom_components/tuya_local/translations/en.json
jianyun8023: I added the lost `'`, I'm sorry.

key: `snow_fungus_and_bird_s_nest`
English: `Snow fungus and bird's nest`
Chinese: `雪耳燕窝`

custom_components/tuya_local/translations/en.json
jianyun8023: Indeed, I also believe using “Built-in” is more appropriate. I have done the same in the Chinese context.

PR #2142, opened 2024-07-21
Els Pet Spaceship Automatic Litter Box

RoadkillUK:
Configuration for the Els Pet Spaceship Automatic Litter Box https://www.elspet.com/products/self-cleaning-cat-litter-box-toilet


RoadkillUK:
I'm not sure what information is required, please point me in the right direction.  Thanks.

RoadkillUK:
I've updated as requested and corrected some errors.

Let me know if I need to do anything else.

make-all:
I don't think your last change reverting mine is a good change. There is only one light exposed, so I think the benefit of reusing the Home Assistant translations for a general light outweighs the benefit of a more specific name. And the custom icon does not really match the type of light, so it is better to avoid interfering with possible theming changes by leaving the icon at default.

RoadkillUK:
> I don't think your last change reverting mine is a good change. There is only one light exposed, so I think the benefit of reusing the Home Assistant translations for a general light outweighs the benefit of a more specific name. And the custom icon does not really match the type of light, so it is better to avoid interfering with possible theming changes by leaving the icon at default.

I agree with what you say, I'm the noob here and I'll submit to your experience.

Lee.

PR #2145, opened 2024-07-22
Update rotenso_ronix_heatpump.yaml

piotrek2555:
Change range temperature from 16 to 31

### DIFF
--- a/custom_components/tuya_local/devices/rotenso_ronix_heatpump.yaml
+++ b/custom_components/tuya_local/devices/rotenso_ronix_heatpump.yaml
@@ -31,7 +31,7 @@ primary_entity:
       unit: C
       range:
         min: 160
-        max: 880
+        max: 310
       mapping:
         - scale: 10
           step: 10


make-all:
@cykus, as you originally submitted this device, do you have any comments on this change?

piotrek2555:
I only copy and paste from other device. In my app smart I have only to 31 degrees celsius, but I not sure or device support Fahrenheit temperature.

make-all:
Thank you for confirming

cykus:
i set it to 88 because of debug info on iot tuya, as @piotrek2555 said in app i can set it only to 31c
`temp_set | Integer | {   "unit": "℃",   "min": 160,   "max": 880,   "scale": 1,   "step": 5 }`

on debug info i see second dps for Fahrenheit, but i don't know have tu use two set_temps
`temp_set_f | Integer | {   "unit": "℉",   "min": 61,   "max": 88,   "scale": 0,   "step": 1 }`

PR #2146, opened 2024-07-22
Implement generic tuya human presence sensor

illuzn:
Not ready for merge.

@make-all I'm pulling my hair out and maybe you can help me... I think i've defined the device correctly but I can't get it to appear in the list of candidate devices (connects fine).

This is the result of tinytuya wizard:

```json
    {
        "name": "Human presence detector",
        "id": "REDACTED",
        "key": "REDACTED",
        "mac": "REDACTED",
        "uuid": "02a339dfa475cd37",
        "sn": "REDACTED",
        "category": "hps",
        "product_name": "Human presence detector",
        "product_id": "e833v6jexwfkjrij",
        "biz_type": 0,
        "model": "RTCZ-03",
        "sub": false,
        "icon": "https://images.tuyaeu.com/smart/icon/ay15427647462366edzT/8293b553146d0a545989f9486f1f9d29.jpg",
        "mapping": {
            "1": {
                "code": "presence_state",
                "type": "Enum",
                "values": {
                    "range": [
                        "none",
                        "presence",
                        "peaceful",
                        "small_move",
                        "large_move"
                    ]
                }
            }
        }
    }
```

and this is the result of a TinyTuya full status dump:

```
DEBUG:TinyTuya [1.15.1]

DEBUG:Python 3.12.3 (main, Apr 10 2024, 05:33:47) [GCC 13.2.0] on linux
DEBUG:Using pyca/cryptography 43.0.0 for crypto, GCM is supported
 > Send Request for Status <
REDACTED
 > Begin Monitor Loop <
REDACTED
Received Payload: {'dps': {'1': 'none', '101': 7, '102': 203, '103': 15, '104': True, '105': 0, '106': 15, '107': 840, '108': 0, '111': 0, '112': 5, '120': 'FDFCFBFA04001201000004030201'}}
```

I'm sure I'm missing something simple but I'm losing my mind...

### DIFF
--- a/custom_components/tuya_local/devices/human_presence_detector.yaml
+++ b/custom_components/tuya_local/devices/human_presence_detector.yaml
@@ -65,7 +65,6 @@ secondary_entities:
           min: 0
           max: 28799
         unit: s
-          - entity: sensor
       - id: 105
         type: integer
         unit: min
--- a/custom_components/tuya_local/devices/human_presence_detector.yaml
+++ b/custom_components/tuya_local/devices/human_presence_detector.yaml
@@ -1,6 +1,6 @@
 name: human presence detector
 products:
-  - id: 02a339dfa475cd37
+  - id: e833v6jexwfkjrij
     name: RTCZ-03 human presence sensor
 primary_entity:
   entity: binary_sensor
--- a/custom_components/tuya_local/devices/human_presence_detector.yaml
+++ b/custom_components/tuya_local/devices/human_presence_detector.yaml
@@ -1,4 +1,4 @@
-name: human presence detector
+name: Human presence detector
 products:
   - id: e833v6jexwfkjrij
     name: RTCZ-03 human presence sensor
@@ -7,7 +7,7 @@ primary_entity:
   class: occupancy
   dps:
     - id: 1
-      type: boolean
+      type: string
       name: sensor
       mapping:
         - dps_val: peaceful
@@ -28,7 +28,7 @@ secondary_entities:
     class: motion
     dps:
       - id: 1
-        type: boolean
+        type: string
         name: sensor
         mapping:
           - dps_val: peaceful
@@ -56,7 +56,7 @@ secondary_entities:
         name: sensor
         unit: lx
   - entity: number
-    name: clear delay
+    name: Clear delay
     dps:
       - id: 103
         type: integer
@@ -65,9 +65,18 @@ secondary_entities:
           min: 0
           max: 28799
         unit: s
+          - entity: sensor
+      - id: 105
+        type: integer
+        unit: min
+        name: minutes
+      - id: 106
+        type: integer
+        unit: s
+        name: seconds
   - entity: light
     category: config
-    name: indicator light
+    translation_key: indicator
     dps:
       - id: 104
         type: boolean
@@ -96,7 +105,7 @@ secondary_entities:
         unit: cm
   - entity: number
     category: config
-    name: Detection Sensitivity
+    name: Detection sensitivity
     dps:
       - id: 111
         type: integer
@@ -106,7 +115,7 @@ secondary_entities:
           max: 10
   - entity: number
     category: config
-    name: Hold Sensitivity
+    name: Hold sensitivity
     dps:
       - id: 112
         type: integer
@@ -114,22 +123,6 @@ secondary_entities:
         range:
           min: 0
           max: 10
-  - entity: sensor
-    category: diagnostic
-    name: hold (min)
-    dps:
-      - id: 105
-        type: integer
-        unit: min
-        name: sensor
-  - entity: sensor
-    category: diagnostic
-    name: hold (sec)
-    dps:
-      - id: 106
-        type: integer
-        unit: s
-        name: sensor
   - entity: sensor
     category: diagnostic
     name: debug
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/human_presence_detector.yaml
```diff
@@ -0,0 +1,139 @@
+name: human presence detector
+products:
+  - id: e833v6jexwfkjrij
+    name: RTCZ-03 human presence sensor
+primary_entity:
+  entity: binary_sensor
+  class: occupancy
+  dps:
+    - id: 1
+      type: boolean
```
make-all, line 10: type is `string`

custom_components/tuya_local/devices/human_presence_detector.yaml
```diff
@@ -0,0 +1,139 @@
...
+          value: false
+    - id: 1
+      type: string
+      name: raw_state
+secondary_entities:
+  - entity: binary_sensor
+    class: motion
+    dps:
+      - id: 1
+        type: boolean
```
make-all, line 31: type is `string`

custom_components/tuya_local/devices/human_presence_detector.yaml
```diff
@@ -0,0 +1,139 @@
+name: human presence detector
```
make-all, line 1: Capitalize Human

custom_components/tuya_local/devices/human_presence_detector.yaml
```diff
@@ -0,0 +1,139 @@
...
+        unit: cm
+  - entity: sensor
+    class: illuminance
+    dps:
+      - id: 102
+        type: integer
+        name: sensor
+        unit: lx
+  - entity: number
+    name: clear delay
```
make-all, line 59: Capitalise Clear

custom_components/tuya_local/devices/human_presence_detector.yaml
```diff
@@ -0,0 +1,139 @@
...
+      - id: 103
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 28799
+        unit: s
+  - entity: light
+    category: config
+    name: indicator light
```
make-all, line 70: Suggest to remove name and use `translation_key: indicator` instead.

custom_components/tuya_local/devices/human_presence_detector.yaml
```diff
@@ -0,0 +1,139 @@
...
+      - id: 108
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 840
+        unit: cm
+  - entity: number
+    category: config
+    name: Detection Sensitivity
```
make-all, line 99: sensitivity should be all lower case

custom_components/tuya_local/devices/human_presence_detector.yaml
```diff
@@ -0,0 +1,139 @@
...
+    dps:
+      - id: 111
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 10
+  - entity: number
+    category: config
+    name: Hold Sensitivity
```
make-all, line 109: sensitivity should be all lower case

custom_components/tuya_local/devices/human_presence_detector.yaml
```diff
@@ -0,0 +1,139 @@
...
+    dps:
+      - id: 112
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 10
+  - entity: sensor
+    category: diagnostic
+    name: hold (min)
```
make-all, line 119: I'm not sure how this relates to the below, but probably just a single sensor with the other value as an extra attribute would be better.
If Hold time is defined by min + sec (eg 1:30 = min:1; sec:30), then make the min one the main sensor, and have seconds as an extra attribute available if users need more resolution.
If seconds and min show the same thing with different resolutions (eg 1:30 = min:1, sec:90), then make the seconds the main sensor and leave the minutes as an extra attribute just in case someone has a use for that.
In both cases, add `class: duration` for better display of values, and following HA naming convention.

custom_components/tuya_local/devices/human_presence_detector.yaml
```diff
@@ -0,0 +1,139 @@
...
+        name: sensor
+  - entity: sensor
+    category: diagnostic
+    name: hold (sec)
+    dps:
+      - id: 106
+        type: integer
+        unit: s
+        name: sensor
+  - entity: sensor
```
make-all, line 133: I don't think this needs to be a sensor, just add it as an extra attribute on one of the other entities (usually the primary entity by convention, unless another one is clearly linked to this value)

custom_components/tuya_local/devices/human_presence_detector.yaml
illuzn: Yeah... I'm not sure why this device is made this way but its the same delay as in DP103 but this one just spits it back out in minutes and seconds in DP105 and DP106.

illuzn:
Any idea why this isn't showing in the list of autodetected devices when I try to add it to Tuya Local? I believe I've got all the DPS correct based upon the TinyTuya call and I've filled in the rest of the product data I believe.

make-all:
Before I would have said the incorrect types on dp 1 would have been stopping the detection, but if it still not detecting I think you need to enable debug logging and look for what it says when matching against this config

illuzn:
Duh... I'm losing my mind... I forgot to restart HA.

Working as expected now and ready for merge.

PR #2152, opened 2024-07-25
Added support for Wada Power Q8 EV charger

MobilettoSoft:
Preliminary support for Wada Energy Q8 EV charger.

- Start/stop charge, set charge current are working
- Actual charging power sensor is working
- Missing phase current and voltage data
- Missing support for time based charge


make-all:
Your product id looks like a device id. These are separate things, the product id is common to all devices of that product, your device id is unique to your device.

MobilettoSoft:
thanks. It should be the right one now. Marked the pull request as draft as I have to make some tests.

custom_components/tuya_local/devices/Wada Energy Q8 Charger.yaml
```diff
@@ -0,0 +1,106 @@
+name: Q8 wall charger
```
make-all, line 1: this top level name should be non-branded, in case other devices match the same config.
Suggest "EV charger" for the name.

MobilettoSoft: done, thanks

PR #2153, opened 2024-07-25
Update madimack_elite_v3_heatpump.yaml

sikkek:
Added dps 15 to secondary entities (fault code), mapping of fault code to actual message is to be researched. Also added 102 (Inlet water temperature) to secondary entities as well, to get a more consistent list in HA, similar to the Tuya app.

### DIFF
--- a/custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
+++ b/custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
@@ -130,22 +130,6 @@ secondary_entities:
             value: C
   - entity: sensor
     category: diagnostic
-    name: Inlet water temperature
-    class: temperature
-    dps:
-      - id: 102
-        name: sensor
-        type: integer
-      - id: 6
-        name: unit
-        type: string
-        mapping:
-          - dps_val: f
-            value: F
-          - dps_val: c
-            value: C
-  - entity: sensor
-    category: diagnostic
     name: Outlet water temperature
     class: temperature
     dps:
--- a/custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
+++ b/custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
@@ -78,6 +78,13 @@ primary_entity:
 secondary_entities:
   - entity: sensor
     category: diagnostic
+    name: Fault code
+    dps:
+      - id: 15
+        name: sensor
+        type: integer
+  - entity: sensor
+    category: diagnostic
     name: Power level
     class: power_factor
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
```diff
@@ -76,6 +76,13 @@ primary_entity:
       name: unknown_107
       type: boolean
 secondary_entities:
+  - entity: sensor
```
make-all, line 79: This would be better as:
```yaml
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 15
        type: bitfield
        name: sensor
        mapping;
          - dps_val: 0
            value: false
          - value: true
      - id: 15
        type: bitfield
        name: fault_code
```

sikkek: Oh, yes that would be better, until I can come up with a way to map the fault codes to actual meaningful messages. I know that 4 equals Error E3 which is insufficient water flow. However, all other possible errors are unknown. I'll try to contact a manufacturer of the heatpump to see if we can get some more info.

sikkek: For now, I've updated the file with your changes

sikkek: OK, I found some additional info. Can you tak a look at https://github.com/make-all/tuya-local/discussions/2154 ? The heatpump sends out bit flags for the errors.

custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
```diff
@@ -76,6 +76,20 @@ primary_entity:
       name: unknown_107
       type: boolean
 secondary_entities:
+  - entity: binary_sensor
```
make-all, line 79: You should remove the "unknown_15" attribute from above, since it is now identified.

custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
```diff
@@ -117,6 +131,22 @@ secondary_entities:
             value: F
           - dps_val: c
             value: C
+  - entity: sensor
```
make-all, line 134: I don't think this is required, since it is duplicating the "current_temperature" of the climate entity above.

custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
sikkek: Removed the unknown entry!

custom_components/tuya_local/devices/madimack_elite_v3_heatpump.yaml
sikkek: This was actually on purpose. In the Tuya app, Inlet and Outlet temperatures are also displayed in the sensor list values, which is quite convenient when adjusting waterflow to get the best efficiency out of the heat pump.

make-all: Home Assistant is not the Tuya App. You can customise the UI however you want it, that does not depend on duplicating the sensor values.

sikkek: Fair enough, that makes sense. I'll work on the faut codes, see if I can come up with something logical to group different errors. Thank for your patience :-)

PR #2162, opened 2024-07-28
Added Doel Cat Litter Box

trasbd:
Configuration for Doel Automatic Litter Box
https://a.co/d/gqnRzUR

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/thoma/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/thoma/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">

DPS | Description
-- | --
6 | Weight kg
7 | Visits
8 | Excretion Duration
22 | Diagnostic
24 | Status
101 | Start Clean
102 | Empty Cycle
103 | Bin Full
104 | Monitoring
105 | Auto
108 | Unit Switch
111 | Infared Switch
114 | Child Lock
117 | Clean Up Wait Time
118 | Clean Up Interval
120 | Unknown (bool)
121 | Cleaning
123 | Capacity Calibration
124 | Detection Sensitivity
125 | Deep Cleaning
126 | Start Level
127 | Soft Mode
128 | Unknown (bool)
129 | Unknown (int)
131 | Litter Type
134 | Weight lbs
135 | dp_developer_mode

</body>

</html>

### DIFF
--- a/custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
+++ b/custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
@@ -1,0 +1,249 @@
+name: Litter Box
+products:
+  - id: ebf44398a24b7b6be3vnml
+    name: DOEL Ti+
+primary_entity:
+  entity: sensor
+  name: Cat Weight
+  icon: "mdi:cat"
+  class: weight
+  dps:
+    - id: 6
+      type: integer
+      name: sensor
+      unit: kg
+      class: measurement
+      mapping:
+        - scale: 10
+    - id: 134
+      type: integer
+      name: sensor
+      unit: lb
+      class: measurement
+      mapping:
+        - scale: 10
+secondary_entities:
+  - entity: sensor
+    name: Visits
+    icon: "mdi:emoticon-poop"
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
+        unit: visits
+  - entity: sensor
+    name: Visits Duration
+    icon: "mdi:paper-roll"
+    class: duration
+    dps:
+      - id: 8
+        type: integer
+        name: sensor
+        unit: seconds
+        class: measurement
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 22
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 22
+        type: bitfield
+        name: fault_code
+  - entity: button
+    name: Clean Cycle
+    icon: "mdi:shimmer"
+    dps:
+      - id: 101
+        type: boolean
+        name: button
+  - entity: button
+    name: Empty Cycle
+    icon: "mdi:delete-empty"
+    dps:
+      - id: 102
+        type: boolean
+        name: button
+  - entity: binary_sensor
+    name: Bin Full
+    icon: "mdi:trash-can"
+    dps:
+      - id: 103
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
+    class: occupancy
+    name: Monitoring
+    icon: "mdi:motion-sensor"
+    dps:
+      - id: 104
+        type: boolean
+        name: sensor
+  - entity: switch
+    name: Auto Clean
+    category: config
+    icon: "mdi:auto-mode"
+    dps:
+      - id: 105
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Infrared detection
+    category: config
+    icon: "mdi:leak"
+    dps:
+      - id: 111
+        type: boolean
+        name: switch
+  - entity: lock
+    translation_key: child_lock
+    name: Child Lock
+    category: config
+    dps:
+      - id: 114
+        type: boolean
+        name: lock
+  - entity: number
+    name: Clean Wait Time
+    category: config
+    icon: "mdi:timer"
+    dps:
+      - id: 117
+        type: integer
+        name: value
+        unit: min
+        range:
+          min: 0
+          max: 60
+  - entity: number
+    name: Clean Interval
+    category: config
+    icon: "mdi:update"
+    dps:
+      - id: 118
+        type: integer
+        name: value
+        unit: min
+        range:
+          min: 0
+          max: 120
+  - entity: binary_sensor
+    name: Smart Cleaning
+    icon: "mdi:refresh-auto"
+    dps:
+      - id: 121
+        type: boolean
+        name: sensor
+  - entity: number
+    name: Capacity Calibration
+    category: config
+    icon: "mdi:trash-can"
+    dps:
+      - id: 123
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 15
+  - entity: number
+    name: Detection Sensitivity
+    category: config
+    icon: "mdi:scale"
+    dps:
+      - id: 124
+        type: integer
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,249 @@
+name: Litter Box
+products:
+  - id: ebf44398a24b7b6be3vnml
```
make-all, line 3: This appears to be a device id (unique to your device) rather than a product id (shared among all Doel Ti+ devices). Normally a product id should be alphanumeric for the entire length, this one is hex except the last 4 characters, which is how device ids are. I may be wrong about this because there is a small chance that an alphanumeric id just happens to be valid hex, but please double check.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,249 @@
...
+    - id: 6
+      type: integer
+      name: sensor
+      unit: kg
+      class: measurement
+      mapping:
+        - scale: 10
+    - id: 134
+      type: integer
+      name: sensor
```
make-all, line 20: Only one dp within the entity can have the name sensor.  The other will need to be renamed and will be an extra attribute (in which case, unit will not be used, so make it clear in the name).

If the intention is to have this dependant on the unit selection, you need to use:
```
    - id: 6
      type: integer
      name: sensor
      class: measurement
      mapping:
        - scale: 10
          constraint: unit
          conditions:
            - dps_val: lb
              value_redirect: weight_lb
    - id: 108
      type: string
      name: unit
    - id: 134
      type: integer
      name: weight_lb
```
I don't remember whether the scale from the original mapping applies to the redirected value, so you may need to add the mapping with `scale:10` under weight_lb as well.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,249 @@
...
+  - entity: sensor
+    name: Visits
+    icon: "mdi:emoticon-poop"
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
+        unit: visits
+  - entity: sensor
+    name: Visits Duration
```
make-all, line 35: Usually this will be either a daily total, or last visit. The name should make that clear.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,249 @@
...
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 22
+        type: bitfield
+        name: fault_code
+  - entity: button
+    name: Clean Cycle
```
make-all, line 59: The word Cycle seems redundant. I think just "Clean" and "Empty" would be better names for these buttons.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,249 @@
...
+  - entity: binary_sensor
+    name: Bin Full
+    icon: "mdi:trash-can"
+    dps:
+      - id: 103
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
+    class: occupancy
+    name: Monitoring
```
make-all, line 81: The name "Monitoring" is a bit unclear. Most likely this would be better without a name, and inheriting the default translations for "occupancy".

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,249 @@
...
+        name: switch
+  - entity: button
+    name: Level Cycle
+    icon: "mdi:arrow-collapse-down"
+    dps:
+      - id: 126
+        type: boolean
+        name: button
+  - entity: switch
+    name: Soft Mode
```
make-all, line 180: "Soft mode" is an unclear name. Is there a better way to describe whatever function this is controlling?

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,249 @@
...
+    category: config
+    icon: "mdi:shimmer"
+    dps:
+      - id: 127
+        type: boolean
+        name: switch
+  - entity: switch
+    name: dp_developer_mode
```
make-all, line 188: This should most likely not be exposed. Or if it is useful, use a better name that makes it clear what is being enabled.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,249 @@
...
+    dps:
+      - id: 131
+        type: string
+        name: option
+        mapping:
+          - dps_val: "mineral_cat_litter"
+            value: "Mineral"
+          - dps_val: "mixed_cat_litter"
+            value: "Mixed"
+  - entity: binary_sensor
```
make-all, line 229: unknown dps should just be listed as extra attributes on another entity, not entities in their own right.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,249 @@
+name: Litter Box
```
make-all, line 1: HA Naming convention: only the first word should be capitalised in device and entity names (applies to quite a few entity names below too)

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -55,37 +59,51 @@ secondary_entities:
       - id: 22
         type: bitfield
         name: fault_code
+      - id: 120
+        type: boolean
+        name: DP 120
```
make-all, line 64: dp name should not have spaces, and by convention attribute names are all lower case.  I suggest unknown_120, though for 127 and 135, you did have a name you can use here, even if the function is not understood.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -0,0 +1,231 @@
...
+  entity: sensor
+  name: Cat weight
+  icon: "mdi:cat"
+  class: weight
+  dps:
+    - id: 6
+      type: integer
+      name: sensor
+      class: measurement
+      unit: kg
```
make-all, line 15: I see you have added the unit back here, but the entity has a dp named "unit", which will always override it. Probably you had a reason to add this back in, but I think it points to a problem elsewhere in that case. If dp 108 is sometimes missing (which I suspect is your reason for adding static units), then it needs to be marked as optional to let the config be matched, and a null mapping for the default case should be added there to avoid the confusion over whether the static unit or the unit dp applies.

PR #2171, opened 2024-07-31
Add new device: Pool-Systems IPS-100max.yaml

nanosonde:
This will add this new device: Pool-Systems IPS-100MAX
https://www.pool-systems.de/Pool-Waermepumpe-mit-Inverter/Pool-Waermepumpe-IPS-100MAX-Inverter-Premium-Silent-MAX-10KW-COP20.html

This works:
* switching between work modes: OFF, HEAT, COOL, AUTO(HEAT/COOL)
* switching between preset modes: SILENT, SMART, BOOST
* switching between °C and °F
* Setting desired temperature for either mode combination (work modes and preset modes) 
* Getting current temperature
* Getting power level in percent
* Getting power in watts
* binary_sensor indicating if water flows through the heat pump

There are two more datapoints for which I have no idea what they represent.

The heatpump is assumed to be hardware/software by Fairland.

### DIFF
--- a/custom_components/tuya_local/devices/poolsystems_ips100max.yaml
+++ b/custom_components/tuya_local/devices/poolsystems_ips100max.yaml
@@ -1,0 +1,92 @@
+name: Pool heatpump
+products:
+  - id: 1ovawewadpk2jiaw
+    name: Pool-Systems IPS-100MAX
+primary_entity:
+  entity: climate
+  dps:
+    - id: 101
+      type: boolean
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          icon: "mdi:hvac-off"
+          icon_priority: 1
+          value: "off"
+        - dps_val: true
+          constraint: work_mode
+          conditions:
+            - dps_val: smart
+              value: heat_cool
+            - dps_val: cool
+              value: cool
+            - dps_val: warm
+              value: heat
+    - id: 102
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: "silence"
+          value: sleep
+        - dps_val: "smart"
+          value: comfort
+        - dps_val: "booster"
+          value: boost
+    - id: 103
+      name: current_temperature
+      type: integer
+    - id: 104
+      name: temperature_unit
+      type: boolean
+      mapping:
+        - dps_val: false
+          value: C
+        - dps_val: true
+          value: F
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/poolsystems_ips100max.yaml
```diff
@@ -0,0 +1,92 @@
...
+        - dps_val: true
+          value: F
+    - id: 106
+      type: string
+      name: work_mode
+      hidden: true
+    - id: 107
+      name: temperature
+      type: integer
+    - id: 108
```
make-all, line 53: These should maybe also have number entities for adjusting them (marked as `category: config`)

custom_components/tuya_local/devices/poolsystems_ips100max.yaml
```diff
@@ -0,0 +1,92 @@
...
+      name: min_temperature
+    - id: 109
+      type: integer
+      name: max_temperature
+secondary_entities:
+  - entity: sensor
+    category: diagnostic
+    name: Power level
+    icon: "mdi:signal"
+    class: power_factor
```
make-all, line 64: Probably better without explicit icon and name.

custom_components/tuya_local/devices/poolsystems_ips100max.yaml
```diff
@@ -0,0 +1,92 @@
...
+  - entity: binary_sensor
+    name: Water flow
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 110
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 4
```
make-all, line 80: Maybe also a general problem sensor to catch all the other non-zero values, unless you know that the water flow is the only fault that is ever indicated..

custom_components/tuya_local/devices/poolsystems_ips100max.yaml
```diff
@@ -0,0 +1,92 @@
...
+            value: true
+          - value: false
+  - entity: sensor
+    category: diagnostic
+    name: Power
```
make-all, line 85: name should be left out, as the class will provide translations.

nanosonde:
@make-all 
Sorry, I have been on summer vacation not being online.
Thanks for considering and fixing!

PR #2180, opened 2024-08-04
Add RainPoint Smart Water Timer

josiasmontag:
Adds the RainPoint Smart Water Timer (TTV103FRF)

image

image

### DIFF
--- a/custom_components/tuya_local/devices/rainpoint_water_timer.yaml
+++ b/custom_components/tuya_local/devices/rainpoint_water_timer.yaml
@@ -1,0 +1,116 @@
+name: RainPoint Smart Water Timer
+products:
+  - id: ew946yrp3pgbaziu
+    name: RainPoint TTV103FRF
+primary_entity:
+  entity: valve
+  class: water
+  icon: "mdi:watering-can"
+  dps:
+    - id: 108
+      type: boolean
+      name: valve
+      mapping:
+        - dps_val: true
+          value: false
+        - dps_val: false
+          value: true
+secondary_entities:
+  - entity: switch
+    name: Weekly Automatic Irrigation
+    category: config
+    icon: "mdi:calendar-clock"
+    dps:
+      - id: 125
+        optional: true
+        type: boolean
+        name: switch
+        mapping:
+          - dps_val: true
+            value: false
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/rainpoint_water_timer.yaml
```diff
@@ -0,0 +1,116 @@
...
+          - dps_val: "1"
+            value: "Manual"
+          - dps_val: "2"
+            value: "Auto"
+          - dps_val: "3"
+            value: "Rain Delay"
+  - entity: binary_sensor
+    category: diagnostic
+    class: problem
+    icon: "mdi:alert"
```
make-all, line 108: problem sensor does not need an icon defined, the base class defines dynamic icons already

custom_components/tuya_local/devices/rainpoint_water_timer.yaml
```diff
@@ -0,0 +1,116 @@
...
+        optional: true
+        type: boolean
+        name: switch
+        mapping:
+          - dps_val: true
+            value: false
+          - dps_val: false
+            value: true
+  - entity: select
+    name: Weekly Irrigation Rain Delay
```
make-all, line 34: This name is a bit long, other similar devices use just "Rain delay".   Also, naming needs to follow HA naming convention.. https://developers.home-assistant.io/docs/core/entity/#:~:text=Entity%20names%20should%20start%20with,a%20capitalized%20abbreviation%20of%20course).

custom_components/tuya_local/devices/rainpoint_water_timer.yaml
```diff
@@ -0,0 +1,116 @@
+name: RainPoint Smart Water Timer
```
make-all, line 1: This top level name should be without branding (and following HA naming convention, see below).

Just "Water timer" should be enough.

josiasmontag:
@make-all Updated, please check

MariusNovak:
Any idea how to add the RainPoint Smart Water Timer Twin way / two zone TWG004WRF ?
using this using this custom component alows only to use one of the two zones ... 
Thanks for adive

Issue #2131, opened 2024-07-18
Request support for ION WIFI Dimmer


Blcktape:
# Log Message

```
Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:483
integration: Tuya Local ([documentation](https://github.com/make-all/tuya-local), [issues](https://github.com/make-all/tuya-local/issues))
First occurred: 12:42:35 (2 occurrences)
Last logged: 12:54:38

Device matches eurom_800_heater with quality of 100%. DPS: {"updated_at": 1722508951.3461888, "1": false, "2": 552, "3": 92, "5": 552, "6": 0}
Device matches eurom_800_heater with quality of 100%. DPS: {"updated_at": 1722509675.1124086, "1": false, "2": 552, "3": 92, "5": 552, "6": 0}
```

# Information about DPS mappings

```json
{
  "result": {
    "category": "tgq",
    "functions": [
      {
        "code": "switch_led_1",
        "desc": "{}",
        "name": "开关1",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "bright_value_1",
        "desc": "{\"min\":10,\"max\":1000,\"scale\":0,\"step\":1}",
        "name": "亮度值1",
        "type": "Integer",
        "values": "{\"min\":10,\"max\":1000,\"scale\":0,\"step\":1}"
      },
      {
        "code": "brightness_min_1",
        "desc": "{\"min\":10,\"max\":1000,\"scale\":0,\"step\":1}",
        "name": "最小亮度1",
        "type": "Integer",
        "values": "{\"min\":10,\"max\":1000,\"scale\":0,\"step\":1}"
      },
      {
        "code": "led_type_1",
        "desc": "{\"range\":[\"led\",\"incandescent\",\"halogen\"]}",
        "name": "光源类型1",
        "type": "Enum",
        "values": "{\"range\":[\"led\",\"incandescent\",\"halogen\"]}"
      },
      {
        "code": "brightness_max_1",
        "desc": "{\"min\":10,\"max\":1000,\"scale\":0,\"step\":1}",
        "name": "最大亮度1",
        "type": "Integer",
        "values": "{\"min\":10,\"max\":1000,\"scale\":0,\"step\":1}"
      },
      {
        "code": "countdown_1",
        "desc": "{\"unit\":\"s\",\"min\":0,\"max\":86400,\"scale\":0,\"step\":1}",
        "name": "开关1倒计时",
[...]

PR #2184, opened 2024-08-06
Add device yaml for ion_wifi_dimmer.yaml

Blcktape:
Create support for the ION Wifi Dimmer
Product: https://www.ionindustries.com/winkel/led-dimmer-wifi/
Ticket: #2131 

Working features:
- Switch: on/off
- Brightness level
- Minimal Brightness
- Maximum Brightness
- Timer

Not controllable trough tuya-local integration:
- Light Type: "led", "incandescent", "halogen"
- Work Mode: "white"

### DIFF
--- a/custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
+++ b/custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
@@ -1,0 +1,58 @@
+name: ION WIFI DIMMER
+products:
+  - id: ceikasvm5p5ymcax
+    name: ION WIFI DIMMER
+primary_entity:
+  entity: light
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+    - id: 2
+      name: brightness
+      type: integer
+      range:
+        min: 10
+        max: 1000
+secondary_entities:
+  - entity: number
+    name: Minimum brightness
+    category: config
+    icon: "mdi:brightness-percent"
+    dps:
+      - id: 3
+        type: integer
+        name: value
+        range:
+          min: 10
+          max: 1000
+        mapping:
+          - step: 1
+  - entity: number
+    name: Maximum brightness
+    category: config
+    icon: "mdi:brightness-percent"
+    dps:
+      - id: 5
+        type: integer
+        name: value
+        range:
+          min: 10
+          max: 1000
+        mapping:
+          - step: 1
+  - entity: number
+    translation_key: timer
+    name: Timer
+    category: config
+    dps:
+      - id: 6
+        type: integer
+        name: value
+        unit: "s"
+        range:
+          min: 0
+          max: 86400
+        mapping:
+          - scale: 0
+          - step: 1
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
```diff
@@ -0,0 +1,58 @@
+name: ION WIFI DIMMER
```
make-all, line 1: The top level name should be generic (without branding).
"Dimmer" would seem to be suitable.

custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
```diff
@@ -0,0 +1,58 @@
...
+    category: config
+    icon: "mdi:brightness-percent"
+    dps:
+      - id: 3
+        type: integer
+        name: value
+        range:
+          min: 10
+          max: 1000
+        mapping:
```
make-all, line 29: No mapping is required, as step 1 is the default.

custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
```diff
@@ -0,0 +1,58 @@
...
+    category: config
+    icon: "mdi:brightness-percent"
+    dps:
+      - id: 5
+        type: integer
+        name: value
+        range:
+          min: 10
+          max: 1000
+        mapping:
```
make-all, line 42: Likewise

custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
```diff
@@ -0,0 +1,58 @@
...
+    dps:
+      - id: 6
+        type: integer
+        name: value
+        unit: "s"
+        range:
+          min: 0
+          max: 86400
+        mapping:
+          - scale: 0
```
make-all, line 57: `scale: 0` is invalid.  There should only be one default entry in the mapping, and since users generally do not need second resolution on a 24h timer (and tuya app also generally does not provide that), I would generally use

```
   unit: min
   mapping:
     - scale: 60
       step: 60
```

custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
Blcktape: Renamed to "Dimmer".

custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
Blcktape: Removed mapping with default step value 1 as this is the default value.

custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
Blcktape: Removed mapping with default step value 1 as this is the default value.

custom_components/tuya_local/devices/ion_wifi_dimmer.yaml
Blcktape: Changed values to recommended values.

Blcktape:
Updated the yaml file with the provided recommendations.

PR #2185, opened 2024-08-07
Add support for generic light sensor

maddijoyce:
I wasn't sure if there was a better way to do this. Using a generic tuya light sensor that works the same as the existing illuminance sensor, but uses different dps ids.

Thanks,
Maddi

### DIFF
--- a/custom_components/tuya_local/devices/light_sensor.yaml
+++ b/custom_components/tuya_local/devices/light_sensor.yaml
@@ -1,0 +1,30 @@
+name: Light sensor
+products:
+  - id: p1632039035856duknw4
+  
+primary_entity:
+  entity: sensor
+  name: Brightness
+  class: illuminance
+  dps:
+    - id: 7
+      type: integer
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/light_sensor.yaml
```diff
@@ -0,0 +1,30 @@
+name: Light sensor
+products:
+  - id: p1632039035856duknw4
+  
+primary_entity:
+  entity: sensor
+  name: Brightness
```
make-all, line 7: There should be no explicit name, so the translations for illuminance can be inherited

custom_components/tuya_local/devices/light_sensor.yaml
```diff
@@ -0,0 +1,30 @@
+name: Light sensor
+products:
+  - id: p1632039035856duknw4
```
make-all, line 3: There should be a name for this product (brand and model)

PR #2189, opened 2024-08-09
Include "name" for "smart_watering_device"

illuzn:
Includes a missing name for smart_watering_device to avoid it being given the default name "None".


illuzn:
On a seperate note, it's not entirely clear to me that valve is the correct domain for this entity (despite the deprecation warnings in log insisting that it is). This turns the device power on/ off it does not control the valve/ pumping state at all.

make-all:
There is something broken about the valve and switch platforms where they do not use the translations provided for class names.
I don't think it should be "fixed" with a workaround like this though.

illuzn:
Thanks for letting me know... will have a look upstream

PR #2215, opened 2024-08-17
Add Support for Faroro Pet Feeder

jon-sanders:
I've been using and tweaking this config for over a year. A mostly normal pet feeder with a a couple quirks-

- Manual Feed has mode set to box, slider resulted in accidental dispensing on tap
- Manual Feed has persist set to false, otherwise brief disconnections would result in food dispensing

### DIFF
--- a/custom_components/tuya_local/devices/faroro_pet_feeder.yaml
+++ b/custom_components/tuya_local/devices/faroro_pet_feeder.yaml
@@ -1,0 +1,135 @@
+name: Wi-Fi Pet Feeder
+product:
+  - id: vglnuveujzyrpku9
+    name: Faroro Automatic Cat Feeder
+primary_entity:
+  entity: number
+  name: Manual Feed
+  icon: "mdi:food-drumstick"
+  mode: box
+  dps:
+    - id: 3
+      name: value
+      type: integer
+      persist: false
+      range:
+        min: 1
+        max: 12
+secondary_entities:
+  - entity: sensor
+    name: Feed Status
+    icon: "mdi:paw"
+    class: enum
+    dps:
+      - id: 4
+        name: sensor
+        type: string
+        mapping:
+          - dps_val: standby
+            value: Standby
+          - dps_val: feeding
+            value: Feeding
+  - entity: button
+    name: Factory Reset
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/faroro_pet_feeder.yaml
```diff
@@ -0,0 +1,135 @@
...
+    category: diagnostic
+    dps:
+      - id: 104
+        name: sensor
+        type: boolean
+  - entity: button
+    name: Reboot
+    icon: "mdi:restart"
+    category: diagnostic
+    optional: true
```
make-all, line 120: Two issues on this line:
- semi-colon should be a colon
- optional belongs on the dp, not the entity.

custom_components/tuya_local/devices/faroro_pet_feeder.yaml
```diff
@@ -0,0 +1,135 @@
+name: Wi-Fi Pet Feeder
```
make-all, line 1: I prefer to leave off terms like "Wi-Fi" or "Smart" from device names, as they apply to all devices that this integration can support.

Also, Home Assistant has a naming policy for devices and entities which says to only captalize the first word, unless there is another reason (trademark, abbreviation) to capitalise them, so "Pet feeder" here, and similar case changes for the entity names below.

custom_components/tuya_local/devices/faroro_pet_feeder.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - id: 19
+        name: switch
+        type: boolean
+        mapping:
+          - dps_val: false
+            icon: "mdi:lightbulb-outline"
+  - entity: sensor
+    name: Battery Level
+    icon: "mdi:battery"
+    category: diagnostic
```
make-all, line 70: This could be made a `class: battery` sensor by changing the mappings to something like 80, 50, 20 (maybe lower for the value "no", I usually avoid 100 and 0 because they can be misleading when the steps are this coarse but "no" does imply that it really means 0).

Then the name and icon can be left to Home Assistant to manage.

jon-sanders:
All changes have been made, thank you for your help!

PR #2221, opened 2024-08-18
Added support for De'Longhi Pinguino PAC

hadicharara:
Hello. This commit adds support for the De'Longhi Pinguino portable air conditioner. Everything seems to work well. I could've been slightly cleaner if nested conditions (or something equivalent to an &&) were supported, but this is the best we can do I think.

image
image
image

Also the auto mode is actually eco, but eco isn't an hvac_mode. I added the dummy Preset section just to clarify this.

Sometimes though, I get this error notification regularly when I'm not touching anything:
Failed to perform the action climate/set_hvac_mode. expected HVACMode or one of 'off', 'heat', 'cool', 'heat_cool', 'auto', 'dry', 'fan_only' for dictionary value @ data['hvac_mode']

I have no idea what would trigger this. Maybe I messed up somewhere. Welcoming any feedback.

### DIFF
--- a/custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
@@ -1,0 +1,130 @@
+name: Portable Air Conditioner
+products:
+  - id: anR7aeh1YPTKAz5a
+    name: De'Longhi Pinguino PACEL140LRFK 3ALBK
+primary_entity:
+  entity: climate
+  dps:
+    - id: 1
+      name: hvac_mode
+      type: boolean
+      mapping:
+        - value: "off"
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: null
+              value: "off"
+            - dps_val: "1"
+              value: dry
+            - dps_val: "2"
+              value: cool
+            - dps_val: "3"
+              value: fan_only
+            - dps_val: "4"
+              value: auto
+    - id: 5
+      name: mode
+      type: string
+    - id: 6
+      type: integer
+      name: temperature
+      range:
+        min: 18
+        max: 32
+      mapping:
+        - constraint: temperature_unit
+          conditions:
+            - dps_val: true
+              range:
+                min: 64
+                max: 90
+    - id: 8
+      type: string
+      name: fan_mode
+      mapping:
+        - value: auto
+        - dps_val: "0"
+          value: auto
+        - dps_val: "1"
+          value: auto
+          constraint: mode
+          conditions:
+            - dps_val: "2"
+              value: low
+            - dps_val: "3"
+              value: low
+        - dps_val: "2"
+          value: auto
+          constraint: mode
+          conditions:
+            - dps_val: "2"
+              value: medium
+            - dps_val: "3"
+              value: medium
+        - dps_val: "3"
+          value: auto
+          constraint: mode
+          conditions:
+            - dps_val: "2"
+              value: high
+            - dps_val: "3"
+              value: high
+    - id: 16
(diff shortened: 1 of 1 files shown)


hadicharara:
Never mind, by fixing the unit tests, I fixed the error.

hadicharara:
I sacrificed hiding the temperature for supporting the correct temp range in F. All done now.

custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
```diff
@@ -0,0 +1,125 @@
+name: Pinguino
```
make-all, line 1: This top level name should be generic for the type of device, not a brand or model name.

custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
```diff
@@ -0,0 +1,125 @@
...
+            - dps_val: "3"
+              value: fan_only
+            - dps_val: "4"
+              value: auto
+    - id: 5
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: "4"
+          value: eco
```
make-all, line 33: I think leave this out.  "auto" HVAC mode covers what an "Eco" mode should be (more complex logic than just switching between heating and cooling based on the temperature, such as taking presence into consideration or a schedule). In that case rename preset_mode and all references to it, so it does not create a separate preset_mode selector.
Leaving it like this will result in invalid values being present, which will interfere with scene support, and possibly fill the logs up with error messages in normal use.

custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
```diff
@@ -0,0 +1,125 @@
...
+      mapping:
+        - value: "off"
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: vertical
+          constraint: hvac_mode
+          conditions:
+            - dps_val: false
+              value: "off"
```
make-all, line 90: Is this really necessary? It is normal for settings to be retained when a device is off. I don't think it is necessary to show swing_mode as off just because the device is off.

custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
```diff
@@ -0,0 +1,125 @@
...
+        name: sensor
+        mapping:
+          - dps_val: "0"
+            value: "off"
+          - dps_val: "1"
+            value: "blue"
+          - dps_val: "2"
+            value: "green"
+          - dps_val: "3"
+            value: "orange"
```
make-all, line 125: There should probably be a select entity for changing the temperature unit.

custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
hadicharara: Do you mean something like "Portable Air Conditioner"?
I chose Pinguino because that's the series of PACs that De'Longhi makes.

custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
hadicharara: Removed.

custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
hadicharara: Removed.

custom_components/tuya_local/devices/pinguino_portable_airconditioner.yaml
hadicharara: Added.

PR #2222, opened 2024-08-19
Add Cecotec ForceClima 14800 Soundless Heating Connected

tv4you2016:
https://cecotec.pt/pt/ares-condicionados-portateis/forceclima-14800-soundless-heating-connected

image


custom_components/tuya_local/devices/forceClima_Soundless.yaml
```diff
@@ -0,0 +1,86 @@
+name: ForceClima Soundless
```
make-all, line 1: This top level name should be generic for the type of device, as Tuya devices are often sold as whitelabeled, so have different brands in different markets.

make-all: That should be ok

tv4you2016: Fix

custom_components/tuya_local/devices/forceclima_soundless.yaml
```diff
@@ -0,0 +1,86 @@
...
+    dps:
+      - id: 103
+        name: switch
+        type: boolean
+  - entity: binary_sensor
+    category: diagnostic
+    class: problem
+    dps:
+      - id: 20
+        type: integer
```
make-all, line 85: Probably this is a bitfield.  It doesn't make much difference until you start trying to decode individual bits, but I like to use the correct types for future reference.

The values should be mapped to booleans for a binary_sensor
```
        mapping:
          - dps_val: 0
            value: false
          - value: true
``` 

As written above, adding the raw fault_code as a second attribute makes it easy to find.

custom_components/tuya_local/devices/forceclima_soundless.yaml
```diff
@@ -0,0 +1,86 @@
+name: ForceClima Soundless
+products:
+  - id: bf122f759ba4882810sdwc
```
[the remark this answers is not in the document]
tv4you2016: @make-all 
How do I get the ID?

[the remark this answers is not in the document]
make-all: You can find it in the Query Device Details tuya cloud API result,  Or it is output when scanning the local network, with tinytuya.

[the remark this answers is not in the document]
tv4you2016: @make-all 
`{
  "result": {
    "active_time": 1723727952,
    "bind_space_id": "34043412",
    "category": "ydkt",
    "create_time": 1723573857,
    "custom_name": "",
    "icon": "smart/icon/ay1573205693675sHuji/5c51f4f072e092afeab68bee1ccdea50.png",
    "id": "bf122f759ba4882810sdwc",
    "ip": "2.80.206.210",
    "is_online": false,
    "lat": "XXX",
    "local_key": "asdadsad",
    "lon": "-7.62",
    "model": "JHS",
    "name": "ForceClima Soundless",
    "product_id": "njvdabnnx51jiqiq",
    "product_name": "ForceClima Soundless",
    "sub": false,
    "time_zone": "+01:00",
    "update_time": 1723842715,
    "uuid": "a2904bc91a74103b"
  },
  "success": true,
  "t": 1724949388956,
  "tid": "d74e3aec662411ef88b02ac421ecbce6"
}`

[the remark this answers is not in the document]
make-all: "product_id": "njvdabnnx51jiqiq" << this one

tv4you2016:
@make-all 
check now

custom_components/tuya_local/devices/forceclima_soundless.yaml
```diff
@@ -48,16 +60,27 @@ primary_entity:
...
-      type: integer
       name: fault_code
+      type: bitfield
+      mapping:
+        - dps_val: 1
+          value: E1
+        - dps_val: 2
+          value: E2
+        - dps_val: 3
+          value: warter_full
```
make-all, line 73: As per previous comment, this should be attached to the problem sensor below, but I would also suggest adding a mapping to the sensor of `dps_val: 4` to `value: false`, and add an additional binary_sensor with `translation_key: tank_full` with a mapping of `dps_val: 4` to `value: true` and default mapping of `value: false`

tv4you2016: This one I don't understand what I have to change

custom_components/tuya_local/devices/forceclima_soundless.yaml
```diff
@@ -28,21 +28,15 @@ primary_entity:
...
-      mapping:
-        - scale: 0
-          step: 1
+        max: 31
       unit: C
 
     - id: 3
       type: integer
       name: current_temperature
       range:
```
make-all, line 37: Range will be ignored for current_temperature anyway, as it is read-only so will just follow what the device sends. But I think temperatures above and below the settable range can probably still be read from the sensor.

custom_components/tuya_local/devices/forceclima_soundless.yaml
make-all: OK, I will take care of this before merge

PR #2224, opened 2024-08-19
Add Duux Whisper Flex (ultimate) fan to devices

juliusvaart:


juliusvaart:
I fixed the YAML

custom_components/tuya_local/devices/duux_whisper_flex_ultimate_fan.yaml
```diff
@@ -0,0 +1,96 @@
...
+        - dps_val: "0"
+          value: normal
+        - dps_val: "1"
+          value: nature
+        - dps_val: "2"
+          value: sleep
+    - id: 3
+      type: integer
+      name: speed
+      mapping:
```
make-all, line 25: You should use range rather than a scale for fan speed.  The speed is now auto-scaled using https://developers.home-assistant.io/blog/2023/12/05/scaling-utils/

custom_components/tuya_local/devices/duux_whisper_flex_ultimate_fan.yaml
```diff
@@ -0,0 +1,96 @@
...
+    dps:
+      - id: 20
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 17
+          max: 28
+  - entity: select
+    name: Timer
```
make-all, line 64: Use `translation_key: timer` rather than name here.

PR #2229, opened 2024-08-21
Updated support for Windcalm fans without lights

hanzwo:
Updated the configuration for Windcalm Fans without lights. It wasn't able to match the device based on the defined entities.


make-all:
The change to expose the timer as an entity is an improvement, but it won't affect matching.

hanzwo:
I thought the same thing, but after changing this configuration locally it matched perfectly.

PR #2231, opened 2024-08-21
Add support for SPD800 (SP800 V2)

sj29fr:
Based on https://forum.hacf.fr/t/integration-micro-onduleur-stepuptech-spd800-via-tuya-local-recuperer-les-donnees-depuis-iot-tuya/44768


custom_components/tuya_local/devices/stepuptech_solar_inverter_v2.yaml
```diff
@@ -0,0 +1,109 @@
...
+      - id: 114
+        type: integer
+        name: sensor
+        unit: V
+        force: true
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: PV1 current
+    class: voltage
```
make-all, line 33: class should be `current`. Also for PV2 below.

PR #2233, opened 2024-08-23
Added Support for Littepets Cat Litter Box

cyrano1966:


custom_components/tuya_local/devices/littepets.yaml
```diff
@@ -0,0 +1,121 @@
...
+    category: diagnostic
+    dps:
+      - id: 24
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: standby
+            value: Standby 
+            icon: "mdi:timer"
+          - dps_val: off
```
make-all, line 93: This needs quotes to prevent yaml from interpreting it as a boolean false.

custom_components/tuya_local/devices/littepets.yaml
```diff
@@ -0,0 +1,121 @@
+name: Cat litter box
+products:
+  - id: 000004cha8
```
make-all, line 3: Is this id the product id? Or is it a model number?

custom_components/tuya_local/devices/littepets.yaml
```diff
@@ -0,0 +1,121 @@
...
+            value: Sleep
+            icon: "mdi:sleep"
+          - dps_val: uv
+            value: UV
+            icon: "mdi:sun-wireless"
+          - dps_val: deodorization
+            value: Deodorisation
+            icon: "mdi:spray"
+  - entity: switch
+    name: "Manual Clean"
```
make-all, line 39: Maybe should be a button?

custom_components/tuya_local/devices/littepets.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+  - entity: switch
+    name: "Sleep Mode"
+    icon: "mdi:sleep"
+    dps:
+      - id: 10
+        type: boolean
+        name: switch
+        
+  - entity: switch
```
make-all, line 54: Maybe should be a button? Or is this a setting for automatically deodorizing?

custom_components/tuya_local/devices/littepets.yaml
```diff
@@ -0,0 +1,121 @@
...
+        
+  - entity: sensor
+    name: Last visit duration
+    icon: "mdi:paper-roll"
+    category: diagnostic
+    dps:
+      - id: 8
+        type: integer
+        name: sensor
+        unit: seconds
```
make-all, line 79: Unit should be `s`, and the entity should be of `class: duration`

custom_components/tuya_local/devices/littepets.yaml
cyrano1966: Thanks for feedback and for being patient (first contribution to a project in Github). I believe it should be a switch as when actioned it remains on "Manual Clean State", while cleaning, and when completed, moves back to off. A button will not give that visual feedback and could be "pressed" several times

custom_components/tuya_local/devices/littepets.yaml
cyrano1966: Same as previous comment on "Manual Clean"

custom_components/tuya_local/devices/littepets.yaml
cyrano1966: Done

custom_components/tuya_local/devices/littepets.yaml
cyrano1966: From the logs:
"model": "{\"modelId\":\"000004cha8\"
I assumed it was the product ID. Any idea where to find the product ID?

make-all: Product id is usually listed in the same tuya cloud api calls that return the local key

cyrano1966: "product_id": "jdxbiz2ca8x2mbfk"
Now added to the yaml

PR #2234, opened 2024-08-23
Create topens_wifi_remote_control.yaml

reniko:
msedge_fbGJkgCbmT
Added configuration for Topens Wifi Remote Control.

https://topens.com/products/tc196-tuya-wifi-remote-control-for-gate-opener

All entities of the device are included and working.

### DIFF
--- a/custom_components/tuya_local/devices/topens_wifi_remote_control.yaml
+++ b/custom_components/tuya_local/devices/topens_wifi_remote_control.yaml
@@ -1,7 +1,7 @@
-name: TOPENS Wi-Fi Remote control
+name: Remote control
 products:
   - id: plgiarn7jndeu6js
-    name: TC196 WiFi Remote
+    name: TOPENS TC196
 primary_entity:
   entity: button
   name: A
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/topens_wifi_remote_control.yaml
```diff
@@ -0,0 +1,36 @@
+name: TOPENS Wi-Fi Remote control
```
make-all, line 1: Outside of the products section, naming should be generic, without branding.

custom_components/tuya_local/devices/topens_wifi_remote_control.yaml
```diff
@@ -0,0 +1,36 @@
...
+        type: boolean
+        name: button
+  - entity: button
+    name: D
+    dps:
+      - id: 104
+        type: boolean
+        name: button
+  - entity: sensor
+    name: Limit Status
```
make-all, line 32: What is this? Should it be `category: diagnostic`? Or even an attribute on another entity if it is unimportant.

PR #2238, opened 2024-08-24
Add support for Medion Life P1002

MarvinBeym:
Adds support for the **Medion Life P1002** portable air conditioner
(https://www.medion.com/de/shop/p/heizen--kuehlen-medion-life-p1002-klimaanlage-md-37733-kuehlen-entfeuchten-und-ventilieren-kuehlleistung-10-000-btu-kuehlmittel-r290-max-34m--app-und-sprachsteuerung-inkl-fenster-kit-50074744A1)

created file was based on the ``monzana_klimaanlage_air_conditioner.yaml`` file but :
- Removed heating capability
- Added swing control
- Added fitting icons to controls
- Added sleep control 
  - The device only allows uses this if it's in cooling mode, otherwise the control is ignored
     (Not sure if something like this can maybe be configured prettier in tuya-local to not even allow changing the state in home-assistant when the device is not in cooling mode?)

Not sure if something has to be changed for the **fault_code** dp

tuya api explorer reports:
```
{
        "code": "fault",
        "dp_id": 22,
        "type": "Bitmap",
        "values": "{\"label\":[\"Rtemp_fault\",\"IPtemp_fault\",\"Water_FULL\"]}"
      },
````

Don't know if those ``"values":`` have to be mapped somewhere?

### CODE REVIEW

File: custom_components/tuya_local/devices/medion_life_p1002.yaml [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: Cool
+              value: cool
+            - dps_val: Dyr
```
make-all, line 20: Just to confirm, this typo is in the tuya implementation?

custom_components/tuya_local/devices/medion_life_p1002.yaml
```diff
@@ -0,0 +1,81 @@
...
+      hidden: true
+    - id: 5
+      type: string
+      name: fan_mode
+      mapping:
+        - dps_val: Low
+          value: low
+        - dps_val: High
+          value: high
+    - id: 22
```
make-all, line 46: Move this to the problem binary_sensor below, as it will be more convenient/obvious to retrieve the fault_code from the same sensor that is indicating a fault.

custom_components/tuya_local/devices/medion_life_p1002.yaml
```diff
@@ -0,0 +1,81 @@
...
+            value: false
+          - value: true
+  - entity: switch
+    name: Sleep
+    icon: "mdi:power-sleep"
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+  - entity: switch
```
make-all, line 75: This can be moved into the climate entity as a swing_mode attribute.

custom_components/tuya_local/devices/medion_life_p1002.yaml
```diff
@@ -0,0 +1,81 @@
...
+    icon: "mdi:arrow-oscillating"
+    name: Swing
+    dps:
+      - id: 15
+        type: boolean
+        name: switch
```
make-all, line 81: Add a newline at the end of this line to complete the file

custom_components/tuya_local/devices/medion_life_p1002.yaml
MarvinBeym: yes

make-all (approved):

PR #2242, opened 2024-08-25
Add support for QCSmart garage door

RGarrido03:
This PR adds support for a garage door opener I bought on AliExpress. Unfortunately the device doesn't give much more information about available DPs (even using DP instruction on the Tuya API), so there is no state like "opening" or "closing".

### DIFF
--- a/custom_components/tuya_local/devices/qcsmart_garage_door.yaml
+++ b/custom_components/tuya_local/devices/qcsmart_garage_door.yaml
@@ -1,0 +1,25 @@
+name: QCSmart garage door
+products:
+  - id: bf68ea3165358a2846kahv
+    name: QCSmart garage door
+primary_entity:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/qcsmart_garage_door.yaml
```diff
@@ -0,0 +1,25 @@
+name: QCSmart garage door
+products:
+  - id: bf68ea3165358a2846kahv
```
make-all, line 3: This appears to be the device id (unique to your device), not the product id.

make-all (commented):
This does not appear to be any different than the generic garage_door_opener, so does not need to be a new config.

RGarrido03:
Oops I didn't see it 😆 sorry

PR #2247, opened 2024-08-27
Update ecopoint_ev_charger.yaml

mat1990dj:
Add support for Nine charger. verified DP IDs with tuya api explorer to match as well as the modes

### DIFF
--- a/custom_components/tuya_local/devices/ecopoint_ev_charger.yaml
+++ b/custom_components/tuya_local/devices/ecopoint_ev_charger.yaml
@@ -2,6 +2,8 @@ name: EV charger
 products:
   - id: jui3mys5c9rfdju2
     name: Ecopoint 32A EV charger
+  - id: bktb3jskdic1ar2t
+    name: Nine 32A EV charger
 primary_entity:
   entity: sensor
   name: Charger State
--- a/custom_components/tuya_local/devices/ecopoint_ev_charger.yaml
+++ b/custom_components/tuya_local/devices/ecopoint_ev_charger.yaml
@@ -37,6 +37,7 @@ primary_entity:
     - id: 22  # meter_id
       type: string
       name: meter_id
+      optional: true
     - id: 23  # system_version
       type: string
       name: system_version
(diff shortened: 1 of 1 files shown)


make-all:
The matching is by dps.  You can test it against the log message of actual dps received using the scripts in util.

`python util/best_match.py 'JSON DPS LOG'` will show the best matches for the device as it was seen in that log message.
`python util/match_against.py ecopoint_ev_charger.yaml 'JSON DPS LOG'` will show where the log is not matching the config.

mat1990dj:
> The matching is by dps. You can test it against the log message of actual dps received using the scripts in util.
> 
> `python util/best_match.py 'JSON DPS LOG'` will show the best matches for the device as it was seen in that log message. `python util/match_against.py ecopoint_ev_charger.yaml 'JSON DPS LOG'` will show where the log is not matching the config.

Do you have a guide on how to run those utils? I tried setting up the project dev env on my computer but it's complaining about No module named 'homeassistant'

mat1990dj:
Ok, I figured it out, I'm now matching correctly at 100% but I've got one issue, if you see the mapping on DPs 6-9 it's a base64 (CXQAUtAT2g== for instance, but this decodes to 09740052d013da, voltage is the first four 0974 but this is in HEX and not directly decimal, 0974 decodes to 2420 which is 242,0V. How can I indicate a mapping to do base64 AND HEX from the result?

feeling001:
Can I suggest an improvement ? 
The scale is incorrect on the "Charging Curent" entity.  scale 10 should be scale 1 (actually 16amp show as 1.6amp)
The range is also incorrect. from 6-80 to 6-32

Here is a correction.

 ```
 - entity: number
    name: Charging current
    category: config
    class: current
    icon: "mdi:ev-plug-type2"
    dps:
      - id: 4  # charge_cur_set
        type: integer
        name: value
        unit: A
        optional: true
        range:
          min: 6
          max: 32
        mapping:
          - scale: 1

```

Best regards.

PR #2254, opened 2024-08-29
Create nine_ev_charger.yaml

mat1990dj:
Sourced from tuya API and other EV chargers yamls in the repo.

### DIFF
--- a/custom_components/tuya_local/devices/nine_ev_charger.yaml
+++ b/custom_components/tuya_local/devices/nine_ev_charger.yaml
@@ -1,0 +1,239 @@
+name: EV charger
+products:
+  - id: bktb3jskdic1ar2t
+    name: Nine 32A EV charger
+primary_entity:
+  entity: sensor
+  name: Charger State
+  icon: "mdi:ev-station"
+  class: enum
+  dps:
+    - id: 3  # work_state
+      type: string
+      name: sensor
+      mapping:
+        - dps_val: charger_charging
+          value: Charging
+        - dps_val: charger_free
+          value: Available
+        - dps_val: charger_insert
+          value: Plugged in
+        - dps_val: charger_free_fault
+          value: Fault
+        - dps_val: charger_wait
+          value: Delaying
+        - dps_val: charger_pause
+          value: Paused
+        - dps_val: charger_end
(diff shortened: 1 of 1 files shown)


mat1990dj:
image

mat1990dj:
image

mat1990dj:
image

custom_components/tuya_local/devices/nine_ev_charger.yaml
```diff
@@ -0,0 +1,262 @@
...
+      - id: 4  # charge_cur_set
+        type: integer
+        name: value
+        unit: A
+        optional: true
+        range:
+          min: 6
+          max: 80
+        mapping:
+          - scale: 1
```
make-all, line 50: Is this a tuya `scale: 1`? Is the range 6-80A, or 0.6-8.0A?

custom_components/tuya_local/devices/nine_ev_charger.yaml
```diff
@@ -0,0 +1,262 @@
...
+    dps:
+      - id: 10  # fault
+        type: bitfield
+        name: sensor
+        optional: true
+        mapping:
+          - dps_val: 0
+            value: false
+          - dps_val: 1
+            value: "Ov2 Cr Fault"
```
make-all, line 75: binary_sensor mappings should be to booleans.
I suggest adding 2 extra attributes: fault_code with the raw integer code, and description with these mapped descriptions. See the recently added Cecotec config https://github.com/make-all/tuya-local/pull/2222/files#diff-8ca651a12d089c6826bf3817b0f031d1edd4938110267451db6d877c6bfab795R90

mat1990dj: Will do, thanks

custom_components/tuya_local/devices/nine_ev_charger.yaml
mat1990dj: Scale 1, from 6 to 80.

make-all: In that case, the mapping can be removed, 1 is the default scale.

PR #2262, opened 2024-08-31
Add Fujicool and Yuzu Heatpump

carmelo42:
This commit allows the usage of Fujicool and Yuzu Heatpump (self installable HVAZC sold in France). It may work with Airton Heatpumps too, but I can’t test it.

### DIFF
--- a/custom_components/tuya_local/devices/fujicool_yuzu_heatpump.yaml
+++ b/custom_components/tuya_local/devices/fujicool_yuzu_heatpump.yaml
@@ -1,0 +1,270 @@
+name: Fujicool-Yuzu Heatpump (F)
+primary_entity:
+  entity: climate
+  translation_only_key: aircon_extra
+  dps:
+    - id: 1
+      type: boolean
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: cold
+              value: cool
+            - dps_val: hot
+              value: heat
+            - dps_val: wet
+              value: dry
+            - dps_val: wind
+              value: fan_only
+            - dps_val: auto
+              value: heat_cool
+    - id: 2
+      type: integer
+      name: temperature
+      range:
+        min: 160
+        max: 260
+      mapping:
+        - scale: 10
+          step: 10
+      unit: C
+    - id: 4
+      type: string
+      name: mode
+      hidden: true
+    - id: 5
+      type: string
+      name: fan_mode
+      mapping:
+        - dps_val: auto
+          value: auto
+        - dps_val: mute
+          value: quiet
+        - dps_val: low
+          value: low
+        - dps_val: mid_low
+          value: medlow
+        - dps_val: mid
+          value: medium
+        - dps_val: mid_high
+          value: medhigh
+        - dps_val: high
+          value: high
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/fujicool_yuzu_heatpump.yaml
```diff
@@ -0,0 +1,270 @@
+name: Fujicool-Yuzu Heatpump (F)
```
make-all, line 1: This should be a generic name (such as Heatpump). The branding belongs in a name under products section, coupled with a product id (available the same place as local key, or when performing a local scan eg with tinytuya scan).

custom_components/tuya_local/devices/fujicool_yuzu_heatpump.yaml
```diff
@@ -0,0 +1,270 @@
...
+      type: boolean
+      name: hot_cool
+    - id: 133
+      type: string
+      name: swing_action
+    - id: 134
+      type: json
+      name: statistics
+secondary_entities:
+  - entity: sensor
```
make-all, line 129: This should be the current_temperature attribute of the climate entity rather than a separate sensor.

It seems you picked the one config to copy that is for a broken device that reports  C for current temperature, but expects F for set temperature.

custom_components/tuya_local/devices/fujicool_yuzu_heatpump.yaml
```diff
@@ -0,0 +1,270 @@
...
+    class: temperature
+    dps:
+      - id: 3
+        name: sensor
+        type: integer
+        unit: C
+        class: measurement
+  - entity: sensor
+    class: humidity
```
make-all, line 138: This can also be current_humidity attribute of the climate entity. Existing devices don't tend to follow this, as current_humidity is not surfaced in the standard climate card, but since recent changes in HA have made it easy to display any attribute in Tile and Entity cards, this isn't really a reason for a separate sensor any more.

custom_components/tuya_local/devices/fujicool_yuzu_heatpump.yaml
```diff
@@ -0,0 +1,270 @@
...
+          value: medlow
+        - dps_val: mid
+          value: medium
+        - dps_val: mid_high
+          value: medhigh
+        - dps_val: high
+          value: high
+        - dps_val: strong
+          value: strong
+    - id: 20
```
make-all, line 59: This can be made into a binary_sensor.

```yaml
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 20
        type: bitfield
        name: sensor
        mapping:
          - dps_val: 0
            value: false
          - value: true
      - id: 20
        type: bitfield
        name: fault_code
```

custom_components/tuya_local/devices/fujicool_yuzu_heatpump.yaml
```diff
@@ -0,0 +1,270 @@
...
+    - id: 119
+      type: string
+      name: electricity_management
+    - id: 120
+      type: string
+      name: gen_mode
+    - id: 123
+      type: hex
+      name: flags_2
+    - id: 125
```
make-all, line 106: This may be better attached as an attribute of the api sensor below.

PR #2277, opened 2024-09-05
Add support Dream Maker Feel Fan DM01

BenJamesAndo:
```
  "data": {
    "device_config": {
      "friendly_name": "Dream Maker Feel Fan",
      "device_id": "bfb4a77d847974992f2mrs",
      "protocol_version": "3.3",
      "model": "",
      "product_key": "lf3n6whotmyd1d1q",
      "dps_strings": [
        "1 ( code: switch , value: False )",
        "2 ( code: mode , value: Straight )",
        "3 ( code: fan_speed , value: 100 )",
        "5 ( code: switch_horizontal , value: True )",
        "7 ( code: fan_horizontal , value: 60 )",
        "13 ( code: fan_beep , value: False )",
        "14 ( code: child_lock , value: False )",
        "15 ( code: light , value: False )",
        "21 ( code: temp_current , value: 27 )",
        "24 ( code: fault , value: 0 )",
        "102 ( code: off_delay_time , value: 0 )",
        "103 ( code: relative_humidity , value: 58 )",
        "104 ( code: fan_left_right , value: NONE )"
      ],
```

image

image

image

### DIFF
--- a/custom_components/tuya_local/devices/dream_maker_feel_fan.yaml
+++ b/custom_components/tuya_local/devices/dream_maker_feel_fan.yaml
@@ -1,0 +1,130 @@
+name: Pedestal Fan
+products:
+  - id: lf3n6whotmyd1d1q
+    name: Dream Maker Feel Fan DM01
+primary_entity:
+  entity: fan
+  translation_only_key: fan_with_presets
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+    - id: 2
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: "Straight"
+          value: normal
+        - dps_val: "Natural"
+          value: nature
+        - dps_val: "AI"
+          value: smart
+    - id: 3
+      type: integer
+      name: speed
+      range:
+        min: 0
+        max: 100
+    - id: 5
+      type: boolean
+      name: oscillate
+
+secondary_entities:
+  - entity: select
+    name: Oscillation Angle
+    icon: mdi:arrow-oscillating
+    category: config
+    dps:
+      - id: 7
+        name: option
+        type: string
+        mapping:
+          - dps_val: 30
+            value: "30°"
+          - dps_val: 60
+            value: "60°"
+          - dps_val: 90
+            value: "90°"
+          - dps_val: 120
+            value: "120°"
+          - dps_val: 140
+            value: "140°"
+  - entity: switch
+    translation_key: indicator
+    name: Beep
+    icon: mdi:bell
+    category: config
+    dps:
+      - id: 13
+        type: boolean
+        name: switch
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 14
+        type: boolean
+        name: lock
+  - entity: switch
+    name: Lights
+    icon: mdi:lightbulb
+    category: config
+    dps:
+      - id: 15
+        type: boolean
+        name: switch
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 21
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/dream_maker_feel_fan.yaml
```diff
@@ -0,0 +1,130 @@
...
+          max: 480
+  - entity: sensor
+    class: humidity
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: select
```
BenJamesAndo, line 116: Ideally I'd have the manual turn left and right as buttons instead of select options, but I couldn't get it to work with
```
  - entity: button
    name: Turn left
    dps:
      - id: 104
        type: string
        name: button
        optional: true
        mapping:
          - dps_val: "LEFT"
            value: "Turn left"
```
It would create a button but clicking it wouldn't do anything.

make-all: value should be `true` for the dps_val you want to send when the button it pressed.

custom_components/tuya_local/devices/dream_maker_feel_fan.yaml
```diff
@@ -0,0 +1,137 @@
...
+        name: switch
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 14
+        type: boolean
+        name: lock
+  - entity: switch
+    name: Lights
```
make-all, line 69: It would be better to use a light entity than a switch called "Lights". No name or icon is then needed.
If the lights are indicators rather than a light used to illuminate the room, then a `category: config` and `translation_key: indicator` would be recommended

custom_components/tuya_local/devices/dream_maker_feel_fan.yaml
```diff
@@ -0,0 +1,137 @@
...
+    dps:
+      - id: 21
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    name: Fault Alarm
```
make-all, line 87: Avoid using a name for the only (or main) problem sensor on a device. Since the value is translated to a true/false boolean in a binary sensor, also include a second attribute with the same id, and the name "fault_code" and no mapping, so the original value is available if needed.

custom_components/tuya_local/devices/dream_maker_feel_fan.yaml
```diff
@@ -0,0 +1,137 @@
...
+    dps:
+      - id: 24
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - dps_val: 1
```
make-all, line 95: Suggest making this the default (no dps_val) to catch other bits also.

custom_components/tuya_local/devices/dream_maker_feel_fan.yaml
BenJamesAndo: I'm not even sure what fault codes it can output. There was no documentation about the fans error states. I've tried physically stopping it from oscillating, force stopping the blade from spinning etc. Maybe it's worth just removing it?

custom_components/tuya_local/devices/dream_maker_feel_fan.yaml
BenJamesAndo: As in
```
        mapping:
          - dps_val: 0
            value: false
          - value: true
```
?

Issue #2239, opened 2024-08-25
Support new Lawnmower RETURNING state


make-all:
From HA 2024.9, the lawn_mower platform will support RETURNING as a LawnMowerActivity state to indicate that the mower is returning to its dock, analogous to the vacuum RETURNING state.

https://developers.home-assistant.io/blog/2024/08/23/lawn-mower-activity-returning

PR #2278, opened 2024-09-05
Support new Lawnmower RETURNING state

chmielowiec:
Fixes https://github.com/make-all/tuya-local/issues/2239


make-all:
This will also force a minimum requirement of HA 2024.9.  I prefer to wait a while before doing this.

PR #2280, opened 2024-09-06
vevor_ev_charger: Add SS_EU15O_M2 support

alexandrebelloni:


custom_components/tuya_local/devices/vevor_ev_charger.yaml
```diff
@@ -116,6 +118,7 @@ secondary_entities:
       - id: 9
         type: integer
         name: sensor
+        optional: true
```
make-all, line 121: Suggest adding to this below (first two lines are already there, provided for context):
```
     ...
        mapping
          - scale: 1000
          - dps_val: null
            value_redirect: decoded_power
      - id: 6
        type: base64
        name: decoded_power
        optional: true
        mapping:
          - scale: 1000
            mask: "0000000000FFFFFF"
```

This should give you the power extracted from dp 6 as a backup when dp 9 is missing

make-all: I see the "Phase A power" sensor is available above already, so this change isn't necessary. You can just hide the non-working sensors in your local install.

alexandrebelloni: Well, I actually implemented your suggested change and I got power after this. For some reason HA doesn't seem to like the Phase A power sensor, it doesn't appear in the list of entities for the device. Shouldn't the class be power and the name "Phase A power"?

make-all: Yes, the sensor was configured incorrectly, which may have been leading to errors.

PR #2291, opened 2024-09-10
Update ble_smart_plant_moisture.yaml

almirus:
fix wrong scale for temp


make-all:
I have some of these SGS01 devices, and the current scale is correct. If there are devices with a different scale, then a new config is needed, not modifying the existing one. Probably the product details are also different.

almirus:
@make-all My mistake. I missed the file—I need to edit a different one. 🤓

PR #2305, opened 2024-09-15
feat[new_device]: add Ballu Aura AC

IATkachenko:


custom_components/tuya_local/devices/ballu_aura.yaml
```diff
@@ -0,0 +1,105 @@
...
+        type: boolean 
+        name: sensor
+  - entity: number
+    category: config
+    translation_key: timer
+    dps:
+      - id: 105
+        name: value 
+        type: integer 
+        unit: m
```
make-all, line 86: m is metres. For minutes, the unit should be "min"

custom_components/tuya_local/devices/ballu_aura.yaml
```diff
@@ -0,0 +1,105 @@
...
+          min: 0
+          max: 7200
+
+  - entity: sensor
+    category: diagnostic
+    name: Alarm
+    dps:
+      - id: 106
+        name: sensor
+        type: integer
```
make-all, line 97: It isn't clear what this sensor is.  If it is a fault code, then use the following standard sensor.

```
  - entity: binary_sensor
    category: diagnostic
    class: problem
    dps:
      - id: 106
        type: bitfield
        name: sensor
        mapping:
          - dps_val: 0
            value: false
          - value: true
      - id: 106
        type: bitfield
        name: fault_code
```

custom_components/tuya_local/devices/ballu_aura.yaml
```diff
@@ -0,0 +1,105 @@
...
+  - entity: sensor
+    category: diagnostic
+    name: Alarm
+    dps:
+      - id: 106
+        name: sensor
+        type: integer 
+  - entity: sensor
+    category: diagnostic
+    name: model
```
make-all, line 100: model should just be an attribute on the climate entity. It is not useful as a sensor I think (it will never change)

custom_components/tuya_local/devices/ballu_aura.yaml
```diff
@@ -0,0 +1,104 @@
...
+      type: string
+      mapping:
+        - dps_val: "1"
+          value: low
+        - dps_val: "2"
+          value: medium
+        - dps_val: "3"
+          value: high
+    - id: 19
+      name: temperature_unit
```
make-all, line 48: I think it is better to leave this out of here unless you know how the temperature unit setting functions. Often devices use different dps for F, or at least the temperature range will need adjusting when it is set to F. But in some devices it only affects the display, in which case setting the unit at entity level is wrong.

custom_components/tuya_local/devices/ballu_aura.yaml
```diff
@@ -0,0 +1,104 @@
+name: Ballu Aura
```
make-all, line 1: This top level name should be generic (Air conditioner).

PR #2313, opened 2024-09-17
Adding alternate version of tonepie cat litter box

vossim:


make-all:
It would be useful to understand what you mean by "alternate version".

In general, I don't want multiple personal preferences for configs for a single device in the repository. There are too many configs for distinct devices as it is.

vossim:
The device I've bought has a completely different set of IDs compared to the tonepie T1 pro that is in the repository already. Since my device was named tonepie T1 pro too,I named it "alternate".

PR #2332, opened 2024-09-25
Update stadlerform_roger_purifier.yaml

arnkit:
Added Child lock switch (ID: 7)
Changed Icon for Air quality (ID: 14) from mdi:air-filter to mdi:air-purifier
Updated timer (ID: 18) values (Added values: 3h, 5h, 6h, 7h)
Changed Icon for Auto-Mode (ID: 101) from mdi:power-sleep to mdi:fan-auto

### DIFF
--- a/custom_components/tuya_local/devices/stadlerform_roger_purifier.yaml
+++ b/custom_components/tuya_local/devices/stadlerform_roger_purifier.yaml
@@ -25,7 +25,16 @@ primary_entity:
         - dps_val: auto
           value: 100
 secondary_entities:
+  - entity: switch
+    name: Child lock
+    class: switch
+    icon: "mdi:lock-outline"
+    dps:
+      - id: 7
+        name: switch
+        type: boolean
   - entity: light
+    name: Light
     category: config
     dps:
       - id: 8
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/stadlerform_roger_purifier.yaml
```diff
@@ -25,7 +25,16 @@ primary_entity:
         - dps_val: auto
           value: 100
 secondary_entities:
+  - entity: switch
```
make-all, line 28: This should be a lock, following the pattern used for other child lock entities in other devices (using translation_key instead of explicit name and icons)

custom_components/tuya_local/devices/stadlerform_roger_purifier.yaml
```diff
@@ -25,7 +25,16 @@ primary_entity:
...
+    name: Child lock
+    class: switch
+    icon: "mdi:lock-outline"
+    dps:
+      - id: 7
+        name: switch
+        type: boolean
   - entity: light
+    name: Light
```
make-all, line 37: Entities should not have the same name as the entity type, as it shadows default translations.

PR #2335, opened 2024-09-27
Add support for Reiga ceiling fan without light

DennisGeisse:
Hi,

I have added a device description for [Reiga smart ceiling fans](https://www.reigafan.com/) that do not have an integrated light. Tested with the fan that I have installed at home ([this model](https://www.reigafan.com/product/65-inch-wood-blade-ceiling-fan/)).

In addition, I propose to change the naming of a similar file, the one that describes fans from the same manufacturer with integrated lights (reiga_52_fan_light.yaml). The naming of the previously existing file suggests that it is only valid for fans with this specific size of the blades (52 inch), which I consider an unnecessary limitation of the scope. Like with the Windcalm fans I think it is fair to just differentiate between fans with or without light. 

Apologies if my edit to DEVICES.md messed up whitespaces in two locations, these were not intended.

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -243,7 +243,7 @@
 - Orison RGB ambient bladeless ceiling fan
 - Ovlaim ceiling fan with cool/warm white dimmable light
 - Princess DC pedestal fan
-- Reiga 52 ceiling fan with light
+- Reiga ceiling fans with and without light
 - Royal Clima RCB 150 ventilation system
 - Skyfan DC fan
 - Skyfan DC fan with light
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/reiga_ceiling_fan_with_light.yaml
make-all, line 1: Renaming files after release requires a migration so that users who already configured the device do not end up with broken configs.  I think it is better to not rename it.

DennisGeisse: I didn't realize that a migration would be need, I have removed the renaming so now only the new definition is added.

PR #2357, opened 2024-09-30
Add support for Windmill AC

ianjsikes:
Older models of the [WIndmill AC](https://www.amazon.com/Windmill-Conditioner-Smart-Home-Unit/dp/B09TM28Q98/ref=sr_1_6) run on Tuya. The current models run on Blynk (I'm not sure when exactly they made the switch, I haven't been able to find model IDs, but the one I bought in 2022 is Tuya). Here's my attempt at creating a config for this device. I got the values by adding the AC as a Simple Switch in the tuya-local integration, and checking the device state in HA while cycling through all possible options.

### DIFF
--- a/custom_components/tuya_local/devices/windmill_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/windmill_airconditioner.yaml
@@ -1,0 +1,53 @@
+name: Windmill Air Conditioner
+products:
+  - id: ebd7ef1c98deabe827aukn
+    name: Windmill Air Conditioner
+primary_entity:
+  entity: climate
+  dps:
+    - id: 1
+      name: hvac_mode
+      type: boolean
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: wind
+              value: fan_only
+            - dps_val: cool
+              value: cool
+            - dps_val: eco
+              value: auto
+    - id: 4
+      name: mode
+      type: string
+      hidden: true
+    - id: 5
+      name: fan_mode
+      type: string
+      mapping:
+        - dps_val: low
+          value: low
+        - dps_val: mid
+          value: medium
+        - dps_val: high
+          value: high
+        - dps_val: auto
+          value: auto
+    # Reported by device, seemingly unused
+    # - id: 22
+    - id: 23
+      name: current_temperature
+      type: integer
+      readonly: true
+    - id: 24
+      name: temperature
+      type: integer
+      range:
+        min: 60
+        max: 86
+      unit: F
+    # Reported by device, seemingly unused
+    # - id: 102


custom_components/tuya_local/devices/windmill_airconditioner.yaml
```diff
@@ -0,0 +1,53 @@
...
+      readonly: true
+    - id: 24
+      name: temperature
+      type: integer
+      range:
+        min: 60
+        max: 86
+      unit: F
+    # Reported by device, seemingly unused
+    # - id: 102
```
ianjsikes, line 53: I left these unknown params in as comments just to have them documented, and in case their use is discovered in the future. But I can take them out no problem.

custom_components/tuya_local/devices/windmill_airconditioner.yaml
```diff
@@ -0,0 +1,53 @@
...
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: wind
+              value: fan_only
+            - dps_val: cool
+              value: cool
+            - dps_val: eco
+              value: auto
```
ianjsikes, line 22: I understand that there is an "eco" preset, and it may make sense to map that to this instead. But for this device, these options (wind, cool, eco) are mutually exclusive, so IDK if it makes sense to have them be separate properties that could be set at the same time. Plus, all the "eco" mode does is automatically turn the cooling off (but keep the fan on) when the temp is at the target temp, so "auto" seems appropriate.

custom_components/tuya_local/devices/windmill_airconditioner.yaml
make-all: If you know the type (from logs), and they are consistently present in the data returned from the device, then it would be better to have them listed as attributes of the primary device (with names unknown_22, unknown_102) because it will help with differentiating this device from others that may be similar but not quite the same. It will also allow monitoring them from HA for anyone motivated to try to figure out what they are for (though getting access to iot.tuya.com to look up the QueryThingsDataModel result for the device is likely to be quicker than guessing by trial and error).

custom_components/tuya_local/devices/windmill_airconditioner.yaml
make-all: This is good. Better to handle it along with the other hvac_mode if it is mutually exclusive with those on the device itself.

custom_components/tuya_local/devices/windmill_airconditioner.yaml
ianjsikes: Makes sense. I updated the config to add those as unknown data points.

custom_components/tuya_local/devices/windmill_airconditioner.yaml
```diff
@@ -0,0 +1,55 @@
+name: Windmill Air Conditioner
+products:
+  - id: ebd7ef1c98deabe827aukn
```
make-all, line 3: THis appears to be in the format used for device id, rather than product id.

custom_components/tuya_local/devices/windmill_airconditioner.yaml
```diff
@@ -0,0 +1,55 @@
+name: Windmill Air Conditioner
```
make-all, line 1: Top level name should be without branding, as often other devices will also match the config.

custom_components/tuya_local/devices/windmill_airconditioner.yaml
ianjsikes: Could you point me to where I could find the product ID? I found this "model ID" in the tuya iot "Query Things Data	Model" response `00000049xm` but I'm not sure if that's right either.

make-all: No, that it also not right (I'm not sure what that is for, and I think it is only available from the cloud API).

The product id is listed in the API calls that return the local key, and is available during device discovery locally, though we currently only use the device discovery via tinytuya library to find IP addresses.

ianjsikes: @make-all found it. Anything else needed to get this merged?

PR #2359, opened 2024-09-30
Add LSC Smart Ceiling Light

a-r-j-a-n:
Initial commit for the LSC Smart Ceiling Light.

Sold in The Netherlands by [Action](https://www.action.com/nl-nl/p/3206306/lsc-smart-connect-slimme-plafondlamp/).

Has some very specific ways of control, because it shows as two separate lights.

### DIFF
--- a/custom_components/tuya_local/devices/lsc_smart_ceiling_light.yaml
+++ b/custom_components/tuya_local/devices/lsc_smart_ceiling_light.yaml
@@ -27,7 +27,7 @@ primary_entity:
         - target_range:
             min: 2700
             max: 6500
-  color_mode: false # Geen RGB-kleuren in deze modus
+  color_mode: white
   supported_features:
     - brightness
     - color_temp
@@ -70,7 +70,6 @@ secondary_entities:
     supported_features:
       - brightness
       - rgb
-      #- effect
 
   - entity: switch
     name: Light off after power restore
--- a/custom_components/tuya_local/devices/lsc_smart_ceiling_light.yaml
+++ b/custom_components/tuya_local/devices/lsc_smart_ceiling_light.yaml
@@ -1,0 +1,236 @@
+name: CCT + RGB Ceiling Light
+products:
+  - id: bwyutsrcsnwf3p5o
+    name: LSC Smart Connect Smart Ceiling Light
+
+primary_entity:
+  entity: light
+  name: Main light
+  icon: mdi:circle
+  dps:
+    - id: 63 # White Light Control Switch
+      type: boolean
+      name: switch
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/lsc_smart_ceiling_light.yaml
```diff
@@ -0,0 +1,228 @@
...
+            bytes: 2
+            range:
+              min: 0
+              max: 1000
+          - name: v
+            bytes: 2
+            range:
+              min: 0
+              max: 1000
+      - id: 24
```
make-all, line 58: I think brightness should not be needed, as this light is only supporting color mode, so the v portion of rgbhsv above will be used for brightness.

custom_components/tuya_local/devices/lsc_smart_ceiling_light.yaml
```diff
@@ -0,0 +1,228 @@
...
+        type: hex
+        name: brightness
+        range:
+          min: 0
+          max: 1000
+        mapping:
+          - mask: "00000000FFFF"
+
+  - entity: switch
```
make-all, line 67: This might be better as a select with translation key initial_state, and mappings true: "off", false: "on" (or false: memory, depending on actual behaviour)

custom_components/tuya_local/devices/lsc_smart_ceiling_light.yaml
a-r-j-a-n: No, the V value is ignored by the light when changed by Home assistant and only works when sending it separately. Took me hours to figure out.

custom_components/tuya_local/devices/lsc_smart_ceiling_light.yaml
a-r-j-a-n: Makes sense, will try this and update the pull req.

a-r-j-a-n: Changed. Had to change it from a boolean to a select. "initial_state" doesn't work on a boolean.

custom_components/tuya_local/devices/lsc_smart_ceiling_light.yaml
make-all: Yes, but the code is already there to map the separate brightness onto the rgbhsv field when a light is in color mode.

a-r-j-a-n: I know, but that part is ignored when changing the brightness. It doesn't affect the lights when changed in Home Assistant.

PR #2367, opened 2024-10-02
Create kesser_infrared_heater_floodlights.yaml

eldi0s:
Hi,

there is a version of the Kesser Infrared Heater with integrated floodlights. I have it working as per the code below.


eldi0s:
Just discovered that the dps_val value was wrong and not working. Not sure if this also applies for the standard infrared heater without the floodlights but very liekely it does.

PR #2374, opened 2024-10-03
Create kogan_kasmgphXXXX_heater.yaml

willieseabrook:
Add support for the white 1.5kw Kogan Smarterhome KASMGPH15YA. This is tested and working on KASMGPH15YA.

While I have not tested, since the user guide appears to be the same for all the series, this would likely work with all KASMGPH series heaters. E.g.

1.5kw - KASMGPH15YA (White) and KASMGPH15YB (Black) 
2.0kw - KASMGPH20YA (White) and KASMGPH20YB (Black) 
2.4kw - KASMGPH24YA (White) and KASMGPH24YB (Black)


willieseabrook:
Updated to try to fix the yaml linter error regarding readonly

Issue #2350, opened 2024-09-28
Siren WiFi STL ESP_65B419


VladimirTuzovGitHub:
### Log message

```text
Device matches zigbee_mir_te100ty_climate_sensor with quality of 100%. DPS: {"updated_at": 1727594590.9474046, "101": "4", "102": "3", "103": 15}

```

### DPS information

```
{"modelId":"0000000dn6","services":[{"actions":[],"code":"","description":"","events":[],"name":"默认服务","properties":[
{"abilityId":101,"accessMode":"ro","code":"BatteryStatus","description":"0：满电\n1：高\n2：中\n3：低\n4：外电","name":"电量状态","typeSpec":{"type":"enum","range":["0","1","2","3","4"]}},
{"abilityId":102,"accessMode":"rw","code":"Alarmtype","description":"每一种代表一种声音","name":"声音类型","typeSpec":{"type":"enum","range":["1","2","3","4","5","6","7","8","9","10"]}},
{"abilityId":103,"accessMode":"rw","code":"AlarmPeriod","description":"","name":"报警时长","typeSpec":{"type":"value","max":60,"min":0,"scale":0,"step":1,"unit":"秒"}},
{"abilityId":104,"accessMode":"rw","code":"AlarmSwitch","description":"","name":"报警开关","typeSpec":{"type":"bool"}}]}]}
```

<details>

```text
{
  "result": [
    {
      "active_time": 1702320185,
      "bind_space_id": "52420419",
      "category": "sgbj",
      "create_time": 1702283264,
      "custom_name": "",
      "icon": "smart/icon/1528364694thahb9sok2l_0.png",
      "id": "4825104784f3eb65b419",
      "ip": "188.124.246.***",
      "is_online": true,
      "lat": "49.8200",
      "local_key": "Zr;E3a_h=`OJ479***",
      "lon": "73.1200",
      "model": "",
      "name": "Siren",
      "product_id": "DYgId0sz6zWlmmYu",
      "product_name": "声光报警器",
      "sub": false,
      "time_zone": "+06:00",
      "update_time": 1712131102,
      "uuid": "4825104784f3eb65b419"
    }
  ],
  "success": true,
  "t": 1727548017173,
  "tid": "3e70a8377dc711efb77736033862a1ba"
}

{
  "result": [
    {
      "id": "4825104784f3eb65b419",
      "status": [
        {
          "code": "BatteryStatus",
          "value": "4"
        },
        {
          "code": "Alarmtype",
[...]

PR #2384, opened 2024-10-11
Update stl_siren.yaml

VladimirTuzovGitHub:
The final integration is fully tested. Please update this code. Thank you very much for your support!

### DIFF
--- a/custom_components/tuya_local/devices/stl_siren.yaml
+++ b/custom_components/tuya_local/devices/stl_siren.yaml
@@ -1,7 +1,7 @@
 name: Siren
 products:
   - id: DYgId0sz6zWlmmYu
-    name: STL siren alarm
+    name: STL siren Type
 primary_entity:
   entity: siren
   dps:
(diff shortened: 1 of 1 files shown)


make-all:
I don't see this as an improvement, sorry.

VladimirTuzovGitHub:
Hello! The improvements are that your code does not have the setting for selecting the siren sound and setting the siren sound time. We discussed this in the topic [https://github.com/make-all/tuya-local/issues/2350](url). I assume that these settings are important in the integration so that any user can configure them from the HA interface and does not need to do it through the Tuya application, which is very inconvenient.

make-all:
The settings are there in the siren entity.

VladimirTuzovGitHub:
I am talking about basic settings that will be available to the average user in a human-readable interface, and not through event or parameter calls in automations. In your configuration, in the very essence of the serena, there are only attributes of melodies and time in the form of a list.

Issue #2349, opened 2024-09-28
Kiturami room thermostat for gas heating boiler and hot water preparation.


VladimirTuzovGitHub:
### Log message

```text
Device matches desk_lamp with quality of 30%. DPS: {"updated_at": 1727538326.4204578, "1": true, "2": "Interior", "3": 60, "5": 22, "6": 23, "20": 0, "101": 20, "102": 2, "103": 53, "104": 0}
```

### DPS information

```text
My tuya iot subscription has expired and I need to write a letter to support to get access, which takes a lot of time. I will try to provide the data that I have, since I have previously integrated it through LocalTuya integration, which I really don't like now.
```

### Product ID

bfc6d1959b288a680ak5ga

### Product Name

Kiturami thermostat

### Information about how the device functions

![image](https://github.com/user-attachments/assets/8131b5e8-8a21-4f1b-ad58-152975754566)
![image](https://github.com/user-attachments/assets/b2c0b1f7-120a-48dc-8a11-f33652ad8882)
![image](https://github.com/user-attachments/assets/28d58bc7-942b-423c-82e0-894756e1eca6)
![image](https://github.com/user-attachments/assets/2c371af5-b8a1-4bab-bb61-c25e8a3c3dd4)
![image](https://github.com/user-attachments/assets/cf1c04cd-6dd3-41e6-bfe0-ffefa13e04cb)
![image](https://github.com/user-attachments/assets/f39eea51-953b-4ef3-a3b8-8853c538eb2f)
![image](https://github.com/user-attachments/assets/a7551e8d-c891-43e4-9930-c21e13aff8e7)
![image](https://github.com/user-attachments/assets/909bc397-cf9a-4a77-9feb-23c988be99be)
![image](https://github.com/user-attachments/assets/d566d57e-731f-4856-8810-0b0df9e7ddbe)
![image](https://github.com/user-attachments/assets/cd370bce-f08e-48df-a72f-5916e8a6d140)
![image](https://github.com/user-attachments/assets/6c882e04-45ad-4c76-a870-feb6b63b24db)
![image](https://github.com/user-attachments/assets/443299b3-01f8-4df3-8378-f3af78da421b)
![image](https://github.com/user-attachments/assets/15ed2baa-be29-4161-bedc-48c17fc19d70)
![Screenshot_20240928-213426](https://github.com/user-attachments/assets/f18240cf-df50-4d9c-ac06-bbea46832fdf)
[...]

PR #2385, opened 2024-10-11
Update kiturami_nctr100_thermostat.yaml

VladimirTuzovGitHub:
The final integration is fully tested. Please update this code. Thank you very much for your support!


make-all:
I don't see this as an improvement. Mostly it is just renaming attributes for no apparent reason, and dismantling the water_heater into primitive entities.

VladimirTuzovGitHub:
Hello! Well, I explained and showed you in the topic [https://github.com/make-all/tuya-local/issues/2349](url) that you implemented it incorrectly and that the device does not correspond to water_heater, separate settings for each function are required there. I explained to you that this is not a room thermostat, this is a remote boiler control module. Your initial configuration does not work correctly and I showed you this and even provided a translation of the settings from the official device manual. You apparently initially misunderstood me, apparently my translator does not translate correctly into English.

PR #2393, opened 2024-10-15
Update faroro_pet_feeder.yaml

jon-sanders:
Corrected value - upon reinstallation this dp had to be set to optional for the device to be recognized.


make-all:
Yes, the persist: false also implies that it is sometimes missing, and should be reflected as unavailable then.

Issue #2108, opened 2024-07-13
Request support for EV charger [AFYEEV 32A 22kw]


jmcollin78:
Hello, here is new device file for adding a AFYEEV 32A 22kw EV charger.
It could be great to integrate it directly into your devices directory.

```
name: EV charger
products:
  - id: aaaealnecwgwerjl
    name: AFYEEV 32A EV charger
primary_entity:
  entity: sensor
  class: enum
  name: Status
  icon: "mdi:ev-station"
  dps:
    - id: 3
      type: string
      name: sensor
      mapping:
        - dps_val: charger_free
          value: Disconnected
        - dps_val: charger_insert
          value: Connected
        - dps_val: charger_free_fault
          value: Fault (disconnected)
        - dps_val: charger_charging
          value: Charging
        - dps_val: charger_wait
          value: Waiting
        - dps_val: charger_end
          value: Charged
        - dps_val: charger_fault
          value: Fault
        - dps_val: charger_pause
          value: Pause
secondary_entities:
  - entity: number
    name: Set current
    icon: "mdi:ev-plug-type2"
    dps:
      - id: 4
        type: integer
        name: value
        unit: A
        range:
          min: 6
          max: 16
  - entity: sensor
    icon: "mdi:flash"
    name: Single phase power
    class: power
    dps:
      - id: 5
        type: integer
        name: sensor
        unit: W
        class: measurement
        precision: 1
        mapping:
          - scale: 3
  - entity: sensor
    icon: "mdi:flash"
    name: Total power
    class: power
    dps:
      - id: 9
        type: integer
        name: sensor
        unit: W
        class: measurement
        precision: 1
        mapping:
          - scale: 1
  - entity: sensor
    name: Fault reason
    icon: 'mdi:alert-outline'
    class: enum
    dps:
      - id: 10
        type: bitfield
        name: sensor
        mapping:
          - dps_val: 0
            value: "Ready"
          - dps_val: 1
            value: "Ov2 Cr Fault"
          - dps_val: 2
            value: "OV Vol"
          - dps_val: 4
            value: "Undervoltage alarm"
          - dps_val: 8
            value: "Contactor adhesion"
          - dps_val: 16
            value: "Contactor fault"
          - dps_val: 32
            value: "Earth fault"
          - dps_val: 64
            value: "Meter Hardware alarm"
          - dps_val: 128
            value: "Scram fault"
          - dps_val: 256
            value: "CP fault"
          - dps_val: 512
[...]

PR #2395, opened 2024-10-15
Add support for AFYEEV 16A portable EV charger

pimpalaputty:
This is very similar to the 32A charger (#2108), but there is a difference in the base64 masking, so I created the specific device file for this model.

The charger is available at [Aliexpress](https://www.aliexpress.com/item/1005006826040824.html).

### DIFF
--- a/custom_components/tuya_local/devices/afyeev_16a_evcharger.yaml
+++ b/custom_components/tuya_local/devices/afyeev_16a_evcharger.yaml
@@ -1,0 +1,199 @@
+name: EV charger
+products:
+  - id: "bfda6008360ed3949fleb0"
+    name: "Afyeev 16A EV Charger"
+primary_entity:
+  entity: sensor
+  name: Charger State
+  icon: "mdi:ev-station"
+  class: enum
+  dps:
+    - id: 3
+      type: string
+      name: sensor
+      mapping:
+        - dps_val: charger_charging
+          value: Charging
+        - dps_val: charger_free
+          value: Available
+        - dps_val: charger_insert
+          value: Plugged in
+        - dps_val: charger_free_fault
+          value: Fault
+        - dps_val: charger_wait
+          value: Delaying
+        - dps_val: charger_pause
+          value: Paused
+        - dps_val: charger_end
+          value: Charged
+        - dps_val: charger_fault
+          value: Fault (plugged in)
+    - id: 23
+      type: string
+      name: system_version
+secondary_entities:
+  - entity: sensor
+    name: Total Forward Energy
+    class: energy
+    dps:
+      - id: 1
+        type: integer
+        name: sensor
+        unit: kWh
+        optional: true
+        class: total_increasing
+        mapping:
+          - scale: 100
+  - entity: number
+    name: Charging current
+    category: config
+    class: current
+    icon: "mdi:ev-plug-type2"
+    dps:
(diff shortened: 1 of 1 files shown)


pimpalaputty:
### Product name
AFYEEV 16A Portable EV Charger

### Product id
bfda6008360ed3949fleb0

### Log message

>   {"updated_at": 1728928026.3962665, "1": 17552, "3": "charger_charging", "4": 16, "5": 0, "9": 0, "10": 0, "14": "charge_now", "18": true, "23": "2.0.0", "24": 23, "25": 1671, "27": "online", "28": 0, "33":"Charge mode"}

### Screenshot
2024-10-15 12 09 58 ha kuvik hu dd51b3c6223b

custom_components/tuya_local/devices/afyeev_16a_evcharger.yaml
```diff
@@ -0,0 +1,198 @@
...
+    dps:
+      - id: 6
+        type: base64
+        name: sensor
+        optional: true
+        unit: A
+        mapping:
+          - mask: "000000FFFF000000"
+            scale: 1000
+  - entity: sensor
```
make-all, line 84: I would just add this as an attribute under the power sensor below, since it is effectively a duplication. If this one provides more reliable data, you can swap them.

custom_components/tuya_local/devices/afyeev_16a_evcharger.yaml
```diff
@@ -0,0 +1,198 @@
...
+    dps:
+      - id: 4
+        type: integer
+        name: value
+        unit: A
+        range:
+          min: 6
+          max: 16
+  - entity: sensor
+    name: Phase A Voltage
```
make-all, line 61: Since this is the only voltage sensor, remove the name so the translations for class: voltage can be used.

custom_components/tuya_local/devices/afyeev_16a_evcharger.yaml
```diff
@@ -0,0 +1,198 @@
...
+      - id: 6
+        type: base64
+        name: sensor
+        optional: true
+        unit: V
+        mapping:
+          - mask: "FFFF000000000000"
+            scale: 10
+  - entity: sensor
+    name: Phase A Current
```
make-all, line 73: Since this is the only current sensor, remove the name so the translations for class: current can be used.

layanto:
@pimpalaputty does this unit allow for current adjustment while charging? I want to set up Home Assistant to automatically adjust current during charging session to use excess solar. Or do I need to stop charging, adjust current, then start charging again everytime I want to change current during charging?

pimpalaputty:
> @pimpalaputty does this unit allow for current adjustment while charging? I want to set up Home Assistant to automatically adjust current during charging session to use excess solar. Or do I need to stop charging, adjust current, then start charging again everytime I want to change current during charging?

I just tested it and you can adjust the current from the Home Assistant during charging. I can see the change immediately in the Smart Life app. However, my charger gave an 'overcurrent' error and stopped charging. I repeated this twice more with the same result. I had to unplug the charger from the car and also from the wall socket.

layanto:
@pimpalaputty Thanks. That's not good. This means the unit doesn't support changing current during charging session. Probably needs to stop charging, change current and restart charging to dynamically adjust charging current limit.

PR #2409, opened 2024-10-17
Add support for Point Panel Heater Series.

coutaq:


custom_components/tuya_local/devices/point_panel_heater.yaml
```diff
@@ -0,0 +1,51 @@
...
+    dps:
+      - id: 7
+        type: boolean
+        name: lock
+  - entity: switch
+    translation_key: display_auto
+    category: config
+    dps:
+      - id: 109
+        type: boolean
```
make-all, line 51: `name: switch`

perhaps even more of your config got cut off here?

custom_components/tuya_local/devices/point_panel_heater.yaml
```diff
@@ -0,0 +1,51 @@
...
+secondary_entities:
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 7
+        type: boolean
+        name: lock
+  - entity: switch
+    translation_key: display_auto
```
make-all, line 47: There is currently no translation key for "display_auto". What does this switch do? Perhaps it is better as a light with translation key `display`?

custom_components/tuya_local/devices/point_panel_heater.yaml
```diff
@@ -0,0 +1,51 @@
+name: Point Panel Heater
```
make-all, line 1: Make this just "Panel heater", in case other brands of device also match.

custom_components/tuya_local/devices/point_panel_heater.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - dps_val: true
+          value: "heat"
+        - dps_val: false
+          value: "off"
+    - id: 2
+      name: temperature
+      type: integer
+      unit: C
+      mapping:
+        - scale: 0.1
```
make-all: This will result in a range of 50 to 500 degrees, with 10 degree steps.  Is it what you intended?

More likely the scale should be 10, and the range min: 50, max: 500

PR #2421, opened 2024-10-20
Add TDS sensor to Yinmik water quality sensor

markxroberts:
2 commits as there were some automatic .vscode changes I have reverted.  Only material changes are to device

### DIFF
--- a/custom_components/tuya_local/devices/yinmik_waterquality_tester.yaml
+++ b/custom_components/tuya_local/devices/yinmik_waterquality_tester.yaml
@@ -16,6 +16,20 @@ primary_entity:
         - scale: 10
 secondary_entities:
   - entity: sensor
+    name: Total dissolved solids
+    icon: "mdi:water-opacity"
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/yinmik_waterquality_tester.yaml
```diff
@@ -15,6 +15,20 @@ primary_entity:
...
+    name: Total dissolved solids
+    icon: "mdi:water-opacity"
+    dps:
+      - id: 1
+        type: integer
+        name: sensor
+        class: measurement
+        unit: ppm
+        optional: true
+        mapping:
```
make-all, line 28: Range should not be inside a mapping. min and max should be inside the range, however.

make-all: Or since this is a read-only sensor, the range is not needed at all

PR #2426, opened 2024-10-22
Update door_peephole_camera.yaml

Jon811:
In case of my camera now has a motion sensor, add it's entity to set.

### DIFF
--- a/custom_components/tuya_local/devices/door_peephole_camera.yaml
+++ b/custom_components/tuya_local/devices/door_peephole_camera.yaml
@@ -143,6 +143,6 @@ secondary_entities:
             value: clear
           - value: detected
       - id: 115
-        type: string
+        type: base64
         name: image_details
         optional: true
--- a/custom_components/tuya_local/devices/door_peephole_camera.yaml
+++ b/custom_components/tuya_local/devices/door_peephole_camera.yaml
@@ -13,12 +13,6 @@ primary_entity:
       type: boolean
       name: record
       optional: true
-    - id: 115
-      type: base64
-      optional: true
-      persist: false
-      sensitive: true
-      name: snapshot
 secondary_entities:
   - entity: switch
     name: Flip image
@@ -136,15 +130,19 @@ secondary_entities:
             value: Event
           - dps_val: "2"
             value: Continuous
-  - entity: binary_sensor
+  - entity: event
     class: motion
     category: diagnostic
     dps:
       - id: 115
-        type: bitfield
-        name: sensor
-        persist: true
+        type: string
+        name: event
+        optional: true
         mapping:
-          - dps_val: Any
-            value: true
-
+          - dps_val: null
+            value: clear
+          - value: detected
+      - id: 115
+        type: string
+        name: image_details
+        optional: true
--- a/custom_components/tuya_local/devices/door_peephole_camera.yaml
+++ b/custom_components/tuya_local/devices/door_peephole_camera.yaml
@@ -136,3 +136,15 @@ secondary_entities:
             value: Event
           - dps_val: "2"
             value: Continuous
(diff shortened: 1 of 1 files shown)


Jon811:
When there is a movement outside, there is a message in Tuya dp:

2024-10-23 11:41:37 | Report | Motion Detection | DlkLTE5NTk1NTIxNi1od3hkOGI5M2I5OTg5M2JmNWNkMi9kZXRlY3QvMTcyOTY3Mjg5My5qcGVnIiwiZDVmY2U0OWFjZmE5ZGRmNyJdXX0= | device itself |  
So it attaches some link or something else for Tuya system. What else do i need to provide? On my system it works fine.

custom_components/tuya_local/devices/door_peephole_camera.yaml
```diff
@@ -136,3 +136,15 @@ secondary_entities:
...
           - dps_val: "2"
             value: Continuous
+  - entity: binary_sensor
+    class: motion
+    category: diagnostic
+    dps:
+      - id: 115
+        type: bitfield
+        name: sensor
+        persist: true
```
make-all, line 146: This may be better as an event than a binary_sensor.
If kept as a binary sensor, what is the mapping for false?
`persist: true` is the default behaviour, and not expected to be specified (this is what the error below is saying).

Jon811: Actually, i am not so good in Tuya, but this works for me. If you can propose something other config, i can check on my side. When something happens outside door Tuya send message like this:
2024-10-23 11:41:37 | Report | Motion Detection | cmFnZTMwLXBpYyIsImZpbGVzIjpbWyIvMDg2ODlkLTE5NTk1NTIxNi1od3hkOGI5M2I5OTg5M2JmNWNkMi9kZXRlY3QvMTcyOTY3Mjg5My5qcGVnIiwiZDVmY2U0OWFjZmE5ZGRmNyJdXX0= | device itself | 
And each time string is different. May be event is better, but how to config it?

make-all: The above decodes to `rage30-pic","files":[["/08689d-195955216-hwxd8b93b99893bf5cd2/detect/1729672893.jpeg","d5fce49acfa9ddf7"]]}`

It looks like a relative link to a camera snapshot (in a JSON structure that seems to be cropped at the start)

It should work like:
```yaml
  - entity: event
    class: motion
    dps:
      - id: 115
        type: string
        optional: true
        name: event
        mapping:
          - dps_val: null
            value: clear
          - value: detected
      - id: 115
        type: base64
        optional: true
        name: image_link
```

If the above spams the event queue with clear events when there is no motion, then changing "clear" to null will fix that, except you won't see any event when motion stops being detected.

Jon811: Actually i don't know how to check for "clear" messages in event queue, but it works fine for movement detection. Could you please add it to next update.

Jon811: And may be to all cameras, that have this function

make-all: I don't know which cameras "have this function".

Some cameras return an encoded image in this motion detect dp, in which case it is better to add it to the camera entity snapshot property. But for cameras that return a relative URL, the rest of the URL and any authentication details is needed to make use of it. I don't know whether this camera runs its own webserver to serve such images, or serves them via the Tuya cloud, so all I can do is make the value of the dp available. Also, since it is cut off JSON, it is a bit difficult to parse and make further use of. I assume the cloud API returns the full JSON, and this is due to some memory constraint on the device.

PR #2436, opened 2024-10-25
Add support for Solar Inverter GTB series

SimonAnnetts:
# Product name:
Smart GTB-800 PV Micro inverter

# Product id:
bf27badb1abd9bfc5beeuq

# Where to buy:
https://www.aliexpress.com/item/1005007385290819.html

# Also known as:
Solar Grid, and with models GTB-300 GTB-350 GTB-400 GTB-500 GTB-600 GTB-700 GTB-800 etc up to GTB-2000

Data values for AC and DC current/voltage/power and frequency cannot be asked for. Must wait for them to be sent by the device when a value changes. DC values use dp:3 AC values use dp:7 and are base 64 encoded range of bytes. Other data is the same as most other solar inverters.

image

Pull request created after dark so energy readings showing 0...

### DIFF
--- a/custom_components/tuya_local/devices/solar_inverter_gtb_series.yaml
+++ b/custom_components/tuya_local/devices/solar_inverter_gtb_series.yaml
@@ -1,7 +1,7 @@
-name: Solar Inverter GTB Series
+name: Solar inverter
 products:
   - id: czu7ynosrkxgbnsb
-    name: GTB Series
+    name: Solar Grid GTB Series
 primary_entity:
   entity: sensor
   class: energy
@@ -28,7 +28,6 @@ secondary_entities:
         type: integer
         name: sensor
         unit: W
-        force: true
         class: measurement
   - entity: sensor
     class: temperature
@@ -38,7 +37,6 @@ secondary_entities:
         type: integer
         name: sensor
         unit: C
-        force: true
         class: measurement
         mapping:
           - scale: 10
--- a/custom_components/tuya_local/devices/solar_inverter_gtb_series.yaml
+++ b/custom_components/tuya_local/devices/solar_inverter_gtb_series.yaml
@@ -1,0 +1,161 @@
+name: Smart Inverter GTB Series
+products:
+  - id: bf27badb1abd9bfc5beeuq
+    name: GTB Series
+primary_entity:
+  entity: sensor
+  class: energy
+  dps:
+    - id: 2
+      type: integer
+      name: sensor
+      unit: kWh
+      class: total_increasing
+      mapping:
+        - scale: 100
+    - id: 15
+      name: model
+      type: string
+    - id: 16
+      name: inverter_id
+      type: string
+secondary_entities:
+  - entity: sensor
+    class: power
+    category: diagnostic
+    dps:
+      - id: 10
+        type: integer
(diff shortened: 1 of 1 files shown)


make-all:
Product id is not the same as device id. Can you check that again?

If you have tinytuya installed, you can use `python -m tinytuya scan` to check (may require device to not be configured in tuya-local and mobile app to not be running if the device only accepts one local connection).

custom_components/tuya_local/devices/solar_inverter_gtb_series.yaml
```diff
@@ -0,0 +1,161 @@
...
+    name: DC voltage
+    class: voltage
+    category: diagnostic
+    dps:
+      - id: 3
+        type: base64
+        name: sensor
+        unit: V
+        optional: true
+        persist: true
```
make-all, line 57: persisting is default behaviour. Only use the persist option if you want to set it to false.

custom_components/tuya_local/devices/solar_inverter_gtb_series.yaml
```diff
@@ -0,0 +1,161 @@
+name: Smart Inverter GTB Series
```
make-all, line 1: Top level name should be generic - such as "Solar inverter".
Stay away from redundant words like "smart" in the naming (they can go in the name under products if they are part of the branding)

custom_components/tuya_local/devices/solar_inverter_gtb_series.yaml
```diff
@@ -0,0 +1,161 @@
...
+    class: power
+    category: diagnostic
+    dps:
+      - id: 10
+        type: integer
+        name: sensor
+        unit: W
+        force: true
+        class: measurement
+        mapping:
```
make-all, line 33: There is no need to include a mapping for a scale of 1 (no scaling). But if you copied that from the tuya docs for the device, it should probably be `scale: 10`, as tuya scales are number of decimal places to shift rather than the divisors that this integration uses.

custom_components/tuya_local/devices/solar_inverter_gtb_series.yaml
```diff
@@ -0,0 +1,161 @@
...
+        name: sensor
+        unit: Hz
+        optional: true
+        persist: true
+        class: measurement
+        mapping:
+          - mask: "0000000000000000FFFF"
+            scale: 10
+  - entity: switch
+    name: Inverter switch
```
make-all, line 155: Since this is the only switch, it probably does not need a name.

SimonAnnetts:
The device uses protocol version 3.4

custom_components/tuya_local/devices/solar_inverter_gtb_series.yaml
```diff
@@ -0,0 +1,150 @@
...
+secondary_entities:
+  - entity: sensor
+    class: power
+    category: diagnostic
+    dps:
+      - id: 10
+        type: integer
+        name: sensor
+        unit: W
+        force: true
```
make-all, line 31: Normally `force: true` should only be needed when the device does not send this regularly. In those cases, `optional: true` would also be needed to ensure the config can be matched reliably.

PR #2438, opened 2024-10-26
Update smartplugv2_energy.yaml

Estcool9:
For the energy sensor, change the class to energy, so that it's compatible with home assistant. I'm a beginner tell me if there are any problems I just want my plug connected to be compatible for energy in home assistant, sorry for my spelling.d

### DIFF
--- a/custom_components/tuya_local/devices/smartplugv2_energy.yaml
+++ b/custom_components/tuya_local/devices/smartplugv2_energy.yaml
@@ -51,11 +51,13 @@ secondary_entities:
   - entity: sensor
     category: diagnostic
     name: Energy
+    class: energy
     dps:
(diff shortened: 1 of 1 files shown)


make-all:
Most Tuya devices do not return useful information in this dp. They reset it to zero every time they report the value to the cloud. It is better for such devices to use an integration helper to calculate the energy based on the power sensor.

Estcool9:
Hello, and thank you for your response and help! I understand now that the direct energy data from the plug may not be reliable. I'll try using an integration helper with the power sensor to calculate the energy. If you have any guidance on setting this up or can point me to any helpful documentation, I'd really appreciate it. Thanks again for your patience as I’m just starting out!

PR #2441, opened 2024-10-28
Add support for Moes thermostat WHT-009

yurkiv:
### Product name:
Moes BHT-009 Smart Knob Thermostat

### Where to buy:
https://moeshouse.com/products/wifi-programmable-thermostat-room-temperature-controller-for-water-electric-gas-boiler-floor-heating

### Product ID:
14bmxdarlb4ravgd

### dps
```
"codes": {
    "1": "Switch",
    "2": "Mode",
    "3": "Working status",
    "4": "Eco mode",
    "16": "Set temperature",
    "18": "Max. set temp. -℉",
    "19": "Max. set temp.",
    "20": "Min. setting temp. -℉",
    "24": "Current temperature",
    "26": "Min. setting temp.",
    "27": "Temp. compensation",
    "36": "Output state",
    "40": "Child lock",
    "71": "Week Program",
    "101": "Floor temp.",
    "102": "Sensor type",
    "103": "Dead zone temp.",
    "104": "High protection temp. setting",
    "105": "Low protection temp. setting",
    "106": "Eco mode temp. setting"
}
```
### Screenshot:
Screenshot 2024-10-28 at 22 42 59

### DIFF
--- a/custom_components/tuya_local/devices/moes_wht009_thermostat.yaml
+++ b/custom_components/tuya_local/devices/moes_wht009_thermostat.yaml
@@ -1,9 +1,10 @@
-name: Moes thermostat WHT-009
+name: Thermostat
 products:
   - id: 14bmxdarlb4ravgd
-    name: Moes thermostat WHT-009
+    name: Moes WHT-009
 primary_entity:
   entity: climate
+  translation_key: thermostat
   dps:
     - id: 1
       name: hvac_mode
@@ -18,9 +19,9 @@ primary_entity:
       type: string
       mapping:
         - dps_val: manual
-          value: Manual
+          value: manual
         - dps_val: auto
-          value: Auto
+          value: program
     - id: 16
       name: temperature
       type: integer
@@ -78,18 +79,6 @@ secondary_entities:
         mapping:
           - scale: 10
             step: 10
-  - entity: sensor
-    name: Current temperature
-    class: temperature
-    dps:
-      - id: 24
-        type: integer
-        name: sensor
-        class: measurement
-        unit: C
-        mapping:
-          - scale: 10
-            step: 5
   - entity: number
     name: Minimum temperature
     class: temperature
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/moes_wht009_thermostat.yaml
```diff
@@ -0,0 +1,226 @@
...
+        name: value
+        optional: true
+        unit: C
+        range:
+          min: 150
+          max: 450
+        mapping:
+          - scale: 10
+            step: 10
+  - entity: sensor
```
make-all, line 81: There is no need to repeat the current_temperature and hvac_action as separate sensors.

custom_components/tuya_local/devices/moes_wht009_thermostat.yaml
```diff
@@ -0,0 +1,226 @@
...
+        optional: true
+        unit: C
+        range:
+          min: 0
+          max: 100
+        mapping:
+          - scale: 10
+            step: 10
+  - entity: number
+    name: Eco mode temperature setting
```
make-all, line 211: The last three names seem a bit long. This one can be "Eco temperature", I am not sure about the two previous ones, as the meaning is not exactly clear, maybe Anti-frost temperature for the low one, if I am understanding it correctly, but I am not sure about what the high one is for.

custom_components/tuya_local/devices/moes_wht009_thermostat.yaml
```diff
@@ -0,0 +1,226 @@
+name: Moes thermostat WHT-009
```
make-all, line 1: This top level name should be without branding (branding belongs under products, as often configs may match multiple devices).

make-all:
Note that several dps listed are not in the config. Notably "4": "Eco mode", which seems necessary to make use of "106" "Eco mode temperate setting".

Since there is no log, and no further info on dps type and range, I cannot change this for you, but it would be better if all dps were included in the config, even if some need to be optional.

Issue #2457, opened 2024-11-04
BUG not detected : Zigbee Thaleos radiator TRV06 AVATTO Tuya-Vanne thermostatique intelligente #2435


ciscolyon69:
### Description

It is not detected https://github.com/make-all/tuya-local/issues/2435

### Steps to reproduce

1) add a new
2) offer only zigbee_waterleak_sensor

### Expected behaviour

avatto_trv06_radiatorvalve.yaml 
        "name": "Bureau",
        "id": "bfee147eaefa55410cba3a",
        "key": "7DA*?c6TN>kkk]#6",
        "mac": "a4:c1:38:99:dc:f0:42:11",
        "uuid": "a4c13899dcf04211",
        "sn": "1001608330253D",
        "category": "wkf",
        "product_name": "TRV06UItra",
        "product_id": "ow09xlxm",
        "biz_type": 18,
        "model": "ME167",

### Additional context

2024-11-04 17:42:06.188 INFO (MainThread) [custom_components.tuya_local.device] Setting protocol version for Test to 3.4
2024-11-04 17:42:11.640 WARNING (MainThread) [custom_components.tuya_local.config_flow] Adding Wireless Zigbee&BLE Gateway device with product id nqs9dfd15xhnykeg
2024-11-04 17:42:11.640 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches zigbee_waterleak_sensor with quality of 50%. DPS: {"updated_at": 1730738526.277745, "4": 80, "5": 170}
2024-11-04 17:42:11.641 WARNING (MainThread) [custom_components.tuya_local.config_flow] Include the previous log messages with any new device request to https://github.com/make-all/tuya-local/issues/

of course I check I have 2024.11.0 and

 ll /config/custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml 
-rw-r--r--    1 root     root         942 Nov  4 17:15 /config/custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml

### Confirmation

- [X] I am sure this is a bug or improvement, that is well enough described that it can be implemented.
- [X] If this is for a specific device, the device config is mentioned above, or diagnostics are included.

- [X] Log messages or diagnostics relevant to the issue are included.

- [X] This is not requesting addition of a new device.

ciscolyon69:
I maid a new test
`mv /config/custom_components/tuya_local/devices/zigbee_waterleak_sensor.yaml /config/billou/`
```
[...]

PR #2465, opened 2024-11-05
avatto_trv06_radiatorvalve.yaml addon …  Issue #2457

ciscolyon69:
addon Battery frost_protection scale_protection local_temperature_calibration Issue #2457
I copied the code from other device.
please check my code and tell me if you see a better way ?

### DIFF
--- a/custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml
+++ b/custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml
@@ -23,6 +23,7 @@ primary_entity:
     - id: 3
       type: string
       name: hvac_action
+      optional: true
       mapping:
         - dps_val: opened
           value: heating
--- a/custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml
+++ b/custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml
@@ -99,7 +99,7 @@ secondary_entities:
         type: integer
         optional: true
         name: value
-        unit: C
+        unit: °
         range:
           min: -9
           max: 9
--- a/custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml
+++ b/custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml
@@ -52,3 +52,53 @@ secondary_entities:
         type: boolean
         optional: true
         name: lock
+  - entity: binary_sensor
+    class: battery
+    category: diagnostic
+    dps:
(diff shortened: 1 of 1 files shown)


ciscolyon69:
I need help to finish 

[custom_components.tuya_local.device] Bureau received {"28": "AwYAANIIAACgDAAA0g4AAKASAADSFgAAoA==", "full_poll": false}
I know the solution = "schedule_monday": "06:00/21.0 08:00/16.0 12:00/21.0 14:00/16.0",    
'HH:MM/C HH:MM/C HH:MM/C HH:MM/C HH:MM/C HH:MM/C'
how to put it in HA?
https://github.com/Koenkk/zigbee-herdsman-converters/issues/8062#issuecomment-2390856949

it is not the solution
```
  - entity: select             
    name: Schedule             
    category: config           
    icon: "mdi:calendar-clock" 
    dps:                       
      - id: 28                 
        type: string           
        name: option           
        optional: true         
        mapping:               
          - dps_val: "1"       
            value: "Weekday+Weekend"
          - dps_val: "2"            
            value: "Mon-Sat+Sun"    
          - dps_val: "3"            
            value: "Daily"          
```

custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml
```diff
@@ -52,3 +53,53 @@ secondary_entities:
...
+  - entity: number
+    name: Temperature calibration
+    category: config
+    icon: "mdi:thermometer-check"
+    dps:
+      - id: 47
+        type: integer
+        optional: true
+        name: value
+        unit: C
```
make-all, line 102: Unit needs to be ° since we need to avoid unit conversion in the case the user has their display preference in HA set to F, so that 0 is always 0.

custom_components/tuya_local/devices/avatto_trv06_radiatorvalve.yaml
```diff
@@ -52,3 +53,53 @@ secondary_entities:
...
+    category: diagnostic
+    dps:
+      - id: 35
+        type: bitfield
+        name: sensor
+        class: measurement
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
```
make-all, line 67: In other devices that use a bitfield to indicate low battery, the low battery indication is a specific bit, and there will often be other bits indicating other things like charging. To avoid falsely indicating a low battery, it would be better to identify which bit is indicating low battery, and map that bit to true, with the default mapping being false.

PR #2479, opened 2024-11-08
Corrected/renamed some names and entities to match with matsee_2way_energymeter.

giacomitti:
Based on matsee_2way_energymeter device
Changed sensor names and entities to match with matsee_2way_energymeter.


make-all:
While I think these names are improvements, once a device is released, changing the entity names can be disruptive and requires a migration to be written.  I would rather do this as part of migrating to use translations for such entity names (but some of these ones require entity parameters, which are not yet supported by this integration).

In any case, there are merge conflicts with this branch, I think it needs to be rebased without your changes there first, and then reapply your changes.

giacomitti:
OK, I will do it.

PR #2481, opened 2024-11-09
Corrected/renamed some names and entities to match with matsee_2way_energymeter.

giacomitti:
Based on matsee_2way_energymeter device
Changed sensor names and entities to match with matsee_2way_energymeter.


giacomitti:
I deleted the device and try to add again, but it´s not showed on the list. Any tip to try to fix it?

image

make-all:
Leaving this open to remind me of the translation change to complete the job.

PR #2482, opened 2024-11-09
Update afyeev_evcharger.yaml - make dp id 1 optional

jmcollin78:
The dpid 1 is not present on my device (I'm the one who create this file)

### DIFF
--- a/custom_components/tuya_local/devices/afyeev_evcharger.yaml
+++ b/custom_components/tuya_local/devices/afyeev_evcharger.yaml
@@ -42,6 +42,7 @@ secondary_entities:
   - entity: sensor
     class: energy
     category: diagnostic
+    optional: true
     dps:
       - id: 1
         type: integer


custom_components/tuya_local/devices/afyeev_evcharger.yaml
```diff
@@ -42,6 +42,7 @@ secondary_entities:
   - entity: sensor
     class: energy
     category: diagnostic
+    optional: true
```
make-all, line 45: It is the dp that needs to be marked optional, not the entity.

jmcollin78: ok, even with that I cannot make it work now. I have to understand what have been changed and what is blocking on my side.

make-all:
It may be easier if you send the log message that lists the dps when it mismatches.

jmcollin78:
> It may be easier if you send the log message that lists the dps when it mismatches.

Curiously there is no log at all. Just all the device is disabled. So i try to restaure my old config step by step until I found the 'optinal: true' but in the wrong place. I though this have fix but finally not. I have to redo the analysis step by step (and no logs ???)

make-all:
Marking a dp as optional only affects the initial configuration when the device is being matched to the config. It seems the problem you are seeing is happening after installation.

make-all:
Closing as no changes remain, and whatever you have done is not improving things. As suggested, it may be better if you delete your device and set it up again from scratch to get the required log message that will show what is coming from the device.

Issue #2483, opened 2024-11-09
Request support for Plikc Neve Pro


Tbaile:
### Log message

```text
{
  "updated_at": 1731018943.3954709,
  "1": true,
  "2": "auto",
  "3": "off",
  "16": 150,
  "19": 210,
  "24": 193,
  "26": 50,
  "27": 0,
  "33": 96,
  "39": true,
  "40": false,
  "101": 1,
  "102": 97,
  "103": 50,
  "104": 3,
  "108": "hc_heat",
  "111": false,
  "118": false
}
```

### Product ID

9myqvomgygt9zgcs

### Product Name

Neve PRO W

### DPS information

```text
{
  "modelId": "000004lekg",
  "services": [
    {
      "actions": [],
      "code": "",
      "description": "",
      "events": [],
      "name": "默认服务",
      "properties": [
        {
          "abilityId": 1,
          "accessMode": "rw",
          "code": "switch",
          "description": "",
          "extensions": {
            "iconName": "icon-dp_power3"
          },
          "name": "开关",
          "typeSpec": {
            "type": "bool"
          }
        },
        {
          "abilityId": 2,
          "accessMode": "rw",
          "code": "mode",
          "description": "自动模式，假日模式，手动模式，暂时模式",
          "extensions": {
            "attribute": "128"
          },
          "name": "工作模式",
          "typeSpec": {
            "type": "enum",
            "range": [
              "auto",
              "holiday",
              "manual",
              "temporary"
            ]
          }
        },
        {
          "abilityId": 3,
          "accessMode": "ro",
          "code": "work_state",
          "description": "继电器状态",
          "extensions": {
            "attribute": "128"
          },
          "name": "工作状态",
          "typeSpec": {
            "type": "enum",
            "range": [
              "off",
              "heat",
              "cool"
            ]
          }
        },
        {
          "abilityId": 16,
          "accessMode": "rw",
          "code": "temp_set",
          "description": "由于防冻模式最新0.5度。 防冻是设置可以0.5",
          "extensions": {
            "iconName": "icon-dp_temp",
            "trigger": "direct"
          },
          "name": "目标温度",
          "typeSpec": {
            "type": "value",
            "max": 350,
            "min": 5,
            "scale": 1,
            "step": 1,
            "unit": "℃"
          }
        },
        {
          "abilityId": 19,
          "accessMode": "rw",
          "code": "upper_temp",
          "description": "",
          "extensions": {
            "iconName": "icon-dp_c",
[...]

PR #2485, opened 2024-11-09
Add support for Plikc Neve Pro W thermostat

Tbaile:
Issue reference: https://github.com/make-all/tuya-local/issues/2483

### DIFF
--- a/custom_components/tuya_local/devices/plikc_neve_pro_thermostat.yaml
+++ b/custom_components/tuya_local/devices/plikc_neve_pro_thermostat.yaml
@@ -1,0 +1,176 @@
+name: Thermostat
+products:
+  - id: 9myqvomgygt9zgcs
+    name: Plikc Neve Pro W
+primary_entity:
+  entity: climate
+  translation_key: thermostat
+  dps:
+    - id: 1
+      type: boolean
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          value: heat_cool
+    - id: 2
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: auto
+          value: program
+        - dps_val: manual
+          value: manual
+        - dps_val: temporary
+          value: temp_override
+          hidden: true
+        - dps_val: holiday
+          value: away
+          constraint: holiday_duration
+          conditions:
+            - dps_val: 0
+              value: "Set holiday length!"
+              hidden: true
+    - id: 3
+      type: string
+      name: hvac_action
+      mapping:
+        - dps_val: heat
+          value: heating
+        - dps_val: cold
+          value: cooling
+        - dps_val: "off"
+          constraint: hvac_mode
+          conditions:
+            - dps_val: true
+              value: idle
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/plikc_neve_pro_thermostat.yaml
```diff
@@ -0,0 +1,176 @@
...
+              value: "off"
+    - id: 16
+      type: integer
+      name: temperature
+      unit: C
+      range:
+        min: 5
+        max: 350
+      mapping:
+        - scale: 1
```
make-all, line 58: Scale I think should be 10.  Tuya uses a different definition than tuya-local, so you cannot just copy from tuya info.
| Tuya | Tuya-local |
|----|----|
| 0 | 1 (default, so not needed) |
| x  | 2 (not representable in tuya format,  but used by some devices) |
| 1 | 10 |
| 2 | 100 | 

etc

custom_components/tuya_local/devices/plikc_neve_pro_thermostat.yaml
```diff
@@ -0,0 +1,176 @@
...
+          constraint: preset_mode
+          conditions:
+            - dps_val: holiday
+              value_redirect: holiday_temp_set
+    - id: 19
+      type: integer
+      name: upper_temperature
```
make-all, line 65: I think this should be `max_temperature`

custom_components/tuya_local/devices/plikc_neve_pro_thermostat.yaml
```diff
@@ -0,0 +1,176 @@
...
+      unit: C
```
make-all, line 66: Range and unit not needed here, as max_temperature is read-only for the climate entity. To change it requires a separate number entity.

custom_components/tuya_local/devices/plikc_neve_pro_thermostat.yaml
```diff
@@ -0,0 +1,176 @@
...
+      range:
+        min: 50
+        max: 350
+      mapping:
+        - scale: 1
+          constraint: preset_mode
```
make-all, line 72: I don't think this should redirect to the holiday mode temperature. It is just the maximum temperature for the thermostat, and applies regardless of mode.

custom_components/tuya_local/devices/plikc_neve_pro_thermostat.yaml
```diff
@@ -0,0 +1,176 @@
...
+            - dps_val: holiday
+              value_redirect: holiday_temp_set
+    - id: 24
+      type: integer
+      name: current_temperature
+      mapping:
+        - scale: 1
+    - id: 26
+      type: integer
+      name: lower_temperature
```
make-all, line 83: I think this should be `min_temperature`. Otherwise, same comments as above apply.

make-all (commented):
I have left some comments for learning purposes.

But regarding the submission, I prefer to merge this support into the other Plikc Neve config, expanding it as necessary.

PR #2497, opened 2024-11-13
Add support for Nince RGB LED Strip

AppelBoomHD:
Added support for the Nince RGB LED Strip. This device uses one dps for both RGB color data and brightness, which is why I've implemented a mask to get the brightness.

Tested and working on local Home Assistant setup ✅


custom_components/tuya_local/devices/nince_rgb_led_strip.yaml
```diff
@@ -0,0 +1,158 @@
...
+          value: "Holi"
+        - dps_val: "265a5a020014006403e800000000464602000003e803e800000000"
+          value: "Victory Day"
+        - dps_val: "275a5a020014006403e800000000464602000003e803e800000000\
+            323202015e01f403e800000000464602011303e803e800000000"
+          value: "Easter"
+        - dps_val: "28464601011303e803e800000000464601001e03e803e800000000"
+          value: "Halloween"
+    - id: 22
+      name: brightness
```
make-all, line 105: This is duplicated with the dp 24 brightness below. If the light strip only supports controlling via dp 24, then I think removing colour_mode and BOTH brightness dp definitions should work.  The bulb will then always be in hs mode (which I think is correct, even when a scene is selected), and the code that automatically controls brightness through the v parameter of rgbhsv will be used.

custom_components/tuya_local/devices/nince_rgb_led_strip.yaml
```diff
@@ -0,0 +1,158 @@
...
+  entity: light
+  dps:
+    - id: 20
+      type: boolean
+      name: switch
+    - id: 25
+      type: string
+      name: color_mode
+      mapping:
+        - dps_val: colour
```
make-all, line 15: This does not seem correct. The other values are all hex encoded, but this one is "colour"?

It seems you have moved dp 25 into the place of dp 21 but left this value here. 
I think dp 25 would be better as a separate "Scene" select entity, similar to other devices or as "effect" attribute of light.

AppelBoomHD:
@make-all Thanks for the feedback!
I've gotten it to work following your advice, with the exception of removing the color_mode dps. Removing it caused effects to not work properly. I changed it back to dps 21, which fixed the issue.
I also took some time to change the effect strings to exactly match the ones from the device, since some scenes weren't fucntioning correctly before.

PR #2521, opened 2024-11-17
Add HLG Infinity Alarm

brunohenriquy:
Added support for an unrecognized Alarm Central (HLG Infinity).

- Product
Screenshot 2024-11-16 at 21 24 07

- Functions
Screenshot 2024-11-16 at 21 27 27
Screenshot 2024-11-16 at 23 11 51


custom_components/tuya_local/devices/hlg_infinity_alarm.yaml
```diff
@@ -0,0 +1,332 @@
...
+    name: Power Line
+    icon: "mdi:power-plug"
+    category: diagnostic
+    dps:
+      - id: 115
+        type: boolean
+        name: sensor
+        mapping:
+          - dps_val: true
+            value: "Power Line"
```
make-all, line 266: Values should be `true` and `false` for binary sensors (no mapping required) and a `class: plug` would be appropriate here. The class defines an icon, and translated name, which might be better than explicitly hardcoding them.

custom_components/tuya_local/devices/hlg_infinity_alarm.yaml
```diff
@@ -0,0 +1,332 @@
...
+          - dps_val: false
+            value: "Battery"
+  - entity: binary_sensor
+    name: Battery Low Sector 1
+    icon: "mdi:battery-alert"
+    category: diagnostic
```
make-all, line 272: `class: battery` would be appropriate here.
The class defines an icon, so the icon can be removed.

custom_components/tuya_local/devices/hlg_infinity_alarm.yaml
```diff
@@ -0,0 +1,332 @@
+name: HLG Infinity Alarm
```
make-all, line 1: In case of other branded devices matching, this should be a generic name, and you should add a `products` section below this with the `id` from product_id (in tuya portal, reported by `python -m tinytuya scan`, or from the logs when using cloud-assisted setup) and the brand and model as `name`.

make-all (commented):
A lot of the sensors here seem to be uninteresting values that would never change. Such values might be better as attributes under another entity to reduce clutter in the UI.

custom_components/tuya_local/devices/hlg_infinity_alarm.yaml
brunohenriquy: This seems to be a very specific and custom project made by a Brazilian company, even the enum values are in Portuguese, that's why I think It won't be used by other companies, and that's why I made it specific. Do you still think it's a good idea to make it generic? If so I will adjust it.

make-all: Yes, all configs have the same pattern of putting branding under products only, only some older ones have not been updated.

brunohenriquy: Ohh ok, gotcha!

brunohenriquy:
> A lot of the sensors here seem to be uninteresting values that would never change. Such values might be better as attributes under another entity to reduce clutter in the UI.

How can I do it? I didn't understand very well!

custom_components/tuya_local/devices/hlg_infinity_alarm.yaml
brunohenriquy: I've used an specific battery icon, the alert one, I would like to keep it, is that ok?

custom_components/tuya_local/devices/hlg_infinity_alarm.yaml
brunohenriquy: Fixed!

custom_components/tuya_local/devices/hlg_infinity_alarm.yaml
make-all: If you look at the docs at https://www.home-assistant.io/integrations/binary_sensor/, the battery class will use a different icon depending on whether it is indicating Low or Normal. So by overriding the icon, you are overriding that built-in logic.

brunohenriquy: Gotcha! Thanks!

brunohenriquy: Fixed!

custom_components/tuya_local/devices/hlg_infinity_alarm.yaml
brunohenriquy: Fixed!

make-all:
To add things as extra attributes, you just add them under the dps list with a name that is not handled specially by the entity.
For sensors, only "sensor" and "unit" have special handling, so any other name will be added as an extra attribute.

brunohenriquy:
> To add things as extra attributes, you just add them under the dps list with a name that is not handled specially by the entity. For sensors, only "sensor" and "unit" have special handling plus the general things like "name", "device_class" etc, so any other name will be added as an extra attribute.

Ohhhh Sweet! Thanks!

brunohenriquy:
@make-all All adjustments have been made, can we proceed to merge it?

make-all:
Thank you for pointing out the file name. It should not be this generic, as this is not a generic device template.

make-all:
The config file name just needs to be unique, so another device matching it does not really matter. But making it too generic when it is not could lead to confusion.

brunohenriquy:
> The config file name just needs to be unique, so another device matching it does not really matter. But making it too generic when it is not could lead to confusion.

Gotcha!! Thanks! xD

PR #2524, opened 2024-11-19
Add support for Friedrich Uni-Fit Air Conditioners (models UCT14A30, UCT12A30, UCT08B10A) and Richelieu Tunable White LED Pucks (166142030)

ncd7:
Add support for
* Friedrich Unit-Fit Air Conditioners (models: UCT14A30, UCT12A30, UCT08B10A)
* Richelieu Tunable White LED Pucks (166142030)

### DIFF
--- a/custom_components/tuya_local/devices/friedrich_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/friedrich_airconditioner.yaml
@@ -63,7 +63,7 @@ primary_entity:
     - id: 102
       type: boolean
       # not settable, and no known effect
-      name: info_temperature_unit
+      name: info_temp_unit
       mapping:
         - dps_val: true
           value: °F


custom_components/tuya_local/devices/friedrich_airconditioner.yaml
```diff
@@ -0,0 +1,136 @@
...
+    icon: "mdi:sleep"
+    translation_key: sleep
+    category: config
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+  - entity: sensor
+    # false is C, true is F
+    name: "Configured Temperature Unit"
```
make-all, line 78: This is not very useful here as an informational sensor. If the value is  C then the climate entity will not be working properly, as the range and unit there is hard coded to F. If it is read-only, then it would be better just including it as the temperature_unit dp on the climate entity, and setting the range conditionally.

custom_components/tuya_local/devices/friedrich_uct08b10a_airconditioner
```diff
@@ -0,0 +1,520 @@
...
+    - id: 3
+      type: integer
+      name: current_temperature_in_c
+      optional: true
+      unit: C
+    - id: 23
+      name: current_temperature
+      type: integer
+      optional: true
+      hidden: true
```
make-all, line 45: You cannot hide an attribute that is part of an entity's official API.

custom_components/tuya_local/devices/friedrich_uct08b10a_airconditioner
```diff
@@ -0,0 +1,520 @@
...
+        # the unit only truly supports 4 speeds: low, med, high, auto
+        - dps_val: auto
+          value: auto
+        - dps_val: low
+          value: low
+        - dps_val: mid
+          value: medium
+        - dps_val: high
+          value: high
+        # the below are all duplicates of some of the above and don't change air speed
```
make-all, line 88: I suggest to map the value to the equivalent value from above, so they don't cause any errors for being unrecognized values, now or in future.

custom_components/tuya_local/devices/friedrich_uct08b10a_airconditioner
```diff
@@ -0,0 +1,520 @@
...
+          value: medhigh
+          optional: true
+          hidden: true
+        - dps_val: strong
+          value: strong
+          optional: true
+          hidden: true
+    - id: 123
+      name: preset_mode
+      # Tuya IOT dev says this is "bool" but the values aren't boolean
```
make-all, line 107: It is possibly hex encoded along with other values, in which case values of 0 and 1 and a mask of 0x0001 might be better to shield it from changes in the other bits.

custom_components/tuya_local/devices/friedrich_uct08b10a_airconditioner
```diff
@@ -0,0 +1,520 @@
...
+        - dps_val: "3"
+          value: through-the-wall
+    - id: 8
+      name: legacy_preset_mode
+      type: boolean
+      # the unit's actual hardware doesn't support this functionality even though the firmware / DP is exposed
+      # we are forced to add it so that Tuya Local gets to 100% DP match and allows us to pick this YAML for this unit
+      hidden: true
+      optional: true
+      mapping:
```
make-all, line 136: I wouldn't bother with the mapping if it is just here for matching purposes. But as it is set to be optional, it is not even useful for that.

custom_components/tuya_local/devices/friedrich_uct08b10a_airconditioner
```diff
@@ -0,0 +1,520 @@
...
+        type: bitfield
+        name: fault_code
+      - id: 122
+        name: fault2
+        type: bitfield
+        optional: true
+  - entity: sensor
+    # unclear what this is, seems like intended to be total power but it's unitless per IOT and doesn't exactly track
+    # what the unit is doing so leaving as a diagnostic-only sensor that should probably be disabled by the user
+    name: totalp
```
make-all, line 496: Not useful or unknown dps should be added as attributes of another entity rather than cluttering up the UI with additional sensors.

PR #2525, opened 2024-11-19
Add HLG Infinity Alarm - Small Adjustments

brunohenriquy:
- Adjusting some names
- Renaming the file name


brunohenriquy:
@make-all You've moved 113(buttons_learned) and 114(sensors_learned) to be attributes, but they are not being shown like the others:

 
- Not showing Attributes
Screenshot 2024-11-19 at 15 11 05
Screenshot 2024-11-19 at 15 06 09

- Showing Attributes
Screenshot 2024-11-19 at 15 06 28

brunohenriquy:
> @make-all You've moved 113(buttons_learned) and 114(sensors_learned) to be attributes, but they are not being shown like the others:

@make-all I guess you haven't seen this comment.

make-all:
Yes, I saw. I think it is because the select entities are appearing as unavailable due to the device not reporting any value.
Please try the latest, which makes Cancel the default for this case.

brunohenriquy:
> Yes, I saw. I think it is because the select entities are appearing as unavailable due to the device not reporting any value. Please try the latest, which makes Cancel the default for this case.

But if you look at  the first screenshot, it's reporting value, and still the attributes doesn't show up.

brunohenriquy:
> I just did a test here, it seems the select entity's "More info" dialog does not show extra attributes, but they are available, and can be seen in the Developer tools and selected as the "State content" to display on a Tile card.

Hmm Gotcha!! Do you think it's a good idea to keep it that way? Some users might not even know it exists, I think it's better exposed!

brunohenriquy:
> I don't think these values are useful enough to be exposed.

All right then!

PR #2526, opened 2024-11-19
Added support for Myuet ME82 thermostat

blitzdose:
I bought a cheap WiFi thermostat on Amazon,thought it would be the ME81, turned out to be an "ME82(LED).03 WifFi" with isn't really supported by any integration. Because I liked yours the most, I added the necessary config :) Awesome work with your integration! Made the configuration so much easier.

### DIFF
--- a/custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
+++ b/custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
@@ -4,6 +4,7 @@ products:
     name: Myuet ME82
 primary_entity:
   entity: climate
+  translation_key: thermostat
   dps:
     - id: 1
       type: boolean
@@ -48,20 +49,15 @@ primary_entity:
           value: heating
         - dps_val: close
           value: idle
+    - id: 2
+      type: string
+      name: preset_mode
+      mapping:
+        - dps_val: auto
+          value: program
+        - dps_val: manual
+          value: manual
 secondary_entities:
-  - entity: select
-    name: Mode
-    icon: "mdi:thermostat-auto"
-    category: config
-    dps:
-      - id: 2
-        type: string
-        name: option
-        mapping:
-          - dps_val: auto
-            value: Auto
-          - dps_val: manual
-            value: Manual
   - entity: switch
     translation_key: anti_frost
     category: config
@@ -192,21 +188,23 @@ secondary_entities:
             value: Six working days
           - dps_val: "no_day_off"
             value: Daily
-  - entity: select
-    name: Backlight
-    icon: "mdi:lightbulb-on"
+  - entity: light
+    translation_key: backlight
     category: config
     dps:
       - id: 52
         type: string
         optional: true
-        name: option
+        name: brightness
         mapping:
           - dps_val: "closed"
-            value: Off
+            value: 0
           - dps_val: "20_bright"
-            value: 20%
+            value: 51
           - dps_val: "50_bright"
-            value: 50%
+            value: 128
           - dps_val: "100_bright"
-            value: 100%
+            value: 255
+          - dps_val: null
+            value: 0
+            hidden: true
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
```diff
@@ -0,0 +1,215 @@
...
+products:
+  - id: 7ingxzrs0wo9rqsd
+    name: Myuet ME82
+primary_entity:
+  entity: climate
+  dps:
+    - id: 1
+      type: boolean
+      name: hvac_mode
+      mapping:
```
make-all, line 11: Is there any reason this does not combine with the mode, the same way the me81 config does?

custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
```diff
@@ -0,0 +1,215 @@
...
+      range:
+        min: 50
+        max: 700
+    - id: 19
+      type: integer
+      name: max_temperature
+      optional: true
+      mapping:
+        - scale: 10
+          step: 10
```
make-all, line 32: step is not relvant here, as the value cannot be changed from here.

custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
```diff
@@ -0,0 +1,215 @@
...
+    - id: 26
+      type: integer
+      name: min_temperature
+      optional: true
+      mapping:
+        - scale: 10
+          step: 10
```
make-all, line 39: step is not relevant here.

custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
```diff
@@ -0,0 +1,215 @@
...
+    - id: 24
+      type: integer
+      name: current_temperature
+      mapping:
+        - scale: 10
+          step: 5
```
make-all, line 45: step is not relevant here

custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
```diff
@@ -0,0 +1,215 @@
...
+        mapping:
+          - dps_val: "off"
+            value: Disabled
+          - dps_val: "weekend"
+            value: Weekend
+          - dps_val: "single_break"
+            value: Six working days
+          - dps_val: "no_day_off"
+            value: Daily
+  - entity: select
```
make-all, line 198: This would probably be better as a light entity.

custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
blitzdose: Yes, for some reason both my thermostats don't respond if I perform more than one action at a time. Changing the mode sets the temperature and the mode and nothing happens, doing both independently works fine

custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
blitzdose: The backlight has only four distinct options: "closed", "20_bight", "50_bright" and "100_bright". I don't think this would make sense if it was a light entity

blitzdose:
Removed unnecessary step properties

custom_components/tuya_local/devices/myuet_me82_thermostat.yaml
```diff
@@ -0,0 +1,212 @@
...
+      type: string
+      name: hvac_action
+      mapping:
+        - dps_val: open
+          value: heating
+        - dps_val: close
+          value: idle
+secondary_entities:
+  - entity: select
+    name: Mode
```
make-all, line 53: Since there is no preset_mode on the climate entity, this can be moved there to keep the main climate controls together as much as possible.

PR #2531, opened 2024-11-21
Add support for Airrobo P20 vacuum

witulu:
This is my first pull request ever! After using your awesome project for some time I discovered that I can integrate my vacuum robot, but it didn't match 100% and was recognised as some other brand vacuum. After some digging and tinkering I was able to create a yaml locally that matches all dps for my vacuum robot.

I hope someone out there has as much fun as I controlling this cheapo vacuum locally without cloud connection.

### DIFF
--- a/custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
+++ b/custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
@@ -1,7 +1,7 @@
 name: Robot vacuum
-products:
-  - id: bfdfcf2fc7921635a7v4ko
-    name: Airrobo P20
+# products:
+#  - id: UNKNOWN
+#    name: Airrobo P20
 primary_entity:
   entity: vacuum
   dps:
@@ -83,8 +83,7 @@ primary_entity:
       optional: true
     - id: 104
       type: string
-      name: Serial No
-##########################################################
+      name: serial_number
 secondary_entities:
   - entity: sensor
     class: battery
@@ -258,7 +257,7 @@ secondary_entities:
           - dps_val: italian
             value: italian
   - entity: switch
-    name: Speed Mode
+    name: Quick clean
     category: config
     dps:
       - id: 101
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -0,0 +1,289 @@
...
+          - dps_val: spanish
+            value: Spanish
+          - dps_val: italian
+            value: Italian
+  - entity: switch
+    name: Speed Mode
+    category: config
+    dps:
+      - id: 101
+        type: string
```
make-all, line 263: need a `name: switch` in here.

custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -0,0 +1,289 @@
...
+          - dps_val: 4096
+            value: Power lowsp
+          - dps_val: 8192
+            value: Middle sweep stop
+          - dps_val: 16384
+            value: Charging error
+          - dps_val: 32768
+            value: Charging no contact
+  - entity: select
+    name: Language
```
make-all, line 239: use `translaton_key: language` here instead of name, and make all values below lowercase.

custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -0,0 +1,289 @@
...
+    dps:
+      - id: 101
+        type: string
+        name: option
+        mapping:
+          - dps_val: careful_clean
+            value: Careful cleaning
+          - dps_val: speed_clean
+            value: Fast cleaning
+  - entity: sensor
```
make-all, line 281: This should never change, so is probably better as an attribute of the vacuum entity rather than a separate sensor.

custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
make-all: But I see this is duplicated with the select entity below, so one should be removed (I suspect you may have added the select to get around the switch not working without a dp named switch).

PR #2538, opened 2024-11-22
Typo in the device brand

PipoCanaja:
Fix yaml file name


make-all:
This would be disruptive for anyone who already has it installed.

PR #2542, opened 2024-11-23
Add device LED Strip WIFI (Deltaco SH-LW5M)

Ladida1:
I created a yaml for this device Deltaco SH-LW5M, gave it a generic name LED Strip WIFI, feel free to change if to generic.
Works fine to me, and wanted to share.

Many years since i did a GIT pull request, so I hope I got it right. If not let me know if I should change something.


custom_components/tuya_local/devices/deltaco_shlw5m_ledstrip.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - id: 3
+      type: integer
+      name: brightness
+      optional: true
+      range:
+        min: 25
+        max: 255
+secondary_entities:
+  - entity: select
+    name: effect
```
make-all, line 20: This can be moved inside the light entity, as the effect dp.

custom_components/tuya_local/devices/deltaco_shlw5m_ledstrip.yaml
```diff
@@ -0,0 +1,43 @@
+name: LED Strip WIFI
```
make-all, line 1: WIFI is superfluous in the device name. And following HA naming convention, strip should be all lower case.

Ladida1: Ok, removed WIFI anc changed strip to lower case.

PR #2548, opened 2024-11-25
Update hydrotherm_dynamic_x8_water_heaterv6 Values to match Unit Manual

Krazy998:
Hello,

Very minor tweak to the display values of each of the operating modes. This is to match what is actually in the manual located here on Page 29: https://www.hydrothermhotwatersystems.com.au/wp-content/uploads/2024/02/HYD_DYNAMIC_USER_MANUAL_WEB.pdf

Helps make sure you select the right value and you know what it actually does.


make-all:
It is more important to be consistent with standard HA modes than with the device manufacturer's chosen names.

PR #2554, opened 2024-11-26
Add Morento Air Purifier and Edge Theory Cold Plunge configurations

tompapajr:
Created device YAMLs for Morento Air Purifier model HY4866-WF and the Edge Theory Cold Plunge Tub Heat Pump.
This is my first pull request ever so I hope I'm doing it right.. :-)


custom_components/tuya_local/devices/edgetheorylabs_coldplunge_heatpump.yaml
```diff
@@ -0,0 +1,77 @@
...
+primary_entity:
+  entity: climate
+  dps:
+    - id: 121
+      name: hvac_mode
+      type: boolean
+      mapping:
+        - dps_val: false
+          value: "off"
+          icon: "mdi:hvac-off"
```
make-all, line 14: I suggest not to override the icons. HA's defaults are getting better all the time, and this will hide any overall improvements.

custom_components/tuya_local/devices/edgetheorylabs_coldplunge_heatpump.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - dps_val: true
+          value: "heat_cool"
```
make-all, line 16: It seems like this is not a mode that automatically switches between heat and cool, but is dependent on the mode dp below. Can the mode be combined here? Some devices do not work with setting two dps at onces, but most do.

make-all: If not, the "mode" needs to be renamed to "preset_mode" in order to make it accessible, as "mode" is not a standard part of the climate entity.

tompapajr: maybe you an help me with this one -- I'll add some comments in my yaml to better describe the dp id's and states. But in comparing all states form tuya's IoT platform API Explorer (Device Control > Query Properties) when the heat pump is off, idle (not heating or cooling target temp reached just circulating), heating, and cooling. I've found the following changes; in fact there was not change to dp id 115 at all - i can probably remove it entirely, its always true.  Unfortuately there isnt much clear documetnation on this unit 

HEATING:
- sw_power (DP 121): true
- temp_set_f (DP 17): 50°F
- temp_current_f (DP 29): 42.4°F
- out_compressor (DP 122): true
- flow_water (DP 102): 140
- sw_water (DP 125): true

COOLING:
- sw_power (DP 121): true
- temp_set_f (DP 17): 39°F
- temp_current_f (DP 29): 42.6°F
- out_compressor (DP 122): true
- flow_water (DP 102): 140
- sw_water (DP 125): true

IDLE:
- sw_power (DP 121): true
- out_compressor (DP 122): false
- flow_water (DP 102): 137
- sw_water (DP 125): true
- temp_current_f (DP 29) ≈ temp_set_f (DP 17)

OFF:
- sw_power (DP 121): false
- out_compressor (DP 122): false
- flow_water (DP 102): 0
- sw_water (DP 125): false
- temp monitoring continues on DP 29

custom_components/tuya_local/devices/morento_HY4866-WF_air_purifier.yaml
```diff
@@ -0,0 +1,50 @@
...
+  entity: fan
+  icon: "mdi:air-purifier"
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+    - id: 3
+      type: string
+      name: preset_mode
+      mapping:
```
md1986, line 15: The mapping is missing the level 4F. See the PR I have open.

custom_components/tuya_local/devices/edgetheorylabs_coldplunge_heatpump.yaml
make-all: So based on this, I think dp id: 121 should be your main "hvac_mode", but is there another one that can switch between heat and cool to combine it with?

dp: 122 looks like it is suitable for use as hvac_action, combined with whatever is indicating whether it is heating or cooling.

make-all:
I guess since PR #2581 also covers the Morento heatpump, this PR should focus just on the Edge Theory Cold Plunge and remove the Morento to avoid conflicting.

custom_components/tuya_local/devices/edgetheorylabs_coldplunge_heatpump.yaml
tompapajr: @make-all - Dug into this, there doesnt seem to be any DP ID that indicates whether the unit is cooling or heating just that the compresor is on.  Went exhaustively through all of them... Can there be logic within a yaml for example if the compressor is on and the target temp is lower than current then its cooling and heating if the target is higher than current?

make-all: Then make dp 122 into a separate sensor with class `running`, and don't have an hvac_action.

make-all: Having read the manual, it seems this is a heat_cool mode, and there is no manual control of whether it is heating or cooling. In that case, the mode below does appear useless, as you observed.

PR #2555, opened 2024-11-26
Create aygrochy_inline_wifi_air_vent.yaml

Kenneth-Audenaert:


custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+    - id: 9
+      readonly: true
+      name: current_temperature
+      type: integer
+      range:
+        min: -40
+        max: 255
+      unit: "F"
+    - id: 106
+      name: target_temp_high
```
make-all, line 36: `target_temp_high` needs to be paired with `target_temp_low` which seems to be available based on the number entities below.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+      name: humidity
+      type: integer
+      range:
+        min: 0
+        max: 100
+      unit: "%"
+
+secondary_entities:
+  - entity: fan
+    name: fan
```
make-all, line 52: Name should be omitted if it is the same as the entity name.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+    dps:
+      - id: 104
+        name: value
+        type: integer
+        unit: s
+        range:
+          min: 0
+          max: 36000
+  - entity: sensor
+    name: Alarm triggered
```
make-all, line 90: Should maybe be an event. But the values are needed (or at least the value when an alarm is not triggered, so it can be filtered out).

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+    dps:
+      - id: 105
+        name: sensor
+        readonly: true
+        type: string
+  - entity: number
+    name: Auto-High Temp
```
make-all, line 97: Seems to be a duplicate of the climate device's target_temp_high above, so should not be needed.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+      - id: 107
+        name: value
+        class: humidity
+        type: integer
+        range:
+          min: 0
+          max: 100
+        unit: "%"
+  - entity: number
+    name: Auto-Low Temp
```
make-all, line 119: This should probably be the target_temp_low for the climate entity above, instead.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+    dps:
+      - id: 123
+        type: integer
+        name: switch
+        mapping:
+          - dps_val: 0
+            value: true
+          - dps_val: 1
+            value: false
+  - entity: sensor
```
make-all, line 278: Probably another one that would be better as an event.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
+name: Aygrochy fan
+primary_entity:
+  entity: climate
+  dps:
+    - id: 2
+      name: preset_mode
```
make-all, line 6: My suggestion is to duplicate this to hvac_mode, and map "ON" to fan_only, "OFF" to "off", "TIMER" to "auto" but marked as hidden, "AUTO" to "auto", and "ALARM" to "fan_only" but marked as hidden.
The target temperatures should have conditional mappings based on this mode to select the appropriate dps for the mode (using `value_redirect`).

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+        max: 255
+      unit: "F"
+    - id: 106
+      name: target_temp_high
+      type: integer
+      range:
+        min: 32
+        max: 194
+      unit: "F"
+    - id: 107
```
make-all, line 42: Since the climate entity does not support high and low target humidity settings, probably this should be removed from here and the separate number entities below used.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+        name: sensor
+        readonly: true
+        type: number
+  - entity: number
+    name: Countdown timer
+    dps:
+      - id: 104
+        name: value
+        type: integer
+        unit: s
```
make-all, line 85: A second resolution time is a bit more precise than the integration can reliably guarantee, so normally I would turn this into a minute timer, using `unit: min` and 
```
mapping:
  - scale: 60
    step: 60
```

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+    dps:
+      - id: 106
+        name: value
+        class: temperature
+        type: integer
+        range:
+          min: 32
+          max: 194
+        unit: "F"
+  - entity: number
```
make-all, line 107: Like with the temperature settings, as the comment in the climate entity says, I think the different humidity settings for different modes should be handled within just two number entities for High and Low, with conditions based on mode using `value_redirect* to select the appropriate dp to change.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
make-all: I did say duplicate.

You will still need it accessible as preset_mode to switch between the modes that have been merged to fit into the limited options of hvac_mode.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,298 @@
...
+        range:
+          min: 0
+          max: 4
+  - entity: sensor
+    name: Current speed
+    dps:
+      - id: 103
+        name: sensor
+        readonly: true
+        type: number
```
make-all, line 79: This one needs to be `integer`

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,298 @@
...
+        - dps_val: "AUTO"
+          value: "auto"
+        - dps_val: "ALARM"
+          value: "fan_only"
+          hidden: true
+    - id: 8
+      readonly: true
+      name: current_humidity
+      type: integer
+      range:
```
make-all, line 25: This range is not needed (and would be wrong to go beyond 100 for %)

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,298 @@
...
+        - dps_val: "TIMER"
+          value: "auto"
+          hidden: true
+        - dps_val: "AUTO"
+          value: "auto"
+        - dps_val: "ALARM"
+          value: "fan_only"
+          hidden: true
+    - id: 8
+      readonly: true
```
make-all, line 22: readonly is not needed on attributes that are read-only by definition

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,298 @@
...
+      type: integer
+      range:
+        min: 0
+        max: 255
+      unit: "%"
+    - id: 9
+      readonly: true
+      name: current_temperature
+      type: integer
+      range:
```
make-all, line 33: This range and readonly are also not needed.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,298 @@
...
+        name: switch
+        type: boolean
+      - id: 102
+        name: speed
+        type: integer
+        range:
+          min: 0
+          max: 10
+  - entity: light
+    translation_key: Backlight brightness
```
make-all, line 65: key is just `backlight`

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,298 @@
...
+          max: 4
+  - entity: sensor
+    name: Current speed
+    dps:
+      - id: 103
+        name: sensor
+        readonly: true
+        type: number
+  - entity: number
+    name: Countdown timer
```
make-all, line 81: This should use `translation_key: timer` instead of the name.

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
Kenneth-Audenaert: fixed

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,314 @@
...
+        name: switch
+        type: boolean
+      - id: 102
+        name: speed
+        type: integer
+        range:
+          min: 0
+          max: 10
+  - entity: light
+    name: Backlight brightness
```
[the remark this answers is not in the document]
Kenneth-Audenaert: fixed

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
Kenneth-Audenaert: sorry misinterpreted, duplicated now

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
Kenneth-Audenaert: fixed

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
```diff
@@ -0,0 +1,301 @@
...
+      - id: 105
+        name: sensor
+        type: string
+  # id 106 see primairy entity
+  - entity: number
+    name: Auto-High Humid
+    dps:
+      - id: 107
+        name: value
+        class: humidity
```
make-all, line 110: This class needs to go up at the entity level

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
Kenneth-Audenaert: adjusted

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
Kenneth-Audenaert: fixed

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
Kenneth-Audenaert: fixed

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
Kenneth-Audenaert: fixed

custom_components/tuya_local/devices/aygrochy_inline_wifi_air_vent.yaml
Kenneth-Audenaert: moved to other entity as suggested

PR #2557, opened 2024-11-26
Add EM3390TF weather station

flapedriza:
Add EM3390TF weather station configuration, only added 1 sensor but the station can have up to 3.

### DIFF
--- a/custom_components/tuya_local/devices/EM3390TF_weather_station.yaml
+++ b/custom_components/tuya_local/devices/EM3390TF_weather_station.yaml
@@ -1,7 +1,7 @@
-name: Weather Station
+name: Weather station
 products:
   - id: cvurn5ynloebrpwy
-    name: EM3390TF
+    name: Viflykoo EM3390TF
 primary_entity:
   entity: sensor
   name: Indoor temperature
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/EM3390TF_weather_station.yaml
```diff
@@ -0,0 +1,50 @@
+name: Weather Station
+products:
+  - id: cvurn5ynloebrpwy
+    name: EM3390TF
```
make-all, line 4: It would be better to include the brand name as well as the model number here.

flapedriza: I also thought about adding the brand at the begginning, but his is one of those generic chinese products that are sold under different brands. I bought it as "VIFLYKOO", but if you search in google you can find listings as "UZOLI " or "JELY"  brand for example, so IDK if it makes sense to add the brand here, since it might be confusing for people with the same model under a different brand.

make-all: Yes I found the Uzoli and Jely branded ones in my search (but not this one).  I am not sure whether the product id will be the same or different for the other brands. But since this is only the name under products, and the top level name is the one that will be most visible, I don't think it is too misleading, and if we later find out the other brands use the same product id, then all 3 brands can be listed.

make-all:
Looking at the device list, another weather station with very similar model number sold under various brands was identified as being made by Emax, so likely the same in this case. Viflykoo seems to be an Amazon exclusive brand.

PR #2581, opened 2024-11-30
add support for Morento HY4866-WF

md1986:


custom_components/tuya_local/devices/morento_hy4866_wf.yaml
```diff
@@ -0,0 +1,90 @@
...
+      name: switch
+      type: boolean
+    - id: 3
+      name: preset_mode
+      type: string
+      mapping:
+        - dps_val: sleep
+          value: sleep
+        - dps_val: Auto
+          value: auto
```
make-all, line 20: For some reason fans don't have a predefined "auto" preset, but they have a "smart" preset which is similar. I don't remember the exact details, but I originally took the translated presets from a list that was in the HA fan implementation at the time, but they have since removed. There is also "strong" predefined, which could replace "very high".

custom_components/tuya_local/devices/morento_hy4866_wf.yaml
```diff
@@ -0,0 +1,90 @@
...
+  - entity: sensor
+    name: Filter used
+    category: diagnostic
+    icon: "mdi:ticket-percent-outline"
+    dps:
+      - id: 5
+        name: sensor
+        type: integer
+        unit: "%"
+  - entity: switch
```
make-all, line 47: For child lock, use a lock entity with translation_key "child_lock".

PR #2583, opened 2024-11-30
fix missing .yaml, fix yaml parsing due to "step" in current temp, fix eco/comfort mode

ncd7:
**Note - I realized this was an issue after I updated to latest tuya-local. I suspect perhaps you kept the file non-yaml due to the issues below so feel free to delete this PR if you have other fixes in mind.**

Fixes for:
* missing .yaml extension
* yaml parsing due to "step" in current temp
* eco/comfort mode not being reflected in HA or controllable


make-all:
I've fixed these issues.

PR #2588, opened 2024-12-01
Add two Tomzn device configurations and misc update for Moes MS-103 device

AlexK98:
This adds two new configuration files:
- tompd_63lw_breaker_v2.yaml
- tompd_63wifi_breaker.yaml

and misc updates to:
- moes_temp_humidity.yaml


AlexK98:
@make-all
I read out the test log, but don't understand where the test fails. Can you help?

PR #2591, opened 2024-12-02
Create cecotec_bigdry_4000_dehumidifier.yaml

darth-hp:
New device. Started discussion here https://github.com/make-all/tuya-local/discussions/2571

### DIFF
--- a/custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
@@ -1,4 +1,4 @@
-name: BigDry 4000 Expert
+name: Dehumidifier
 products:
   - id: d44nan1uokr1hlwf
     name: Cecotec BigDry 4000 Expert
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
```diff
@@ -0,0 +1,145 @@
...
+primary_entity:
+  entity: humidifier
+  class: dehumidifier
+  name: Dehumidifier
+  dps:
+    - id: 1
+      name: switch
+      type: boolean
+    - id: 3
+      type: integer
```
make-all, line 15: dp 3 above has quotes around it. This suggests that it is actually an string (enum) type, and will only accept a small number of specific values, not the full range between 40 and 80. The implementation below for a duplicate select entity seems to follow this, but to move it up here, you will need to remove the quotes from the `value` side of the mappings.

custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
```diff
@@ -0,0 +1,145 @@
...
+        type: string
+        name: option
+        mapping:
+          - dps_val: low
+            value: Low
+          - dps_val: high
+            value: High
+  - entity: sensor
+    class: temperature
+    name: Room Temperature
```
make-all, line 70: Remove the name so that translations are used (there is no other temperature sensor attached to the device, so the default "Temperature" is unambiguous.

custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
```diff
@@ -0,0 +1,145 @@
...
+            value: "60"
+          - dps_val: "65"
+            value: "65"
+          - dps_val: "70"
+            value: "70"
+          - dps_val: "75"
+            value: "75"
+          - dps_val: "80"
+            value: "80"
+  - entity: select
```
make-all, line 57: Use a fan entity, or even the humidifier entity's mode for this.

darth-hp: Thanks Jason for the hint. Indeed I decided to put it as extra selection because I do not like that wheel as selector and sometimes it was in read-only mode for no known reason.

custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
darth-hp: For me the question was, is that the devices temperature or the room temperature? You are right, it's unambiguous and I could rename it later.

Issue #1913, opened 2024-05-12
ZY-M100-WIFI24G - not recognized as ZY-M100 in tuya local


Motya-HA:
This form is for reporting bugs.  If you have encountered a bug, you should
generally have some log messages or device diagnostics to show something is
going wrong.  If your problem is more of a question or a request for help,
then [Discussions](https://github.com/make-all/tuya-local/discussions) is
probably a more suitable venue.

**Describe the bug**
For some reason Tuya presence sensor ZY-M100-WIFI is not recognized as such during new device addition

**To Reproduce**
During "add device" process, ZY-M100-WIFI24G presence sensor device has following options:

![Screenshot 2024-05-12 at 4 19 57 PM](https://github.com/make-all/tuya-local/assets/143926968/c329ba4a-b07a-42ae-accc-6169fbd4629e)

**Expected behavior**
Expected device been recognized as ZY-M100_WIFI

**Additional context**
Parameters results from Tuya IoT platform

{
  "result": [
    {
      "active_time": 1715474703,
      "bind_space_id": "166925674",
      "category": "hps",
      "create_time": 1715474703,
      "custom_name": "Laundry Entry Presence Sensor ",
      "icon": "smart/icon/ay15431316572284AcPZ/bc0c8e066c89476de379df5cf1538fee.png",
      "id": "ebed371ad78c7473e84uu4",
      "ip": "xxxxxxxxxx",
      "is_online": true,
      "lat": "",
      "local_key": "xxxxxxxxxxxx",
      "lon": "",
      "model": "1.0.1",
      "name": "ZY-M100-WIFI24G存在传感器",
      "product_id": "4cjzlliucdjaanvl",
      "product_name": "ZY-M100-WIFI24G存在传感器",
      "sub": false,
      "time_zone": "-07:00",
      "update_time": 1715474751,
      "uuid": "11a9664989c658a4"
    }
  ],
  "success": true,
  "t": 1715555227412,
  "tid": "5aafb1a110b411efa6ed16994efa91b4"
}

If the bug involves a device, then please include device diagnostics from
Settings / Devices & Services / Tuya Local (your device) / "1 device"

Please check if there are messages from Tuya Local in the Home
Assistant log (Settings / System / Logs), and include those too.

make-all:
I cannot diagnose the issue without logs or diagnostics

Motya-HA:
But I can't get logs or diagnostics if I can't add the device?
What you would suggest to do?

[...]

PR #2611, opened 2024-12-04
ZYM100W presence: fix min/max distance range

felipecrs:
@make-all, I have this device, and these new values are what both the app and tinytuya show as min/max.

```json
             "3": {
                "code": "near_detection",
                "type": "Integer",
                "values": {
                    "unit": "M",
                    "min": 0,
                    "max": 825,
                    "scale": 2,
                    "step": 75
                }
            },
            "4": {
                "code": "far_detection",
                "type": "Integer",
                "values": {
                    "unit": "M",
                    "min": 75,
                    "max": 900,
                    "scale": 2,
                    "step": 75
                }
            },
```


make-all:
Rather than simply changing from the original values in #1913 to your device's values, it would be better to reduce the limits to cover both. Since the other variant has a lower limit of 150 for both, don't change the lower limit, but since your upper limits are lower, bring down the upper limit to 900/825 instead of 975/900.  That way neither device will be put into a setting that it cannot handle.

felipecrs:
Oh. That's weird. Maybe our devices are not the same after all.

The box of mine says:

Loginovo (brand) Smart Human Presence Sensor M100

- [x] Wifi
- [x] Wall version
- [x] 24G

On one label, it says:

Product name: Smart Sensor
Model: C3007

On another label, it says:

TY-WS-24G

https://aliexpress.com/item/1005007586670159.html

Variant: 24G Wifi Wall

This is Tinytuya's device data for it:

Note that this is the only place where a model like ZYM100W is mentioned (under `product_name`):

```json
      {
        "name": "Presen\u00e7a do escrit\u00f3rio",
        "id": "eb1846afdec6131f8ecjsn",
        "key": "hidden",
        "mac": "hidden",
        "uuid": "f7b5c6295a6df923",
        "sn": "100128046005CA",
        "category": "hps",
        "product_name": "ZY-M100-WIFI24G\u5b58\u5728\u4f20\u611f\u5668V2",
        "product_id": "5lld8pgsoynvctqa",
        "biz_type": 0,
        "model": "",
        "sub": false,
        "icon": "https://images.tuyaus.com/smart/icon/ay15431316572284AcPZ/51bdad4264cd4894214d790a99c39ca8.png",
        "mapping": {
            "1": {
                "code": "presence_state",
                "type": "Enum",
                "values": {
                    "range": [
                        "none",
                        "presence"
                    ]
                }
            },
            "2": {
                "code": "sensitivity",
                "type": "Integer",
                "values": {
                    "unit": "",
                    "min": 1,
                    "max": 10,
                    "scale": 0,
                    "step": 1
                }
            },
            "3": {
                "code": "near_detection",
                "type": "Integer",
                "values": {
                    "unit": "M",
                    "min": 0,
                    "max": 825,
                    "scale": 2,
                    "step": 75
                }
            },
            "4": {
                "code": "far_detection",
                "type": "Integer",
                "values": {
                    "unit": "M",
                    "min": 75,
                    "max": 900,
                    "scale": 2,
                    "step": 75
                }
            },
            "9": {
                "code": "target_dis_closest",
                "type": "Integer",
                "values": {
                    "unit": "M",
                    "min": 0,
                    "max": 100,
                    "scale": 1,
                    "step": 1
                }
            },
            "103": {
                "code": "illuminance_value",
                "type": "Integer",
                "values": {
                    "unit": "lux",
                    "min": 0,
                    "max": 10000,
                    "scale": 0,
                    "step": 1
                }
            }
        },
        "ip": "192.168.1.69",
        "version": "3.3"
    },
```

@make-all maybe the best would be to add a new profile just for it. What do you think?

make-all:
Do you really have a use case for setting the detection limits less than 1.5m?

If we make a duplicate with different limits, how does the user know which one to choose?

felipecrs:
> If we make a duplicate with different limits, how does the user know which one to choose?

Maybe we can distinguish it by firmware version.

> Do you really have a use case for setting the detection limits less than 1.5m?

No, that's fine. It's not big deal. I can live with it.

To be honest I'll probably flash this device with an alternative firmware as soon as possible (but LibreTiny/ESPHome doesn't support its board yet), because I am not able to extract relying results out of this device. It has lots of false detections and false un-detections.

make-all:
Firmware version is not readable locally. For now, just restore the min back to 150 for both sensors, and we'll merge the reductions for max to avoid going out of range on either sensor.

felipecrs:
> Firmware version is not readable locally.

I meant to present the user two choices of profiles when setting up the device, and then user can choose whichever they want.

But in fact, when the user has enabled cloud assisted setup, you can fetch the dp information with tinytuya and then use that data (the correct min/max) to select the right profile.

But I agree it's probably not worth the time spent implementing that.

> For now, just restore the min back to 150 for both sensors, and we'll merge the reductions for max to avoid going out of range on either sensor.

To be honest I don't like that option either. Right now, the profile is perfect for one device. If we change that, it's gonna be unperfect for both.

make-all:
Unfortunately the tuya cloud does not allow fetching local dp information with the QR code authentication flow.

felipecrs:
Got it. Thank you anyway then. I suggest you wait for someone else to complain before considering doing anything.

felipecrs:
@make-all I changed my mind and applied your suggestion. Can you please take a look again?

Reason is: I now need to change my minimum distance beyond the minimum value currently allowed.

make-all:
I think this change goes the other direction. Now both variants can be set out of range.

felipecrs:
But also mean both can reach all their supported values.

Do you mean I should limit their maximum values?

If the minimum can go out of limit, why not let maximum be too?

make-all:
I mean both minimum and maximum should not be able to go outside the limits.

felipecrs:
Then there's no solution.

I need to set my minimum distance to 0, so that the sensor activates as soon as movement is detected regardless of how close it is.

That's because my sensor is located close to the door, in the wall, rather than in the ceiling. When in the ceiling, it would make sense to set a minimum distance higher than zero.

make-all:
I think the solution must be to create a new config file, as the incompatible ranges is not possible to handle within the same file without some flag telling us which device is which.

felipecrs:
I agree. Even though Tuya Local cannot automatically infer which one is correct for the device, I can still just select it during the profile selection.

felipecrs:
Do you want me to submit a PR for it?

felipecrs:
There we go: https://github.com/make-all/tuya-local/pull/2744

PR #2612, opened 2024-12-05
Add support for Dr. Heater DR-008 Electric Radiant Floor Thermostat

ThinkDualBrain:
Added support for Dr. Heater DR-008 Electric Radiant Floor Thermostat

### DIFF
--- a/custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
+++ b/custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
@@ -1,6 +1,10 @@
-name: Dr. Heater DR-008 Electric Radiant Floor Thermostat
+name: Floor thermostat
+# products:
+#   - id: UNKNOWN
+#     name: Dr. Heater DR-008
 primary_entity:
   entity: climate
+  translation_key: thermostat
   dps:
     - id: 101
       type: boolean
@@ -9,14 +13,7 @@ primary_entity:
         - dps_val: false
           value: "off"
         - dps_val: true
-          constraint: mode
-          conditions:
-            - dps_val: "Smart"
-              value: auto
-            - dps_val: "Manual"
-              value: heat
-            - dps_val: "Anti_frozen"
-              value: cool
+          value: heat
     - id: 102
       type: integer
       name: temperature
@@ -35,8 +32,14 @@ primary_entity:
                 max: 1220
     - id: 103
       type: string
-      name: mode
-      hidden: true
+      name: preset_mode
+      mapping:
+        - dps_val: Smart
+          value: program
+        - dps_val: Manual
+          value: manual
+        - dps_val: Anti_frozen
+          value: away
     - id: 105
       type: integer
       name: current_temperature
@@ -89,17 +92,12 @@ primary_entity:
         max: 9
 secondary_entities:
   - entity: lock
-    name: Screen lock
+    translation_key: child_lock
     category: config
     dps:
       - id: 108
         type: boolean
         name: lock
-        mapping:
-          - dps_val: true
-            icon: "mdi:hand-back-right-off"
-          - dps_val: false
-            icon: "mdi:hand-back-right"
   - entity: sensor
     name: Room temperature
     class: temperature
@@ -139,17 +137,17 @@ secondary_entities:
           - dps_val: true
             value: F
   - entity: sensor
-    name: Energy
+    name: Runtime
+    class: duration
+    category: diagnostic    
     dps:
       - id: 117
         type: integer
         name: sensor
-        unit: kWh
-        mapping:
-          - scale: 10
+        unit: min
       - id: 116
         type: integer
-        name: calibration
+        name: wattage
   - entity: select
     category: config
     translation_key: temperature_unit
--- a/custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
+++ b/custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
@@ -1,0 +1,229 @@
+name: Dr. Heater DR-008 Electric Radiant Floor Thermostat
+primary_entity:
+  entity: climate
+  dps:
+    - id: 101
+      type: boolean
+      name: hvac_mode
+      mapping:
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: "Smart"
+              value: auto
+            - dps_val: "Manual"
+              value: heat
+            - dps_val: "Anti_frozen"
+              value: cool
+    - id: 102
+      type: integer
+      name: temperature
+      range:
+        min: 200
+        max: 500
+      mapping:
+        - scale: 10
+          step: 5
+          constraint: temperature_unit
+          conditions:
+            - dps_val: true
+              step: 10
+              range:
+                min: 680
+                max: 1220
+    - id: 103
+      type: string
+      name: mode
+      hidden: true
+    - id: 105
+      type: integer
+      name: current_temperature
+      mapping:
+        - constraint: sensor
+          conditions:
+            - dps_val: "0"
+              scale: 10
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
```diff
@@ -0,0 +1,229 @@
...
+        - dps_val: false
+          value: "off"
+        - dps_val: true
+          constraint: mode
+          conditions:
+            - dps_val: "Smart"
+              value: auto
+            - dps_val: "Manual"
+              value: heat
+            - dps_val: "Anti_frozen"
```
make-all, line 18: These are all heat modes, so I don't think mapping them incorrectly to hvac_mode is the right thing to do. 103 can just be preset_mode, if translation_key is set to thermostat, "manual", "program" and "away" are available as presets.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
```diff
@@ -0,0 +1,229 @@
...
+        max: 9
+    - id: 113
+      type: integer
+      name: floor_calibration_offset
+      range:
+        min: -9
+        max: 9
+secondary_entities:
+  - entity: lock
+    name: Screen lock
```
make-all, line 92: You could just use the child_lock translation here, which is the same thing.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
```diff
@@ -0,0 +1,229 @@
...
+      - id: 107
+        name: unit
+        type: boolean
+        mapping:
+          - dps_val: false
+            value: C
+          - dps_val: true
+            value: F
+  - entity: sensor
+    name: Energy
```
make-all, line 142: If this reliably reports useful data, you can use `class: energy` with `class: total_increasing` on the sensor dp below, then this will be available for tracking in the Energy dashboard.
If it doesn't report useful data, then probably it should just be an attribute of another entity rather than an entity of its own.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
```diff
@@ -0,0 +1,229 @@
...
+          - dps_val: "2"
+            value: "Both"
+            icon: "mdi:thermometer"
+  - entity: select
+    category: config
+    name: Auto schedule
+    icon: "mdi:calendar-sync"
+    dps:
+      - id: 110
+        type: integer
```
make-all, line 219: Is it really an integer, and not a string?

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
```diff
@@ -0,0 +1,229 @@
...
+        name: option
+        mapping:
+          - dps_val: 0
+            value: "7"
```
make-all, line 223: Usual values used for this type of Schedule are "Daily", "Mon-Sat+Sun", "Weekday+Weekend". These indicate one schedule for every day, one schedule for Mon-Sat and a different schedule for Sunday, and one schedule Mon-Fri and a different schedule for Sat-Sun, respectively.

But your "Adaptive" and also the "5+1+1" seems to suggest something slightly different than the normal "7", "6+1", "5+2" from Tuya. But anyway maybe there is wording similar to the above that will make the meaning clearer.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
```diff
@@ -0,0 +1,229 @@
+name: Dr. Heater DR-008 Electric Radiant Floor Thermostat
```
make-all, line 1: This name should be under the products section, along with the product id.

The top level name should be generic, without any branding, as the same device is often sold in different markets with different branding, and this is the one that is used for device naming (which is also a reason to keep it short).

I think "Floor thermostat" would be appropriate for the top level name, and "Dr. Heater DR-008" for the product specific name.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
ThinkDualBrain: True, this is an electric floor heater, so it only provides heat. I can change this if you think that is preferrable?

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
ThinkDualBrain: Most of this was copied from [Jiahong ET-72W thermostat](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml), which was the closest match I was able to find. I only tweaked the DPS values, so I'm not entirely sure how much of this is accurate.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
ThinkDualBrain: Good question. As I mentioned above, this started as a clone of [Jiahong ET-72W thermostat](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml), and I only changed what I was familiar with. Everything else seemed alright, so I didn't mess with it.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
ThinkDualBrain: True, this does appear different than normal, but it matches what is shown in the Tuya app
Screenshot_20241207-212752

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
ThinkDualBrain: Okay, I'll update

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
make-all: Yes, I think hvac_mode just off and heat, and using preset_mode to switch between the modes is preferable.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
make-all: Just check what is actually output on that sensor, and whether it is something useful or not. Jianhong is an older device config, probably pre-dating the Energy dashboard in HA, so I don't know whether that one is not marked as energy class because it was unreliable, or it is just not updated.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
make-all: Usually these "select between a limited list of options" are what Tuya calls "enum" type, which is a string. The string may be numeric values, but has the quotes around it.

Because this has been confusing in the past, the current code is designed to work if you make them integer, but at the expense of some false matches during device detection, as the config will be matching both integers and quoted integers, while if you list them as string it will only match the quoted ones. Also it isn't guaranteed that this leniency will stay around forever, if we ever find out it causes a real problem somewhere.

ThinkDualBrain: Makes sense. I'll try and reconcile it with actual values.

custom_components/tuya_local/devices/dr_heater_dr008_thermostat.yaml
make-all: According to the original author, the Jianhong device actually reports minutes of runtime on that dp, and it is multiplied by another config value that depends on which heating element is fitted to the specific heater to get kWh.

Maybe your heater is similar, in which case this should be changed accordingly.

PR #2618, opened 2024-12-06
add tompd_63wifi_breaker

AlexK98:
Add configuration for Tomzn TOMPD-63WIFI circuit breaker.

### DIFF
--- a/custom_components/tuya_local/devices/tompd_63wifi_breaker.yaml
+++ b/custom_components/tuya_local/devices/tompd_63wifi_breaker.yaml
@@ -149,9 +149,9 @@ secondary_entities:
         name: sensor
         optional: true
         unit: V
+        mask: "FFFF000000000000"
         mapping:
-          - mask: "FFFF000000000000"
-            scale: 10
+          - scale: 10
   - entity: sensor
     name: Current
     class: current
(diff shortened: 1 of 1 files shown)


make-all:
This seems very similar to existing TOMP 63 config.
Differences:
- Refresh sensors button (dp 106) was removed. As this is already marked optional, it should not interfere with detecting your device, and if not functional, you can hide it.
- Earth leak test button (dp 21) was removed. As above, it is already marked optional and can just be ignored if it is not working.
- config options for dp 104 and 105 were added. As above, we could just add these as optional to the existing config.
- some renaming, which I agree with except the names should have been removed completely to let class translations be used.
- removal of diagnostic category for some sensors. Since these are not the primary sensors for the device, I think diagnostic is appropriate. The only affect this has is that they are not automatically added to generated dashboards, or voice assistants, but they can be manually added if you want them there.
- removal of one optional marking on a dp. This doesn't have a major effect, except filtering the config out from some false matches perhaps.

My preference for this device is to modify the existing config with the two new entities you added rather than accept a new config for it.

make-all:
Seeing the conflicts in dps 104 and 105 between this and the tompd 63lw v2 config you also submitted, I'm reconsidering which should be merged with the original tompd 63lw config. Since the name matches, I think it is probably better for the other one to merge (it also only really has additional dps which can be optional), and this one to be new. I will need to clean up the main branch now to rebase this PR so it can be merge.

PR #2634, opened 2024-12-10
Support for new device Eberg Rico R12E2 dehumidifier

nerodoxus:
Support for new device Eberg Rico R12E2 dehumidifier
https://www.amazon.pl/EBERG-Elektryczny-RICO-pod%C5%9Bwietlanym-wy%C5%9Bwietlaczem/dp/B0C85PCM88

### DIFF
--- a/custom_components/tuya_local/devices/eberg_rico_r12e2_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/eberg_rico_r12e2_dehumidifier.yaml
@@ -1,4 +1,4 @@
-name: Eberg Rico R12E2
+name: Dehumidifier
 products:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/beok_tol47_thermostat.yaml
```diff
@@ -43,7 +43,7 @@ primary_entity:
           value: program
         - dps_val: wind
           value: temp_override
-          hidden: true
+          hidden: false
```
make-all, line 46: Please keep changes to other configs separate.

I believe this was hidden because temporary override is not something you set by changing the mode, but is automatically set if you change another setting. So it is indicated here, but not available to select.

PR #2636, opened 2024-12-11
Add support for Xiaomi Mijia Temperature & Humidity sensor

markoceri:
Added support for Xiaomi Mijia temperature and humidity sensor, model LYWSD03MC. I have one converted to zigbee.

xiaomi-mijia-zigbee


make-all:
This integration is not adding support for non-Tuya devices via Tuya hubs, as that would make the maintenance workload spiral out of control.

markoceri:
I understand, you are right!

Collected in a different repository 👍 
https://github.com/markoceri/tuya-local-hub-devices

PR #2652, opened 2024-12-16
Add xoca_dac2121c_energy_meter.yaml

AlexK98:
XOCA DAC2121C BI Energy Meter

### DIFF
--- a/custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
+++ b/custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
@@ -1,4 +1,4 @@
-name: XOCA DAC2121C BI Energy Meter
+name: BiDirectional Energy Meter with Multi Tariff functions
 products:
   - id: opdhepy0p7eyeurz
     name: DAC-2121C BI
@@ -13,9 +13,14 @@ primary_entity:
       type: base64
       optional: true
       force: true
+    - id: 6
+      name: alarm_set_2
+      type: string
+      optional: true
+      force: true
 secondary_entities:
   - entity: sensor
-    name: Total Forward Energy
+    translation_key: energy_consumed
     class: energy
     # category: diagnostic
     dps:
@@ -28,7 +33,7 @@ secondary_entities:
         mapping:
           - scale: 100
   - entity: sensor
-    name: Total Reverse Energy
+    translation_key: energy_produced
     class: energy
     # category: diagnostic
     dps:
@@ -80,15 +85,6 @@ secondary_entities:
         mask: "0000000000FFFFFF"
         mapping:
           - scale: 1000
-  - entity: sensor
-    name: Alarm Set 2
-    category: diagnostic
-    dps:
-      - id: 17
-        type: string
-        name: sensor
-        optional: true
-        force: true
   - entity: sensor
     name: Meter ID
     category: diagnostic
--- a/custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
+++ b/custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
@@ -23,6 +23,7 @@ secondary_entities:
         type: integer
         unit: kWh
         class: total_increasing
+        force: true
         name: sensor
         mapping:
           - scale: 100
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
```diff
@@ -0,0 +1,158 @@
...
+  - entity: sensor
+    name: Alarm Set 2
+    category: diagnostic
+    dps:
+      - id: 17
+        type: string
+        name: sensor
+        optional: true
+        force: true
+  - entity: sensor
```
make-all, line 92: I don't think the Meter ID needs to be a sensor, it never changes, and is only useful if you have multiple meters and need to distinguish which is which.  It could just be an attribute of the primary entity.

custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
```diff
@@ -0,0 +1,158 @@
+name: XOCA DAC2121C BI Energy Meter
```
make-all, line 1: The branding should be under the products section below. This top level name should be generic in case of rebranded devices (as is common with Tuya devices sold in different parts of the world).

custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
```diff
@@ -0,0 +1,158 @@
...
+      name: switch
+      type: boolean
+    - id: 6
+      name: phase_a
+      type: base64
+      optional: true
+      force: true
+secondary_entities:
+  - entity: sensor
+    name: Total Forward Energy
```
make-all, line 18: Use `translation_key: energy_consumed` instead of the name.

custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
```diff
@@ -0,0 +1,158 @@
...
+      - id: 1
+        type: integer
+        unit: kWh
+        class: total_increasing
+        force: true
+        name: sensor
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Total Reverse Energy
```
make-all, line 31: Use `translation_key: energy_produced` instead of the name.

custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
```diff
@@ -0,0 +1,158 @@
...
+        optional: true
+        force: true
+        type: base64
+        unit: kW
+        name: sensor
+        mask: "0000000000FFFFFF"
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: Alarm Set 2
```
make-all, line 84: This is likely an encoded binary field, so not useful as a sensor. Instead include it as an attribute of the primary entity.

custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
AlexK98: For starters, putting it into the attributes makes it not really obvious where to find it.

custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
AlexK98: I own rebranded one which reports itself in Tuya as XOCA DAC2121C BI. Have no idea what would generic look like.

make-all: Generic would be "Energy meter"

custom_components/tuya_local/devices/xoca_dac2121c_energy_meter.yaml
```diff
@@ -13,9 +13,14 @@ primary_entity:
       type: base64
       optional: true
       force: true
+    - id: 6
```
make-all, line 16: The id changed from `17` to `6` in the move up to here. Was it intentional?

AlexK98: ooops. usual copy/paste issue :)

PL4STER:
Is it possible to add support for "XOCA-DAC7321C" ??

PR #2655, opened 2024-12-16
Add t34_smart_plug_plus.yaml

AlexK98:
T34-Smart-Plug+ configuration


make-all:
This is essentially the same as Aubess 1-gang, so a new config is not needed

PR #2659, opened 2024-12-17
Adding Mirabella Genio door/window sensor config

PHeonix25:
Adding support for Mirabella Genio Door/Window Sensor. 

It's quite similar to the Tuya Generic Zigbee sensor, but reports on tampering (_or "tempering" as Tuya describes it_) and is not Zigbee, but instead uses an XR809/XR3 controller which is quite slow & laggy.


make-all:
I think this is best handled by modifying the existing config. There are approaching 1000 configs now, and it will have a performance impact to make separate configs for every minor variation.

PR #2665, opened 2024-12-18
Add configuration for Devola Intelligent Heater

jaimbo:
Adding support for Devola Intelligent Heater - https://www.devola.co.uk/products/devola-designer-2kw-smart-glass-panel-heater-with-timer-black-dvpw2000b

This also supports the 1kw model and presumably the 1.5kw model as well


make-all:
This config is identical to the Andersson GSH config, only with added bugs.

PR #2666, opened 2024-12-18
Add device ids 

chris-garbas:
- Add device ID for ips pro heatpump
- Add new device id for klarstein dryfy pro connect


chris-garbas:
why is the linting check failed?

make-all:
The failure is unrelated

PR #2667, opened 2024-12-19
fix(akai_dryer): drying time should be writable

YodaDaCoda:
this DP is used when the Drying Mode is Time

the string values I've chosen are based on what's displayed on the dryer's LED display

I initially tried to provide values in minutes so they could be displayed in HASS and formatted appropriately but I couldn't seem to get that to work. Should I remove `class: duration` and `unit: min`?

### DIFF
--- a/custom_components/tuya_local/devices/akai_dryer.yaml
+++ b/custom_components/tuya_local/devices/akai_dryer.yaml
@@ -113,6 +113,53 @@ secondary_entities:
             value: "Medium"
           - dps_val: 3
             value: "High"
+  - entity: select
+    name: Drying time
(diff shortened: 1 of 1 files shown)


make-all:
The duration class is only available to sensor entities. I guess they do not implement it for number entities because it makes input more complicated than a simple number if the values are formatted as ddd hh:mm:ss. Unit is only used by number and sensor entities, or the temperature on climate and water_heater entities, so both should be removed if you leave this as a select (which is probably easiest, given there is an offset).

YodaDaCoda:
Done, thanks for the quick response!

PR #2678, opened 2024-12-21
Added more entities

jannisan:
Some more useful entities and one entity name clarified.


custom_components/tuya_local/devices/diivoo_dwv010.yaml
```diff
@@ -25,22 +25,46 @@ secondary_entities:
...
+    name: Daily water usage
+    category: diagnostic
+    dps:
+      - id: 6
+        type: integer
+        name: sensor
+        unit: "l"
   - entity: sensor
     class: battery
+    name: Battery
```
make-all, line 47: Name should not be needed here, why was it added?

custom_components/tuya_local/devices/diivoo_dwv010.yaml
```diff
@@ -25,22 +25,46 @@ secondary_entities:
...
+    name: Cumulative water usage
     category: diagnostic
     dps:
       - id: 112
         type: integer
         name: sensor
         unit: "l"
+  - entity: sensor
+    class: volume
+    name: Daily water usage
```
make-all, line 38: This can just be an attribute of the water use sensor (unrenamed).  There doesn't seem to be a lot of value in having two separate sensors showing different versions of the same measurement.
DP 112 would benefit from having `class: total_increasing` added, so it can be added to the energy dashboard where all the time derived versions can be automatically available from Home Assistant's own data recording.

custom_components/tuya_local/devices/diivoo_dwv010.yaml
```diff
@@ -25,22 +25,46 @@ secondary_entities:
...
   - entity: sensor
     class: battery
+    name: Battery
     category: diagnostic
     dps:
       - id: 7
         type: integer
         name: sensor
         unit: "%"
+  - entity: sensor
```
make-all, line 54: It isn't obvious what information these two are providing. I suspect it is not really useful enough to justify sensors, as free text cannot be tracked in any meaningful way, only numeric values and fixed sets of discrete values.

This is why I originally changed them to attributes, and I see no reason to change that.

custom_components/tuya_local/devices/diivoo_dwv010.yaml
```diff
@@ -91,3 +115,47 @@ secondary_entities:
...
+        type: integer
+        name: value
+        unit: min
+        optional: true
+        range:
+          min: 0
+          max: 1440
+  - entity: sensor
+    class: time
+    name: Remaining irrigation time 2
```
make-all, line 146: Time remaining is the standard wording used for such sensors.
The class should be `duration`

custom_components/tuya_local/devices/diivoo_dwv010.yaml
```diff
@@ -25,7 +25,7 @@ secondary_entities:
         type: string
         name: operation
   - entity: sensor
-    class: volume
+    class: total_increasing
```
make-all, line 28: Sorry, it is confusing but there are two separate `class` attributes.
Device class (volume) goes here. State class (total_increasing) should go on the dp definition below.

custom_components/tuya_local/devices/diivoo_dwv010.yaml
```diff
@@ -44,21 +44,20 @@ secondary_entities:
...
     class: battery
-    name: Battery
     category: diagnostic
     dps:
       - id: 7
         type: integer
         name: sensor
         unit: "%"
-  - entity: sensor
+  - entity: attribute
```
make-all, line 53: To add an extra attribute to an entity, you just need to list it under the entity it should be an attribute of. These two dps are already listed under the valves above, so the definitions here can be removed.

PR #2702, opened 2024-12-28
Fresco hydrate ultra alternative

Fuechslein:
The current device configuration didn't work for my Fresco Hydrate Ultra pet fountain as it seems to provide different DPS's. So I created one that works with my model/revision.

image


custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+    dps:
+      - id: 101
+        type: integer
+        name: button
+        mapping:
+          - dps_val: 0
+            value: true
+
+  - entity: sensor
+    name: Last Filter change
```
make-all, line 121: As with above, it might be better to invert the value so it is a countdown and rename to "Filter change due" or "Filter remaining"

The unit should be `min` not `m` (meters)

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+    dps:
+      - id: 102
+        type: integer
+        name: button
+        mapping:
+          - dps_val: 0
+            value: true
+
+  - entity: sensor
+    name: Last pump maintenance
```
make-all, line 167: As above, invert and rename to "Pump maintenance due", to eliminate the need for a duplicate "percentage" sensor to give a better idea when it comes due.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+          max: 86400
+        mapping:
+          - target_range:
+              min: 100
+              max: 0
+        precision: 1
+        unit: "%"
+
+  - entity: button
+    name: Reset Pump health
```
make-all, line 200: Rename to "Pump maintenance done"

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+
+  - entity: sensor
+    name: Working Time
+    icon: mdi:briefcase-clock
+    category: diagnostic
+    dps:
+      - id: 104
+        type: integer
+        name: sensor
+        range:
```
make-all, line 219: You don't need a range on a read-only value, unless you are inverting it or shifting it into a different target range.

Suggest adding a unit and `class: duration`.

And maybe a more standard name is "Runtime".

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+          min: 0
+          max: 124002400
+
+  - entity: sensor
+    name: Bright Time
```
make-all, line 224: It isn't clear what this is.  Is it something to do with the indicator?

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
+name: Pet fountain
+products:
+  - id: avhcsv4u4nj05otr
+    name: Petoneer Fresco Hydrate Ultra alternative
```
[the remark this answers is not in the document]
Fuechslein: done.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+
+primary_entity:
+  entity: switch
```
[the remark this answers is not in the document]
Fuechslein: done.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+
+  - entity: sensor
+    name: Water quality
+    icon: "mdi:water"
+    category: diagnostic
+    dps:
+      - id: 13
+        type: integer
+        name: sensor
+        unit: TDS
```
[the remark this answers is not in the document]
Fuechslein: changed to ppm.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+
+  - entity: button
```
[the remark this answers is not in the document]
Fuechslein: It only starts an one-time-run that takes 180s. It doesn't enable or disable UV sterilization, so I would keep this as a button.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+    dps:
+      - id: 101
+        type: integer
+        name: sensor
+        range:
+          min: 0
+          max: 7200
+        unit: min
+
+  - entity: sensor
```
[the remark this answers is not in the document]
Fuechslein: removed it in all 3 instances.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+          max: 7200
+        mapping:
+          - target_range:
+              min: 100
+              max: 0
+        precision: 1
+        unit: "%"
+
+  - entity: button
+    name: Reset Water health
```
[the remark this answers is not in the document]
Fuechslein: changed accordingly.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+    dps:
+      - id: 102
+        type: integer
+        name: sensor
+        range:
+          min: 0
+          max: 43200
+        unit: "m"
+
+  - entity: sensor
```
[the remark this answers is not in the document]
Fuechslein: removed.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
```diff
@@ -0,0 +1,233 @@
...
+        mapping:
+          - target_range:
+              min: 100
+              max: 0
+        precision: 1
+        unit: "%"
+
+
+  - entity: button
+    name: Reset Filter health
```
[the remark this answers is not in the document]
Fuechslein: changed.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
Fuechslein: inverted the value and fixed unit.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
Fuechslein: changed as above.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
Fuechslein: changed.

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
Fuechslein: This is a bit a tricky one since it is technically r/w but the format is not self explanatory. As per the docs:

> "The highest digit (0/1) indicates rule enable. The following 8 digits indicate start and end time. 20102320 means 20:10~23:20"

So it doesn't really have a unit. How would you deal with this? Is the `class: duration` neccessary and fitting?

custom_components/tuya_local/devices/fresco_hydrateultra_petfountain_alt.yaml
Fuechslein: Basically LED runtime and same issue as above.

PR #2704, opened 2024-12-29
Support for Mirabella Genio Fairy Lights

Cameron-D:
Have this working specifically for this product: https://www.mirabellagenio.com.au/product-range/mirabella-genio-wi-fi-1-8m-190l-led-multicolour-easy-tree-lights/
But the controller itself is a very generic plug that I believe will work for all their non-addressable fairy lights.


make-all:
This is an exact match for the already existing deltaco_shlwm_ledstrip.yaml

If you can find the actual product id (it will be logged if you use the cloud assisted flow to add the device), then that can be added to the products section, I think the one you have specified is not the product id, but some other id.

Issue #2713, opened 2024-12-30
Request support for ZigBee Energy Meter PJ-1203A


markoceri:
### Log message

```text
2024-12-31 00:13:33.406 DEBUG (MainThread) [custom_components.tuya_local.device] Bidirectional energy meter received {"1": 1197, "2": 38, "101": 11963, "102": "FORWARD", "104": "FORWARD", "105": 0, "106": 1072, "107": 19, "108": 125, "109": 19, "110": 94, "111": 4896, "112": 2292, "113": 5580, "114": 0, "115": 11963, "116": 1000, "117": 1014, "118": 1000, "119": 1000, "121": 100, "122": 1000, "123": 818, "124": 1000, "125": 1000, "127": 1000, "128": 1000, "129": 3, "full_poll": true}
```

### Product ID

bf334d83f1258a715bgzyi

### Product Name

Tuya PJ-1203A

### DPS information

```text
{
  "modelId": "ebiz5o",
  "services": [
    {
      "actions": [],
      "code": "",
      "description": "",
      "events": [],
      "name": "默认服务",
      "properties": [
        {
          "abilityId": 1,
          "accessMode": "ro",
          "code": "forward_energy_total",
          "description": "",
          "extensions": {
            "attribute": "1152",
            "trigger": "direct"
          },
          "name": "正向总有功电量",
          "typeSpec": {
            "type": "value",
            "max": 99999999,
            "min": 0,
            "scale": 2,
            "step": 1,
            "unit": "kW·h"
          }
        },
        {
          "abilityId": 2,
          "accessMode": "ro",
          "code": "reverse_energy_total",
          "description": "",
          "extensions": {
            "attribute": "1152",
            "trigger": "direct"
          },
          "name": "反向总有功电量",
          "typeSpec": {
            "type": "value",
            "max": 99999999,
            "min": 0,
            "scale": 2,
            "step": 1,
            "unit": "kW·h"
          }
        },
        {
          "abilityId": 3,
          "accessMode": "rw",
          "code": "month_energy",
[...]

PR #2714, opened 2024-12-31
Add support for Tuya Zigbee PJ-1203A bidirectional energy meter

markoceri:
I tried to add the support for this sensor Tuya Zigbee PJ-1203A bidirectional energy meter (#2713).

When I tried to add this sensor the wizard suggested using this mapping file 

> matsee_2way_energymeter

However, the power factor was in the wrong type, and some configurations were missing (all the configurations for calibrating the sensors, resets, ecc).

Added features:

- fixed the power factor for channel A, now it is correctly scaled.
- added support for all calibration setpoints (current, power, voltace, ecc)

The reset buttons seem configured correctly but don't seem to work (in any case they don't seem to work even when clicked from the tuya application)

image
image
image

### DIFF
--- a/custom_components/tuya_local/devices/zigbee_pj-1203a_bidirectional_energymeter.yaml
+++ b/custom_components/tuya_local/devices/zigbee_pj-1203a_bidirectional_energymeter.yaml
@@ -87,7 +87,6 @@ secondary_entities:
       - id: 104
         optional: true
         type: string
-        optional: true
         name: direction
       - id: 124
         optional: true
--- a/custom_components/tuya_local/devices/zigbee_pj-1203a_bidirectional_energymeter.yaml
+++ b/custom_components/tuya_local/devices/zigbee_pj-1203a_bidirectional_energymeter.yaml
@@ -423,7 +423,7 @@ secondary_entities:
     category: config
     dps:
       - id: 120
-        type: bool
+        type: boolean
         name: button
         optional: true
   - entity: button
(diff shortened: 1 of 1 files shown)


make-all:
Regarding your comments on the matsee config: the power factor is correct, it is just expressed as a % instead of unitless.

The factory calibration coefficients are not present in the log on your corresponding issue, making the matsee a 100% match. No other device exposes the calibration coefficients like this, because the device usually needs to be in a special factory mode to set them, even for devices which send the current values out on the local protocol, which this one doesn't. So I am not inclined to accept this PR.

markoceri:
The matsee mapping file has the power factor with scale 100 for channel B but it doesn't have scale 100 for channel A, and this could create an inconsistency in the data.

Furthermore matsee has the calibration coefficients as alternate values for current, energy, power ad so on (for read only), without the possibility to set them.

According tuya cloud DPS the power factor has a scale of 100 without unit. What are the differences between tuya cloud DPS and the tuya local logs?

make-all:
Sorry, I didn't scroll down far enough to see the inconsistency in B. That is incorrect, and should be fixed in the original config.

I don't think it makes sense to expose calibration dps to allow adjustment by end users. Calibrating measurement equipment is not a general end user requirement, and most users will not have the equipment or knowledge to do that. Exposing such dps opens the door to curious users messing up their equipment in a way that they cannot easily recover from.

PR #2718, opened 2025-01-02
added jiahong ET-44W Thermostat

yaro24:
added a template for the ET-44W electric underfloor heating thermostat, based on the existing ET-72W

### DIFF
--- a/custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
+++ b/custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
@@ -1,80 +1,79 @@
 name: Jiahong ET-44W thermostat
-primary_entity:
-  entity: climate
-  dps:
-    - id: 1
-      type: boolean
-      name: hvac_mode
-      mapping:
-        - dps_val: false
-          value: "off"
-        - dps_val: true
-          constraint: mode
-          conditions:
-            - dps_val: "smart"
-              value: auto
-            - dps_val: "hold"
-              value: heat
-            - dps_val: "holiday"
-              value: cool
-    - id: 16
-      type: integer
-      name: temperature
-      range:
-        min: 50
-        max: 400
-      mapping:
-        - scale: 10
-          step: 5
-          constraint: temperature_unit
-          conditions:
-            - dps_val: true
-              step: 30
-              range:
-                min: 120
-                max: 750
-    - id: 2
-      type: string
-      name: mode
-      hidden: true
-    - id: 24
-      type: integer
-      name: current_temperature
-      mapping:
-        - scale: 10
-    - id: 107
-      name: temperature_unit
-      type: boolean
-      mapping:
-        - dps_val: false
-          value: C
-        - dps_val: true
-          value: F
-    - id: 111
-      type: string
-      name: sensor
-      hidden: true
-    - id: 3
-      type: string
-      name: hvac_action
-      mapping:
-        - dps_val: "heating"
-          value: heating
-        - dps_val: "warming"
-          value: idle
-    - id: 104
-      type: integer
-      name: unknown_104
-    - id: 109
-      type: integer
-      name: holiday_days
-    - id: 112
-      type: integer
-      name: unknown_112
-    - id: 113
-      type: integer
-      name: unknown_113
-secondary_entities:
+entities:
+  - entity: climate
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: "smart"
+                value: auto
+              - dps_val: "hold"
+                value: heat
+              - dps_val: "holiday"
+                value: cool
+      - id: 16
+        type: integer
+        name: temperature
+        range:
+          min: 50
+          max: 400
+        mapping:
+          - scale: 10
+            step: 5
+            constraint: temperature_unit
+            conditions:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
```diff
@@ -0,0 +1,194 @@
...
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: "smart"
+                value: auto
+              - dps_val: "hold"
+                value: heat
+              - dps_val: "holiday"
```
make-all, line 18: I don't think this is correct. Holiday mode is not cool. It is probably better to make these preset_modes and have hvac_mode just switch between heat and off.

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
```diff
@@ -0,0 +1,194 @@
+name: Jiahong ET-44W thermostat
```
make-all, line 1: Branding should be under a `products` section (requires product id).
For top level name, I suggest just Thermostat.

Reasoning:
 - HA prefixes device names in various places in the UI, so it is better to keep it short.
 - Other devices may match the same config (likely same device sold under different brands around the world), so default name without branding is less likely to confuse.
 - if a user has multiple thermostats, they can always rename the device (they would need to anyway if they are the same brand and model).

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
```diff
@@ -0,0 +1,194 @@
...
+        range:
+          min: 50
+          max: 400
+        mapping:
+          - scale: 10
+            step: 5
+            constraint: temperature_unit
+            conditions:
+              - dps_val: true
+                step: 30
```
make-all, line 32: This is a bit strange. I would expect a step of 10 for Fahrenheit, and a range close to what the Celsius range is (5-40C = 41-104F, so I expect maybe 400-1000 or 410-1040 for the range)

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
```diff
@@ -0,0 +1,194 @@
...
+            value: heating
+          - dps_val: "warming"
+            value: idle
+      - id: 104
+        type: integer
+        name: unknown_104
+      - id: 109
+        type: integer
+        name: holiday_days
+      - id: 112
```
make-all, line 71: 112 and 113 are room temperature calibration and floor temperature calibration for ET72W (recently discovered). Maybe the same for this model, but 103 is different (integer vs boolean) so not guaranteed.

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
```diff
@@ -0,0 +1,194 @@
...
+        type: integer
+        name: unknown_112
+      - id: 113
+        type: integer
+        name: unknown_113
+  - entity: lock
+    name: Screen lock
```
make-all, line 78: Use `translation_key: child_lock` is better I think, then icon mappings can be removed.

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
```diff
@@ -0,0 +1,194 @@
...
+    dps:
+      - id: 108
+        type: boolean
+        name: lock
+        mapping:
+          - dps_val: true
+            icon: "mdi:hand-back-right-off"
+          - dps_val: false
+            icon: "mdi:hand-back-right"
+  - entity: sensor
```
make-all, line 89: Don't duplicate the temperature here, it is already available in the climate entity.

ET72 does it because there are two sensors for room temperature and floor temperature (and also it is an old config and not following up to date best practice)

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
AlexLIte: I have the same device but it's called "TheroLife ET-44(w)" and find it like "ET-44W x4.44 WiFi Thermostat" at our marketplaces so it's better to give it short name  ET-44W

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
AlexLIte: `{\"abilityId\":16,\"accessMode\":\"rw\",\"code\":\"temp_set\",\"description\":\"range in  C with step 5,  5----40℃\\n range in F with step 10, 41--104℉\\nholiday range in  C with step  5,  5----15℃\\n     range in F with step 10 , 41---59℉\",\"extensions\":{\"iconName\":\"icon-dp_temp\",\"attribute\":\"2\"},\"name\":\"target temperature\",\"typeSpec\":{\"type\":\"value\",\"max\":400,\"min\":50,\"scale\":1,\"step\":5,\"unit\":\"℃\"}},`
I got this from Tuya  API. Descriptions were translated from Chinese by google.  And "holiday" likely cool mode or somehow to keep floor temperature above 0 C

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
make-all: This is precisely why I am suggesting above to make the top level name just "Thermostat". Someone else might also find the same product in their market being sold as DS-33X.

PR #2726, opened 2025-01-04
Add support for HISB80A Energy meter

estevao90:
I didn't find a good link for reference, so linking the manual: https://www.geonav.com.br/wp-content/uploads/2022/04/HIS80A_Manual.pdf

HA demo:
image


custom_components/tuya_local/devices/hisb80a_energymeter.yaml
```diff
@@ -0,0 +1,168 @@
...
+entities:
+  - entity: switch
+    class: outlet
+    dps:
+      - id: 16
+        type: boolean
+        name: switch
+  - entity: sensor
+    name: Total Energy
+    class: energy
```
make-all, line 14: If the total_increasing is removed, you also need to remove the `energy` class too avoid complaints from HA. Unfortunately it is not possible to implement a nett energy sensor that goes up and down without also having a last reset timestamp so HA can tell the difference between a drop due to excess solar production and a reset then consumption.
But since you have separate consumed and produced sensors, HA should be able to derive the nett use from those instead.

estevao90: That makes sense. Thank you. I've fixed it and it looks like it's working fine.

PR #2728, opened 2025-01-04
Add device Ledvance Smart+PlaBL100x25B

akop:
First try to add a new device. To avoid the duplicate light id in hass I had to add both lights as secondary device.


custom_components/tuya_local/devices/ledvance_smart_plabl100x25b.yaml
```diff
@@ -0,0 +1,79 @@
+name: Ledvance Smart+PlaBL100x25B
+products:
+  - id: bfaf1efff4a825ac2c8aly
```
make-all, line 3: This is a device id, not a product id

custom_components/tuya_local/devices/ledvance_smart_plabl100x25b.yaml
```diff
@@ -0,0 +1,79 @@
...
+      name: switch
+secondary_entities:
+  - entity: light
+    name: RGB LED
+    category: config
+    dps:
+      - id: 20
+        type: boolean
+        name: switch
+      - id: 26
```
make-all, line 48: Are you sure this is brightness? In the standard light dp layout, 26 is timer. RGB light brightness should be taken from the v component of the hsv fields.

custom_components/tuya_local/devices/ledvance_smart_plabl100x25b.yaml
akop: true, updated with product id from tinytuya:
`Unknown v3.3 Device   Product ID = key9cdjdeqjph5sh  [Valid Broadcast]:

Address = 192.168.0.XXX   Device ID = bfaf1efff4a825ac2c8aly (len:22)`

custom_components/tuya_local/devices/ledvance_smart_plabl100x25b.yaml
akop: yep and now I known why it was working. Thanks for this input ;)

PR #2731, opened 2025-01-04
Update breville_lad208_dehumidifier.yaml

neilmartin83:
Remove deprecated entity for humidity sensor.

Due to related log message:

```
2025-01-04 11:36:56.103 WARNING (MainThread) [custom_components.tuya_local.helpers.config] The use of sensor for Dehumidifier is deprecated and should be replaced by humidifier.current_humidity.
```


make-all:
The deprecation message was only added in October, it is still early to remove the entity completely.

PR #2732, opened 2025-01-04
Update ustellar_ut99911_rgb.yaml

neilmartin83:
Change incorrectly named ID for DP 21 `color_mode` ID to `work_mode`. This stops the following log entry:

```
2025-01-04 16:30:55.116 WARNING (MainThread) [homeassistant.components.light] light.liam_s_monitor_light_bar_rear_rgb_led (<class 'custom_components.tuya_local.light.TuyaLocalLight'>) set to unsupported color mode white, expected one of {<ColorMode.HS: 'hs'>}, this will stop working in Home Assistant Core 2025.3, please create a bug report at https://github.com/make-all/tuya-local/issues
```

DP 21 is a string that changes when you select a different tab in the Smart Life app for the light. With this change it now shows as an attribute properly (screenshot below) and updates when I page through the different tabs in the app:

```
{\"abilityId\":21,\"accessMode\":\"rw\",\"code\":\"work_mode\",\"description\":\"调色场景、动态模式、场景模式、音乐场景，白光模式\",\"extensions\":{\"iconName\":\"icon-dp_mode\"},\"name\":\"模式\",\"typeSpec\":{\"range\":[\"color\",\"dynamic_mod\",\"scene_mod\",\"music\",\"white\"],\"type\":\"enum\",\"typeDefaultValue\":\"color\"}},
```

image

### DIFF
--- a/custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
+++ b/custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
@@ -35,16 +35,18 @@ entities:
         name: switch
       - id: 21
         type: string
-        name: color_mode
+        name: work_mode
         mapping:
           - dps_val: dynamic_mod
             value: Dynamic
           - dps_val: color
-            value: hs
+            value: Color
           - dps_val: scene_mod
             value: Scene
           - dps_val: music
             value: Music
+          - dps_val: white
+            value: White
       - id: 24
         name: rgbhsv
         type: hex


neilmartin83:
Thinking about it - I'll move the entity to its own text sensor rather than it being an attribute of one of the light entities (as this has 2). Will have a play and submit another commit shortly.

neilmartin83:
Moved it to a select entity in the configuration section. It can be set in HA and does change device behaviour. The Smart Life app also changes the selected tab automatically too.

image

neilmartin83:
Also made a change to the attribute represented by DP 103 so it's more descriptive/useful:

image

custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
```diff
@@ -65,8 +53,21 @@ entities:
               min: 0
               max: 1000
       - id: 103
-        name: line_sequence_adjustment
-        type: integer
+        type: string
+        name: RGB Order
+        mapping:
+          - dps_val: 1
```
make-all, line 59: Numbers used as a string should have "quotes"

make-all (commented):
If the light supports white, then white should be added to colour modes. Otherwise, what is the behaviour of the light when white is selected?

If white is not supported. then color_mode is not really necessary, as it is always hs, but in that case renaming to effect keeps it within the light, and "color" in that case should map to "off" (for no effect selected), and if "white" basically does nothing, then map that to "off" as well and mark it hidden as there is no reason to select it from HA, it is just there because of a bug when controlled by the app.

neilmartin83:
DP 21 doesn't look to relates with HA's `color_mode`. It's related to the page/tab you have open in the Smart Life app and if you're using a Dynamic Mode or Scene. From more experimentation we can remove it entirely as it's not useful to change it from HA and doesn't offer any more functionality for the light bar beyond the provided entities. Updating other entities in HA also change values of this one.

Out of curiosity, I observed the following:

Changing this to `white` selects the White page in the Smart Life app if it's open:

86CA7BA3-DFD7-495C-A30F-08645F107E93_1_102_o

Light behaviour does not change.

Changing this to `color` selects this page in the Smart Life app:

8648A482-F4CD-4043-9B21-C6406C650536_1_102_o

No change to the light.

`dynamic_mod` and `scene_mod` both select this page in the Smart Life app but with different functionality being available.  The light also changes behaviour to switch between a mode or scene respectively.

`dynamic_mod`:

You can change the sliders and select one of the modes in the blue circle at the top. The modes correspond to the Modes select entity in HA so if you change that, this also changes to `dynamic_mod`. 

6E32A00B-5DE5-491C-9C38-5FED124AB103_1_102_o

`scene_mod`:

You can select one of the scenes at the bottom (Party etc). Sliders are greyed out/not selectable. The scenes correspond to the Scene select entity in HA so if you change that, this DP also changes to `scene_mod`. 

A42D6D83-E4C1-4E13-86A1-B30AE374D462_1_102_o

make-all:
Maybe it is best then to rename the attribute to `app_mode` to clearly show it is linked to the app. But leave it as an attribute, not as a separate select. Just make sure it is possible to switch between scene, dynamic and plain color modes without that dp (ie adjusting other dps will automatically switch in all cases).

custom_components/tuya_local/devices/ustellar_ut99911_rgb.yaml
neilmartin83: Have set `type` back to `integer` as the values are always numbers from this DP. 👍🏼

neilmartin83:
Yeah - have added it back as an attribute with the name `app_mode` but with values mapped to their corresponding tabs in the app.

Thank you for the guidance!

neilmartin83:
👍😊

PR #2736, opened 2025-01-05
Cecotec BigDry 4000 dehumidifier - Fix tank full error

Alokin90:
This device uses "19":2 to report a full or removed water tank.

### DIFF
--- a/custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
@@ -52,15 +52,6 @@ entities:
         type: integer
         class: measurement
         unit: C
-  - entity: switch
-    translation_key: uv_sterilization
-    # Not supported by the device, still shown by Tuya
-    category: config
-    dps:
-      - id: 13
-        type: boolean
-        name: switch
-        optional: true
   - entity: binary_sensor
     translation_key: tank_full
     dps:
@@ -80,7 +71,7 @@ entities:
         mapping:
           - dps_val: 0
             value: false
-          - dps_val: 1
+          - dps_val: 2
             value: false
           - dps_val: null
             value: false
@@ -95,7 +86,7 @@ entities:
           - dps_val: 0
             value: OK
           - dps_val: 1
-            value: "E2 Compressor Fault"
-            # Unknown. Might be compressor fault
+            value: "Error 1"
+            # Unknown. Compressor or temperature/humidity sensor fault?
           - dps_val: 2
             value: "Water Tank Full or Removed"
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
```diff
@@ -67,7 +68,7 @@ entities:
         type: bitfield
         name: sensor
         mapping:
-          - dps_val: 1
+          - dps_val: 2
```
make-all, line 71: Line 83 also needs updating to stay in sync.

custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
```diff
@@ -94,6 +95,7 @@ entities:
           - dps_val: 0
             value: OK
           - dps_val: 1
-            value: "Water Tank Full or Removed"
-          - dps_val: 2
             value: "E2 Compressor Fault"
```
make-all, line 98: It may be better to delete this if unsure

custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
```diff
@@ -54,6 +54,7 @@ entities:
         unit: C
   - entity: switch
     translation_key: uv_sterilization
+    # Not supported by the device, still shown by Tuya
```
make-all, line 57: It may be better to move this to an attribute rather than having an entity that does not work as advertised

Alokin90:
Honestly, reading #2571 I think it'd be better to check with other owners of this device. Just in case there are variants of the same model name with different features.
Tagging @darth-hp in case he has it.

The Tuya app shows UV sterilization, but it doesn't work. This product isn't even advertised with a UV sterilization feature.
Also how the empty/removed water tank is reported might be different.

custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
Alokin90: Sorry! missed that

custom_components/tuya_local/devices/cecotec_bigdry_4000_dehumidifier.yaml
Alokin90: Renamed to "Error 1" as the original JSON reports E1 and E2.
Might be a compressor fault or temperature/humidity sensor issue, or something else.

darth-hp:
I can confirm, 2 is Water tank removed/full. It's a reed relais triggered by a magnet attached to a swimmer.

As for the UV part I replicated what the Tuya app was offering. I have no idea if that is really used. Once switched on, it goes off the next few seconds. That function is not mentioned in the docs or product offering though.
Cecotec

Unfortunately, I wasn't able to mimic the timer function.

Issue #2733, opened 2025-01-05
Request support for <productname>


vytux-com:
### Log message

```text
2025-01-05 15:48:50.074 WARNING (MainThread) [custom_components.tuya_local.config_flow] Adding Fan lamp switch device with product id hqq73kftvzh8c92u
2025-01-05 15:48:50.074 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches zigbee_smart_bulb with quality of 60%. DPS: {"updated_at": 1736052527.1404722, "1": true, "3": "2", "6": "0", "7": 0, "9": true}
2025-01-05 15:48:50.074 WARNING (MainThread) [custom_components.tuya_local.config_flow] Include the previous log messages with any new device request to https://github.com/make-all/tuya-local/issues/
```

### Product ID

hqq73kftvzh8c92u

### Product Name

3A Nue Light Fan Switch 

### DPS information

```text
{
  "result": {
[...]

PR #2737, opened 2025-01-05
Add 3A Nue Smart AC Fan 3 Speed & Lamp Controller

vytux-com:
#2733 Add support for 3A Nue Smart AC Fan 3 Speed & Lamp Controller


vytux-com:
@make-all not quite sure what the test failure means.... sorry.

```
=========================== short test summary info ============================
FAILED tests/test_device_config.py::TestDeviceConfig::test_config_files_parse - AssertionError: Validation error in nue_smart_fan_3speed_light.yaml: extra keys not allowed @ data['primary_entity']
========================= 1 failed, 25 passed in 5.87s =========================
```

vytux-com:
Is this something very new? As I tried that before creating this PR and I was getting an error stating that the primary entity is missing.

Unless this is to be changed in the upcoming release?

make-all:
Yes, it is in the upcoming release. The old configs will work (with deprecation warnings), but the CI is no longer considering the old format valid because the ones in the repository should be updated to the new format already.

PR #2757, opened 2025-01-10
Consolidate dependencies and settings into pyproject.toml

slonopotamus:
Motivation: pyproject.toml is a modern replacement for setup.py/setup.cfg/requirements.txt that aims to fix architectural issues

While the benefit for tyua-local is not very big, at least this change allows to remove duplicate dependencies between requirements.txt and requirements-dev.txt.


.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "Tuya Local Dev",
 	"image": "mcr.microsoft.com/devcontainers/python:1-3.13",
-	"postCreateCommand": "pip3 install --user -r requirements.txt",
```
slonopotamus, line 4: For some reason dev dependencies were not installed here, I assume it was a bug.

make-all: Probably. This file was an old contribution from another user, I do not use vscode myself. Probably requirements.txt is enough to get completion working, at least for some imports, but dev is required for unit testing and probably for pulling in the HA imports.

make-all:
I don't see a clear description of what actual problem this is solving.

The new syntax for commands and config files seems more obscure than the simple syntax of the requirements.txt, so I am not sure this is an improvement if it is just to keep up with the latest fashion.

slonopotamus:
Okay, no problem

PR #2773, opened 2025-01-13
Added a device YAML file for the lightless Casafan Eco Genuino

tschiex:
The YAML file has been tested with the corresponding model (1.8 meters diameter model, it's possible that other diameters have different model ids). The information has been extracted from Tuya Cloud Platform.

### DIFF
--- a/custom_components/tuya_local/devices/casafan_ceiling_fan.yaml
+++ b/custom_components/tuya_local/devices/casafan_ceiling_fan.yaml
@@ -1,0 +1,42 @@
+name: Ceiling fan
+products:
+  - id: 000001nbqb
+    name: Casafan Eco Genuino
+primary_entity:
+  entity: fan
+  translation_only_key: fan_with_presets
+  dps:
+    - id: 1
+      type: boolean
+      name: switch
+    - id: 2
+      type: string
+      name: preset_mode
+    - id: 3
+      type: integer
+      name: speed
+      range:
+        min: 1
+        max: 6
+    - id: 8
+      type: string
+      name: direction
+secondary_entities:
+  - entity: select
+    translation_key: timer
+    category: config
+    dps:
+      - id: 22
+        type: string
+        name: option
+        mapping:
+          - dps_val: "off"
+            value: "off"
+          - dps_val: "1hour"
+            value: "1h"
+          - dps_val: "2hour"
+            value: "2h"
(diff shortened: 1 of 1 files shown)


make-all:
This is identical to the already existing reiga_ceiling_fan config.

I would add the new product in, but the product id does not appear to be correct (I think it is taken from "model id" returned by one of the cloud API functions, judging by the format, which is not the same as product id).

tschiex:
Ok. I thought the Cloud model id in the answer of the PAI Query was the correct source. 
The Tuya device id (to be used with the local key) is "bffeb165510fdf5a48vjrq".

make-all:
The product id is neither the device id, nor the model id.

You can find the product id in the same Tuya cloud API functions that return the local key. You can also get it from the local discovery using `python -m tinytuya scan` from a PC on the same network (you may need to `pip install tinytuya` first, and you may need to disable the device in tuya-local to unblock local connections if it is a device that only accepts one local connection at a time). Sometimes the local and cloud product ids are different (the local one seems to always start with "key" in these cases)

tschiex:
Sincere apologies for not RTFM'ing. Did it with tinytuya:

        "product_name": "Ceiling  Fan",
        "product_id": "aktxh7558udrex8a"

Hope I'm doing it the right way now :-)

make-all:
The product id is also identical to the Reiga one, so they are obviously just rebadged.

tschiex:
Thanks. Sorry for generating time losses (four both of us).

BTW, it would be nice if available YAMLs that precisely match the device (same DPs, same Product Id) could be clearly presnted as "matching". When I added the fan, I was presented with a list of several fans models, with just their stripped  filenames.

I selected the casafan one (with lights) hoping it would work based on the brand.
Then, it didn't work and there was no other casafan models. So I decided to create a YAML.

Having matching DPs and ProductId could be enough to select the YAML 'automatically' possibly?

Tuya Local is great! 
I will uninstall Local Tuya this evening.

make-all:
Usually the closest match is highlighted by default.  In this case, there is also another fan with exactly the same dps but only 5 speeds, which cannot be distinguished by a snapshot of current values alone, so it is possible that the reiga one is not selected by default. The product id probably should be taken into account for this process, I guess it could be done by giving a product_id match a score of 101% instead of 100%, so it gets boosted above a perfect dps only match.

PR #2787, opened 2025-01-15
Improvements to the Axen heat pump configuration

amantia:
- added rest of DPS
- commented some actions
- added comments about problems

### DIFF
--- a/custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
+++ b/custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
@@ -1,5 +1,3 @@
-# Most DPS above 128 not implemented, as they do not appear in the log
-# see #2561 for details if they show up in future.
 name: Combo heat pump
 products:
   - id: l01fseshej0xmz8x
@@ -14,7 +12,6 @@ entities:
         name: switch
   - entity: binary_sensor
     class: problem
-    name: Errors
     category: diagnostic
     dps:
       - id: 6
@@ -27,8 +24,76 @@ entities:
       - id: 6
         type: bitfield
         name: fault_code
+      - id: 6
+        type: bitfield
+        name: description
+        mapping:
+          - value: unknown
+          - dps_val: 0
+            value: OK
+          - dps_val: 1
+            value: e1
+          - dps_val: 2
+            value: e2
+          - dps_val: 4
+            value: e3
+          - dps_val: 8
+            value: e4
+          - dps_val: 16
+            value: e5
+          - dps_val: 32
+            value: e6
+          - dps_val: 64
+            value: e7
+          - dps_val: 128
+            value: e8
+          - dps_val: 256
+            value: e9
+          - dps_val: 512
+            value: e10
+          - dps_val: 1024
+            value: e11
+          - dps_val: 2048
+            value: e12
+          - dps_val: 4096
+            value: e13
+          - dps_val: 8192
+            value: e14
+          - dps_val: 16384
+            value: e15
+          - dps_val: 32768
+            value: e16
+          - dps_val: 65536
+            value: p1
+          - dps_val: 131072
+            value: p2
+          - dps_val: 262144
+            value: p3
+          - dps_val: 524288
+            value: p4
+          - dps_val: 1048576
+            value: p5
+          - dps_val: 2097152
+            value: p6
+          - dps_val: 4194304
+            value: p7
+          - dps_val: 8388608
+            value: p8
+          - dps_val: 16777216
+            value: p9
+          - dps_val: 33554432
+            value: p10
+          - dps_val: 67108864
+            value: p11
+          - dps_val: 134217728
+            value: p12
+          - dps_val: 268435456
+            value: p13
+          - dps_val: 536870912
+            value: p14
+
   - entity: switch
-    name: Silent mode
+    translation_key: silent_mode
     icon: "mdi:volume-mute"
     icon_priority: 2
     category: config
@@ -135,44 +200,50 @@ entities:
           - dps_val: auto_dhw
             value: auto
             available: water_on
-      - id: 165
+      - id: 117
         type: boolean
-        optional: true
-        name: away_mode
-        mapping:
-          - dps_val: null
-            value: false
-            hidden: true
-
-  - entity: water_heater
-    name: Water heater
-    dps:
-      - id: 1
-        type: boolean
-        name: operation_mode
-        # FIXME: selected mode reverts to the original setting, while
-        # modifying the DPS 109 directly works.
-        # It follows the setting made in another instance/Tuya app though
+        name: hvac_action
         mapping:
           - dps_val: false
-            value: "off"
+            value: idle
           - dps_val: true
-            constraint: work_mode
+            constraint: hvac_mode
             conditions:
-              - dps_val: cool
-                value: Cooling
-              - dps_val: heat
-                value: Heating
-              - dps_val: auto
-                value: Auto
-              - dps_val: hot_water
-                value: Hot water
-              - dps_val: cool_hot_water
-                value: Hot water and cooling
-              - dps_val: heat_hot_water
-                value: Hot water and heating
-              - dps_val: auto_dhw
-                value: Auto and hot water
+              - dps_val: ["cool", "cool_hotwater"]
+                value: cooling
+              - dps_val: ["heat", "heat_hotwater"]
+                value: heating
+              - dps_val: "hot_water"
+                value: idle
+              - dps_val: ["auto", "auto_dhw"]
+                value: null
+      - id: 106
+        type: integer
+        name: current_temperature
+        mapping:
+          - scale: 10
+
+  - entity: water_heater
+    translation_key: water_heater
+    dps:
+      - id: 109
+        type: string
+        name: operation_mode
+        mapping:
+          - dps_val: cool
+            value: cooling
+          - dps_val: heat
+            value: heating
+          - dps_val: auto
+            value: auto
+          - dps_val: hot_water
+            value: hot_water
+          - dps_val: cool_hot_water
+            value: cool_hot_water
+          - dps_val: heat_hot_water
+            value: heat_hot_water
+          - dps_val: auto_dhw
+            value: auto_dhw
       - id: 104
         type: integer
         name: temperature
@@ -202,7 +273,8 @@ entities:
             hidden: true
   # Added because of the above problem in the water_heater operation_mode
   - entity: select
-    name: Operation mode
+    translation_key: axen_heat_pump_operation_mode
+    hidden: true
     category: config
     dps:
       - id: 109
@@ -210,22 +282,23 @@ entities:
         name: option
         mapping:
           - dps_val: cool
-            value: Cool
+            value: cooling
           - dps_val: heat
-            value: Heat
+            value: heating
           - dps_val: auto
-            value: Auto
+            value: auto
           - dps_val: hot_water
-            value: Hot water
+            value: hot_water
           - dps_val: cool_hot_water
-            value: Cool + Hot Water
+            value: cool_hot_water
           - dps_val: heat_hot_water
-            value: Heat + Hot Water
+            value: heat_hot_water
           - dps_val: auto_dhw
-            value: Auto + Hot Water
+            value: auto_dhw
   - entity: sensor
-    name: Inlet temperature
+    translation_key: inlet_temperature
     class: temperature
+    category: diagnostic
     dps:
       - id: 106
         type: integer
@@ -235,8 +308,9 @@ entities:
         mapping:
           - scale: 10
   - entity: sensor
-    name: Outlet temperature
+    translation_key: outlet_temperature
     class: temperature
+    category: diagnostic
     dps:
       - id: 107
         type: integer
@@ -246,7 +320,7 @@ entities:
         mapping:
           - scale: 10
   - entity: binary_sensor
-    name: Protection
+    translation_key: protection
     class: running
     category: diagnostic
     dps:
@@ -255,7 +329,7 @@ entities:
         optional: true
         name: sensor
   - entity: binary_sensor
-    name: Frost protection
+    translation_key: antifreeze
     class: running
     category: diagnostic
     icon: "mdi:snowflake-off"
@@ -264,7 +338,7 @@ entities:
         type: boolean
         name: sensor
   - entity: binary_sensor
-    name: Defrosting
+    translation_key: defrost
     class: running
     category: diagnostic
     icon: "mdi:snowflake-melt"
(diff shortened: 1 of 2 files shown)


amantia:
I own an Axel/Dotels 10KW split pump where I wrote a configuration just before the Axen configuration was added.
I migrated the missing parts here, but I still have issues with the current config that I couldn't figure out yet myself, so help is welcome.
Those are mostly marked in the file, but the most important would be the problem with the operation mode of the water heater. 
Personally I find the way how the water heater and climate setting are separated not ideal, due to the fact that the device actually just heats the water based on certain configurations and whatever you do with that later is up to the user.
Eg. the climate values are not for room temperatures, but for the heating agent.

Some DPS i renamed slightly based on what the Tuya app, the documentation and the DPS query returns.

Furthermore I still see this output that I couldn't yet fix:

2025-01-15 09:36:40.304 ERROR (MainThread) [custom_components.tuya_local.helpers.config] Error adding binary_sensor_protection for axen_ks100w_heatpump.yaml: 'sensor'


I attach also the (auto-translated)  JSON file describing the DPS.
[r32_heat_pump_query_english.json](https://github.com/user-attachments/files/18420707/r32_heat_pump_query_english.json)

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
```diff
@@ -134,53 +135,44 @@ entities:
...
+        # It follows the setting made in another instance/Tuya app though
         mapping:
           - dps_val: false
             value: "off"
           - dps_val: true
             constraint: work_mode
             conditions:
               - dps_val: cool
-                value: air_cooling
+                value: Cooling
```
make-all, line 163: It seems I set this up ready for translations, but did not provide the translations. It would be better to finish the translations rather than change the names.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
```diff
@@ -357,3 +412,579 @@ entities:
...
+  #     - id: 180
+  #       type: integer
+  #       name: option
+  #       mapping:
+  #         - dps_val: 0
+  #           value: "0"
+  #         - dps_val: 1
+  #           value: "1"
+
+  # Below items are reported by the device.
```
make-all, line 553: If they are not useful enough to be sensor entities on their own, they can be added as attributes to another entity with which they are linked.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
```diff
@@ -357,3 +412,579 @@ entities:
...
+            value: "Zone 1"
+            icon: "mdi:home-floor-1"
+          - dps_val: 2
+            value: "Zone 2"
+            icon: "mdi:home-floor-2"
+          - dps_val: 3
+            value: "Zone 1 & 2"
+            icon: "mdi:home"
+  # TODO: Find the meaning of the values and provide a user friend description
+  - entity: sensor
```
make-all, line 582: This is possibly a bitfield that defines which modes are available in the specific model, in which case it should be used in `available` attributes in the appropriate entity.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
```diff
@@ -357,3 +412,579 @@ entities:
...
+        unit: C
+        range:
+          min: 0
+          max: 1000
+        mapping:
+          - scale: 10
+
+  # Indoor means the room where the internal unit is located
+  - entity: sensor
+    name: Indoor temperature
```
make-all, line 875: Possibly this might be a better dp to use for current_temperature in the climate entity if the previously used ambient temperature dp is not suitable.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
```diff
@@ -357,3 +412,579 @@ entities:
...
+    dps:
+      - id: 156
+        type: integer
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: "Manual setting"
+          - dps_val: 1
+            value: "Room temperature"
+  # TODO: This needs a double check as the manual is not clear.
```
make-all, line 973: It seems to be selecting between a small number of options, the meaning of each option is needed.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
```diff
@@ -134,53 +135,44 @@ entities:
...
+        optional: true
+        name: away_mode
         mapping:
-          - scale: 10
+          - dps_val: null
+            value: false
+            hidden: true
+
   - entity: water_heater
+    name: Water heater
```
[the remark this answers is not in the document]
amantia: I am not yet very familiar with how the strings are translated. I changed the names because otherwise all I got was the lower case and with underscore texts in the frontend and i see similar code in other devices, where names were provided as English strings.  I will look up though how to do it properly.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
```diff
@@ -208,10 +200,32 @@ entities:
...
+          - dps_val: cool_hot_water
+            value: Cool + Hot Water
+          - dps_val: heat_hot_water
+            value: Heat + Hot Water
+          - dps_val: auto_dhw
+            value: Auto + Hot Water
   - entity: sensor
     name: Inlet temperature
     class: temperature
-    category: diagnostic
```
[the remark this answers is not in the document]
amantia: I don't really mind and was not really aware what should go under diagnostics and what is under sensors. My reasoning was to have those fields that give variable data throughout the functioning, like temperature values as not part of diagnostic as the user could use them for automations.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
amantia: Sure, I would gladly do that way.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
amantia: I think both are not good choice. The ambient temperature reported is the outside temperature, which makes no sense to report. The internal temperature is the temperature of the room where the unit is, again not that useful for the end user.  The only thing that I see valuable are one the below:
- temperature of the heating water
- temperature of the returned water
- temperature of the water tank (for the domestic hot water)

As I wrote in the first comment, I am not sure using water_heater or even more climate is really good choice for such devices. Personally what I am doing is controlling fully the device from HA, thus the best is to directly access all the DPS values, not through an intermediate device, like water_heater or climate. There might be other use cases though, that is why I didn't push for that.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
```diff
@@ -134,53 +135,44 @@ entities:
           - dps_val: auto_dhw
             value: auto
             available: water_on
-      - id: 117
+      - id: 165
         type: boolean
-        name: hvac_action
```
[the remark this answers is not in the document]
amantia: 117 t```
ells the state of the option 4 way walve .
                        {
                            "abilityId": 117,
                            "accessMode": "ro",
                            "code": "valve",
                            "description": "",
                            "name": "Four-way valve status",
                            "typeSpec": {
                                "type": "bool"
                            }
                        },
```
I don't see how this is related to the climate entity. It is something that gets turned on/off while the pump is running, if you have it installed. I don't even have it.

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 2 above)
Comment (amantia) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:553]: Some min/max values indeed might be linked, but they don't match very well, as there can be different  max value for heating temperature depending on other configurations, like how many zones there are defined. The heat pump can deal with no zones, one zone or two zones and a different set of combination how the heating behaves per zone. 
I was mostly referring that some of these cannot be modified remotely, you need to modify on the device itself, which makes it less useful for normal use case, but could be useful for advanced users to remotely see the configuration, especially that all this information is very much hidden in the official app as well behind Chinese names.

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 3 above)
Comment (amantia) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:582]: It is 
```
                        {
                            "abilityId": 131,
                            "accessMode": "ro",
                            "code": "mode_valid",
                            "description": "",
                            "name": "valid mode",
                            "typeSpec": {
                                "type": "enum",
                                "range": [
                                    "0",
                                    "1",
                                    "2",
                                    "3",
                                    "4",
                                    "5",
                                    "6",
                                    "7"
                                ]
                            }
                        },
```

but couldn't correlate yet to any real functionality, hence the comment.

**Review by amantia — COMMENTED** [review]
Thanks for the comments, I will try to improve the file and if you can give some hints to do the translations/naming properly I will gladly do that,

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 5 above)
Comment (make-all) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:973]: What I would normally do in this case is to put them as attributes, so they are not directly exposed in the UI, but are still available for observation so users can figure out the meanings if motivated.

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 8 above)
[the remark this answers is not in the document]
Comment (make-all) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:139]: The four way valve is what controls the flow of coolant through a heat pump. The state is is in is absolutely linked to the status of the climate entity. When it is turned off when the device is running, it means the device is idle (not actively heating or cooling, though it may leave the fan/pump running so it can monitor the air/water temperature.

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 6 above)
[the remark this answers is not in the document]
Comment (make-all) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:148]: What language is your HA front end set to? It may be that we still need translations for that language.

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 7 above)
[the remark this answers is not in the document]
Comment (make-all) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:214]: You can use diagnostic sensors in automations. The only effect of marking them as diagnostic is that they don't get exported by default to voice assistants, and auto-generated UIs will not include them. But they are still there and can always be explicitly used.

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 2 above)
Comment (make-all) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:553]: I have now also added the possibility to mark entities as hidden, in which case they will not be included in HA by default, but are available for the user to enable if they want them. This may also be suitable for these sensors.

**Comment by amantia** [comment]
FYI, I am reworking based on the suggestion, just need more time ...

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 8 above)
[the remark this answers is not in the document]
Comment (amantia) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:139]: You might be right indeed, sorry I confused with the optional 3 way valve that is outside of the pump. Indeed, it is turned on/off based on the fact if the pump is running (as in heating/cooling)  or not. I revert to the original code for now.

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 2 above)
Comment (amantia) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:553]: I added hidden: true for those sensors, but I don't see a change, they are still fully visible.  I saw for other devices a possibility to have by default hidden entities, e.g it show `+4 entities not shown`

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml (hunk 6 above)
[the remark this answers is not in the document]
Comment (amantia) [reply custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml:148]: I use English for the HA frontend for now.
I tried to add the translations and it works for the separate entity for operation mode, but not when used inside the water heater:
Screenshot_20250119_191419
Screenshot_20250119_191619

File: custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml [hunk 9]
```diff
@@ -134,53 +135,44 @@ entities:
...
+          - dps_val: null
+            value: false
+            hidden: true
+
   - entity: water_heater
+    name: Water heater
     dps:
       - id: 1
         type: boolean
         name: operation_mode
```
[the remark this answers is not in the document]
amantia: Seems to work, minus the missing translation for the operation_mode. I couldn't figure out how to solve that, tried also adding things like
```
    "water_heater": {
      "operation_mode": {
        "state": {
          "cooling": "Cooling",
          "heating": "Heating",
          "auto": "Auto",
          "hot_water": "Hot water",
          "cool_hot_water": "Hot water and cooling",
          "heat_hot_water": "Hot water and heating",
          "auto_dhw": " Auto and hot water"
        }
      }
    }
```
or
```
    "water_heater": {
      "state_attributes": {
        "operation_mode": {
          "state": {
            "cooling": "Cooling",
            "heating": "Heating",
            "auto": "Auto",
            "hot_water": "Hot water",
            "cool_hot_water": "Hot water and cooling",
            "heat_hot_water": "Hot water and heating",
            "auto_dhw": " Auto and hot water"
          }
        }
      }
    }
```

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
amantia: FTR, I set the water heater to show the water tank temperature, the climate to show the inlet water temperature as that is what you can control with.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
amantia: I honestly not sure how to do it and won't have time to deal with it for at least a week. I am fine if they are hidden by default.

custom_components/tuya_local/translations/en.json
amantia, line 1: The large diff is probably because of whitespace changes.  I should have not removed anything, just added.

custom_components/tuya_local/devices/axen_ks100w_heatpump.yaml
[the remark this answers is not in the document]
make-all: You need an extra layer which is the translation_key. For the default case, "_" is used.

make-all:
Partially added some of the changes. Sensors above 128 do not seem to be reported by the device according to available info in the original report, so I skipped those since they need a complete rewrite to overcome the excessive translations added without actual translation.

andraskdab:
Thanks for handling this and sorry for being silent, was busy with other stuff.

PR #2804, opened 2025-01-17
Updated Italian translation

Alokin90:
Corrected and completed setup stages.
Corrected various strings.


Alokin90:
I don't use it but others might still find this useful :)

PR #2812, opened 2025-01-18
Add support for Gosund Dimmer SW2

jcconnell:
Adds support for the [Gosund SW2 Dimmer switch](https://us.ghomesmart.com/products/app-smart-dimmer-switch-sw2).

Tuyas API Explorer output:

### Device details:

```json
{
  "result": {
    "active_time": 1736548924,
    "bind_space_id": "xxx",
    "category": "dj",
    "create_time": 1588141247,
    "custom_name": "Light",
    "icon": "smart/icon/ay1517974017716dABhC/8d518c970a152a978d8c3513bfcdb5bb.png",
    "id": "xxx",
    "ip": "xxx",
    "is_online": true,
    "lat": "xxx",
    "local_key": "xxx",
    "lon": "xxx",
    "model": "",
    "name": "SW2 2",
    "product_id": "qcgkaqmaivuwfwz4",
    "product_name": "SW2-AR",
    "sub": false,
    "time_zone": "-10:00",
    "update_time": 1736548926,
    "uuid": "xxx"
  },
  "success": true,
  "t": 1737243160296,
  "tid": "xxx"
}
```

### Properties:

```json
{
  "result": {
    "properties": [
      {
        "code": "led_switch",
        "custom_name": "",
        "dp_id": 1,
        "name": "",
        "time": 1737241202185,
        "type": "bool",
        "value": true
      },
      {
        "code": "bright_value",
        "custom_name": "",
        "dp_id": 3,
        "name": "",
        "time": 1737182288352,
        "type": "value",
        "value": 167
      },
      {
        "code": "temp_alarm",
        "custom_name": "",
        "dp_id": 101,
        "name": "",
        "time": 1736548932942,
        "type": "bool",
        "value": false
      }
    ]
  },
  "success": true,
  "t": 1737243283469,
  "tid": "xxx"
}
```

### Instruction set

```json
{
  "result": {
    "category": "dj",
    "functions": [
      {
        "code": "switch_led",
        "desc": "switch led",
        "name": "switch led",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "bright_value",
        "desc": "bright value",
        "name": "bright value",
        "type": "Integer",
        "values": "{\"min\":25,\"scale\":0,\"unit\":\"\",\"max\":255,\"step\":1}"
      }
    ]
  },
  "success": true,
  "t": 1737243342447,
  "tid": "xxx"
}
```


tests/devices/test_gosund_SW2_dimmer_switch.py
```diff
@@ -0,0 +1,46 @@
+"""Tests for the switch entity."""
```
make-all, line 1: There is no need to add specific tests for such a simple device. Every additional test added takes additional time, so the practice of adding tests for every device was ended long ago. Now it is only done when there is new features added in the python code.

custom_components/tuya_local/devices/gosund_SW2_dimmer_switch.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - entity: binary_sensor
+    name: Temperature Alarm
+    class: problem
+    category: diagnostic
+    icon: "mdi:thermometer-alert"
+    dps:
+      - id: 101
+        type: boolean
+        name: sensor
+        readonly: true
```
make-all, line 27: readonly is redundant for sensors.

custom_components/tuya_local/devices/gosund_SW2_dimmer_switch.yaml
```diff
@@ -0,0 +1,28 @@
+name: Gosund SW2 Dimmer Switch
```
make-all, line 1: This name should be generic. With such a simple device, there is high likelihood that the config will match other brands.

jcconnell: Got it - what's a good generic choice? `Dimmer Switch`?

custom_components/tuya_local/devices/gosund_SW2_dimmer_switch.yaml
```diff
@@ -0,0 +1,28 @@
...
+        type: boolean
+        name: switch
+      - id: 3
+        type: integer
+        name: brightness
+        range:
+          min: 25
+          max: 255
+  - entity: binary_sensor
+    name: Temperature Alarm
```
make-all, line 19: Naming: HA naming convention is to only capitalise the first word of a device or entity name.

Also, the class may be better as `heat` rather than the more generic `problem`

custom_components/tuya_local/devices/gosund_SW2_dimmer_switch.yaml
make-all: Yes, but following HA naming convention, "Dimmer switch"

PR #2820, opened 2025-01-19
Fix missing device_class: energy

cdalexndr:
Needed to add to HA energy monitoring: https://www.home-assistant.io/docs/energy/faq/#troubleshooting-missing-entities

Example of missing device_class:
image

### DIFF
--- a/custom_components/tuya_local/devices/smartplugv2_energyv3.yaml
+++ b/custom_components/tuya_local/devices/smartplugv2_energyv3.yaml
@@ -53,6 +53,7 @@ entities:
   - entity: sensor
     category: diagnostic
     name: Energy
+    class: energy
     dps:
       - id: 17
         name: sensor


make-all:
This is left off deliberately, because 99% of Tuya smart plugs do not return usable data in dp 17 (add_ele).

cdalexndr:
@make-all any workaround for this?

My device is a MOGODA circuit braker:
res_5a98c9419cbfe7114c6cb7de93186373

make-all:
The workaround is to use an integration helper on the power sensor.

If you think the energy sensor is returning useful data (46 is higher than smartplugs usually get before resetting, but still quite low, so unless this is a brand new device it is probably still resetting too frequently to be usable, like at 0.1kWh) then you could try setting up a template sensor (helper) to track it and set the energy device class and total_increasing state class on there.

PR #2822, opened 2025-01-20
Adding two smart WiFi RCBOs

rbb-dev:
Adding tob9l-63m_breaker.yaml for TOMZN 63A 1P+N WIFI Smart Metering RCBO switch.
Product id: q4btghzwwksqec2p

Adding tongou_stb3l125wj_breaker.yaml for STB3L-125/WJ 63A 3P+N WIFI Smart Metering RCBO switch
Product id: sh45wydgedautxda

Things Models for both switches attached.
[tob9l-63m_breaker_things_model.txt](https://github.com/user-attachments/files/18476707/tob9l-63m_breaker_things_model.txt)
[tongou_stb3l125wj_breaker_things_model.txt](https://github.com/user-attachments/files/18476708/tongou_stb3l125wj_breaker_things_model.txt)

Both devices tested and working.
Screenshots attached.
tob9l-63m_breaker
tongou_stb3l125wj_breaker

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v3
 
-      - name: Create GitHub Release
+      - name: Create Release
         id: create_release
         uses: actions/create-release@v1
         with:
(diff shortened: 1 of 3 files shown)


make-all:
It really would be better as two separate PRs. These are quite close to already existing configs, so there is a high chance I will reject one (or both) and modify an existing config to better support it instead.

rbb-dev:
> It really would be better as two separate PRs. These are quite close to already existing configs, so there is a high chance I will reject one (or both) and modify an existing config to better support it instead.

Good with me either way, if you want to update existing configs it would probably be better, cleaner.

PR #2826, opened 2025-01-21
Add support for OGACFO 9L Floor Ultrasonic Humidifier

ShawnBuckley:
Adds support for OGACFO 9L Floor Ultrasonic Humidifiers. Tested working for several days.

Output from TinyTuya:
```
Upstairs Humidifier   Product ID = 4wilpf0amoade7i3  [Valid Broadcast]:
    Address = xxx   Device ID = xxx (len:22)  Local Key = xxx  Version = 3.4  Type = default, MAC = 
    Status: {'1': True, '5': True, '14': 30, '16': False, '19': 'cancel', '22': 0, '23': 'level_3', '101': '00', '102': 0}
```

image

### DIFF
--- a/custom_components/tuya_local/devices/ogacfo_ultrasonic_floor_humidifier.yaml
+++ b/custom_components/tuya_local/devices/ogacfo_ultrasonic_floor_humidifier.yaml
@@ -1,0 +1,102 @@
+name: Humidifier
+products:
+  - id: 4wilpf0amoade7i3
+    manufacturer: OGACFO
+    model: LFHM055
+entities:
+  - entity: humidifier
+    class: humidifier
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+      - id: 101
+        name: humidity
+        type: integer
+        range:
+          min: 30
+          max: 80
+        mapping:
+          - step: 5
+      - id: 14
+        name: current_humidity
+        type: integer
+      - id: 23
+        type: string
+        name: mode
+        mapping:
+          - dps_val: level_1
+            value: eco
+          - dps_val: level_2
+            value: normal
+          - dps_val: level_3
+            value: boost
+  - entity: switch
+    translation_key: uv_sterilization
+    category: config
+    dps:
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/ogacfo_ultrasonic_floor_humidifier.yaml
```diff
@@ -0,0 +1,108 @@
...
+entities:
+  - entity: humidifier
+    class: humidifier
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+        mapping:
+          - dps_val: true
+            icon: "mdi:air-humidifier"
```
make-all, line 15: Don't override the icons here. The defaults in HA should be sufficient.

custom_components/tuya_local/devices/ogacfo_ultrasonic_floor_humidifier.yaml
```diff
@@ -0,0 +1,108 @@
...
+            value: "9h"
+          - dps_val: "10H"
+            value: "10h"
+          - dps_val: "11H"
+            value: "11h"
+          - dps_val: "12H"
+            value: "12h"
+  - entity: sensor
+    translation_key: time_remaining
+    category: diagnostic
```
make-all, line 102: `class: duration`

and below, unit should be `h`

custom_components/tuya_local/devices/ogacfo_ultrasonic_floor_humidifier.yaml
```diff
@@ -0,0 +1,108 @@
...
+    translation_key: tank_empty
+    category: diagnostic
+    dps:
+      - id: 22
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+            icon: "mdi:water"
```
make-all, line 63: Don't override the icons here. The tank_empty translation_key should already do it for you.

Issue #751, opened 2023-06-01
Request support for EMS Heatpump water heater


nk00li:
# Log Message

```
Please paste logs here
```

# Information about DPS mappings

```
Please paste the output here.
```
{
  "result": {
    "category": "rs",
    "functions": [
      {
        "code": "Power",
        "dp_id": 1,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "dp_id": 2,
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":10,\"max\":70,\"scale\":0,\"step\":1}"
      },
      {
        "code": "mode",
        "dp_id": 4,
        "type": "Enum",
        "values": "{\"range\":[\"auto\",\"mix\",\"electric\"]}"
      }
    ],
    "status": [
      {
        "code": "Power",
        "dp_id": 1,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "dp_id": 2,
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":10,\"max\":70,\"scale\":0,\"step\":1}"
      },
      {
        "code": "temp_current",
        "dp_id": 3,
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":-10,\"max\":100,\"scale\":0,\"step\":1}"
      },
      {
        "code": "mode",
        "dp_id": 4,
        "type": "Enum",
        "values": "{\"range\":[\"auto\",\"mix\",\"electric\"]}"
      },
      {
        "code": "fault",
        "dp_id": 9,
        "type": "Bitmap",
        "values": "{\"label\":[\"P01\",\"P02\",\"P03\",\"P04\",\"P05\",\"P06\",\"E01\",\"E02\",\"E03\",\"PA\",\"E08\",\"EA8\",\"EA9\",\"CS\"],\"maxlen\":14}"
      }
    ]
  },
  "success": true,
  "t": 1687261538138,
  "tid": "f9b767c10f5f11eea68636790d793901"
}

# Product ID

nqlqesmkrje7vdhm

# Information about how the device functions

[EMS Thermal 200L Operation Manual.pdf](https://github.com/make-all/tuya-local/files/11760383/EMS.Thermal.200L.Operation.Manual.pdf)

nk00li:
[...]

PR #2832, opened 2025-01-22
Update ems_waterheater.yaml

Waz-Cpt:
According to the owners manual, the max temp is 60 degrees C

### DIFF
--- a/custom_components/tuya_local/devices/ems_waterheater.yaml
+++ b/custom_components/tuya_local/devices/ems_waterheater.yaml
@@ -27,7 +27,7 @@ entities:
         name: temperature
         range:
           min: 10
-          max: 70
+          max: 60
         unit: C
       - id: 4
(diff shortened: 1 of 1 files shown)


make-all:
According to the platform info on #751 the maximum is 70, so I would expect the app to also offer that. Unless your model is not the same as this config was originally made for.

Waz-Cpt:
> According to the platform info on #751 the maximum is 70, so I would expect the app to also offer that. Unless your model is not the same as this config was originally made for.

The SmartLife APP has a 60 degree upper limit and the owners manual. Tnhe EMS Holding website also lists a max of 60 degrees max.

https://ems-holdings.com/product/200l-thermo-tank-hot-water-heat-pump/

Yes, the Tuyo Cloud shows a max of 70 degrees. However, the app does not let you set higher than 60.

Issue #2830, opened 2025-01-21
Request support for Sunred SMART Triangle Dark patio heater


gurtjun:
### Log message

```text
Device matches rgbww_lightbulb with quality of 75%. DPS: {"updated_at": 1737487187.9082558, "1": false, "3": 24, "5": "0", "11": "0"}
```

### Product ID

f2sycol9awtabrsp

### Product Name

取暖器

### DPS information

```text
{
  "result": {
    "model": {
      "modelId": "0000003fhw",
      "services": [
        {
          "actions": [],
          "code": "",
          "description": "",
          "events": [],
          "name": "默认服务",
          "properties": [
            {
              "abilityId": 1,
              "accessMode": "rw",
              "code": "switch",
              "description": "",
              "extensions": {
                "attribute": "4"
              },
              "name": "开关",
              "typeSpec": {
                "type": "bool"
              }
            },
            {
              "abilityId": 3,
              "accessMode": "ro",
              "code": "temp_current",
              "description": "",
              "name": "当前温度",
              "typeSpec": {
                "type": "value",
                "max": 50,
                "min": -20,
                "scale": 0,
                "step": 1,
                "unit": "°C"
              }
            },
            {
              "abilityId": 5,
              "accessMode": "rw",
              "code": "level",
              "description": "",
              "extensions": {
                "attribute": "4"
              },
              "name": "档位",
              "typeSpec": {
                "type": "enum",
                "range": [
                  "0",
                  "1",
                  "2",
                  "3",
                  "4",
                  "5",
                  "6",
                  "7",
                  "8",
                  "9"
                ]
              }
            },
            {
              "abilityId": 11,
              "accessMode": "rw",
              "code": "countdown",
              "description": "0：取消，1,2，....h计时",
              "extensions": {
                "attribute": "4"
              },
              "name": "倒计时",
              "typeSpec": {
                "type": "enum",
                "range": [
                  "0",
                  "1",
                  "2",
                  "3",
                  "4",
                  "5",
                  "6",
                  "7",
                  "8",
                  "9",
                  "10",
                  "11",
                  "12",
                  "13",
                  "14",
                  "15",
                  "16",
                  "17",
                  "18",
                  "19",
                  "20",
                  "21",
                  "22",
                  "23",
                  "24"
                ]
              }
            }
          ]
        }
      ]
    }
  },
  "success": true,
  "t": 1737488259758,
  "tid": "2ca8dfa6d82f11ef9e5f9e9247cde4aa"
[...]

PR #2837, opened 2025-01-22
Add support for Sunred SMART Triangle Dark patio heater

gurtjun:
Tried to setup basic YAML for Sunred SMART Triangle Dark patio heater (related to #2830).

Working in Home Assistant (see image below)

Please let me know what improvements can be made.

image

### DIFF
--- a/custom_components/tuya_local/devices/sunred_smart_triangle_heater.yaml
+++ b/custom_components/tuya_local/devices/sunred_smart_triangle_heater.yaml
@@ -1,9 +1,9 @@
-name: heater
+name: Heater
 products:
   - id: f2sycol9awtabrsp
     manufacturer: Sunred
-    model: ZHQ2582
-    name: SMART Triangle Dark Wall Model
+    model_id: ZHQ2582
+    model: Smart Triangle Dark Wall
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/sunred_smart_triangle_heater.yaml
```diff
@@ -0,0 +1,109 @@
...
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 3
+        name: sensor
+        type: integer
+        unit: C
+        class: measurement
+        readonly: true
+  - entity: select
```
make-all, line 22: I think this would be better converted to a number entity, so a slider can be used. The type of the dp needs to remain string, but the `value` attributes of the mapping should have the quotes removed.

make-all: Is the slider working as expected?

gurtjun: Added range in last commit, now I'm able to turn it on and set the level correctly

PR #2882, opened 2025-01-31
Add support for multiple Kogan devices

Mantrid42:
These devices have all been tested to work with my Kogan devices. Some caveats:
1. I'm ignoring the bitfield reporting for air_quality (125) since it reports data; mostly to see if it ever changes
2. Ignoring the bitfield reporting for left_right_freeze (127) because the bitfield says it's not supported but it definitely is and works
3. I barely have any "optional" tags, so the feature set would have to match exactly
4. I don't have any "hidden: unavailable" tags because they never get hidden when adding the devices


make-all:
It would be better to split these into individual PRs. The smartplug configs seem identical to existing configs, and both air conditioners seem to be broadly compatible with the fisher_summer config, but I haven't looked into the details like presets available. Quite likely the air purifier is an easy addition, the vacuum is very close to both the lx8 and lx10 and possibly could be merged with one of them, or the liectrex model that is also very close.

PR #2892, opened 2025-02-01
Add ID for Kogan dual smart plug

Mantrid42:
Sorry about before. Just tested with the updated ID and it auto-matches and works


Mantrid42:
Argh. I think you just added it anyway. Sorry. Noob :(

PR #2898, opened 2025-02-02
Create support for PC321-W-TY (Bidirectional) Energy Meter

FreemanJD:
this code has been running for my PC321-W-TY (Bidirectional) energy Meter. The difference between this code and pc321ty (which also runs well for my device) is the support for "Total Generation." The pc321zty_energy_meter (which has the same product code) does not work for my device as the dp's are different.


make-all:
Since it is just one additional sensor, I think it is better to add it to the existing config, which I have done.  I did notice though that the product id is currently attached to the config for the zigbee model, which does not match as well as the WiFi one which also had a config already, just without any reverse energy sensors mapped.

PR #2899, opened 2025-02-02
Update smart_powerstrip_3_outlet_energy.yaml

norfolkmustard:
Added additional product for better match of Antela power strip. 

Without this, the best match was dual_power_monitor_smartplugv2 with quality of 88%, which only has 2 outlets while the device has 3


norfolkmustard:
product link for ref https://www.amazon.co.uk/dp/B0D1XSVMPG

PR #2907, opened 2025-02-03
Update jiahong_et44w_thermostat.yaml

AlexLIte:
This configuration does not work for my device. Could you change this configuration or add it as a new one? My configuration is tested and it works. All dps are found and added in comments in case of improvement
image
image

### DIFF
--- a/custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
+++ b/custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
@@ -1,11 +1,12 @@
 name: Thermostat
+legacy_type: et_44w_thermostat
 products:
   - id: e4e7k8
-    manufacturer: TheroLife
-    model: ET44W
+    model: E-44W
 entities:
+####### Thermostat HUD section ########
   - entity: climate
-    translation_key: thermostat
+    translation_only_key: thermostat
     dps:
       - id: 1
         type: boolean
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
```diff
@@ -14,36 +15,78 @@ entities:
           - dps_val: false
             value: "off"
           - dps_val: true
-            value: heat
+            constraint: mode
```
make-all, line 18: This is incorrect, you should use preset_mode for "modes" that do not fit the HA hvac_mode model, not fake the hvac_modes.

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
```diff
@@ -1,11 +1,12 @@
 name: Thermostat
+legacy_type: et_44w_thermostat
```
make-all, line 2: There was never a legacy config by this name. Why are you adding this?

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
```diff
@@ -1,11 +1,12 @@
...
 products:
   - id: e4e7k8
-    manufacturer: TheroLife
-    model: ET44W
+    model: E-44W
```
make-all, line 5: Why remove the manufacturer that another user reported for this product id? If you want to add another manufacturer, that is fine too. Also, dropping the T from the model id when multiple users have reported the model as "ET" does not seem correct.

make-all (changes requested):
There is too much unnecessary churn here which hides the actual changes so I cannot see if there is anything additional worth including in the config.

If you can revert the unnecessary moving of entities around in the config, changing of names etc, I can take another look.

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
AlexLIte: I tried this before, but then HA creates two device control buttons: On/Off and a drop-menu button with a selection of modes. To automate you must use each button. The option to eliminate the name "preset_mode" allows you to simplify this, and HA creates just 1 select button
image

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
AlexLIte: I don’t know how correct this is, I decided to remove the name of any manufacturer, since iot tuya will only report the base name of ET-44W. Again, I don’t quite understand how tuya local looks for similar configurations so I thought I’d simplify it by throwing out the extra names. The missing T is a typo i'll fix it
image

custom_components/tuya_local/devices/jiahong_et44w_thermostat.yaml
AlexLIte: I don’t yet fully understand how tuya local searches for similar devices, because earlier it saw my device as beok_tgr81_thermostat, so I decided to write my device config based on beok_tgr81_thermostat, replacing the name with et-44w. Including legacy_type: avatto_me81_thermostat. If it is not necessary, I will fix it, the main thing is that it finds et-44w, because the jiahong_et44w_thermostat.yaml config is not compatible with my device. Perhaps because in "et44w" the "-" is missing

AlexLIte:
> There is too much unnecessary churn here which hides the actual changes so I cannot see if there is anything additional worth including in the config.
> 
> If you can revert the unnecessary moving of entities around in the config, changing of names etc, I can take another look.

In my configuration, I enable all the dps that Tuya Cloud shows for this device, arrange them in ascending order, give names close to the names in tuya cloud, explaining what they are responsible for, to give the opportunity to improve it if desired. Enable dp for calibrating sensors and power limits. The thermostat has dps for C and F readings, I show how to implement it, only there may be problems with the native translation of HA, so I commented them out. In addition, the "holiday" mode has its own temperature limit, which I also tried to set. It will be difficult for me to navigate the original configuration now, where all the dp are scattered in a chaotic order. Could you just add my config as a separate other device? For examle et_44w_thermostat.yaml. I can store it now in a local folder and it works, but it is annoying to create it again after each tuya local update.

PR #2917, opened 2025-02-04
Update gratkit_filament_dryer.yaml - revised code

MikN313:
Added two missing entities to the Gratkit Filament Dryer device (The speaker switch and the LCD backlight switch)


MikN313:
Tested locally on my HA instance and it works

PR #2918, opened 2025-02-05
Add Coredy R750 to compatible vacuums

TheOneOgre:
Added the Coredy R750 to the list as I just added and tested my own and it works perfectly with all variables and controls showing as they should under the "Mellerware City Move" vacuum config when setting up the device.

chrome_2025-02-05_01-31-16

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -744,6 +744,7 @@ of device.
 
 - Abir X8 vacuum cleaner
 - Airrobo P20 vacuum cleaner
+- Coredy R750 vacuum cleaner (shows as Mellerware City Move in config)
 - iHome Autoac Nova vacuum cleaner
 - Kabum Smart 500 vacuum cleaner
 - Kabum Smart 700 vacuum cleaner (also sold as Liectroux XR500[T2] and maybe others)


make-all:
It would be more useful to get the product id to add to the mellerware config file to ensure matching.

In general we don't list all the compatible models in the devices file, as it would balloon out of control.

TheOneOgre:
> It would be more useful to get the product id to add to the mellerware config file to ensure matching.
> 
> In general we don't list all the compatible models in the devices file, as it would balloon out of control.

Apologies, I just moved here to test from LocalTuya and saw the devices list and thought it was all inclusive.

If the Product ID is the one from the TuyaIoT page, it is: R3871_R750(Coredy-三档水量)

make-all:
That is not the product id, it is the product name. The id should look something like `sayafukjeqp9kupy` (the product id for Mellerware)

TheOneOgre:
> That is not the product id, it is the product name. The id should look something like `sayafukjeqp9kupy` (the product id for Mellerware)

Ah, apologies, I just found the correct place for it. Tuya API can be odd to navigate.

Product ID: eg0tdmbmozbtlzyg

PR #2983, opened 2025-02-19
Create poolex_silverline_heatpump_with_fault_codes.yaml

dorianmartinez31:


make-all:
Extra config files should not be created to add extra entities. Is Environmental Temperature Protection really a common enough occurence to justify a separate sensor from the general problem sensor?

PR #2984, opened 2025-02-19
Create smart_gate_opener.yaml

dorianmartinez31:


make-all:
Merged with dongguan_garage_door_opener.yaml, which is actually also a gate.

Issue #2997, opened 2025-02-22
Request support for Grow Planter


Levanterman:
### Log message

<details>

```text
2025-02-22 15:04:25.157 DEBUG (SyncWorker_17) [tinytuya.core.XenonDevice] status() entry (dev_type is device22)
2025-02-22 15:04:25.157 DEBUG (SyncWorker_17) [tinytuya.core.XenonDevice] final payload_dict for 'bfa8553d586062138du3wt' ('v3.2'/'device22'): {1: {'command': {'gwId': '', 'devId': '', 'uid': '', 't': ''}}, 7: {'command': {'devId': '', 'uid': '', 't': ''}}, 8: {'command': {'gwId': '', 'devId': ''}}, 9: {'command': {'gwId': '', 'devId': ''}}, 10: {'command': {'devId': '', 'uid': '', 't': ''}, 'command_override': 13}, 13: {'command': {'devId': '', 'uid': '', 't': ''}}, 16: {'command': {'devId': '', 'uid': '', 't': ''}}, 18: {'command': {'dpId': [18, 19, 20]}}, 64: {'command': {'reqType': '', 'data': {}}}}
2025-02-22 15:04:25.158 DEBUG (SyncWorker_17) [tinytuya.core.XenonDevice] building command 10 payload=b'{"devId":"bfa8553d586062138du3wt","uid":"bfa8553d586062138du3wt","t":"1740236665","dps":{"1":null,"2":null,"3":null,"4":null,"5":null,"6":null,"7":null,"8":null,"9":null,"10":null}}'
2025-02-22 15:04:26.188 DEBUG (SyncWorker_9) [tinytuya.core.XenonDevice] status() entry (dev_type is default)

2025-02-22 15:04:22.078 DEBUG (SyncWorker_2) [tinytuya.core.XenonDevice] building command 10 payload=b'{"devId":"bfa8553d586062138du3wt","uid":"bfa8553d586062138du3wt","t":"1740236662","dps":{}}'
2025-02-22 15:04:23.517 DEBUG (SyncWorker_3) [tinytuya.core.XenonDevice] status() entry (dev_type is default)
[...]

PR #3006, opened 2025-02-24
Grow Planter

Levanterman:
Pulled Request for Grow Planter Issue / New Device #2997


Levanterman:
Fixed the Errors which were being produced.
A couple of formatting errors in the Grow Planted YAML
And it looks like my local git was out of date for the Swing Arm Window YAML from 2023, so I just updated by local repo with a copy and paste of the Swing Arm from your main / devices folder. All good now. All checks passed.

PR #3016, opened 2025-02-27
Add Support for Drexma WiStat thermostat ET7AW

Kevin-0u:
Add support for Drexma WiStat thermostat (model: ET-7AW), which is used for floor heating. Product details can be found [here](https://www.amazon.com/WarmAll-Thermostat-Electric-Heating-Network/dp/B0CP9ZNBRC). I also added its document and DPs retrieved via the Tuya IoT API in [docs folder](https://github.com/Kevin-0u/tuya-local/tree/docs/custom_components/tuya_local/devices/docs) in my fork so it can be referenced if anyone wants to modify or develop this farther.

### DIFF
--- a/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
+++ b/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
@@ -165,7 +165,6 @@ entities:
         name: lock
   - entity: select
     name: Location
-    type: string
     category: config
     icon: "mdi:map-marker-multiple-outline"
     hidden: true
@@ -215,7 +214,6 @@ entities:
             icon: "mdi:thermometer"
   - entity: number
     category: config
-    type: integer
     name: Room temperature calibration
     icon: "mdi:tune-vertical-variant"
     dps:
@@ -245,7 +243,6 @@ entities:
             value: °F
   - entity: number
     category: config
-    type: integer
     name: Floor temperature calibration
     icon: "mdi:tune-variant"
     dps:
--- a/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
+++ b/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
@@ -282,7 +282,7 @@ entities:
       - id: 116
         type: integer
         name: value
-        unit: w
+        unit: W
         range:
           min: 0
           max: 4500
--- a/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
+++ b/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
@@ -274,9 +274,9 @@ entities:
           - dps_val: true
             value: °F
   - entity: number
-    category: config
     name: Heater Wattage
-    icon: "mdi:lightning-bolt-circle"
+    class: power
+    category: config
     hidden: true
     dps:
       - id: 116
--- a/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
+++ b/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
@@ -238,9 +238,9 @@ entities:
         type: boolean
         mapping:
           - dps_val: false
-            value: °C
+            value: °
           - dps_val: true
-            value: °F
+            value: °
   - entity: number
     category: config
     name: Floor temperature calibration
@@ -267,9 +267,9 @@ entities:
         type: boolean
         mapping:
           - dps_val: false
-            value: °C
+            value: °
           - dps_val: true
-            value: °F
+            value: °
   - entity: number
     name: Heater Wattage
     class: power
--- a/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
+++ b/custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
@@ -1,8 +1,9 @@
-name: Drexma WiStat Thermostat
+name: Thermostat
 products:
   - id: 1oi7spyvsihkwrrr
     manufacturer: Drexma Industries inc.
-    model: "ET-7AW"
+    model: "WiStat"
+    model_id: "ET-7AW"
 entities:
   - entity: climate
     dps:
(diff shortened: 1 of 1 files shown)


Kevin-0u:
FYI, this is primarily based on jiahong_et72w_thermostat.yaml with a few additions and modifications of the DPs.

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,324 @@
...
+    dps:
+      - id: 116
+        type: integer
+        name: value
+        unit: w
+        range:
+          min: 0
+          max: 4500
+  - entity: sensor
+    name: Runtime
```
make-all, line 285: add `class: duration`

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,324 @@
...
+    icon: "mdi:chart-timeline-variant-shimmer"
+    hidden: true
+    dps:
+      - id: 117
+        type: integer
+        name: sensor
+        unit: day
```
make-all, line 292: `unit: d`

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,324 @@
...
+        mapping:
+          - scale: 2
```
make-all, line 294: Note that Tuya uses a different definition of scale, if this is copied from the tuya info, it should be 100. But if you want the value halved, this is correct (and it the reason why I don't just follow the Tuya convention, since that cannot represent doubled values that some devices use to represent values with steps of 0.5)

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,324 @@
+name: Drexma WiStat Thermostat
```
make-all, line 1: Top level name should be generic, as other brands may also match this config (eg: just "Thermostat")

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,324 @@
...
+products:
+  - id: 1oi7spyvsihkwrrr
+    manufacturer: Drexma Industries inc.
+    model: "ET-7AW"
```
make-all, line 5: I would use `model: WiStat`, and change this to `model_id: ET-7AW`

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,324 @@
...
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: "Smart"
+                value: auto
+              - dps_val: "Manual"
+                value: heat
+              - dps_val: "Anti_frozen"
+                value: away
```
make-all, line 23: `away` is not a valid hvac_mode.

Probably it is better to just use a simpler mapping for dp 101 where true is always mapped to `heat`, and change mode to `preset_mode`.
If you add `translation_key: thermostat` to the entity definition above (between name and dps), then you can use `manual`, `program` and `away` as values in the mapping to get suitable translated preset names.

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,324 @@
...
+        type: boolean
+        mapping:
+          - dps_val: false
+            value: C
+          - dps_val: true
+            value: F
+  - entity: number
+    category: config
+    name: Heater Wattage
+    icon: "mdi:lightning-bolt-circle"
```
make-all, line 274: You can use `class: power` here instead of the explicit icon.

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,324 @@
...
+    hidden: true
+    dps:
+      - id: 116
+        type: integer
+        name: value
+        unit: w
```
make-all, line 280: W is capital.

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
Kevin-0u: This is intentional. The Tuya DPs information doesn’t specify a unit for this value. From my observations, the runtime number increments by 1 every 12 hours, beginning at 0 when the thermostat is first registered with Tuya.

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,329 @@
...
+                step: 10
+                range:
+                  min: 410
+                  max: 1040
+      - id: 103
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: "Smart"
+            value: Program
```
make-all, line 40: Use all lower case here to pick up the translations.

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,329 @@
...
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 108
+        type: boolean
+        name: lock
+  - entity: select
+    name: Location
+    type: string
```
make-all, line 168: Type is not valid here.

custom_components/tuya_local/devices/drexma_et7aw_thermostat.yaml
```diff
@@ -0,0 +1,324 @@
...
+            step: 5
+            constraint: unit
+            conditions:
+              - dps_val: true
+                step: 10
+                range:
+                  min: -90
+                  max: 90
+      - id: 107
+        name: unit
```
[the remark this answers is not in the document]
Kevin-0u: The dynamic unit `dp` is retained, with the unit hardcoded to `°` and the `class: temperature` attribute removed. This prevents HA from performing conversions, allowing the calibration range and step to update automatically as needed.

PR #3028, opened 2025-03-02
Expose room humidity as a sensor. Fix typo in device name

dpap76:
While the room temperature is exposed as a sensor in HA, humidity does not has it's own sensor. This patch exposes a similar sensor.

Also: fix typo in the device name: atomsphere -> atmosphere. I choose to ommit the 'dehumidifier' string to avoid HA creating very long entity names


make-all:
humidity is already exposed in the humidifier entity, so there is no need for a separate sensor.

PR #3031, opened 2025-03-02
Add support for Amantii Bespoke R1-2-AP1 fireplace

Kevin-0u:
Document about the Tuya app for this fireplace can be found [here](https://github.com/Kevin-0u/tuya-local/blob/docs/custom_components/tuya_local/devices/docs/Amantii%20Mobile%20App%20Guide.pdf) and DPs info can be found [here](https://github.com/Kevin-0u/tuya-local/blob/docs/custom_components/tuya_local/devices/docs/amantii_r12ap1_fireplace.yaml).

### DIFF
--- a/custom_components/tuya_local/devices/amantii-r12ap1-fireplace.yaml
+++ b/custom_components/tuya_local/devices/amantii-r12ap1-fireplace.yaml
@@ -6,7 +6,6 @@ products:
     model_id: R1-2-AP1
 entities:
   - entity: switch
-    name: On/Off
     icon: "mdi:power"
     category: config
     dps:
@@ -24,8 +23,10 @@ entities:
             value: "off"
           - dps_val: H1
             value: heat
+            hidden: true
           - dps_val: H2
             value: heat
+            hidden: true
           - dps_val: Auto
             value: heat
       - id: 108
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/amantii-r12ap1-fireplace.yaml
```diff
@@ -0,0 +1,340 @@
...
+    icon: "mdi:power"
+    category: config
+    dps:
+      - id: 109
+        name: switch
+        type: boolean
+  - entity: climate
+    icon: "mdi:fireplace"
+    dps:
+      - id: 108
```
make-all, line 19: Why not use dp 109 here, with only off and heat as options? Then the on/off control will work from the climate entity. You anyway have this duplicated in the preset_mode to get control of the full range of values.

Kevin-0u: The distinction lies in their specific functions: DP 108 directly controls the heating element, whereas DP 109 governs the power for the entire appliance. For example, activating DP 109 powers on the device - enabling features like the embers, flame and sound - while keeping the heating off. You can then activate the heat separately using DP 108. Conversely, turning off DP 109 shuts down everything, including heat if it was turned on. This separation allows for more granular operation, which is why I didn’t limit DP 109 to just 'off' and 'heat'

PR #3036, opened 2025-03-04
Added Support for FERCO GN1 RF Heat_Cold thermostat for aerothermal systems

pacorola:
This change adds support for FERCO GN1 (Tuya) heat/cold RF thermostats, the ones that are usually supplied with aerothermal systems.

Tested at home with such thermostats, ([link](https://tucalentadoreconomico.es/accesorios-calderas/2410-cronotermostato-sin-cable-frio-calor-ferco-gn1r.html)), although I have seen identically looking thermostats branded aso as "Avidsen" ([link](https://www.elcorteingles.es/electronica/A43033585-termostato-inalambrico-wifi-conectado/?awc=13075_1740318172_7eb0f5591da27ea99962fd6036d9b984&aff_id=2118094&dclid=COGZ17P22YsDFfCaJwIdIFMGTg)) and generic alibaba RF thermostats ([link](https://etopcontrol.en.made-in-china.com/product/cJirybkUCDVS/China-Smart-WiFi-Room-Best-Thermostat-for-Underfloor-Heating-System-Thermostat.html?pv_id=1ikpgha3q1c3&faw_id=1ikpghl93890))

In any case, these are rather generic looking RF thermostats, compatible with TUYA, connected through a RF-WIFI bridge.

termostato

This modification adds support and permits setting temperatures, check if it's working or idle, change mode from heat to cold to off seamlessly and child lock them.  Works flawlessly (and much better than in the official tuya app, where setting heat/cold mode is an obscure mislabeled setting).

termostato
entidad

Already tested and working successfully at home with 8 thermostats running from 2 RF-WIFI bridges.


pacorola:
Yes, but when reviewing Tuya DPs and Device Info it is called a rather generic "HeatCold Thermostat", which drove my assumption. ¿Do you want me to remove Alibaba mention? ¿Or to just name it FERCO? Anyway, Tuya Local will match the other devices if they share the same DPs

make-all:
The original manufacturer seems to be ETOP based on the model number, but if there is not enough info about the original model number to differentiate it from the existing ETOP supported models, then using the Ferco branding for the filename is OK. Generally first to report gets naming rights, even if it is just a market specific rebranding.

pacorola:
Thanks a lot @make-all !!!!!!!!! I am soooo happy to contribute to your great project! And also thanks a lot since your project is the key to making my smart home more sustainable in the future. ¡Cheers!

PR #3054, opened 2025-03-15
Added support for MT15/MT29 AirQuality Box

Paxy:
Support for AirQuality box MT15/MT29
Screenshot from 2025-03-20 09-25-05

### DIFF
--- a/custom_components/tuya_local/devices/mt15_mt29_box.yaml
+++ b/custom_components/tuya_local/devices/mt15_mt29_box.yaml
@@ -1,0 +1,101 @@
+name: AirBox
+products:
+  - id: bf2cd2e61dc121e657jsbj
+    model: MT15/MT29
+entities:
+  - entity: sensor
+    translation_key: air_quality
+    class: enum
+    dps:
+      - id: 1
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: level_1
+            value: excellent
+          - dps_val: level_2
+            value: good
+          - dps_val: level_3
+            value: poor
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 2
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+  - entity: sensor
+    class: humidity
+    dps:
+      - id: 3
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    class: carbon_monoxide
+    dps:
+      - id: 4
+        type: integer
+        name: sensor
+        unit: ppm
+        class: measurement
+  - entity: sensor
+    name: HCHO
+    class: volatile_organic_compounds
+    dps:
+      - id: 5
+        type: integer
+        name: sensor
+        unit: µg/m³
+        class: measurement
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    class: pm25
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
+        unit: µg/m³
+        class: measurement
+  - entity: sensor
+    class: pm1
+    dps:
+      - id: 8
+        type: integer
+        name: sensor
+        unit: µg/m³
+        class: measurement
+  - entity: sensor
+    class: pm10
+    dps:
+      - id: 9
+        type: integer
+        name: sensor
+        unit: µg/m³
+        class: measurement
+  - entity: sensor
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 22
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: binary_sensor
+    class: plug
+    category: diagnostic
+    dps:
+      - id: 23
+        type: boolean
+        name: sensor
+  - entity: switch
+    icon: "mdi:sprout"
+    name: Enable Alarm
+    dps:
+      - id: 106
+        type: boolean
+        name: switch
--- a/custom_components/tuya_local/devices/mt15_mt29_airbox.yaml
+++ b/custom_components/tuya_local/devices/mt15_mt29_airbox.yaml
@@ -1,7 +1,7 @@
 name: AirBox
-products:
-  - id: bf2cd2e61dc121e657jsbj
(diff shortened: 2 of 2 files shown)


custom_components/tuya_local/devices/mt15_mt29_box.yaml
```diff
@@ -0,0 +1,101 @@
+name: AirBox
+products:
+  - id: bf2cd2e61dc121e657jsbj
```
make-all: This appears to be a device id, not product id.

1920x1200:
Hello, I have a MT29 and excited to see that support was added, but I'm having troubles getting it to work. What combination of settings do I need to select to expose all sensors? So far the closest I got is to use protocol version 3.5 and then choose the type of "co2_box" but this only exposes one of the many sensors. Am I doing something wrong, or is only one sensor supported?
Screenshot 2025-04-02 at 11 38 37 PM

Paxy:
Hmm, I had the same issue when develping this yaml file. In the list device was missing until I add:
`products:
  - id: bf2cd2e61dc121e657jsbj`
 @make-all, can you guess what could be the reason? Only after adding product id, MT15/MT19 was showned in the list.

make-all:
Since there are no logs available, I decline your invitation to play a guessing game.

make-all:
> > Hmm, I had the same issue when develping this yaml file. In the list device was missing until I have added: `products: -id: bf2cd2e61dc121e657jsbj` @make-all, can you guess what could be the reason? Only after adding product id, MT15/MT19 was showned in the list.
> 
> Where do I put that product id?

It won't help, that is not a product id, it is a device id, and is unique to @Paxy's device.

Paxy:
Ok, sorry, product id is probably: "MT15/MT29".

1920x1200:
I tried adding it again with the debugging log on. Please see screenshots below and advise where I can change the product type so that it's correct. 

Screenshot 2025-04-05 at 2 24 10 PM

Screenshot 2025-04-05 at 2 24 21 PM

Screenshot 2025-04-05 at 2 24 43 PM

Screenshot 2025-04-05 at 2 26 43 PM

make-all:
I would need the full log message that starts "Device matches zym100_presence_sensor with quality 29%.". Also the product_id that you greyed out in the screenshot above (there is nothing sensitive about that, it should be the same for all devices of this type).

1920x1200:
`product id: owmkja70doamcxkh`


`Device matches zym100_presence_sensor with quality of 29%. DPS: {"updated_at": 1743877465.847518, "1": "level_1", "2": 22, "3": 50, "4": 402, "5": 6, "7": 19, "8": 20, "9": 24, "22": 100, "23": true, "28": "middle", "101": 24, "102": 0, "104": 1500, "105": 10, "106": false, "107": 19, "108": 2, "109": 1, "110": 2, "111": 0, "112": "f", "113": 200, "114": 100, "115": 61, "116": false, "117": false, "118": false}`

make-all:
Based on the log message, it does seem to match the config, but probably you do not have the config yet because it is not released. There does seem to be a substantial amount of dps not covered by the config though.

> mt15_mt29_airbox matched 39%
>   Missing 112: f
>   Missing 114: 100
>   Missing 113: 200
>   Missing 109: 1
>   Missing 101: 24
>   Missing 105: 10
>   Missing 102: 0
>   Missing 115: 61
>   Missing 104: 1500
>   Missing 108: 2
>   Missing 110: 2
>   Missing 117: False
>   Missing 28: middle
>   Missing 111: 0
>   Missing 116: False
>   Missing 107: 19
>   Missing 118: False

make-all:
The MT29 seems to be advertised as a 9-in-1 air quality sensor, and from the config we have 7 numeric sensor values, and the subjective "air quality" sensor which they probably count, so we seem to be missing at least one sensor (unless they count battery level as one of the 9 sensors). The rest of the "missing" dps are probably related to alarm configuration, which other users may find useful even if you yourself just want the readings.

The Query Things Data Model cloud API function should give full info about all of these dps.

make-all:
The current config represents the information that has been supplied. 

The Query Things Data Model cloud API result from platform.tuya.com would probably reveal the extra sensors and config options that are not in the current config.

1920x1200:
I followed the process [listed here](https://www.zigbee2mqtt.io/advanced/support-new-devices/03_find_tuya_data_points.html#_8-find-your-data-point) and shown in [this video](https://www.youtube.com/watch?v=apWfjargTpA) and here is what I found:
```
1	AQI		
2	TEMP
3 	HUM
4	CO2
5	HCHO
7	PM2.5
8	PM1.0
9 	PM10
22	Battery Level
23	Charging
28	Alarm Volume
101 	TVOC
102	CO
103	Brightness
104	CO2 Alarm Value
105	Sleep
106	Alarm
107	PM0.3
108	Timer
109	Clock1
110	Clock2
111	Clock3
112	Temp Unit Convert
113	CO Alarm Value
114	PM2.5 Alarm Value
115	HCHO Alarm Value
116	CLOCK1_EN
117	CLOCK2_EN
118	CLOCK3_EN
```

Is this the data that is needed?

make-all:
The full info for this device has already been provided under #3209, and the config expanded to cover all the sensors.

1920x1200:
I removed and re-added it, but I'm not seeing any differences.

CO and CO2 are still reversed, and TVOC and PM0.3 are missing.

1920x1200:
Another update came through and things are improving!

Fixed:
- CO and CO2 that were flipped are now correct. 
- TVOC and PM0.3 which were missing are now available. 

New issues:
- Carbon monoxide, PM 0.3, and Volatile organic compounds (TVOC) are at categorized as "Diagnostic" instead of "Sensors"
- When the temperature unit is set to Fahrenheit, the value for the temperature is actually in Celsius but says ºF. To fix it, I have to change the temperature unit to Celsius and then the correct temperature is displayed in Fahrenheit.

Screenshot 2025-05-16 at 6 22 02 PM

PR #3057, opened 2025-03-15
Add support for Prosto WFU-T12PM smartplug

zastrixarundell:
Add support for the Prosto WFU-T12PM smartplug with energy-monitoring capabilities.

image

The energy entity is provided by an integral sensor helper.

Data exported from smartlife 3.6.1: [tuya.json](https://github.com/user-attachments/files/19265525/tuya.json)


make-all:
Merged with smartplugv2_energy.yaml

zastrixarundell:
Nevermind. Moved back to the previous file the issue seems the issue is due to my WiFi.

make-all:
There are already 1000 config files, and it now takes around 10s to scan for a new device because of the shear number of files to scan through. Adding new configs for every minor variation of smartplug (and lightbulb and other common devices) that exists would very quickly balloon out the number of configs to take minutes rather than seconds to identify new devices.

zastrixarundell:
Alright. First time contributing and I thought each device generally should have its' own config. I've opened another PR: #3092 as the committed changes don't show current nor power.

PR #3062, opened 2025-03-17
Adds support for Aquagem Inverpro pool pump

tomcastleman:
## Adds support for Aquagem Inverpro pool pump

This pump appears to be a cut down version of the following:

- [madimack_inverflow_pro](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/madimack_inverflow_pro.yaml)
- [aquaforte_vsp_poolpump](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/aquaforte_vsp_poolpump.yaml)

This pump does not have the **AI - auto inverter mode** feature and so the associated dps are missing.

I used the `madimack` config as a base with the unused dps removed.


make-all:
Merged all 3 models mentioned into the first added Madimack Inverflow config.

PR #3067, opened 2025-03-18
Add support for Manta Windy MT0200B (Weather Station)

MichalSieciechowicz:
[official website](https://manta.com.pl/produkt/stacja-pogody-z-zewnetrznym-czujnikiem-manta-windy-smart-mt0200b/)

Preview:
image


custom_components/tuya_local/devices/manta_windy_MT0200B.yaml
```diff
@@ -0,0 +1,94 @@
+name: Weather Forecast
+products:
+  - id: bf9fc4b6e5c2c91014elwe
```
make-all, line 3: This appears to be a device id rather than the required product id.

PR #3078, opened 2025-03-22
Add support for Morento MR7566-WF air purifier

ShawnBuckley:


custom_components/tuya_local/devices/morento_mr7566_wf.yaml
```diff
@@ -0,0 +1,144 @@
...
+  - entity: select
+    translation_key: timer
+    category: config
+    dps:
+      - id: 18
+        type: string
+        name: option
+        mapping:
+          - dps_val: 0
+            value: continuous
```
make-all, line 70: This should be "cancel" to make use of available translations.
Also when mapping strings that are numbers, it is better to enclose them in double quotes so they are read into Python as strings, not as integers.

PR #3082, opened 2025-03-23
Fancy Leds: add support 3 HDMI 2.1

latargere:
Add support for the fancy leds 3 HDMI 2.1 Fancy Sync box
I've named the different scene based on my smart life app.
Also I had to create another light just for the scene color, otherwise with the implementation like the one for lytmi, the brightness was not working due to the fact the light was not in color mode white and was sending only color.

### DIFF
--- a/custom_components/tuya_local/devices/fancy_led_hdmisync_backlight.yaml
+++ b/custom_components/tuya_local/devices/fancy_led_hdmisync_backlight.yaml
@@ -1,0 +1,153 @@
+name: HDMI Fancy Leds backlight
+products:
+  - id: s9tscadzzt5s2cz3
+    manufacturer: Fancy Leds
+    model: 3 HDMI 2.1 Fancy Sync Box
+entities:
+  - entity: light
+    dps:
+      - id: 20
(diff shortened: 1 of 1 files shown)


channafam:
@latargere  - any chance you can share a preview similar to below for this pull request? I have created a pull request #3089 and I am trying to evaluate whether v2.0 and v3.0 are similar or not. 


image

latargere:
Sure here is it.
IMG_7122

make-all:
Lytmi config has been modified to prevent the hsv dp from being used to change brightness. I think this change should be correct for that device also, since it makes no sense to even have the brightness dp if the brightness is supposed to be changed through the hsv v component like other Tuya lights when in color mode, since this device does not support separate white only mode.

For the scene differences, you can enable the text entity to send whatever scene data you want, so I don't consider differences in scenes to be enough to justify a new config, since these are somewhat arbitrary and there is already a fairly big performance impact caused by having 1000 config files that I need to keep under control especially when it comes to all the minor variations among lights and smartplugs there are in the market.

PR #3084, opened 2025-03-24
Sefaul Q9 Dehumidifier

peng1can:
Unfortunately this isn't the most intuitive device - neither via the Smartlife App nor the device's control panel.  The device doesn't have a humidity sensor, so the dehumidifier cant natively be triggered that way, and it has a separate air purifier component that gets automatically turned on, so I couldn't really make it work as a dehumidifier the way HA expects.  Perhaps someone with more experience would have a different approach.  But, this config does work the same way as the device's control panel.  

https://www.amazon.com/dp/B0D7959FJX

(Also my first time contributing to a project using github, so apologies if I've done something wrong)

### DIFF
--- a/custom_components/tuya_local/devices/sefaul_q9_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/sefaul_q9_dehumidifier.yaml
@@ -1,0 +1,68 @@
+name: Dehumidifier
+products:
+  - id: e8c9gukirorvyhk8
+    manufacturer: Sefaul
+    model: Q9 Dehumidifier
+
+entities:
+  - entity: light
+    class: light
+    name: Light
+    dps:
+      - id: 11
+        name: switch
+        type: boolean
+  - entity: fan
+    class: fan
+    name: Purifier
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+      - id: 4
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: "low"
+            value: "Low"
+          - dps_val: "high"
+            value: "High"
+  - entity: switch
+    name: Dehumidifier
+    dps:
+      - id: 101
+        name: switch
+        type: boolean
+  - entity: select
+    name: Timer Set
+    dps:
+      - id: 17
+        name: option
+        type: string
+        mapping:
+          - dps_val: "cancel"
+            value: "Cancel"
+          - dps_val: "2h"
+            value: "2H"
+          - dps_val: "4H"
+            value: "4H"
+          - dps_val: "8H"
+            value: "8H"
+  - entity: sensor
+    translation_key: time_remaining
+    class: duration
+    dps:
+      - id: 18
+        type: integer
+        name: sensor
+        unit: min
+        class: measurement
+  - entity: binary_sensor
+    class: problem
+    dps:
+      - id: 19
+        type: bitfield
+        name: sensor
+        mapping:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/sefaul_q9_dehumidifier.yaml
```diff
@@ -0,0 +1,68 @@
...
+        class: measurement
+  - entity: binary_sensor
+    class: problem
+    dps:
+      - id: 19
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 2
+            value: "Tank Full"
```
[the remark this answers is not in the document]
make-all: Use `translation_key: tank_full` in place of `class: problem`, that will take care of the name, icon and words used to display the true and false values.

And add a problem sensor to capture any other bit values from that dp.

custom_components/tuya_local/devices/sefaul_q9_dehumidifier.yaml
```diff
@@ -0,0 +1,68 @@
+name: Dehumidifier
+products:
+  - id: e8c9gukirorvyhk8
+    manufacturer: Sefaul
+    model: Q9 Dehumidifier
+
+entities:
+  - entity: light
+    class: light
```
make-all, line 9: There is no class "light", and no need to add a name that is the same as the entity type either, it is better to let it default to the translated default name of the entity.

custom_components/tuya_local/devices/sefaul_q9_dehumidifier.yaml
```diff
@@ -0,0 +1,68 @@
...
+entities:
+  - entity: light
+    class: light
+    name: Light
+    dps:
+      - id: 11
+        name: switch
+        type: boolean
+  - entity: fan
+    class: fan
```
make-all, line 16: Likewise, there is no class "fan".

custom_components/tuya_local/devices/sefaul_q9_dehumidifier.yaml
```diff
@@ -0,0 +1,68 @@
...
+    name: Purifier
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+      - id: 4
+        type: string
+        name: preset_mode
```
make-all, line 24: If you use `translation_key: fan_with_presets`, then change these low and high to all lower case the values will get translated.

custom_components/tuya_local/devices/sefaul_q9_dehumidifier.yaml
```diff
@@ -0,0 +1,68 @@
...
+          - dps_val: "high"
+            value: "High"
+  - entity: switch
+    name: Dehumidifier
+    dps:
+      - id: 101
+        name: switch
+        type: boolean
+  - entity: select
+    name: Timer Set
```
make-all, line 37: Remove the name and use `translation_key: timer`, then change all the mappings to lower case and the values and entity name will be translated.

make-all (approved):

peng1can:
Just wanted to say thanks for fixing this up and merging it.  My programming skills are pretty mediocre, and I am still getting hang of yaml, so I'd been trying to sort out all the changes when I had time, but wasn't get very far very quickly.  I loaded this updated file into the current release and everything seems to be working properly (relative to how wonky the controls are on this device is in the first place).

Thanks again.

PR #3092, opened 2025-03-26
Move WFU-T12PM to polled config

zastrixarundell:
Based on #3057

It needs polling to display the current and power usage.


make-all:
Please just use smartplugv2_polled_power if you need polling for the sensors. I am not adding any more smartplug configs.

PR #3105, opened 2025-03-30
Added Ambient temperature to powerworld_pw58330_waterheater.yaml

comcowo:
Removed almevo_hpht_waterheatpump.yaml since it is now included in powerworld_pw58330_waterheater.yaml
Added Ambient temperature to powerworld_pw58330_waterheater.yaml


make-all:
Ambient temperature is already available from the climate entity.

PR #3108, opened 2025-03-31
Added unit % to 'Coefficient of performance' of powerworld_pw58330_waterheater

comcowo:
powerworld_pw58330_waterheater: Added unit % to 'Coefficient of performance' to have a propper graph and not only a changing string. Actualy it is a factor but I didn't find a unit for that.

### DIFF
--- a/custom_components/tuya_local/devices/powerworld_pw58330_waterheater.yaml
+++ b/custom_components/tuya_local/devices/powerworld_pw58330_waterheater.yaml
@@ -452,7 +452,7 @@ entities:
         type: base64
         optional: true
         name: sensor
-        unit: '%'
+        class: measurement
         mapping:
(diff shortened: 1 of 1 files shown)


make-all:
Are you sure this change has the effect you expect? A factor is unitless, if you want to make it a percentage, you will need to change the scale. But usually it is `class: measurement` on the dp that makes the difference in whether HA keeps history for the sensor for use in graphs etc.

comcowo:
Thank you for your advice. I changed the 'unit: %' to  'class: measurement' which works as well and is less misleading.
However, a lot of sensors doesn't have 'class: measurement' but 'unit: xxx' and they have a graph.
It seems like both entries are making a string to number convertion.
But I am not ab expert :-)

PR #3117, opened 2025-04-06
MPPT Solar Charge Controller LM 4048/6048

Katawengineering:


Katawengineering:
Hello
I just saw that this yaml did't pass two test
I made the needed corrections (and hope I got all)

Do we have possibilities to do these 4 workflow, before opening a pull request ?? these space and "N"ame mistake are very embarrassing, I would like to deliver works which are free of that kind of mistake

Katawengineering:
I still have some warning about comments... can we let these comments or I have to clean it too?

make-all:
The tests can be run from the root directory of the project using `pytest`. For just an added yaml file, `pytest tests/test_device_config.py' is a lot quicker and covers everything needed for the new file.

The lint is using [yamllint](https://www.yamllint.com), other checks are the general HA and HACS checks which may fail for forked repos, so are best left to the CI to run.

Katawengineering:
very thanks for the info's about testing 

OK I will made the correction for the two space-mistake 
my question is still for the 4 warnings about the "comments" can I let it (for me it is an help, I'm not a programmer and sometime I don't touch codes for months... so I try to put some comments to remember what I did some months ago LOL

make-all:
Warnings will not fail the build once you get rid of the errors.

Katawengineering:
Ok the two errors should be gone 

about testing with YAMLint and pytest... I guess I will need to create a new virtual machine with Ubuntu (all this is very new for me, I'm from the windows-world) I will do it for the next integration... I'm sure I will have much more hehee....

Katawengineering (commented):
YAML got 4 "7:1 [comments-indentation] comment not indented like content"
# energy
# configuration
# Diagnostics
# base64 Diagnostic

I like to use "title" when a code is so long, to see faster where I am

but if it's not welcome to the community, then I have no problem to remove it :-)

I just saw that the comment aren't consistent formatted (some are with capital) I will change this but intend to let the comments along nobody complain about it

PR #3119, opened 2025-04-06
Add Support for Genio Galaxy Starship Projector I004545

channafam:
Add Support for Genio Galaxy Starship Projector I004545

Preview:
image


channafam:
@make-all  - please note that I have had a look at contents of galaxy_projector_light.yaml however it seems to be different. Also looked at the product online and it seems to be different.

make-all:
Galaxy projector light is the same, it was just missing Music mode, and has a different list of scenes. Rather than proliferate more configs for these sorts of arbitrary differences, I have added text entities that will allow setting any scene or music mode rather than having to select from a predefined list.

PR #3125, opened 2025-04-07
Support for Generic Tuya ceiling fan and multiple Dehumidifiers

omersht:


make-all:
There seems to be some problem with your conflict resolution when rebasing, as one of these configs is already merged and was renamed, none of the commits in this PR touches the file so I am not sure why it is showing up in the PR reinstating it under the old name.

The other two devices seem to be made by Israeli manufacturers so adding them at this time while Israel's violation of international law is ongoing is not an option I will consider.

PR #3128, opened 2025-04-07
add support for the Greenmigo Alpha Q13 pro Dehumidifier

forksiceberg:
add support for the Greenmigo Alpha Q13 pro Dehumidifier


make-all:
Merged with Eeese Otto config.

forksiceberg:
Please note that this device has a 1 humidity steps while the  Eeese Otto has 5 humidity steps

PR #3136, opened 2025-04-10
feat(device): add ElectriQ EcoSilent12HPW config

XargsUK:
This pull request includes updates to the `DEVICES.md` file and the addition of a new device configuration in the `custom_components/tuya_local` directory. The most important changes include adding a new device to the list and providing a configuration for it.

### Documentation Updates:

* Added `ElectriQ EcoSilent 12HPW` to the list of supported devices in the `DEVICES.md` file. https://www.electriq.co.uk/p/ecosilent12hpw/electriq-ecosilent12hpw

### Device Configuration:

* Added configuration for `ElectriQ EcoSilent12HPW` in `custom_components/tuya_local/devices/electriq_ecosilent12hpw_aircon.yaml`, including entities for climate control, temperature settings, fan modes, and additional features like sleep mode and timer. Note the difference with this model to the EcoSilent12WAP is the additional heat pump.

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -95,6 +95,7 @@
 - ElectriQ Airflex 15W
 - ElectriQ EcoSilent 14HPW
 - ElectriQ EcoSilent 12WAP
+- ElectriQ EcoSilent 12HPW
 - Fersk Vind 2
 - Fisher Summer air conditioner
 - Friedrich Uni-Fit air conditioner (models: UCT14A30, UCT12A30, UCT08B10A)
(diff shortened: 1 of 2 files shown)


XargsUK:
> Device is the same as already supported 14HPW

@make-all  I can take a look at a way to help present options to the user, as it's not intuitive to users that when they have a EcoSilent12HPW, then they shouldn't be selecting EcoSilent12WAP but rather the EcoSilent14HPW without the user having prerequisite knowledge of how the vendor structures their model numbers. It feels as though having brand selection, then model, and having a map in the codebase with references to generic profiles would be smoother UX than researching model numbers for products you don't own to determine similarities to your own and bruteforcing to determine if it will work.

I understand the DRY principle with this approach, but leaving out models from the model selection simply because there's a profile with the same config in the codebase but under a different model number seems problematic.

PR #3147, opened 2025-04-13
Add support for Atorch S1TW thermostat/smart socket

vvv2542:


custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
```diff
@@ -0,0 +1,396 @@
...
+                value: "off"
+      - id: 118
+        type: string
+        name: temperature_unit
+        mapping:
+          - dps_val: c
+            value: C
+          - dps_val: f
+            value: F
+      - id: 0
```
make-all, line 43: This is guaranteed to prevent matching to this config. Rather than defining a fake dp here, it would be better to make one of the c or f dps the default, and only conditionally switch on the other.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
```diff
@@ -0,0 +1,396 @@
...
+    icon: mdi:cash-multiple
+    dps:
+      - id: 125
+        type: integer
+        name: sensor
+        class: total_increasing
+        mapping:
+          - scale: 1000
+  - entity: binary_sensor
+    name: Sensor detection flag
```
make-all, line 369: Is this useful as an entity? I cannot imagine what it is for from the name.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
```diff
@@ -0,0 +1,396 @@
...
+    category: diagnostic
+    dps:
+      - id: 126
+        type: boolean
+        name: sensor
+        class: measurement
+  - entity: sensor
+    name: Add ele
```
make-all, line 377: This is not useful as a sensor, it is better to make it an attribute of some other entity.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
```diff
@@ -0,0 +1,396 @@
...
+      - id: 118
+        type: string
+        name: unit
+        mapping:
+          - dps_val: c
+            value: C
+          - dps_val: f
+            value: F
+  - entity: sensor
+    class: current
```
make-all, line 231: These sensors are not really the primary function of the device, so should be categorized as diagnostic I think.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
vvv2542: Done.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
vvv2542: I cannot imagine too. The name was taken from Tuya Developer Platform. Maybe someone knows how to use this sensor.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
vvv2542: Do you have any idea what this sensor is for? I just took the name from Tuya Developer Platform. I watched it for a long time, but I still didn’t understand what kind of meaning it showed.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
vvv2542: Atorch **S1TW** is a multifunctional device: power/energy meter, smart socket, thermostat. It's based on **S1W/S1WP**. It uses the same board, case and display. Features an additional temperature sensor and modified firmware with thermostat functionality and without timers of **S1W/S1WP**. Since sensors such as 'current' are not placed in the diagnostics category in **atorch_s1wp.yaml**, I did not do this either.

As for me, most of the time I use this device as voltage/power/energy meter and thermometer.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
make-all: It signals changes in energy usage. Since this device has a total_forward_energy sensor, it is redundant as that is more convenient to use, so it can be added as an attribute on the energy sensor.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
make-all: Marking them as diagnostic does not prevent them being used, it just makes the default generated UI cleaner. The majority of smart switches mark current, voltage and power as diagnostic, just the example you chose is an outlier and I will probably fix that now you have pointed it out.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
vvv2542: Done.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
vvv2542: Done: current, power and voltage categorized as diagnostic.

custom_components/tuya_local/devices/atorch_s1tw_thermostat.yaml
make-all: My guess would be that it either indicates which type of probe is connected, or whether the probe is connected or missing. Maybe as an attribute of the climate entity is appropriate based on the assumption it is related to the temperature probe.

vvv2542: This is an inverse probe presence sensor:
true - disconnected
false - connected
Added as 'available' attribute of temperature sensor.

PR #3162, opened 2025-04-18
Added "configured current" to goodcell_ev_charger.yaml

muhhh0:
Added DP 125
For the Feyree charging cable, DP 125 is described as:
设置电流显示 → "Set Current Display" – it displays the configured current.

According to the Tuya API Explorer, the code for 125 is "set60a", but this does not make much sense (also, it is read-only).

This value is useful because you can lower the charging limit below 8 A, which is not officially supported. If you try to do so using "set charge current" (and set it to 6 A or 7 A), the value jumps back to 8 A.
Therefore, you can use "set charge current" to set the current and validate the actual setting via this new value.

### DIFF
--- a/custom_components/tuya_local/devices/goodcell_ev_charger.yaml
+++ b/custom_components/tuya_local/devices/goodcell_ev_charger.yaml
@@ -346,4 +346,4 @@ entities:
         unit: A
         class: measurement
         mapping:
-          - scale: 1 
+          - scale: 1
--- a/custom_components/tuya_local/devices/goodcell_ev_charger.yaml
+++ b/custom_components/tuya_local/devices/goodcell_ev_charger.yaml
@@ -335,3 +335,15 @@ entities:
         type: boolean
         name: switch
         optional: true
+  - entity: sensor
+    name: Configured Current
+    class: current
+    category: diagnostic
+    dps:
+      - id: 125
+        type: integer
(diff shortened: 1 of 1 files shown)


make-all:
I think a sensor is overkill for this, an attribute makes it available for use by automations etc without adding extra noise to the UI.

dooblem:
Thanks for adding the attribute.

Having played with this device and tinytuya, I found that reading the value from DP 125 is always reliable. However the amps value is sometimes wrong when red from DP 115. @muhhh0  gave an example, but I think there are others, though I'm not 100% sure.

From my scripts I set the value using DP 115, but I read it using DP 125. 

In HA it's certainly misleading for the user, because as described by @muhhh0 , when set to 6A, the value displayed is the old value, though the setting worked.

muhhh0:
The device is actually limiting the current.
When you use the "Set charge current" slider to set a value below 8 A, the slider jumps back to the previous value, but the charger accepts the value. See screenshot - it contains "Configured Current" from my PR which displays DP 125 - slider is still at 10 A (setting before my change to 6A).
ev_charger
Don't know if there is a better way to display the actual setting.

dooblem:
I think the best would be a single entity in HA : when changed it writes the value to DP 115, but gets the value from DP 125.
I don't know if it's easy though.

PR #3168, opened 2025-04-20
Moebot S: add Start and Pause mowing

Peck07:
I have added a "Start mowing" and "Pause mowing" button, as at least for the Parkside PMRDA mower, these are the most common commands, and were missing.
(sidenote: fixed mowing is basically the "spot mode" where you need to place the mower somewhere on the lawn, it will not exit the docking station)

### DIFF
--- a/custom_components/tuya_local/devices/moebot_s_mower.yaml
+++ b/custom_components/tuya_local/devices/moebot_s_mower.yaml
@@ -88,6 +88,28 @@ entities:
           - dps_val: StartFixedMowing
             value: true
   - entity: button
+    name: Start mowing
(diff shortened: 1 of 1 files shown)


make-all:
Start, Pause and Dock commands are provided by the mower entity already, so not required as separate buttons.

Peck07:
> Start, Pause and Dock commands are provided by the mower entity already, so not required as separate buttons.

but not as simple buttons, which is... kind of annoying, but I can live with it.

make-all:
It is easy to create buttons that perform an action in the UI. The intent of the integration is to make the functionality of devices available in HA, not to cover every potential UI possibility that users may want.

PR #3177, opened 2025-04-22
Implement Idra Skiv Airconditioner

Sapd:
Implements control for the "Idra Skiv" Indoor units, often used in conjunction with the IDRA Next Compressor (using Water instead of Air): https://cms.bconsole.com/documents/1699282517-productsUnita_Interne_inv_Fra_TED_2.pdf/teknopoint

The indoor unit is also technically the same as "SKYWORTH VELA", as the Idra Skiv is a re-brand of that one.

The error codes are from the manual

### DIFF
--- a/custom_components/tuya_local/devices/idra_skiv_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/idra_skiv_airconditioner.yaml
@@ -335,7 +335,9 @@ entities:
           - dps_val: 1048576
             value: Outside and inside unit not compatible (PA)
           - dps_val: 2097152
-            value: Conflict between indoor units - Not all on either heat or cool (PC)
+            value: >
+              Conflict between indoor units -
+              Not all on either heat or cool (PC)
           - dps_val: 4194304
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/idra_skiv_airconditioner.yaml
```diff
@@ -0,0 +1,347 @@
...
+            conditions:
+              - dps_val: 0
+                value: false
+          - value: true
+      - id: 20
+        type: bitfield
+        name: fault_code
+      - id: 20
+        type: bitfield
+        name: Fault Description
```
make-all, line 291: Attribute names should not contain capital letters or spaces. Normally "description" is used for this. Icon normally is not overridden when a class or translation_key defines an icon already, as this can interfere with user theming.

Sapd:
Thank you! Also great recommendations regarding the positions. Tuya on default just displays 1,2 etc.  in its API and App which is quite annoying

PR #3183, opened 2025-04-24
Add support for Smart Camera SC116-WZ3A

wuhgit:


make-all:
You can update a PR by simply pushing new commits to your branch, it is not necessary to close the PR and open a new one (which becomes a problem once a review has started, as any comments will get lost).

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
```diff
@@ -338,6 +338,7 @@ entities:
       - id: 160
         type: integer
         name: value
+        optional: true
```
make-all, line 341: The one the message was complaining about was the `optional: false` (false is the default, so adding it explicitly is flagged as a possible mistake, you may have meant true otherwise remove it).

wuhgit:
> You can update a PR by simply pushing new commits to your branch, it is not necessary to close the PR and open a new one (which becomes a problem once a review has started, as any comments will get lost).

Sorry for my stupid operation.

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
wuhgit: I'm confused, I've been using this config file locally for a long time and it's working fine, so why does it not pass the test?

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
```diff
@@ -0,0 +1,511 @@
+name: Camera
+products:
+  - id: b60fclvud3xtcufm
+    manufacturer: Tuya
```
make-all, line 4: Don't list Tuya as a manufacturer. Tuya does not generally manufacture end devices, only the WiFi/Zigbee/Bluetooth/Matter module in it, and they may provide reference designs for manufacturers to copy, but the end manufacturer is another company.
If you don't know the manufacturer, leave it out and let it default to showing Tuya.

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
```diff
@@ -0,0 +1,511 @@
...
+      - id: 150
+        type: boolean
+        name: record
+        optional: true
+      - id: 115
+        type: base64
+        optional: true
+        persist: false
+        sensitive: true
+        name: snapshot
```
make-all, line 34: Does this work? There seem to be two types of camera: one that sends base64 encoded jpeg images in the dp, and one that sends JSON data with a relative link to a snapshot file to be downloaded from the cloud or maybe an HTTP server on the camera itself.  If you are not getting still images when there is a motion event. then the format may not be suitable for use as a snapshot, in which case it will be better to use a separate event entity to pass through the JSON so it can be parsed by an external script that can maybe do the download if all the other details (server address, authentication etc) can be found.

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
```diff
@@ -0,0 +1,511 @@
...
+  - entity: switch
+    name: Indicator
```
make-all, line 36: This should be a light entity, then there is a translation_key for indicator to use instead of name and icon.

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
```diff
@@ -0,0 +1,511 @@
...
+  - entity: switch
+    name: Flip image
+    category: config
+    icon: "mdi:flip-vertical"
+    dps:
+      - id: 103
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Time OSD
```
make-all, line 52: Normally "Watermark" is used, along with `icon: "mdi:watermark"`. It will make later translation expansion easier if a common name is used.

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
```diff
@@ -0,0 +1,511 @@
...
+    category: config
+    icon: "mdi:clock-time-three-outline"
+    dps:
+      - id: 104
+        type: boolean
+        name: switch
+  - entity: switch
```
make-all, line 59: This is already covered by the camera switch, no need to duplicate it.

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
```diff
@@ -0,0 +1,511 @@
...
+    dps:
+      - id: 159
+        type: boolean
+        name: switch
+        optional: true
+        mapping:
+          - dps_val: null
+            value: false
+            hidden: true
+      - id: 185
```
make-all, line 329: This should be an event entity. See kerui_200w_camera

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
wuhgit: I will move this to separate event.

This device get the json data through dp_id 212, and then parse it to get the snapshot stored in the cloud.

custom_components/tuya_local/devices/smart_camera_sc116-wz3a.yaml
wuhgit: I have a question, what should I do if I want to turn off the camera temporarily? It seems more convenient to have it as a switch alone, as I didn't find the `camera switch` you mentioned.

make-all: I don't have a camera, so I don't know how the UI looks, but camera has turn_on and turn_off actions, so I am sure you can make the on/off switch appear in the UI somehow - maybe the entities card or tile card will let you do this.

Issue #3163, opened 2025-04-18
Neo-A Duoqu Cat Litter box


Airey001:
### Log message

```text
[33m2025-04-18 14:58:39.697 WARNING (MainThread) [custom_components.tuya_local.config_flow] Adding NEO-A duoqu device with product id ah3vg6gvftgdl0sb[0m
[33m2025-04-18 14:58:40.079 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device specification:
[{"id": 1, "name": "switch", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 2, "name": "work_mode", "type": "Enum", "format": "{\"range\":[\"auto_clean\",\"manual_clean\"]}", "enumMap": {}}, {"id": 3, "name": "start", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 4, "name": "auto_clean", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 5, "name": "delay_clean_time", "type": "Integer", "format": "{\"unit\":\"\",\"min\":1,\"max\":60,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 6, "name": "cat_weight", "type": "Integer", "format": "{\"unit\":\"\",\"min\":0,\"max\":500,\"scale\":1,\"step\":1}", "enumMap": {}}, {"id": 7, "name": "excretion_times_day", "type": "Integer", "format": "{\"unit\":\"times\",\"min\":0,\"max\":60,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 8, "name": "excretion_time_day", "type": "Integer", "format": "{\"unit\":\"\",\"min\":0,\"max\":600,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 9, "name": "manual_clean", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 23, "name": "factory_reset", "type": "Boolean", "format": "{}", "enumMap": {}}][0m
[...]

PR #3185, opened 2025-04-24
Neo-A Duoqu Cat Litter Box Support

Airey001:
This litter box is very simliar to the Vevor 76L configurations.   

I have made some changes to adjust

1.  Removed Factory Reset - This option doesn't seem to work in the Tuya app there is a confirmation after pressing. 
2.  For the backlight feature the api has green spelled as greed adjusted for this
3.  Renamed 'Cleaning' to 'Work Mode' to match Tuya App
4.  Renamed 'Start' to match Tuya App ('Emptry')
5.  Removed Child lock as this is not an option with this model
6.  Set label name to the Main Switch as this is actual the 'Reset' Litter option in the Tuya App
7.  Got rid of duplicate congurations for dps 7 and relabeled to 'Daily Usage' to match Tuya App
8.  Added label for Backlight
9.  Removed Notification and Problem DPS as mine don't seem to update in the API.  The api always says there is a pressure issue but the app doesn't.  
10. Backlight corected to this devices DPS
11. added 'Flow Set' option - Not sure on this it does nothing to my device but the changes show up in the app

### DIFF
--- a/custom_components/tuya_local/devices/duoqu_selfcleaning_litterbox.yaml
+++ b/custom_components/tuya_local/devices/duoqu_selfcleaning_litterbox.yaml
@@ -11,7 +11,7 @@ entities:
         type: boolean
         name: switch
   - entity: select
-    name: Working Mode
+    name: Mode
     icon: "mdi:broom"
     category: config
     dps:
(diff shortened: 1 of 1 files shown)


Airey001:
#3185

Airey001:
#3163

PR #3213, opened 2025-05-03
ET-72W update

craibo:
- Configure default values for temperature unit and temperature sensor
- Move dps 109 to configs with add correct configurations for Location
- Move dps 116 to configs with correct configurations for Sensor Type
- Move temperature calibration dps 112 and 113 values under the temperature correlating temperature sensors

### DIFF
--- a/custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
+++ b/custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
@@ -60,6 +60,7 @@ entities:
         type: boolean
         mapping:
           - dps_val: false
+            default: true
             value: C
           - dps_val: true
             value: F
@@ -78,15 +79,6 @@ entities:
       - id: 104
         type: integer
         name: unknown_104
-      - id: 109
-        type: boolean
-        name: unknown_109
-      - id: 112
-        type: integer
-        name: floor_temp_calibration
-      - id: 113
-        type: integer
-        name: room_temp_calibration
   - entity: lock
     translation_key: child_lock
     category: config
@@ -113,6 +105,9 @@ entities:
             value: C
           - dps_val: true
             value: F
+      - id: 113
+        type: integer
+        name: room_temp_calibration
   - entity: sensor
     name: Floor temperature
     class: temperature
@@ -132,16 +127,17 @@ entities:
             value: C
           - dps_val: true
             value: F
+      - id: 112
+        type: integer
+        name: floor_temp_calibration
   - entity: sensor
     name: Runtime
+    icon: "mdi:update"
     dps:
       - id: 117
         type: integer
         name: sensor
         unit: min
-      - id: 116
-        type: integer
-        name: load_w
   - entity: select
     category: config
     translation_key: temperature_unit
@@ -151,6 +147,7 @@ entities:
         type: boolean
         mapping:
           - dps_val: false
+            default: true
             value: celsius
           - dps_val: true
             value: fahrenheit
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
```diff
@@ -60,6 +60,7 @@ entities:
         type: boolean
         mapping:
           - dps_val: false
+            default: true
```
make-all, line 63: default: true is only used by the siren entity to select a siren tone to use when none is provided in the action parameters and there is no current value being reported by the device.

To define a read-only default value when there is no current value being reported in general, you should just remove the `dps_val: false`. If this is required, it is probably required for the other uses (except the select entity where it needs to be read-write).

custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
```diff
@@ -113,6 +105,9 @@ entities:
             value: C
           - dps_val: true
             value: F
+      - id: 113
+        type: integer
+        name: room_temp_calibration
```
make-all, line 110: Since this is attached now to the Room temperature sensor, the attribute can just be "calibration". I am not sure if it might be worth splitting this out to a number entity though, as this is usually for the user to adjust based on compensation for the installation location, unlike the calibration options for smartswitches which are more for factory calibration and take specialist equipment to tune.

custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
```diff
@@ -132,16 +127,17 @@ entities:
             value: C
           - dps_val: true
             value: F
+      - id: 112
+        type: integer
+        name: floor_temp_calibration
   - entity: sensor
     name: Runtime
+    icon: "mdi:update"
```
make-all, line 135: Adding `class: duration` is probably also a good idea here.

custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
```diff
@@ -219,3 +217,34 @@ entities:
...
             value: "5+1+1 (Adaptive)"
+  - entity: select
+    category: config
+    name: Sensor Type
+    icon: "mdi:heating-coil"
+    dps:
+      - id: 116
+        type: integer
+        name: option
+        unit: W
```
make-all, line 228: Unit is not used by the select entity. It would be better to add it as a suffix on the values. Rather than "Sensor type" this seems to be "Heating power".

custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
craibo: Thanks, I've removed the `default` configs. I was attempting to resolve and issue where one of my devices was having wifi connectivity issues and as a result it was showing error messages and not defining the temperature measurement unit C/F at all.

Correct fix is to sort out my wifi here.

custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
craibo: Yes, having calibration dps as config sensors makes sense. I was planning to do that in a seperate PR.

custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
craibo: Done

custom_components/tuya_local/devices/jiahong_et72w_thermostat.yaml
craibo: Renamed the sensor to align with the app naming and aligned the icon

craibo:
@make-all Thank you

PR #3236, opened 2025-05-11
Added support for HomeMate and Conbre smart fan regulators

guruprasadah:
Reviewed the logs and added basic support (on/off and speed control) for these devices


make-all:
These configs are identical, you do not need two of them

guruprasadah:
I've now removed the identical config and renamed it as Generic Fan Regulator. Is it fine now?

PR #3245, opened 2025-05-15
New device ElectriQ SupremeCool SC16HPW Air Conditioner (XKA-G02)

bputtick:


custom_components/tuya_local/devices/electriq_sc16hpw_aircon.yaml
```diff
@@ -0,0 +1,216 @@
+name: ElectriQ SupremeCool SC16HPW Air Conditioner (XKA-G02)
```
make-all, line 1: Branding belongs under a products section. This name should be generic in case of other products matching the same config.

custom_components/tuya_local/devices/electriq_sc16hpw_aircon.yaml
```diff
@@ -0,0 +1,216 @@
...
+        mapping:
+          - dps_val: true
+            value_mirror: current_hvac_mode
+          - dps_val: false
+            value: 'off'
+      - id: 2
+        name: current_temperature
+        type: integer
+        mapping:
+          - scale: 1
```
make-all, line 18: A scale of 1 is redundant, it is only needed inside a conditional block to override a different scale in special cases

custom_components/tuya_local/devices/electriq_sc16hpw_aircon.yaml
```diff
@@ -0,0 +1,216 @@
...
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Quiet Mode
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+  - entity: binary_sensor
+    name: Water Full
```
make-all, line 211: Use `translation_key: tank_full` instead of this name and class

PR #3249, opened 2025-05-17
Added new device: PIR sensor with alarm functionality

EmilianoMorello:
PIR sensor with alarm function. The integration allows arming and disarming the device, selecting the operating mode, showing presence detection, setting the alarm duration, and checking the power and battery status.


custom_components/tuya_local/devices/pir_sensor_alarm.yaml
```diff
@@ -0,0 +1,69 @@
...
+            value: false
+  - entity: switch
+    name: Armed
+    icon: "mdi:shield-home"
+    dps:
+      - id: 104
+        type: boolean
+        name: switch
+  - entity: binary_sensor
+    translation_key: charge_energy
```
make-all, line 25: There is no translation key by this name, but maybe `class: battery_charging` is what you really want here.

custom_components/tuya_local/devices/pir_sensor_alarm.yaml
```diff
@@ -0,0 +1,69 @@
...
+    translation_key: motion
+    class: motion
+    dps:
+      - id: 1
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: pir
+            value: true
+          - dps_val: "none"
```
make-all, line 15: It is probably more failsafe to just make false the default (remove the dps_val filter from this list item)

custom_components/tuya_local/devices/pir_sensor_alarm.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - entity: binary_sensor
+    translation_key: charge_energy
+    class: plug
+    category: diagnostic
+    dps:
+      - id: 103
+        type: boolean
+        name: sensor
+  - entity: sensor
+    translation_key: battery_level
```
make-all, line 33: translation_key does not exist, and is not needed

custom_components/tuya_local/devices/pir_sensor_alarm.yaml
```diff
@@ -0,0 +1,69 @@
...
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 4
+        type: integer
+        name: sensor
+        unit: "%"
+  - entity: number
+    name: Alarm time
+    category: config
```
make-all, line 43: class: duration

Issue #3211, opened 2025-05-02
Request support for Treatlife DS03


com6056:
### Log message

```text
Adding Fan Dimmer Switch device with product id ea88zei7auxu48s6
Device specification: [{"id": 1, "name": "switch_fan", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 6, "name": "bright_value", "type": "Integer", "format": "{\"min\":10,\"max\":1000,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 102, "name": "fan_countdown", "type": "Integer", "format": "{\"unit\":\"\u5206\u949f\",\"min\":0,\"max\":1440,\"scale\":0,\"step\":1}", "enumMap": {}}]
Device matches gosund_SW2_dimmer_switch with quality of 43%. DPS: {"updated_at": 1746213438.72302, "1": true, "3": "1", "6": 100, "101": false, "105": 100, "106": "mode1", "108": 1}
```

### Product ID

ea88zei7auxu48s6

### Product Name

Treatlife DS03

### DPS information

```text
{
  "result": {
    "model": "{\"modelId\":\"000003ibqy\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"默认服务\",\"properties\":[
{\"abilityId\":1,\"accessMode\":\"rw\",\"code\":\"switch_fan\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-dp_power\",\"attribute\":\"1665\"},\"name\":\"风扇开关\",\"typeSpec\":{\"type\":\"bool\"}},
{\"abilityId\":3,\"accessMode\":\"rw\",\"code\":\"fan_speed_enum\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-FanSpeed\",\"attribute\":\"1152\"},\"name\":\"风速\",\"typeSpec\":{\"type\":\"enum\",\"range\":[\"1\",\"2\",\"3\",\"4\"]}},
{\"abilityId\":6,\"accessMode\":\"rw\",\"code\":\"bright_value\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-dp_half\",\"attribute\":\"1152\"},\"name\":\"风扇灯亮度值\",\"typeSpec\":{\"type\":\"value\",\"max\":1000,\"min\":10,\"scale\":0,\"step\":1}},
[...]

PR #3273, opened 2025-05-23
Adds more product_ids observed for Treatlife DS03 devices

falquaddoomi:
I own a couple three Treatlife DS03 switches; one of them associates correctly with the `treatlife_ds03_fan_light.yaml` profile even though its `product_id` is `gvfmw8c8n92umpax`, I assume because its productKey is somehow matching the profile.

My other two switches that have a `product_id` of `ea88zei7auxu48s6` don't associate with the profile, so I can't make them be recognized as a DS03 when I add them. This ID has also been observed in #3211.

This PR adds the two product IDs I've observed, `gvfmw8c8n92umpax` and `ea88zei7auxu48s6`, to the profile for the DS03.

### DIFF
--- a/custom_components/tuya_local/devices/treatlife_ds03_fan_light.yaml
+++ b/custom_components/tuya_local/devices/treatlife_ds03_fan_light.yaml
@@ -6,9 +6,6 @@ products:
   - id: gvfmw8c8n92umpax
     manufacturer: Treatlife
     model: DS03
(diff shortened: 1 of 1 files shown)


make-all:
Based on the logs in #3211, the `ea88zei7auxu48s6` device is incompatible with the config, so simply adding the product id is not the correct solution.

PR #3294, opened 2025-05-30
Added sensors and some fixes to feyree_ev_portable_charger.yaml

georgekucher:
Added: "Alarm event", "Current", "Energy", "Charging time" sensors and "Charging duration" input number.

Fixed: Voltage scale and Status typo ("no_connet" is what comes from the device)
Screenshot_1
image

### DIFF
--- a/custom_components/tuya_local/devices/feyree_ev_portable_charger.yaml
+++ b/custom_components/tuya_local/devices/feyree_ev_portable_charger.yaml
@@ -75,7 +75,7 @@ entities:
           - scale: 2
   - entity: binary_sensor
     name: Alarm event
-    device_class: problem
+    class: problem
     icon: 'mdi:alert-outline'
     dps:
       - id: 10


custom_components/tuya_local/devices/feyree_ev_portable_charger.yaml
```diff
@@ -73,6 +73,49 @@ entities:
           max: 999999
         mapping:
           - scale: 2
+  - entity: sensor
```
make-all, line 76: This would maybe be better as an event entity, or as a problem binary_sensor, with the description in extra attributes

georgekucher: Hi,
Thanks for the comment. I'd prefer option 2, but I'm not sure how to map an extra attributes to a binary sensor.
Could you please give an example of such implementation?

make-all:
```
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 10
        type: bitfield
        name: sensor
        mapping:
          - dps_val: 0
            value: false
          - value: true
      - id: 10
        type: bitfield
        name: fault_code
      - id: 10
        type: bitfield
        name: description
        mapping:
          # as existing mapping
```

make-all: This will make it easier to use in automations, also you will get highlighting in the UI when there is an actual problem. On a Tile card, you should be able to display the description attribute as secondary info easily, for other cards you maybe need to edit the yaml to get attributes displayed.

PR #3299, opened 2025-05-31
Dowell EV Charger - update 01

VulpesFurvus:


custom_components/tuya_local/devices/dowell_dchs-k7_32_7kw_ev_charger.yaml
```diff
@@ -234,6 +234,7 @@ entities:
     translation_placeholders:
       x: B
     class: voltage
+    hidden: true
```
make-all, line 237: I wonder if the phase A sensors should be also hidden, since we have total's, and if there is only one phase, they will be identical.

custom_components/tuya_local/devices/dowell_dchs-k7_32_7kw_ev_charger.yaml
```diff
@@ -338,7 +338,7 @@ entities:
         mapping:
           - scale: 100
   - entity: number
-    name: energy_charge
+    name: Total daily charge
```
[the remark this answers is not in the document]
VulpesFurvus: Looking too the description in chinese, it is not a limit it is the daily total energy consumption. And the Tuya Smart EV Charger app is giving me the same hint. It must be a sensor yes, I'll test it.

custom_components/tuya_local/devices/dowell_dchs-k7_32_7kw_ev_charger.yaml
VulpesFurvus: I'll share the printscreens of the app for you to help me confirm. In my smartphone the app do not fully present the data because of the theme. Some of the fields are simply partially shown. Is there a  way I can run the application in my PC?

Issue #3262, opened 2025-05-21
Request support for Korex AX-WF306N kettle


thd6n75:
### Log message

```text
2025-05-20 21:12:30.588 DEBUG (MainThread) [custom_components.tuya_local.device] Korex smart kettle received
{
"1": false, 
"5": 81, 
"10": 0, 
"103": false, 
"104": "1", 
"105": "f", 
"106": "0", 
"107": "0",
 "full_poll": true
}
```

<details>
```text
2025-05-20 21:12:19.382 DEBUG (SyncWorker_3) [custom_components.tuya_local.helpers.device_config] Loaded device config kogan_glass_1_7l_kettle.yaml
2025-05-20 21:12:20.461 INFO (MainThread) [custom_components.tuya_local.config_flow] Pausing existing device to test new connection parameters
2025-05-20 21:12:25.463 DEBUG (MainThread) [custom_components.tuya_local.device] Refreshing device state for Test
2025-05-20 21:12:25.463 DEBUG (MainThread) [custom_components.tuya_local.device] Setting protocol version for Test to 3.3
2025-05-20 21:12:25.463 DEBUG (SyncWorker_7) [tinytuya.core.XenonDevice] status() entry (dev_type is default)
2025-05-20 21:12:25.463 DEBUG (SyncWorker_7) [tinytuya.core.XenonDevice] final payload_dict for 'eb941bd4593057d61ci6a1' ('v3.3'/'default'): {1: {'command': {'gwId': '', 'devId': '', 'uid': '', 't': ''}}, 7: {'command': {'devId': '', 'uid': '', 't': ''}}, 8: {'command': {'gwId': '', 'devId': ''}}, 9: {'command': {'gwId': '', 'devId': ''}}, 10: {'command': {'gwId': '', 'devId': '', 'uid': '', 't': ''}}, 13: {'command': {'devId': '', 'uid': '', 't': ''}}, 16: {'command': {'devId': '', 'uid': '', 't': ''}}, 18: {'command': {'dpId': [18, 19, 20]}}, 64: {'command': {'reqType': '', 'data': {}}}}
2025-05-20 21:12:25.463 DEBUG (SyncWorker_7) [tinytuya.core.XenonDevice] building command 10 payload=b'{"gwId":"eb941bd4593057d61ci6a1","devId":"eb941bd4593057d61ci6a1","uid":"eb941bd4593057d61ci6a1","t":"1747789945"}'
[...]

PR #3301, opened 2025-05-31
New device: Korex AX-WF306N kettle

thd6n75:
This adds support for Korex AX-WF306N kettle.

https://github.com/make-all/tuya-local/issues/3262

### DIFF
--- a/custom_components/tuya_local/devices/korex_axwf306n_smart_kettle.yaml
+++ b/custom_components/tuya_local/devices/korex_axwf306n_smart_kettle.yaml
@@ -101,7 +101,7 @@ entities:
     dps:
       - id: 105
         type: string
-        name: sensor
+        name: option
         mapping:
           - dps_val: c
             value: celcius


thd6n75:
I think I fixed the linting errors, but I'm not sure how to rerun the approval workflows

thd6n75:
Hi, I installed this .yaml file that you fixed up- https://raw.githubusercontent.com/make-all/tuya-local/75b5b50539c5212be4fe0825c44c4a714eac954b/custom_components/tuya_local/devices/korex_axwf306n_smart_kettle.yaml  But it seems to have broken most of the functionality compared to the one I PR'd.

Here's the device when deleted and freshly added with the .yaml from my PR:

kettle-pr

Here's the device when deleted and freshly added from https://raw.githubusercontent.com/make-all/tuya-local/75b5b50539c5212be4fe0825c44c4a714eac954b/custom_components/tuya_local/devices/korex_axwf306n_smart_kettle.yaml :

kettle-main


I can't turn the kettle on! :)

make-all:
There should be a water_heater entity, which allows turning on and seeing the current temperature. Since it is missing, there may be some problem indicated in the logs?

Also, you do not seem to be using the latest version of the file. I made some changes after reviewing the info in the issue that was filed for this same device.

thd6n75:
Fresh attempt with yaml from https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/korex_axwf306n_smart_kettle.yaml



```
2025-06-04 08:09:54.305 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches korex_axwf306n_smart_kettle with quality of 80%. DPS: {"updated_at": 1749038991.9191644, "1": false, "5": 75, "10": 0, "101": "4", "102": "0", "103": fals
e, "104": "1", "105": "f", "106": "0", "107": "0"}
2025-06-04 08:09:54.305 WARNING (MainThread) [custom_components.tuya_local.config_flow] Include the previous log messages with any new device request to https://github.com/make-all/tuya-local/issues/
2025-06-04 08:09:55.542 DEBUG (SyncWorker_1) [custom_components.tuya_local.helpers.device_config] Loaded device config korex_axwf306n_smart_kettle.yaml
2025-06-04 08:09:56.438 DEBUG (MainThread) [custom_components.tuya_local] Setting up entry for device: eb941bd4593057d61ci6a1
2025-06-04 08:09:56.439 INFO (SyncWorker_19) [custom_components.tuya_local.device] Creating device: eb941bd4593057d61ci6a1
2025-06-04 08:09:56.446 DEBUG (SyncWorker_16) [custom_components.tuya_local.helpers.device_config] Loaded device config korex_axwf306n_smart_kettle.yaml
2025-06-04 08:09:56.450 DEBUG (SyncWorker_0) [custom_components.tuya_local.helpers.device_config] Loaded device config korex_axwf306n_smart_kettle.yaml
2025-06-04 08:09:56.450 DEBUG (MainThread) [custom_components.tuya_local.helpers.config] Adding sensor for sensor_status
2025-06-04 08:09:56.451 DEBUG (MainThread) [custom_components.tuya_local.device] Starting monitor loop for Kettle
2025-06-04 08:09:56.451 DEBUG (MainThread) [custom_components.tuya_local.device] Setting protocol version for Kettle to 3.3
2025-06-04 08:09:56.455 DEBUG (SyncWorker_3) [custom_components.tuya_local.helpers.device_config] Loaded device config korex_axwf306n_smart_kettle.yaml
2025-06-04 08:09:56.463 DEBUG (SyncWorker_12) [custom_components.tuya_local.helpers.device_config] Loaded device config korex_axwf306n_smart_kettle.yaml
2025-06-04 08:09:56.467 DEBUG (SyncWorker_13) [custom_components.tuya_local.helpers.device_config] Loaded device config korex_axwf306n_smart_kettle.yaml
2025-06-04 08:09:56.468 DEBUG (MainThread) [custom_components.tuya_local.helpers.config] Adding binary_sensor for binary_sensor_temperature_unit_mismatch
2025-06-04 08:09:56.468 DEBUG (MainThread) [custom_components.tuya_local.helpers.config] Adding binary_sensor for binary_sensor_problem
2025-06-04 08:09:56.471 DEBUG (SyncWorker_15) [tinytuya.core.XenonDevice] status() entry (dev_type is default)
2025-06-04 08:09:56.472 DEBUG (SyncWorker_15) [tinytuya.core.XenonDevice] final payload_dict for 'eb941bd4593057d61ci6a1' ('v3.3'/'default'): {1: {'command': {'gwId': '', 'devId': '', 'uid': '', 't': ''}}, 7: {'command': {'devId': '', 'uid': '', 't': ''}}, 
8: {'command': {'gwId': '', 'devId': ''}}, 9: {'command': {'gwId': '', 'devId': ''}}, 10: {'command': {'gwId': '', 'devId': '', 'uid': '', 't': ''}}, 13: {'command': {'devId': '', 'uid': '', 't': ''}}, 16: {'command': {'devId': '', 'uid': '', 't': ''}}, 18:
 {'command': {'dpId': [18, 19, 20]}}, 64: {'command': {'reqType': '', 'data': {}}}}
2025-06-04 08:09:56.472 DEBUG (SyncWorker_15) [tinytuya.core.XenonDevice] building command 10 payload=b'{"gwId":"eb941bd4593057d61ci6a1","devId":"eb941bd4593057d61ci6a1","uid":"eb941bd4593057d61ci6a1","t":"1749038996"}'
2025-06-04 08:09:56.483 DEBUG (MainThread) [custom_components.tuya_local.helpers.config] Adding select for select_heat_to_temperature
2025-06-04 08:09:56.483 DEBUG (MainThread) [custom_components.tuya_local.helpers.config] Adding select for select_keep_warm_temperature
2025-06-04 08:09:56.485 DEBUG (MainThread) [custom_components.tuya_local.helpers.config] Adding water_heater for water_heater_kettle
2025-06-04 08:09:56.485 ERROR (MainThread) [homeassistant.components.water_heater] Error adding entity None for domain water_heater with platform tuya_local
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 657, in _async_add_entities
    await self._async_add_entity(
        entity, False, entity_registry, config_subentry_id
    )
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 877, in _async_add_entity
    capabilities=entity.capability_attributes,
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/water_heater/__init__.py", line 211, in capability_attributes
    self.hass, self.min_temp, self.temperature_unit, self.precision
               ^^^^^^^^^^^^^
  File "/config/custom_components/tuya_local/water_heater.py", line 225, in min_temp
    return r[0]
           ~^^^
TypeError: 'NoneType' object is not subscriptable
2025-06-04 08:09:56.487 DEBUG (SyncWorker_15) [tinytuya.core.XenonDevice] sending payload
2025-06-04 08:09:56.487 DEBUG (SyncWorker_15) [tinytuya.core.XenonDevice] payload encrypted=b'000055aa000000010000000a00000088e5b5a3dcfe331e473cc1d91ff13a0832c235142c37883554567ab13d95e5ce2e58f0d6f0f6e5d9e33572ef0edd6c4c900b916f89a51b3e728761ad3b748490838551593132be2c8ad73a58c997db594dc235142c37883554567ab13d95e5ce2e479daad681c00a0bc40ccfe707c46ff1f3c7af8485900ec23821e3738f262cf93b12d1a30000aa55'
2025-06-04 08:09:56.498 DEBUG (SyncWorker_15) [tinytuya.core.XenonDevice] received data=b'000055aa000000010000000a0000007c00000000b0d721960ae9b0c2c6c0c89b2a2716397ea82a5ce50837837655dc48a130b92299ffd3779a62b0e28b26c39998608d1b011890393325eeb2f921b1ff7c31678eec8b46b14cc65391a836b5df4688f31657ed2915e4e1efcaacedd6ddd878b35679dc5966c00917525a131f52efeafb7a89da4c340000aa55'

```



kettle-main2

thd6n75:
https://github.com/make-all/tuya-local/commit/75342c56fe608617d6efe4bd8d351008e5264fe9

```
                range:
                  min: 113
                  max: 212
```

Will that go in increments of 1?  The kettle doesn't support 1-degree increments, unfortunately.  It only supports the temperatures enumerated in the .yaml (which is why I thought a dropdown/option/select would be appropriate?)

make-all:
Yes, you are right, the step needs to be explicitly specified even when there are mapped values. The automatic behaviour of mapped values is only that it will snap to nearest, so with a default step of 1 you will be unable to change the value with up/down arrows, only with a card that shows it as a slider or a text entry box.

thd6n75:
https://github.com/make-all/tuya-local/commit/7d5c2e4e34a0c37b625faa0cfd197027fad6b414

```
step: 9
```

The kettle supports temperature units in both C and F (based on a physical button on the kettle).  Is this going to work for both?  The unit is DPS 105 in the [original PR](https://github.com/make-all/tuya-local/pull/3301/commits/cba4b7e22832bc810f51376a0d78d52b589cc6a1#diff-c15ebf336a5aada7e7dd369ce1903c475bbbd2d19d36959225956de83bcbe7d8R48).  Maybe it's possible to have step==5 for C and step==9 for F, based on the value of that DPS?

make-all:
The water heater cannot work for both, the logic is too complex with the temperatures needing to be mapped and separate settings for heat and boil already. There is a sensor which will indicate a problem if the unit is set to C.

thd6n75:
When I run this script (the intention being: heat the water to 185 and keep it that way):

kettle-script

It appears to actually heat it to the next step up: 194.  And it sets the "heat to temperature" select option to 194, too:

kettle-heattotemp

Array index off by 1 somewhere? :)

thd6n75:
We're getting closer!  The previous issue (off by 1) issue seems fixed but the current_temperature is way off now.  For example when the kettle reads 140F on the physical display, the current_temp is 293F in HA.

Here are some pics after running that same script that sets to 185.  Note also that the target temperature in the gauge (first pic) reads 235 - that is of course not the target temp (185 is).

kettle-gauge

And note the correct selection of 185 in the dropdown:

kettle-device

Also the water goes to a boil instead of stopping at 185.

thd6n75:
I'll also note that, weirdly, there are 2 distinct methods of heating.

1. Heat to target temperature.  So in my case I want it to go from room temp to 185 and stop.  (This is I think what most folks would want and intuitively expect).  This is DPS 101.
2. Boil (go all the way to to 212) and then cool down to the target temp.  So it goes: room temp -> 212 -> target.  I have no idea why this feature exists.  This is DPS 102.  

You have [DPS 102](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/korex_axwf306n_smart_kettle.yaml#L67) labeled as "keep warm temperature" which is not exactly what it does.

make-all:
Boiling then keeping at/above the keep warm temperature seems exactly what it does.

make-all:
I think HA must be defaulting to C when the units are unrecognised. I changed the mapping the wrong way around when copying into the water_heater temperature unit so the current temperature will display correctly when the units are not matching the way the config is written (for F).

thd6n75:
> Boiling then keeping at/above the keep warm temperature seems exactly what it does.

Yes that's what method 2 above does.  Method 1 does not boil - it just goes up to the target point.

thd6n75:
Can we switch the water heater entity's keep_warm function to `heat up to target temp and stop, and then keep warm` (method 1, DPS 101) instead of `boil to 212F, then cool down to target temp, then keep warm` (method 2, DPS 102)? :)

make-all:
I think "heat" and "boil" may be better names if both options keep the water warm at the end.

thd6n75:
>  if both options keep the water warm at the end.

They only do so if the boolean keep_warm is true.  (This is a physical button on the kettle).  That's why the orig PR had 3 distinct things:

1. select/option dropdown for boil and cool to target temp.  DPS 101
2. select/option dropdown for heat to target temp.  DPS 102
3. input_boolean for keep_temp. DPS 103

thd6n75:
kettle-smartlife

If this helps to visualize, this is how the official app presents the controls.  The "timing" bit is a scheduler - i.e. "boil water at 630am on Tuesdays and toggle keep warm" - I think we can ignore this and instead rely on HA's native automation capabilities.

**"boil to tempset" is a poor English translation.  It doesn't boil unless 212F/100C is the target temperature**

thd6n75:
I made some changes, I'll submit a PR and let me know what you think.

thd6n75:
https://github.com/make-all/tuya-local/pull/3342

thd6n75:
It's still not working correctly.

Screenshot at 2025-06-07 11-26-30


I used the dropdown box to set the temp to 185.  Then I used the water_heater entity to turn it on (switch to "Electric").  Then I toggled keep warm.

It heated up to boiling (note the screenshot, current temp is hotter than the set point) and is presently cooling.  Looks like somewhere is still using DPS 102 when it should be using DPS 101.

make-all:
I see the source of the confusion. "0" was "cancel" in your original mapping, to make it fit on a numeric scale, I made it 113 (below the minimum by 9F), but in fact the cancel for dp 101 means to boil the water instead of heating. So while the current temperature of 190 appears to be above the target of 113, it is actually below the target of boiling point, meaning the kettle has either not reached boiling yet, or is cooling down to the keep warm target temperature of 185. 

I expect there will be some hysteresis around the keep warm temperature, so it could also be caused by that.

thd6n75:
Still the same behavior.  See 1st screenshot- set point is 194, and it goes to a boil, and then cools.



Screenshot at 2025-06-07 12-58-34

In this 2nd screenshot, notice that Status near the bottom has changed to "Cooling".  I don't want it to ever get to this state, because I don't ever want the water temperature to go above the set point.

Screenshot at 2025-06-07 12-59-51

And because it bears repeating, here's the SmartLife app again and how the kettle is modeled there:

kettle-smartlife


Fundamentally there are 2 different behaviors as I've described before.

1. Method 1.  Heat to the target temperature.  Go no further.  **Do NOT boil**, do NOT cool.  This is the top-most mode in the SmartLife app, and corresponds to DPS 101 `boiltotempset` in the data model API.  Disregard the word "boil" in "boiltotempset" - it's just a poor English translation.  I believe this behavior is what all users would intuitively expect from a water kettle.
2. Method 2.  Regardless of what the target temp is, go to **boil**.  And **then cool** to target temperature. This is the 2nd mode in the SmartLife app.  Temperature 203 is the highest target temperature available in this mode.  This is presently the behavior **I'm observing in Homeassistant**, and is the behavior demonstrated in screenshots 1 and 2 in this post.  From the data model API, this method is DPS 102, `boilandcooltotempset`.  In this case, "boil" in "boilandcooltotempset" _actually_ means boil :)

I think both methods should be modeled just out of principle.  However, if only 1 behavior is to be modeled, I think it should be method 1 and not method 2.

I hope this makes sense.  I do agree with one of the commit messages I saw you make: this kettle is confusing and weird. :)

thd6n75:
By the way, I haven't described how to use the kettle in the SmartLife app.  Unsurprisingly, it's klunky.

First you have to turn the kettle ON by pushing the power button under the gauge.  This puts the kettle into "boil to tempset" (Method 1) with a target temperature of 212.  If you just want boiling water, stop here.  Here's a screenshot of the UI after simply pushing power:

smartlife-poweredon

But if you want water that's only 194, then you click "boil to tempset" (Method 1) and you select your temperature:

smartlife-selecttemp

Not sure if this gives any insight into how the SmartLife folks had to design the UI in light of how the API works.  In my original PR I attempted to mimic this functionality and UI by making a power switch, a keepwarm switch, and 2 select/dropdowns (for method 1 and method 2).

make-all:
Your method 1 seems to be what the water heater temperature setting will give you.

thd6n75:
Whoops, you're right!  Calling `water_heater.set_temperature(185, Electric)` does the Right Thing (method 1)!  It must have changed at some point, because I changed my script to call the `select_option` against the Keep Warm Temperature thing instead since `water_heater.set_temp` previously wasn't using method 1.

I think everything is good now.

As a purely cosmetic thing, I think the dropdown box being called "Keep warm temp" is a misnomer.  It doesn't keep warm unless the keep warm toggle is on.  "Boil and cool" I think would be more accurate.

THANK YOU! <3

thd6n75:
The most recent commit I think broke the select_option box?

Screenshot at 2025-06-08 07-38-57

Screenshot at 2025-06-08 07-40-42

make-all:
The intention of the last change was to disable the "Keep warm" temperature when it is not valid to use, to try to avoid the confusion you have had until now. It is only enabled when the water heater temperature is set to boil (221 F) and the Keep warm switch is on.

thd6n75:
Boiling in F is 212, not 221 :)

make-all:
Unfortunately the value 212 is already a value that can be selected from the heating temperature list without setting it to boil (Cancel).

thd6n75:
Ack.  Thanks again for everything.  As far as I can tell, this is the first kettle obtainable in the USA that is compatible with HA. 

Check your discord when you get a chance.

PR #3307, opened 2025-06-01
New Device: PetsPride Petfeeder OO15

Gotanius:
Petspride.nl is a NL-based webshop that offers multiple tuya related pet appliances. I've created a yaml so that their OO15 petfeeder is compatible with tuya local.


custom_components/tuya_local/devices/petspride_petfeeder_OO15.yaml
```diff
@@ -0,0 +1,257 @@
...
+          - dps_val: "2"
+            value: "On"
+  - entity: number
+    name: Sound volume
+    icon: "mdi:volume-high"
+    category: config
+    dps:
+      - id: 160
+        type: integer
+        name: volume
```
make-all: name must be `value` for a number entity.

Gotanius: Changed :)

PR #3309, opened 2025-06-02
Two new goldair heaters

horse-pants:
goldair_gpph900_heater
goldair_gpcv340_heater

Some more goldair devices, I've been using these configs for a few months now, they're working well

### DIFF
--- a/custom_components/tuya_local/devices/goldair_gpcv340_heater.yaml
+++ b/custom_components/tuya_local/devices/goldair_gpcv340_heater.yaml
@@ -1,0 +1,62 @@
+name: Goldair Panel heater (GPCV340)
+products:
+  - id: ynjanlglr4qa6dxf
+    manufacturer: Goldair
+    model: NPH-2000G-WIFI
+entities:
+  - entity: climate
+    translation_key: heater
+    dps:
+      - id: 1
+        name: hvac_mode
+        type: boolean
+        mapping:
+          - dps_val: true
+            value: "heat"
+          - dps_val: false
+            value: "off"
+      - id: 3
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 5
+          max: 40
+      - id: 7
+        name: preset_mode
+        type: string
+        mapping:
+          - dps_val: Low
+            value: eco
+          - dps_val: High
+            value: boost
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 2
+        type: boolean
+        name: lock
+  - entity: number
+    translation_key: timer
+    class: duration
+    category: config
+    dps:
+      - id: 5
+        type: integer
+        name: value
+        unit: min
+        range:
+          min: 60
+          max: 1440
+        mapping:
+          - step: 60
+            scale: 1
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 4
+        type: integer
+        name: sensor
+        class: measurement
+        unit: C
--- a/custom_components/tuya_local/devices/goldair_gpph900_heater.yaml
+++ b/custom_components/tuya_local/devices/goldair_gpph900_heater.yaml
@@ -1,0 +1,172 @@
+name: Goldair Panel heater (GPPH770/GPPH780; GPPH900/GPPH910)
+products:
+  - id: xr6jsgylldbpkaz9
+    manufacturer: Goldair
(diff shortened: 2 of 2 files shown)


custom_components/tuya_local/devices/goldair_gpcv340_heater.yaml
```diff
@@ -0,0 +1,62 @@
+name: Goldair Panel heater (GPCV340)
+products:
+  - id: ynjanlglr4qa6dxf
+    manufacturer: Goldair
+    model: NPH-2000G-WIFI
```
make-all: Is this correct? It does not match the filename, but it does match the Kogan heater that this config is identical to.

custom_components/tuya_local/devices/goldair_gpph900_heater.yaml
```diff
@@ -0,0 +1,172 @@
...
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 5
+        mapping:
+          - step: 1
+            scale: 1
+  - entity: sensor
+    class: temperature
```
make-all: This should be integrated into the climate entity as current_temperature.

custom_components/tuya_local/devices/goldair_gpph900_heater.yaml
```diff
@@ -0,0 +1,172 @@
...
+    category: config
+    dps:
+      - id: 110
+        type: integer
+        name: value
+        unit: min
+        range:
+          min: 1
+          max: 1440
+        mapping:
```
make-all: Mapping can be removed, as no scaling and a step of 1 is default

custom_components/tuya_local/devices/goldair_gpph900_heater.yaml
```diff
@@ -0,0 +1,172 @@
...
+  - entity: switch
+    name: Heat mode
+    icon: "mdi:heat-wave"
+    category: config
+    dps:
+      - id: 107
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Open window
```
make-all: If this is a switch for the feature, it should be "Open window detection", but if it is a sensor to indicate that an open window is detected, then it should be a binary_sensor.

custom_components/tuya_local/devices/goldair_gpph900_heater.yaml
```diff
@@ -0,0 +1,172 @@
+name: Goldair Panel heater (GPPH770/GPPH780; GPPH900/GPPH910)
```
make-all: This name should be generic (just "Panel heater"), branding belongs under products.

custom_components/tuya_local/devices/goldair_gpph900_heater.yaml
```diff
@@ -0,0 +1,172 @@
...
+  - entity: switch
+    name: Auto off LCD
+    icon: "mdi:monitor"
+    category: config
+    dps:
+      - id: 104
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Sound
```
make-all: What sound?

If it is beeps for user action, you can use `translation_key: keytone` instead of name and icon.

custom_components/tuya_local/devices/goldair_gpph900_heater.yaml
```diff
@@ -0,0 +1,172 @@
...
+          min: 3
+          max: 7
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 7
+        type: boolean
+        name: lock
+  - entity: switch
```
make-all: Should maybe be a light entity, and if translation key is specified, name and icon are not needed. I am not sure whether nightlight, indicator or display would be the more appropriate. If the logic is reversed, you can reverse it back with a mapping:
```
mapping:
  - dps_val: true
    value: false
  - dps_val: false
    value: true
```

make-all:
The goldair_gpcv340_heater will be rejected, as it is identical down to the product id to the existing kogan heater with the same model number. So if you are able to rebase the change to exclude that file, it will help speed things up.

horse-pants:
I made some changes (including product ID) to the 340 heater and tested them today, if you'd still prefer I removed the file I can though. I just figured it might be better to have it match the brand on the heater rather than say kogan when someone's setting it up

PR #3313, opened 2025-06-03
Add support for Quoya QL500 Curtain

alex-red:
Adding support for these smart curtains from Amazon. (https://www.quoyaliving.com/products/smart-electric-curtains-curtain-track-3m).

The config seems to be exactly the same as the Abalon bcm700d but I wasn't sure if you wanted a new device config or to just add it under the Abalon config file. Been running this with no issues for many months.

Also more information here: https://github.com/home-assistant/core/issues/128911


make-all:
If it is exactly the same as an existing config, then adding the product details to the existing config is better. There are already over 1000 configs in the repository, and this is starting to have a performance impact when adding new devices so I don't want to add more just for branding differences.

PR #3327, opened 2025-06-04
Add support for "wifi_dual_channel_smart_meter" dual channel smart meter (product_id: v5jlnn5hwyffkhp3) – full entity mapping

TommasoPino:
## Description
This PR adds support for a dual-channel smart meter (clamp meter, 2 channels), commonly found on Tuya, with product_id: v5jlnn5hwyffkhp3.

### Device features

The device exposes several measurement channels (A/B), providing:
- Instantaneous total power (W)
- Voltage (V)
- Frequency (Hz)
- Current per channel (A/B)
- Power per channel (A/B)
- Power factor per channel (A/B)
- Imported/exported energy (total and per channel)

### Configuration highlights

- All core entities are mapped using the new `primary_entity` and `secondary_entities` syntax.
- All `dps` are configured with `name: sensor` in accordance with [tuya-local documentation](https://github.com/make-all/tuya-local/wiki/Device-configuration-files) to avoid "missing a sensor dps" errors.
- Optional and calibration DPs are set with `optional: true` for robust device matching and compatibility.
- The `reporting rate` entity (`dps 126`) is commented out or marked as optional, as it is not consistently reported by all firmware versions, preventing false negatives during device matching.

### Validation

- Tested with Home Assistant 2024.x and tuya-local.
- Device is correctly detected, and all entities are working as expected.

### Notes

- Configuration is based on Tuya Cloud API property mapping and real device testing.
- Further calibration or diagnostic entities can be added as secondary entities, provided they are marked `optional: true` if not always present.

---

Thank you for maintaining this excellent integration!

### DIFF
--- a/custom_components/tuya_local/devices/wifi_dual_channel_smart_meter.yaml
+++ b/custom_components/tuya_local/devices/wifi_dual_channel_smart_meter.yaml
@@ -1,0 +1,165 @@
+name: Smart Meter
+products:
+  - id: v5jlnn5hwyffkhp3
+    name: Smart Meter
+    manufacturer: Tuya
+    model: 更换互感器
+
+primary_entity:
+  entity: sensor
+  name: Total Power
+  class: power
+  dps:
+    - id: 103
+      name: sensor
+      type: integer
+      unit: W
+      class: measurement
+      mapping:
+        - scale: 10
+
+secondary_entities:
+  - entity: sensor
+    name: Voltage
+    class: voltage
+    dps:
+      - id: 101
+        name: sensor
+        type: integer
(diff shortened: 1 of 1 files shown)


make-all:
- This device is already supported in matsee_2wayv2_energymeter.yaml.

TommasoPino:
In this case, i am sorry for the unnecessary pull request. Can I advice to add the product_id: v5jlnn5hwyffkhp3 and to mark the dps 126 as optional?

Thank you very much! :)

make-all:
Both these are already the case. I will review for other differences though.

Issue #3102, opened 2025-03-29
Request support for SR208C (Solar collector pump relay with temperature sensors)


BeriBGood:
### Log messages

```text
Adding SR208C device with product id hyrswocbx6i1rveo

Device specification: [{"id": 1, "name": "switch", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 2, "name": "mode", "type": "Enum", "format": "{\"range\":[\"cold\",\"heating\",\"auto\"]}", "enumMap": {}}, {"id": 4, "name": "temp_set", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":0,\"max\":100,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 6, "name": "temp_unit_convert", "type": "Enum", "format": "{\"range\":[\"c\",\"f\"]}", "enumMap": {}}, {"id": 8, "name": "instant_heating", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 14, "name": "countdown_left", "type": "Integer", "format": "{\"unit\":\"min\",\"min\":0,\"max\":1440,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 21, "name": "temp_top", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":0,\"max\":4000,\"scale\":1,\"step\":1}", "enumMap": {}}, {"id": 22, "name": "temp_bottom", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":0,\"max\":4000,\"scale\":1,\"step\":1}", "enumMap": {}}, {"id": 30, "name": "pump_state", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 32, "name": "ele_heating_state", "type": "Boolean", "format": "{}", "enumMap": {}}]

Device matches zigbee_smart_bulb with quality of 40%. DPS: {"updated_at": 1743313612.7234359, "1": false, "2": "heating", "4": 60, "6": "c", "8": false, "21": 1090, "22": 419, "26": 155, "30": false, "32": false}
###
```

### Product ID

 hyrswocbx6i1rveo

### Product Name

SR208C

### DPS information

```text
QueryThingsDataModel:

{
  "result": {
    "model": "{\"modelId\":\"000003kw5f\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"默认服务\",\"properties\":[
[...]

PR #3335, opened 2025-06-06
Create Sr223

basdelouw78:
Heatpipe controller sr223.


make-all:
This device has the same product id as the sr208c from #3102 added in 2025.6.0 release.

basdelouw78:
Sorry for the late action, this is the first time i do this and i'm not sure how it works on github.The tuya import says it has id bf193a921225e3b0dekzvm

make-all:
That is a device id, which varies from device to device, so is not interesting to put into a config file. It should only be used for your own device configuration.

PR #3336, opened 2025-06-06
Create CN EV charger 22kw.yaml

Anzic23:
I checked with myself. The data appeared in the home assistant. Charging is controlled. If you need any more tests or edits, please let us know, I will be happy to help.

### DIFF
--- a/custom_components/tuya_local/devices/CN EV charger 22kw.yaml
+++ b/custom_components/tuya_local/devices/CN EV charger 22kw.yaml
@@ -1,0 +1,503 @@
+name: CN EV charger # Название, которое будет отображаться в Home Assistant
+products:
+  - id: s2wublk5n2pycus1 # product_id, который вы предодоставили
+    manufacturer: cn ev charger # Производитель, который вы предоставили
+    model: 22kw # Модель, которую вы предоставили
+    model_id: 000004uqrl # modelId из данных вашей зарядки
+entities:
+  # DP 109: charging_state (Boolean) - Переключатель состояния зарядки
+  # accessMode: rw
+  - entity: switch
+    icon: "mdi:ev-station"
+    name: Состояние Зарядки
+    dps:
+      - id: 109
+        type: boolean
+        name: switch # Имя DP для Home Assistant
+
+  # DP 103: voltage_phase_a (Value / integer, scale 2) - Сенсор напряжения Фазы A
+  # accessMode: ro
+  - entity: sensor
+    name: Напряжение Фазы A
+    class: voltage
+    unit: V
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        mapping:
+          - scale: 100 # Делим на 100 для корректного значения
+
+  # DP 104: voltage_phase_b (Value / integer, scale 2) - Сенсор напряжения Фазы B
+  # accessMode: ro
+  - entity: sensor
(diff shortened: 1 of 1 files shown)


Anzic23:
image

custom_components/tuya_local/devices/CN EV charger 22kw.yaml
```diff
@@ -0,0 +1,510 @@
+name: Моя Зарядная Станция  # Название, которое будет отображаться в Home Assistant
+products:
+  - id: s2wublk5n2pycus1  # product_id, который вы предодоставили
+    manufacturer: cn ev charger  # Производитель, который вы предоставили
```
make-all: If it is this one, then Bluesky would be more appropriate, as it seems to be their marketing brand
https://cnevcharger.com/
These seem to have model numbers that can go into the model field also.

custom_components/tuya_local/devices/CN EV charger 22kw.yaml
```diff
@@ -0,0 +1,510 @@
...
+    name: Состояние Зарядки
+    dps:
+      - id: 109
+        type: boolean
+        name: switch  # Имя DP для Home Assistant
+
+  # DP 103: voltage_phase_a (Value / integer, scale 2)  # Сенсор напряжения Фазы A
+  # accessMode: ro
+  - entity: sensor
+    name: Напряжение Фазы A
```
make-all: I suggest instead of this Russian name to use
```yaml
  translation_key: voltage_x
  translation_parameters:
    x: A
```

custom_components/tuya_local/devices/CN EV charger 22kw.yaml
Anzic23: I'm not sure, charging looks different. Here is a photo of charging and a photo of the sticker on charging:
image
photo_2025-06-09_16-25-40

make-all:
Thank you, I think there may be an English manual which will help my understanding here https://www.evlaidai.lt/images/Userfiles/K055TA-AC-FG.pdf

Anzic23:
> Thank you, I think there may be an English manual which will help my understanding here https://www.evlaidai.lt/images/Userfiles/K055TA-AC-FG.pdf

I also found the [manual](https://kalininauto.ru/wp-content/uploads/2023/10/ta-ac-al-tary-ev-charger-manual.pdf), and ChatGPT suggests that this is the manufacturer of Anhui Tary Tongda Mechanical & Electrical Co., Ltd., which produces equipment under the TARY trademark.
I entered this data in the config.

Anzic23:
I decided to try updating the config data, a lot of entities appeared. Then I decided to delete and re-add charging, but now it is not added. How can this be fixed? At the same time, in Iobroker, I see that the phase parameters are updated and the charging status is online. Integration issues the following log::
This error occurred in the user integration.

Logger: custom_components. tuya_local. device
Source: custom_components/tuya_local/device. py:630
integration: Tuya Local (documentation, issues)
First recorded: 18: 11: 27 (8 cases)
Last recorded: 18: 13: 58

Failed to refresh device state for Test.

I tried all versions of the protocols and auto too. I also checked the "Survey only" box, but it didn't help either.

Anzic23:
Everything is fine, I downloaded the integration again, added the file manually , and everything started. You can add it to the stable release!

PR #3342, opened 2025-06-07
Updates to water heater entity

thd6n75:
- water heater to reference DPS 101 instead of 102
- keep warm switch

https://github.com/make-all/tuya-local/pull/3301


make-all:
This is obsolete, as the config has changed substantially

PR #3360, opened 2025-06-11
Added Fral FSC08 Air Conditioner

LunarEclipse363:
This should be a full configuration for Fral Super Cool FSC08 WiFi, however when I tried to connect my unit it would not appear in the search results, and I couldn't find any information on debugging this sort of problem.

I don't know what's wrong with this PR but it felt like it would be a waste to not share this code, if anyone knows how to figure out what's wrong here please help!

For now I will be trying localtuya personally because I really want to have a working HA integration for this and already spent the whole day trying to get it to work, but if this is some simple mistake I'm willing to try tuya-local again.

## Extra Device Information

Things Data Model from the Tuya developer portal: [fral_fsc08_tuya-things_data_model.json](https://github.com/user-attachments/files/20697146/fral_fsc08_tuya-things_data_model.json)
Manual for the device: [FSC08_Instrukcja_obsługi_PL_EN_DE_01.02.2024_wer.1.pdf](https://github.com/user-attachments/files/20697178/FSC08_Instrukcja_obslugi_PL_EN_DE_01.02.2024_wer.1.pdf)

Notably, the device will only be actually pairable a single time after it's plugged into power. If pairing fails, you must unplug it from power, plug it in again, and then hold the fan button for 3s while it's in standby to enter pairing mode. It seems to only support the "slow flashing" mode.

The device uses API version 3.4 (doesn't work with any other, and that was reported by tinytuya.

Tinytuya device information: 
```json
[
    {
        "name": "mobile air conditioning",
        "id": "[REDACTED]",
        "key": "[REDACTED]",
        "mac": "4c:a9:19:ff:ff:ff",
        "uuid": "[REDACTED]",
        "sn": "[REDACTED]",
        "category": "kt",
        "product_name": "mobile air conditioning",
        "product_id": "pkudk18s3u1dshxn",
        "biz_type": 18,
        "model": "PCR-QMA(FRAL)",
        "sub": false,
        "icon": "https://images.tuyaeu.com/smart/icon/bay1650340653090jw4U/a57ff2c2062387852f592f619d788ea8.jpg",
        "mapping": {
            "1": {
                "code": "switch",
                "type": "Boolean",
                "values": {}
            },
            "2": {
                "code": "temp_set",
                "type": "Integer",
                "values": {
                    "unit": "\u00b0C",
                    "min": 17,
                    "max": 30,
                    "scale": 0,
                    "step": 1
                }
            },
            "3": {
                "code": "temp_current",
                "type": "Integer",
                "values": {
                    "unit": "\u00b0C",
                    "min": 0,
                    "max": 50,
                    "scale": 0,
                    "step": 1
                }
            },
            "4": {
                "code": "mode",
                "type": "Enum",
                "values": {
                    "range": [
                        "cold"
                    ]
                }
            },
            "5": {
                "code": "fan_speed_enum",
                "type": "Enum",
                "values": {
                    "range": [
                        "high",
                        "low",
                        "auto"
                    ]
                }
            },
            "19": {
                "code": "temp_unit_convert",
                "type": "Enum",
                "values": {
                    "range": [
                        "c",
                        "f"
                    ]
                }
            },
            "20": {
                "code": "countdown_set",
                "type": "Enum",
                "values": {
                    "range": [
                        "cancel",
                        "1h",
                        "2h",
                        "3h",
                        "4h",
                        "5h",
                        "6h"
                    ]
                }
            },
            "21": {
                "code": "countdown_left",
                "type": "Integer",
                "values": {
                    "unit": "min",
                    "min": 0,
                    "max": 1440,
                    "scale": 0,
                    "step": 1
                }
            },
            "23": {
                "code": "temp_current_f",
                "type": "Integer",
                "values": {
                    "unit": "\u00b0F",
                    "min": 32,
                    "max": 99,
                    "scale": 0,
                    "step": 1
                }
            },
            "24": {
                "code": "temp_set_f",
                "type": "Integer",
                "values": {
                    "unit": "\u00b0F",
                    "min": 63,
                    "max": 86,
                    "scale": 0,
                    "step": 1
                }
            },
            "33": {
                "code": "switch_horizontal",
                "type": "Boolean",
                "values": {}
            }
        }
    }
]
```

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -90,7 +90,7 @@
 - ElectriQ SupremeCool SC16HPW
 - Fersk Vind 2
 - Fisher Summer air conditioner
-- Fral Super Cool FSC08 WiFi portable air conditioner
+- Fral Super Cool FSC08 WiFi portable air conditioner (uses protocol version 3.4)
 - Fral Super Cool FSC14.2 DH portable air conditioner
 - Friedrich Uni-Fit air conditioner (models: UCT14A30, UCT12A30, UCT08B10A)
 - Fujicool Yuzu heat pump
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -90,6 +90,7 @@
 - ElectriQ SupremeCool SC16HPW
 - Fersk Vind 2
 - Fisher Summer air conditioner
+- Fral Super Cool FSC08 WiFi portable air conditioner
 - Fral Super Cool FSC14.2 DH portable air conditioner
 - Friedrich Uni-Fit air conditioner (models: UCT14A30, UCT12A30, UCT08B10A)
 - Fujicool Yuzu heat pump
--- a/custom_components/tuya_local/devices/fral_supercool_airconditioner_fsc08.yaml
+++ b/custom_components/tuya_local/devices/fral_supercool_airconditioner_fsc08.yaml
@@ -1,0 +1,126 @@
+name: Portable Air Conditioner
+products:
+  - id: pkudk18s3u1dshxn
+    manufacturer: Fral
+    model: Super Cool FSC08 WiFi
+    model_id: FSC08
+entities:
+  - entity: climate
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: cold
+                value: cool
+              - dps_val: dry
+                value: dry
+              - dps_val: fan
+                value: fan_only
+      - id: 2
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 17
+          max: 30
+      - id: 3
+        type: integer
+        name: current_temperature
+        readonly: true
+        unit: C
+        range:
+          min: 0
+          max: 50
+      - id: 4
+        type: string
+        name: mode
+        hidden: true
+      - id: 5
+        type: string
+        name: fan_mode
+        mapping:
+          - dps_val: high
+            value: high
+          - dps_val: low
+            value: low
+          - dps_val: auto
+            value: auto
+      - id: 19
+        type: string
+        name: temperature_unit
+        mapping:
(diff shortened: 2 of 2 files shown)


make-all:
The clue for why it will not detect the correct config is likely to be in the logs. At least the log of what it does detect will show the dps received from the device, but if you turn on debug logging before adding it, you will also get detailed matching info for every config tried, so you can search for the config name to find it.

Likely some dps need to be marked optional. But if you are using the cloud config option, then if the product ids match it should match anyway if dps are missing, so would have to be a type mismatch in the dps defined in the config with what is received from the device.

LunarEclipse363:
Thanks! I checked the logs and it looks like it varies which temperature DPs it sends depending on the temperature unit setting, I've marked those as optional and will check now.

I'm not sure how to enable debug logs for an integration which hasn't been added yet but the base logs have proven enough to figure that part out.

LunarEclipse363:
I've also found it very convenient that HACS lets you add a custom repository, makes it much easier to test stuff than copying tarballs to my server c:

LunarEclipse363:
Okay, now it partially works!

image
image

The only thing missing is the shutoff timer, which I'm guessing I messed up the YAML for (i.e. I assumed that you don't need to explicitly set names for the states if they are the same in the DP as in HA but I guess I was wrong on that)

image

LunarEclipse363:
Ah that should have been on the dps not the entity.

LunarEclipse363:
Okay, so I have a problem - when you set a timer, and then cancel it, the device still reports the previous value but in reality the timer is cancelled (the timer indicator LED on the panel turns off).

I don't know how to fix that, but I will probably remove the persist/force options since they didn't fix anything.

LunarEclipse363:
It looks like cancelling the timer, whether online or via the panel on the device, never changes the values available via the Tuya API until the AC is turned off completely, so I think that's a firmware issue rather than something wrong with this PR.

LunarEclipse363:
Okay, it seems to work about as well as it can now:
image
image

PR #3377, opened 2025-06-16
New Device: Earu Multi Circuit Power Meter

BevDan:
The new device Earu Multi Circuit Power Meter uses several raw data types that are little endian but were also 2's complemented, I didn't find a way of dealing with 2's complement or signed=true via the yaml file. I have added some lines of code to the device_config.py in the helpers directory that allows a new entry in the yaml file
The new entry in the yaml file is signed_bits: to allow entry of the number of bits in the mask to be used to calculate the signed value
The alternative could be to use signed: True as an input in the yaml file and modify device_config.py to count the bits in the mask


BevDan:
Hello Jason, I guessing it is to be expected that my yaml will fail some of the tests as I have added the 'signed_bits' to the yaml to handle the numbers being offered by the device, am I going about adding this functionality the right way?
Regards,
Bevan

custom_components/tuya_local/helpers/device_config.py
```diff
@@ -473,7 +480,14 @@ def get_value(self, device):
         if mask and isinstance(bytevalue, bytes):
             value = int.from_bytes(bytevalue, self.endianness)
             scale = mask & (1 + ~mask)
-            return self._map_from_dps((value & mask) // scale, device)
+            raw_result = (value & mask) // scale
+
+            # Insert signed interpretation here
+            signed_bits = self._config.get("signed_bits")
```
make-all: I think it would be better to just use a true/false setting in the config, and use mask.bit_count() to get the number of bits. Probably a name `mask_signed` is best to get across that it is related to the use of mask.

custom_components/tuya_local/helpers/device_config.py
```diff
@@ -483,10 +483,11 @@ def get_value(self, device):
             raw_result = (value & mask) // scale
 
             # Insert signed interpretation here
-            signed_bits = self._config.get("signed_bits")
-            if signed_bits:
-                raw_result = to_signed(raw_result, signed_bits)
-
+            if self._config.get("mask_signed", False):
+            # Count how many bits are set in the mask
+                bit_count = bin(mask).count("1")
```
make-all: `mask.bit_count()` should do the same more efficiently. Is there are reason you changed from what I proposed?

BevDan: Sorry I missed the suggestion you had made, I will change it to mask.bit_count()

tests/test_device_config.py
```diff
@@ -119,6 +119,7 @@
         vol.Optional("format"): [FORMAT_SCHEMA],
         vol.Optional("mask"): str,
         vol.Optional("endianness"): vol.In(["little"]),
+        vol.Optional("signed_bits"): int,
```
make-all: The test should also change to `vol.Optional("mask_signed"): True` (only true is considered valid, as it is better to be omitted in the false case).

BevDan:
Thank you for all your work on this Jason, I can see a lot of effort has gone into this project. 
Hopefully I have sorted out the changes you asked for, I will have to wait until my solar kicks in again in the morning to check these changes have been successful on the live system
Regards,
Bevan

BevDan:
Hello Jason,
Can now confirm that the changes made are working on my live system, I had missed changing some of the 'signed_bits' entries to 'mask_signed', believe I have found them all now.
Regards,
Bevan

PR #3381, opened 2025-06-17
Added product ID in airrobo_p20_vacuum.yaml

GianlucaUlivi:


GianlucaUlivi:
17_09-57-51_chrome-Tuya_Developer_Platform_-_Google_Chrome

GianlucaUlivi:
Tested this change locally: setup with "Smart Life cloud-assisted device setup" correctly recognized and suggest the correct devices, while "Manually provide device connection information" do not suggest this device but 2 other similar vacuums.

custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -1,6 +1,6 @@
 name: Robot vacuum
 # products:
-#  - id: UNKNOWN
+#  - id: fptshnsx98tak0zo
```
make-all: Please also remove the commenting from these lines

GianlucaUlivi: Sorry my bad, done now.  
Thanks

Issue #2235, opened 2024-08-23
Request support for <Cleverio PF100>


FredrikM97:
### Log message

```
DPS: {  "3": 1, "11": 90, "12": false, "14": 0, "15": 1, "19": true, "101": "ac_power_on", "103": "real_time_mode", "112": 3, "116": "97%" }
```

<details>

```json
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2024.8.1",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.12.4",
    "docker": true,
    "arch": "aarch64",
    "timezone": "Europe/Stockholm",
    "os_name": "Linux",
    "os_version": "6.6.31-haos-raspi",
    "supervisor": "2024.08.0",
    "host_os": "Home Assistant OS 13.0",
    "docker_version": "26.1.4",
    "chassis": "embedded",
    "run_as_root": true
  },
  "custom_components": {
    "localtuya": {
      "documentation": "https://github.com/rospogrigio/localtuya/",
      "version": "5.2.1",
      "requirements": []
    },
    "browser_mod": {
      "documentation": "https://github.com/thomasloven/hass-browser_mod/blob/master/README.md",
      "version": "2.3.1",
      "requirements": []
    },
    "tuya_local": {
      "documentation": "https://github.com/make-all/tuya-local",
      "version": "2024.7.1",
      "requirements": [
        "tinytuya==1.15.1",
        "tuya-device-sharing-sdk>=0.1.9,<=0.2.99"
      ]
    },
    "sureha": {
      "documentation": "https://github.com/benleb/sureha",
      "version": "0.2.0",
      "requirements": [
        "surepy>=0.9.0"
      ]
    },
    "hacs": {
      "documentation": "https://hacs.xyz/docs/configuration/start",
      "version": "1.34.0",
      "requirements": [
        "aiogithubapi>=22.10.1"
      ]
    }
  },
  "integration_manifest": {
    "domain": "tuya_local",
    "name": "Tuya Local",
    "codeowners": [
      "make-all"
    ],
    "config_flow": true,
    "dependencies": [],
    "documentation": "https://github.com/make-all/tuya-local",
    "integration_type": "device",
    "iot_class": "local_push",
    "issue_tracker": "https://github.com/make-all/tuya-local/issues",
    "requirements": [
      "tinytuya==1.15.1",
      "tuya-device-sharing-sdk>=0.1.9,<=0.2.99"
[...]

PR #3384, opened 2025-06-17
Clarify fault code and separate mode and meal_plan from Manual feed f…

FredrikM97:
The following is changed:
* Meal plan as separate configuration and be able to modify it.
* Mode to tell if the device use eco or real time changes.
* Problem specify what problem it is and extra attribute for the name of error

image

image

### DIFF
--- a/custom_components/tuya_local/devices/cleverio_pf100_petfeeder.yaml
+++ b/custom_components/tuya_local/devices/cleverio_pf100_petfeeder.yaml
@@ -14,13 +14,29 @@ entities:
         range:
           min: 1
           max: 20
-      - id: 1
-        type: string
-        optional: true
-        name: meal_plan
+  - entity: sensor
+    name: Mode
+    class: enum
+    category: diagnostic
+    dps:
       - id: 103
         type: string
-        name: mode
+        name: sensor
+        mapping:
+          - dps_val: 1
+            value: real_time_mode
+          - dps_val: 2
+            value: eco_mode
+
+  - entity: text
+    name: Meal plan
+    category: config
+    hidden: true
+    dps:
+      - id: 1
+        type: base64
+        name: value
+
   - entity: button
     translation_key: factory_reset
     category: config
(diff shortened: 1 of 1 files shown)


FredrikM97:
One thing that could be improved is when Unknown (assume it is None?) would be displayed which would happen if the network goes down. Cant find any filter parameter to make it ignore those responses and keep current value. Any suggestions?

custom_components/tuya_local/devices/cleverio_pf100_petfeeder.yaml
```diff
@@ -48,20 +64,30 @@ entities:
       - id: 12
         type: boolean
         name: sensor
-  - entity: binary_sensor
-    class: problem
+  - entity: sensor
```
make-all: This breaks with the standard pattern of such sensors. Instead of changing this to a numeric sensor with "fault_name" as an attribute, you should just add one more attribute called "description".

custom_components/tuya_local/devices/cleverio_pf100_petfeeder.yaml
```diff
@@ -14,13 +14,29 @@ entities:
         range:
           min: 1
           max: 20
-      - id: 1
-        type: string
-        optional: true
-        name: meal_plan
+  - entity: sensor
```
make-all: Is it only a sensor? It seems to make more sense if the user can control it.

As a sensor, it might make more sense as a binary_sensor of class connectivity. Otherwise it can be a switch (name "Realtime updates")

FredrikM97: binary_sensor make more sense, I will update the pr.

custom_components/tuya_local/devices/cleverio_pf100_petfeeder.yaml
FredrikM97: Sorry about that, fixing

custom_components/tuya_local/devices/cleverio_pf100_petfeeder.yaml
FredrikM97: Unsure if there is a control for it. Cant find any functions for it.

make-all: The function would be how frequently the device returns data. It is likely hidden in the app and functions based on whether you are looking at the device details or not. The way to tell if it is a sensor or a switch is to look at the Query Things Data Model output for a "rw" or "ro".

custom_components/tuya_local/devices/cleverio_pf100_petfeeder.yaml
```diff
@@ -14,13 +14,26 @@ entities:
...
-      - id: 1
-        type: string
-        optional: true
-        name: meal_plan
-      - id: 103
+  - entity: binary_sensor
+    name: Eco mode
+    category: diagnostic
+    dps:
+      - id: 12
```
make-all: This is the same dp id as the battery charging sensor. It can only be one of these (boolean or string)

custom_components/tuya_local/devices/cleverio_pf100_petfeeder.yaml
make-all: Checking the details on #2235, it appears to be read-only.

custom_components/tuya_local/devices/cleverio_pf100_petfeeder.yaml
make-all: It seems this should be 103, not 12. The details on #2235 match with these two values.

PR #3400, opened 2025-06-21
New device: Tadiran Wind40 Inverter AC

deliad2:
product-id pdwzhqdjh6ppnjpb


make-all:
I am not accepting PRs for products from this company that retains board level ties to the former military division even after splitting the company a few years back.

deliad2:
What’s the point? Why are you concerned with the company’s internal affairs? This is a widely used product — what really matters is the user experience, not the politics behind these corporations.

PR #3408, opened 2025-06-24
New device: Arlec ceh243ha ceramic heater

hugglesfox:
Tested locally


custom_components/tuya_local/devices/arlec_ceh243ha_heater.yaml
```diff
@@ -0,0 +1,78 @@
+name: Heater
+products:
+  - id: bfc14552e7e985d62expty
```
make-all: This does not appear to be a product id, it looks like a device id

hugglesfox: Thanks for picking that up. Fixed

Issue #3405, opened 2025-06-23
De EV charger - controlling amps


shifthk:
### Description

When adjusting the charging current (amps) downward via the slider UI in Home Assistant, the maximum value of the slider becomes incorrectly capped at the previously selected (higher) amperage value.

https://github.com/user-attachments/assets/dffb8177-3717-4b35-bc38-bef112cb31f6

[tuya_local-01JXPK39TD702NF8GZ0NF5AYA5-EV charger-708eff34c5185fb0b2aefae9f509f257.json](https://github.com/user-attachments/files/20866852/tuya_local-01JXPK39TD702NF8GZ0NF5AYA5-EV.charger-708eff34c5185fb0b2aefae9f509f257.json)

### Steps to reproduce

1. Open the EV charger control panel in Home Assistant.

2. Use the slider to set a lower amperage (e.g., from 32A down to 10A).

3. Observe that the slider's maximum becomes locked to 10A, and it is no longer possible to drag it back up to a higher value.

### Expected behaviour

The amperage slider should always:

Reflect the correct allowable range (e.g. 6A-32A).

Allow both increasing and decreasing amps within this range, regardless of prior value.

### Confirmation

- [x] I am sure this is a bug or improvement, that is well enough described that it can be implemented.
- [x] This report is for tuya-local, **not** for localtuya.
- [x] If this is for a specific device, the device config filename is mentioned above, or diagnostics are included.

- [x] Log messages or diagnostics relevant to the issue are included.

- [x] This is not requesting addition of a new device.

PR #3409, opened 2025-06-24
Update dewall_evcharger.yaml

shifthk:
Fixed current (amps) selector bar


make-all:
I'm not sure what the intention was with the additional entity, but I have implemented the change requested on #3405 which is the topic of the commit message.

PR #3424, opened 2025-06-28
Aesthetic changes earu_multicircuit_energymeter.yaml

BevDan:
Hello Jason,
Was wanting to do some Aesthetic changes to the earu_multicircuit_energymeter.yaml file, so I downloaded the version that is due to be released, installed it on my live HA, the sub circuit power and current measurements failed, the debug report gave this message
` File "/config/custom_components/tuya_local/helpers/device_config.py", line 469, in mask
    return int(mask, 16)
ValueError: invalid literal for int() with base 16: '000000000000000000000000000000000000000000000000000000\\ 000000000000000000000000000000000000000000000000000000\\ 000000000000000000000000000000000000000000FFFFFFFF0000'
2025-06-28 15:51:05.341 ERROR (MainThread) [homeassistant] Error doing job: Exception in callback Entity._async_write_ha_state_from_call_soon_threadsafe() (None)
`
It looks like the continuation notation is corrupting the mask

Changed the yaml file so that measurement show up as Sensors instead of Diagnostics, Also removed the continuation notation to avoid errors

Regards,
Bevan

### DIFF
--- a/custom_components/tuya_local/devices/earu_multicircuit_energymeter.yaml
+++ b/custom_components/tuya_local/devices/earu_multicircuit_energymeter.yaml
@@ -9,7 +9,6 @@ entities:
     translation_key: voltage_x
     translation_placeholders:
       x: A
-    category: diagnostic
     dps:
       - id: 101
         optional: true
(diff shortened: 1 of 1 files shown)


make-all:
I don't think having so many non-diagnostic sensors on a device is a good idea, the primary function is an energy meter, so other measurements are of secondary importance and many users will not want them cluttering up the generated UI.

The issue with the masks can be fixed by adding double quotes, so I have installed that change.

PR #3444, opened 2025-07-01
Add support for Howeall Register Booster Fan

dlee1j1:
**Device:** Howeall Register Booster Fan  
**Product ID:** zbhikxau56accfsp

**Features supported:**
- Fan power control (DP 1)
- Fan speed 1-10 (DP 12)  
- Operating modes: Fan/Cool/Heat (DP 2)
- Temperature triggers for cooling/heating (DP 45, 106)
- Current temperature sensor (DP 9)
- Child lock (DP 14)
- Light control (DP 102)
- Various status sensors (DP 101, 103-105)

**Changes:**
- Added device configuration file: `howeall_register_booster_fan.yaml`
- Updated `DEVICES.md` to include the new device in the Fans section

**Testing:** 
tested on the 4x10" and 4x12'" versions of the register booster fans bought from Amazon. 
https://a.co/d/50yyX08. 
Runs on my home assistant instance at home with automation that syncs the fan operation with the thermostat kicking in for both heating and cooling

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -264,6 +264,7 @@
 - Hiper T3 bladeless fan
 - Hoenofly Smart Amari ceiling fan with light
 - Homebase 12" oscillating fan
+- Howeall Register Booster Fan
 - Hunter Pacific 9 speed ceiling fan with light
 - HYD WeAir Plus bladeless fan with heating function
 - Immax Neo Lite Vento ceiling fan with light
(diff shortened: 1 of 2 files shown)


dlee1j1:
I just redid this to be a fan with speed controls. Let me know what other rework is needed. Thanks,

make-all:
Fan entity is an improvement, but I think this device fits the climate entity better.

PR #3457, opened 2025-07-04
feat: restore text on restart and add property to ignore None values

FredrikM97:
This change introduces a mechanism to retain the meal_plan value after a restart. Since meal_plan only updates when explicitly modified, it's reasonable to exclude it from automatic resets. This avoids the need to redefine the value after every reload, restart, or temporary loss of access to the device.


make-all:
I don't agree with using undocumented internal classes of HA to workaround issues in device firmware.

The class you changed is also noted as a mixin in the comment, so does not seem to be intended to be used as a base class like this. I cannot find any examples of this class being used in the HA codebase outside of the unit test for it, so there is no example of what correct usage would be.

 This will also change the behaviour of text entities in ways that may not be compatible with all uses across the 1000+ config files that exist. In general, we don't know how long a device has been offline for and what was done while it was offline, so it is misleading at best to show the status as anything other than Unknown or Unavailable.

FredrikM97:
Maybe we can simplify it then?

Since it is required to be base64 we should not allow other patterns to avoid this error:
```
Pattern attribute value [-A-Za-z0-9+/]*={0,3} is not a valid regular expression: Uncaught SyntaxError: Invalid regular expression: /[-A-Za-z0-9+/]*={0,3}/v: Invalid character in character class
```
value was `unknown`

PR #3482, opened 2025-07-11
Device Support - Herschel XLS T-PL Plugin Wifi Thermostat

Keith-H-T:
Added support for a new device.

Manufacturer: Herschel
Device: XLS T-PL Plugin Wifi Thermostat

Device Webpage: https://www.herschel-infrared.com.au/product/t-pl-plugin-thermostat/
Device Manual: https://drive.google.com/file/d/1x3qmKyOxJx0ugtU__A7yZegjx-SRQLGt/view

Includes HVAC_STATE monitoring on ID109.

### DIFF
--- a/custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
+++ b/custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
@@ -132,20 +132,20 @@ entities:
         range:
           min: 0
           max: 99
-  - entity: select
+  - entity: switch
     name: Open window detection
     icon: "mdi:window-open"
     category: config
     dps:
       - id: 103
-        name: option
         type: string
+        name: switch
         optional: true
         mapping:
           - dps_val: continue
-            value: "On"
+            value: true
           - dps_val: close
-            value: "Off"
+            value: false
   - entity: sensor
     name: Up time
     class: duration
--- a/custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
+++ b/custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
@@ -1,6 +1,6 @@
 name: IR Heater
 products:
-  - id: cl2wgyhohpjjjuzb
+  - id: eb7636b53a97009d63e1ur
     manufacturer: Herschel
     model: Herschel XLS T-PL Plugin Wifi Thermostat 
 entities:
--- a/custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
+++ b/custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
@@ -1,0 +1,159 @@
+name: IR Heater
+products:
+  - id: cl2wgyhohpjjjuzb
+    manufacturer: Herschel
+    model: Herschel XLS T-PL Plugin Wifi Thermostat 
+entities:
+  - entity: climate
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
```diff
@@ -0,0 +1,160 @@
+name: IR Heater
```
make-all: If it is a thermostat, then put Thermostat here, not IR heater.

custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
```diff
@@ -0,0 +1,160 @@
...
+products:
+  - id: 7o1bzm3omsqau6w6
+    manufacturer: Herschel
+    model: XLS-T-PL
+    name: Herschel plugin Wifi thermostat
```
make-all: Name not needed here, it is just a combination of manufacturer, model and top level name

custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
```diff
@@ -0,0 +1,160 @@
...
+        range:
+          min: 0
+          max: 370
+        unit: "C"
+      - id: 3
+        name: current_temperature
+        type: integer
+        mapping:
+          - scale: 10
+            step: 10
```
make-all: No step needed on current_temperature, as the user cannot change it, you just want to display the data that comes from the device.

custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
```diff
@@ -0,0 +1,160 @@
...
+    dps:
+      - id: 101
+        name: value
+        type: integer
+        unit: d
+        optional: true
+        range:
+          min: 0
+          max: 99
+  - entity: select
```
make-all: Since there are only on and off options, this can be a switch.

custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
Keith-H-T: Amended

custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
Keith-H-T: Name removed

custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
Keith-H-T: Step removed

custom_components/tuya_local/devices/herschel_xls_tpl_thermostat.yaml
Keith-H-T: id 101 is number of days in holiday mode, value 0 - 99

make-all: My comment was in reference to the select entity for open window detection that follows that.

Keith-H-T: Done. Sorry took me awhile to get the sytnax right

PR #3484, opened 2025-07-11
feat(devices): add support for Tuya MINI-W1 3 gang switch

rigibe:
Added support for generic Tuya MINI-W1 3 gang switch.
Tested with real device

### DIFF
--- a/custom_components/tuya_local/devices/tuya_3_gang_switch.yaml
+++ b/custom_components/tuya_local/devices/tuya_3_gang_switch.yaml
@@ -1,0 +1,121 @@
+name: Triple switch with timer
+products:
+  - id: 3njf1fgcsbuvkbhw
+    manufacturer: Tuya
+    model: MINI-W1
+    name: Triple switch with timer
+entities:
+  - entity: switch
+    translation_key: switch_x
+    translation_placeholders:
+      x: "1"
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+      - id: 42
+        type: string
+        name: random_schedule
+      - id: 43
+        type: string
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/tuya_3_gang_switch.yaml
```diff
@@ -0,0 +1,121 @@
+name: Triple switch with timer
+products:
+  - id: 3njf1fgcsbuvkbhw
+    manufacturer: Tuya
```
make-all: Manufacturer seems to be Aubess from the info I can find online.

custom_components/tuya_local/devices/tuya_3_gang_switch.yaml
```diff
@@ -0,0 +1,121 @@
...
+    model: MINI-W1
+    name: Triple switch with timer
```
make-all: Name does not need to be duplicated here.

make-all (commented):
filename should be less generic. If Aubess is the correct manufacturer, aubess_3gang_switch.yaml would fit nicely with the existing configs.

custom_components/tuya_local/devices/tuya_3_gang_switch.yaml
rigibe: Yes, it seems so. My device was unbranded, though.

custom_components/tuya_local/devices/tuya_3_gang_switch.yaml
rigibe: Removed

PR #3485, opened 2025-07-11
Add support for wireless level sensor ME201WZS

camillePernoud:
tested with sensor at home

### DIFF
--- a/custom_components/tuya_local/devices/morray_level_sensor.yaml
+++ b/custom_components/tuya_local/devices/morray_level_sensor.yaml
@@ -1,0 +1,109 @@
+name: Level sensor
+products:
+  - id: ywcgq
+    manufacturer: MORRAY
+    model: ME201WZS
+entities:
+  - entity: sensor
+    translation_key: status
+    class: enum
+    dps:
+      - id: 1
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: normal
+            value: normal
+          - dps_val: lower_alarm
+            value: empty
+          - dps_val: upper_alarm
+            value: full
+  - entity: sensor
+    name: Depth
+    class: distance
+    dps:
+      - id: 2
+        type: integer
+        name: sensor
+        unit: m
+        class: measurement
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Batterie
+    class: voltage
+    dps:
+      - id: 5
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/morray_level_sensor.yaml
```diff
@@ -0,0 +1,109 @@
+name: Level sensor
+products:
+  - id: ywcgq
```
make-all, line 3: Product id is not usually this short, even for Zigbee devices. Is it missing some characters, or a different id?

custom_components/tuya_local/devices/morray_level_sensor.yaml
```diff
@@ -0,0 +1,109 @@
...
+    manufacturer: MORRAY
```
make-all, line 4: The brand seems to be "Moray", with one r and not an abbreviation so no need for all-caps.

custom_components/tuya_local/devices/morray_level_sensor.yaml
```diff
@@ -0,0 +1,109 @@
...
+      - id: 1
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: normal
+            value: normal
+          - dps_val: lower_alarm
+            value: empty
+          - dps_val: upper_alarm
+            value: full
```
make-all, line 20: I think low and high are better here than empty and full, as the threshold is not 100 and 0, but whatever was set.

custom_components/tuya_local/devices/morray_level_sensor.yaml
```diff
@@ -0,0 +1,109 @@
...
+    dps:
+      - id: 2
+        type: integer
+        name: sensor
+        unit: m
+        class: measurement
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Batterie
```
make-all, line 33: Battery.

make-all:
Apart from the additional Battery voltage reading, this is identical to ept_level_sensor.yaml. It might be better to add the voltage as an optional sensor there.

make-all:
I think between the EPT level sensor and the newly added MW201W with different limits in some of the config options, this should be covered now and the PR not needed.

camillePernoud:
thanks a lot

PR #3490, opened 2025-07-13
feat: add support for Dura Comfort DH50PWM Dehumidifier

ShawnBuckley:
Adds support for the Dura Comfort DH50PWM Dehumidifier.
https://www.amazon.com/dp/B09NBBM1PS

Is a switch required to implement a fan entity? I've used DP-1 in this case for both the dehumidifier and the fan since it seemed like it was required to get it working.

Sample DPs from tinytuya:
```
{'1': True, '5': '0', '6': 40, '8': '0', '15': 32, '18': 53, '20': False, '101': 1}
```

I'm not sure what `101` maps to.

The fault code accounts for P1 which is tank full/removed. E1, E2, E3 exist based on looking at some of the Amazon reviews and look to be related to things such as critical failure, and ambient temperature too hot or shut down due to coils being frosted.

image
image

### DIFF
--- a/custom_components/tuya_local/devices/duracomfort_dh50pwm_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/duracomfort_dh50pwm_dehumidifier.yaml
@@ -2,6 +2,7 @@ name: Dehumidifier
 products:
   - id: keyfa7qemdvvux5c
     manufacturer: Dura Comfort
+    model: DH50PWM
 entities:
   - entity: humidifier
     class: dehumidifier
@@ -33,7 +34,6 @@ entities:
         name: current_humidity
         type: integer
   - entity: fan
-    translation_only_key: fan_with_presets
     dps:
       - id: 1
         type: boolean
@@ -55,6 +55,8 @@ entities:
         name: sensor
         mapping:
           - dps_val: 0
+            value: false
+          - dps_val: 32
             value: false
           - value: true
   - entity: binary_sensor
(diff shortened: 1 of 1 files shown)


make-all:
Switch is recommended in the fan entity if the speed does not go down to zero (stopped). Some UI cards will only give easy access to the switch, so it is inconvenient if there is no turn_on/off capability.

custom_components/tuya_local/devices/duracomfort_dh50pwm_dehumidifier.yaml
```diff
@@ -0,0 +1,77 @@
...
+        range:
+          min: 35
+          max: 80
+        mapping:
+          - step: 5
+      - id: 18
+        name: current_humidity
+        type: integer
+  - entity: fan
+    translation_only_key: fan_with_presets
```
make-all: There are no presets, the translation key is not needed.

custom_components/tuya_local/devices/duracomfort_dh50pwm_dehumidifier.yaml
```diff
@@ -0,0 +1,77 @@
...
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 15
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
```
make-all, line 61: You should also make the "tank_full" value false, as that has its own sensor, and is not a "problem" in the same way an actual hardware fault is.
It is useful to also add a fault_code attribute to the problem sensor with the raw code in it, otherwise it is lost.

PR #3508, opened 2025-07-17
PlantsIO Ivy smart planter - add touch buttons

thewade:
This PR adds the various touch pads on the [PlantsIO Ivy Smart Planter](https://store.plantsio.com/products/ivy-smart-planter) to use as buttons to trigger automations.

This was requested as per https://github.com/make-all/tuya-local/discussions/2791#discussioncomment-13262702.

Because event entity is stateless, I only implemented the touch event. I noticed other implementations also had a release state, but I think that event would fire every time the device is loaded not just on release, which does not make sense to me. If you disagree let me know.

Touching both the left and right pad at the same creates a unique hug state so I have mapped that as well.

Example automation using the hug event:

```yaml
alias: Love
description: ""
triggers:
  - entity_id:
      - event.smart_planter_hug
    trigger: state
conditions:
  - condition: state
    entity_id: event.smart_planter_hug
    attribute: event_type
    state: pressed
actions:
  - target:
      entity_id: tts.home_assistant_cloud
    data:
      media_player_entity_id: media_player.bedroom_display
      message: I love you!
    action: tts.speak
mode: single
```

### DIFF
--- a/custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
+++ b/custom_components/tuya_local/devices/plantsio_ivy_smart_planter.yaml
@@ -141,6 +141,55 @@ entities:
           - dps_val: 2
             value: "Too Dry"
           - value: "Unknown state"
+  # Button entities
+  - entity: event
+    name: Left touch
+    class: button
+    dps:
+      - id: 127
+        type: boolean
+        name: event
+        optional: true
+        mapping:
+          - dps_val: true
+            value: pressed
+          - value: null
+  - entity: event
+    name: Right touch
+    class: button
+    dps:
+      - id: 128
+        type: boolean
+        name: event
+        optional: true
+        mapping:
+          - dps_val: true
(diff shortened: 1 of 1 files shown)


make-all:
I think due to the fact that the Tuya protocol does not guarantee real-time delivery, we cannot rely on a boolean key-down / key-up event to derive short and long press events ourselves. If the device is not doing the timing for multiple events itself, then a single press event is the best we can reliably support, so I think the current implementation is good.

Maybe some icons would be useful? `mdi:hand-pointing-right`, `mdi:hand-pointing-left`, `mdi:hand-pointing-up`, `mdi:hand-okay`, for example?

thewade:
Sounds good, I’ll have a look through the icons this evening.

thewade:
I added some icons.

image

I think the set works well. The direction of the pointing is counter intuitive reading the yaml, but it represents how you would interact with the actual device.

PR #3517, opened 2025-07-18
Update makeskyblue_mppt_charger.yaml

alexeytm82:
add description fault code


alexeytm82:
I don't understand where the mistake is.

Run ruff check --select I --diff .
--- custom_components/tuya_local/time.py
+++ custom_components/tuya_local/time.py
@@ -3,7 +3,7 @@
 """
 
 import logging
-from datetime import time, timedelta, datetime
+from datetime import datetime, time, timedelta
 
 from homeassistant.components.time import TimeEntity
 

Would fix 1 error.
Error: Process completed with exit code 1.

make-all:
The lint error is unrelated to this change.

PR #3519, opened 2025-07-18
fix: Separate Meal Plan into text config

FredrikM97:
Allow edit of the meal_plan for multiple devices


custom_components/tuya_local/devices/arlec_5l_petfeeder.yaml
```diff
@@ -15,8 +15,11 @@ entities:
         range:
           min: 1
           max: 12
-      - id: 1
-        type: string
+  - entity: text
+    name: Meal plan
+    category: config
+    optional: true
```
make-all: optional: true should go on the dp definition.
Generally text entities like this should have `hidden: true` to hide them by default, as their use requires knowledge of the format that the dp uses, and often base64 decoding and encoding, so it is an advanced feature that most users will not want cluttering up the UI.

FredrikM97: Sounds good. Will remove optional and replace with hidden.

make-all: But if the `optional: true` was on the dp originally, then it will still be required (on the dp).

FredrikM97: What differs optional and hidden? I thought hidden was for password/sensitive info? Is there any docs for this to explain the properties

make-all: There are different uses of `hidden: true`:
 - entity level (here): instruct HA to make the entity hidden (and disabled) by default. Used for advanced entities that most users will not be interested in, or will find difficult to use. Also `hidden: unavailable` is used together with an `available` dp to cover different variants of devices in a single config.
 - dp level: do not include the dp in the extra attributes on the entity (for when the dp is used in the entity to implement some conditional logic, or as a value_redirect, but it is not necessary to show the raw value)
 
`optional: true` is only valid at dp level, and is used when a device is being added to still match the config if some dps are not provided by the device.

For password/sensitive info, there is `sensitive: true`

All of these default to false, so you don't generally need to set them to false.

FredrikM97: I see so we want both optional and hidden. I will update the later today.

make-all: You want:
```yaml
   - entity: text
     name: ...
     hidden: true
     dps:
       - id: N
         optional: true
```

PR #3521, opened 2025-07-18
Add pre-push validation to indicate early syntax errors

FredrikM97:


.yamllint.yaml
```diff
@@ -4,3 +4,5 @@ extends: default
 rules:
   document-start: disable
   new-lines: disable
+  comments:
```
make-all: What is the purpose of this change from the standard rules?

custom_components/tuya_local/devices/feyree_ev_portable_charger.yaml
```diff
@@ -248,20 +248,20 @@ entities:
...
-      #     min: 8
-      #     max: 60
-      # - id: 126
-      #   type: integer
-      #   optional: true
-      #   name: value_80a
-      #   range:
-      #     min: 24
-      #     max: 80
+          # - id: 125
```
FredrikM97, line 251: These changes are due to yamllint complaining. Best would be to remove but Indent also work.

.yamllint.yaml
FredrikM97: Allow comment to not need double spaces to reduce number of required changes.

make-all: But there are no required changes currently for this rule.

FredrikM97: Yamllint default to two space before comment. Currently most code uses one space. This change allow to keep current code as is.

Unclear why yamllint not complain on it in github actions.

FredrikM97:
If we don't want the prettier changes I suggest to remove it from this pr or I modify the configuration to suit the current structure.
It is also possible to change to only check/fix the edited files.

.yamllint.yaml
make-all: Yamllint does complain on github actions about this. But there are currently no instances of this in the repo (at least the devices directory, where yamllint is limited to)

The reason for keeping this rule is that it can potentially catch unintentional comments,  when # happens to appear in an unquoted string. In unintentional cases, 0 or 1 spaces are likely, but 2 or more spaces is less likely except when intentionally appending a comment to the line.

make-all:
Yes, I think removing prettier from this would be better. It is OK on new files, but having it even for only edited files will lead to actual changes getting lost in the noise of whatever this month's prettier rules happen to be.

.yamllint.yaml
FredrikM97: Running it locally there was a few instances but could old older files. Could be due to it checking all files and not just the commited files. Can review if possible to only validate the staged files.

FredrikM97:
I modified it now to run all the pre-commit on commit only (prettier too) this will only change the modified files/new files and not a global overwrite. How about it?

.yamllint.yaml
FredrikM97: Removed it for now.

FredrikM97: Removed it for now. Can be fixed later if we see any issues.

PR #3522, opened 2025-07-18
Fix syntax with prettier

FredrikM97:
Followup to #3521 to  format all files the same.
Command used: `pre-commit run --all-files`


make-all:
What is the advantage of adding another tool to take over this job from ruff?

FredrikM97:
Ruff for python, prettier for none python files (like yaml and md). It fixed some of complaints from yamllint which did not have a "write" to fix the code.

make-all:
I see, but it appears to have changed almost 20% of the repository to suit its own inflexible mostly whitespace rules (in some cases it is impossible to see what it has actually done). I don't think this is a good tool to force on the project

FredrikM97:
> I see, but it appears to have changed almost 20% of the repository to suit its own inflexible mostly whitespace rules (in some cases it is impossible to see what it has actually done). I don't think this is a good tool to force on the project

Understandable.

PR #3524, opened 2025-07-19
add support for Raddy PT-5 pool thermometer

sportchucks:


custom_components/tuya_local/devices/pool_thermometer.yaml
```diff
@@ -0,0 +1,36 @@
+name: Raddy PT-5 pool thermometer
+products:
+  - id: qirx90k6hg4qzpbz
+    model: EM3406B-TY
+entities:
+  - entity: sensor
+    name: indoor temperature
```
make-all: Is it really indoor? Maybe "Ambient temperature" or "Air temperature" would be better to handle both indoor and outdoor installations.

Entity names should have the first letter of the first word capitalised.

custom_components/tuya_local/devices/pool_thermometer.yaml
```diff
@@ -0,0 +1,36 @@
...
+    dps:
+      - id: 101
+        name: sensor
+        type: integer
+        class: measurement
+        unit: C
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: indoor humidity
```
make-all: I suggest just remove the name and allow HA to use the built in translations for humidity class.

custom_components/tuya_local/devices/pool_thermometer.yaml
```diff
@@ -0,0 +1,36 @@
...
+    class: humidity
+    dps:
+      - id: 102
+        name: sensor
+        type: integer
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    name: pool temperature
```
make-all: As this is the primary function of the device, I suggest to remove the name and allow HA to use the translations for temperature class.

PR #3530, opened 2025-07-22
Add support for 36 and 40 inch Holmes Smart Tower Fans

demartinoaj:
Add support for 36 and 40 inch Holmes Smart Tower Fans


custom_components/tuya_local/devices/holmes_tower_fan.yaml
```diff
@@ -0,0 +1,71 @@
+name: Tower Fan
+products:
+  - id: eb6ec964408209669ajeca
```
make-all: This is a device id, not a product id

custom_components/tuya_local/devices/holmes_tower_fan.yaml
```diff
@@ -0,0 +1,71 @@
...
+    manufacturer: Holmes
+    model: 32510020
```
make-all: Numeric values need to be quoted where strings are expected.

PR #3543, opened 2025-07-26
Create Relay_Module_Smart_16CH.yaml

stepsolar:


custom_components/tuya_local/devices/rcu16_relaymodule.yaml
```diff
@@ -0,0 +1,264 @@
...
+        name: option
+        mapping:
+          - dps_val: "off"
+            value: "off"
+          - dps_val: "on"
+            value: "on"
+          - dps_val: "fixed"
+            value: "memory"
+  - entity: select
+    name: 1-8 IO set
```
make-all: It is better not to start names with numbers, it may cause issues elsewhere in HA.

Perhaps "I/O mode 1-8".

It seems to match this description in the linked webpage:
> 3 Working Modes: Switch among interlock/self-locking/inching modes.

Maybe interlock/self-lock/inching would be clearer options, though it isn't 100% clear how they should be mapped: eturn maybe inching, sensor interlock?

custom_components/tuya_local/devices/rcu16_relaymodule.yaml
```diff
@@ -0,0 +1,264 @@
...
+        name: option
+        mapping:
+          - dps_val: "eturn"
+            value: "eturn"
+          - dps_val: "switch"
+            value: "switch"
+          - dps_val: "sensor"
+            value: "sensor"
+  - entity: select
+    name: Mode 1
```
make-all: Based on the app screenshots on the web page, Mode 1 seems to apply to switches 11 and 12, mode 2 to 13 and 14, and mode 3 to 15 and 16. So maybe the naming should show this.

custom_components/tuya_local/devices/rcu16_relaymodule.yaml
```diff
@@ -0,0 +1,264 @@
...
+    name: Scene
+    translation_key: scene
+    category: config
+    dps:
+      - id: 123
+        type: string
+        name: option
+        mapping:
+          - dps_val: "ready"
+            value: "ready"
```
make-all: Maybe "off" or "none" would be better?

custom_components/tuya_local/devices/rcu16_relaymodule.yaml
```diff
@@ -0,0 +1,264 @@
...
+          - dps_val: "scene4"
+            value: "scene4"
+          - dps_val: "scene5"
+            value: "scene5"
+          - dps_val: "scene6"
+            value: "scene6"
+          - dps_val: "scene7"
+            value: "scene7"
+  - entity: number
+    name: Number 1-30
```
make-all: What is this number? In the app screenshots it is just labelled "number", in which case the name can be left off to use HA default translations for number. But it isn't really clear to users - if it is a timer, then it should have a translation_key timer and class duration with approriate units.

stepsolar:
Even the app is not that clear, here are the photos
Screenshot_2025-07-31-15-01-54-496_com tuya smartlife
Screenshot_2025-07-31-15-01-39-273_com tuya smartlife
Screenshot_2025-07-31-15-01-30-629_com tuya smartlife

Issue #2594, opened 2024-12-02
Request support for Klarta Forste 4


dzejdzej1:
### Log message

```text
Dear All,

I could see on a list of your supported devices one of Klarta Air Purifier, however it was not allowing me to select that when adding mine model: Klarta Forste 4. I could only select 2 types of a regular fan which allowed me to only turn on/off the device.

Nevertheless I was playing around with .yaml files comparing a regular fan and your already existing Klarta Air Purifier model.

I combined a part from the regular fan and added from already existing Klarta device .yaml file what I thought my device is capable of retrieving and it worked with the newly created file: I can retrieve and control more data from my device now.

As I am an amateur I am including the configuration from .yaml file that is eventually working for me in "Information about how the device functions" section. Feel free to review that and if you think that any part of it is worth something, maybe you can consider to add Klarta Forste 4 to your standard list of .yaml configuration file
```

### Product Name

Klarta Forste 4

### DPS information

```text
{
  "result": {
[...]

PR #3546, opened 2025-07-26
Add support for Klarta Forste 4 air purifier

rimmaciej:
This PR adds support for [Klarta Forste 4](https://klarta.pl/oczyszczacze-powietrza/klarta-forste-4/) air purifier.
It's based on the already existing config for Klarta Stor 2: https://github.com/make-all/tuya-local/pull/1365.

There are slight changes to the speed values and some missing features compared to Stor 2, but this works for my setup, I've been including this config manually for a while, and it's worked flawlessly for the purifier.

image

### DIFF
--- a/custom_components/tuya_local/devices/klarta_forste4_purifier.yaml
+++ b/custom_components/tuya_local/devices/klarta_forste4_purifier.yaml
@@ -1,0 +1,97 @@
+name: Air purifier
+products:
+  - id: zxpmhropoeoigxav
+    manufacturer: Klarta
+    model: Forste 4
+entities:
+  - entity: fan
+    translation_only_key: fan_with_presets
+    icon: "mdi:air-purifier"
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 3
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: auto
+            value: smart
+          - dps_val: manual
+            value: normal
+          - dps_val: sleep
+            value: sleep
+          - dps_val: screenoff
+            value: displayoff
+      - id: 4
+        type: string
+        name: speed
+        mapping:
+          - dps_val: 8
+            value: 100
+          - dps_val: 6
+            value: 75
+          - dps_val: 4
+            value: 50
+          - dps_val: 2
+            value: 25
+  - entity: sensor
+    class: pm25
+    dps:
+      - id: 2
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/klarta_forste4_purifier.yaml
```diff
@@ -0,0 +1,97 @@
...
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    translation_key: air_quality
+    dps:
+      - id: 21
+        type: string
+        name: sensor
```
make-all, line 98: It may be better to make this sensor `class: enum` and list the possible values in a mapping (in which you can also remap any that do not exactly match the translations).

custom_components/tuya_local/devices/klarta_forste4_purifier.yaml
```diff
@@ -91,7 +91,17 @@ entities:
...
     translation_key: air_quality
+    class: enum
     dps:
       - id: 21
         type: string
         name: sensor
+        mapping:
+          - dps_val: great
+            value: excellent
+          - dps_val: Good
```
make-all, line 102: Are the values from the device for Good and Poor capitalised? The Klarta Stor may also need modifying in that case, as they are likely to be similar.

Issue #2594 which previously requested this device (without logs) lists `{"type":"enum", "range":["great","good","mild","poor"]}` which is all lower case.

custom_components/tuya_local/devices/klarta_forste4_purifier.yaml
rimmaciej: Alright, I've changed it to an enum. I tried triggering all the states, and it gave me `great`, `Good`, `mild` and `Poor`, so I've mapped those to excellent,good,moderate and poor. I see that there are 5 possible values, but this purifier gives only those 4 as far as I can see. 

The tuya app also includes only these 4.

custom_components/tuya_local/devices/klarta_forste4_purifier.yaml
rimmaciej: Yeah, they are. I'm not sure I'd assume that they would be similar :) 
This seems very inconsistent in general. I don't have the Stor though, so I can't verify.
image

PR #3549, opened 2025-07-27
Update its_45hd_heatpump to correct heating / cooling

dmostert:
Current definition always shows "Idle" even when unit is actually heating the water.
This update changes this

### DIFF
--- a/custom_components/tuya_local/devices/its_45hd_heatpump.yaml
+++ b/custom_components/tuya_local/devices/its_45hd_heatpump.yaml
@@ -33,6 +33,9 @@ entities:
       - id: 14
         type: integer
         name: EEV
+      - id: 17
+        type: string
+        name: work_mode
       - id: 27
         type: boolean
         name: hvac_action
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/its_45hd_heatpump.yaml
```diff
@@ -33,20 +33,18 @@ entities:
       - id: 14
         type: integer
         name: EEV
-      - id: 17
-        type: string
+      - id: 27
```
make-all, line 39: I'm not sure what dp 17 (work_state) is actually indicating, but you might want to keep the dp as an extra attribute for monitoring. It says it has output "normal", "NO" and "ERR", but maybe it is always stuck on "NO" rather than indicating "normal" when running as was assumed.

dmostert: Thanks - this looks like a good idea

PR #3556, opened 2025-07-27
Add Hoenofly Smart Wood Low Profile Ceiling Fan with Lights

fender4645:
**Amazon Product Link:** https://a.co/d/3PUya5i

There a few other Hoenofly smart fans that already have support but this particular one hasn't seemed to be added and has a different set of DPS settings.

This gives support for both the fan and light entities. For the fan, there are 6 speeds and will be exposed both in the fan speed slider (as percentages) as well as scenes (where you a select the speed number directly). The light is pretty straight forward, allowing on/off, brightness, and color temperature.


fender4645:
Will file a new PR

PR #3583, opened 2025-08-03
New Device - PC473 3 phase energy monitor / EARU

JakubMadro:


make-all:
This is the same device as #3612, but some of the dps ids are different.
It would help resolve the differences if one of you could provide the QueryThingsDataModel result from the cloud API Explorer.

PR #3593, opened 2025-08-05
Fix Detection Because Optional Flag is missing inkbird_ibs_m2

LordGuenni:
Hey there just a Heads Up, the latest change broke the matching because some of the values only get transmitted after a time, the initial  connection only includes those:  
{"updated_at": 1754338572.520221, "9": "c", "113": 0, "114": 0, "115": 0, "116": 0, "117": 0, "118": 0, "119": 0, "120": 0, "121": 0, "122": 0, "124": 1, "125": 1, "126": 0} of which only the 9 is relevant.  I added the needed optional flags to the DPS.

Thx. Florian

### DIFF
--- a/custom_components/tuya_local/devices/inkbird_ibs_m2.yaml
+++ b/custom_components/tuya_local/devices/inkbird_ibs_m2.yaml
@@ -15,6 +15,7 @@ entities:
             value: celsius
           - dps_val: "f"
             value: fahrenheit
+
   - entity: sensor
     name: Temperature probe 1
     class: temperature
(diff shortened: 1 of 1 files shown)


LordGuenni:
Also when a device doesnt support humidity it transmits the humidity Value as this: 65535 , based on the code this should then disable the value, but doesnt appear to be working, i have  tried locally with different configs and conditions but didnt get it working, maybe you have a better Idea ?

Payload: CA4B////////RQ==
Length: 10 bytes
Raw bytes: [8, 14, 1, 255, 255, 255, 255, 255, 255, 69]
Temperature: 27.0 °C
Humidity:    65535 %
Battery:     69 %
Available:   True

PR #3594, opened 2025-08-05
Create yoolax_ble_roller_v2.yaml

ZacTyAdams:
Adding support for Yoolax BLE Roller V2 shades:
https://www.yoolax.com/collections/roller-shades

### DIFF
--- a/custom_components/tuya_local/devices/yoolax_ble_roller_v2.yaml
+++ b/custom_components/tuya_local/devices/yoolax_ble_roller_v2.yaml
@@ -1,0 +1,59 @@
+name: BLE Roller V2
+products:
+  - id: hldq12vsaqalcfer
+    manufacturer: Yoolax
+    name: BLE Roller V2
+entities:
+  - entity: cover
+    class: shade
+    dps:
+      - id: 1
+        name: control
+        type: string
+        mapping:
+          - dps_val: open
+            value: open
+          - dps_val: close
+            value: close
+          - dps_val: stop
+            value: stop
+          - invert: true
+      - id: 2
+        name: position
+        type: integer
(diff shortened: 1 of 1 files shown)


ZacTyAdams:
Just a heads up for anyone looking at this, onboarding this device didn't work with standard Moes BLE gateways. The manufacturer had to send their BLE gateway to get this working.

make-all:
A new config is not needed for this device, the graywind_shades.yaml already matches this product id, though with the position uninverted. If the inverted position is really needed, the Loonas curtain config should also match.

ZacTyAdams:
@make-all No it doesn't, there are other issues with the mappings that causes the device to desync entirely from the local tuya integration when using any other device configs. I only used the same product ID as I am unsure of where to retrieve the one for this particular device. If this cannot be merged in how can I have my config not overwritten by subsequent releases in my local environment?

make-all:
What are the issues with the mappings in the updated loonas and graywind configs?
Your config is materially no different than the loonas_curtain config after the latest update if you ignore the product ids, which you have just told me you randomly copied instead of getting the correct product id. This just reinforces that it would not be appropriate to accept your contribution.

ZacTyAdams:
image

@make-all There are many missing and broken bits when using another config. I have a job though and can't find the time to continue mapping and troubleshooting the rest of this but for myself and other individuals purchasing these blinds they won't know loonas vs. greywind vs. yoolax because they are all different blinds, this commit is an incremental change and  foundational stop gap=. If you don't want to accept it or provide guidance that is fine, then how do I keep my custom config in my local instance so I don't keep losing my work and can fix this later down the road if you're unable to.

PR #3599, opened 2025-08-07
New device: Thermoval Aqua TI Water Heater

szupi-ipuzs:
I have recently installed [this water heater](https://waterm24.pl/produkt/thermoval-elektryczny-ogrzewacz-pojemnosciowy-aqua-ti-100l-w-wifi/) in my house and it works with Tuya Smart. Using the (attached) schema obtained from Tuya IOT I have created this yaml. Seems to work fine in HA :)
The only minor thing is that the device lets you change the temperature in the "Manual" mode only (in other modes the temperature is hardcoded) - is there a way to model this behavior in yaml?

Here are the values that the heater reported initially:

`
DPS: {"updated_at": 1754514991.5316627, "1": true, "2": "Eco", "9": 55, "10": 48, "13": "Heating", "20": 0}
`

[schema.txt](https://github.com/user-attachments/files/21660685/schema.txt)

### DIFF
--- a/custom_components/tuya_local/devices/thermoval-aquati-waterheater.yaml
+++ b/custom_components/tuya_local/devices/thermoval-aquati-waterheater.yaml
@@ -1,4 +1,4 @@
-name: Thermoval Aqua Ti Water Heater
+name: Electric Water Heater
 products:
   - id: jxma5t0bc0sagddh
     manufacturer: Thermoval
@@ -10,19 +10,19 @@ entities:
     dps:
       - id: 1
         type: boolean
-        name: operation_mode
+        name: water_heater_mode
         mapping:
           - dps_val: false
-            value: Off
+            value: off
           - dps_val: true
             constraint: work_mode
             conditions:
               - dps_val: Manual
-                value: Manual
+                value: manual
               - dps_val: Eco
-                value: Eco
+                value: eco
               - dps_val: Turbo
-                value: Turbo
+                value: turbo
       - id: 2
         type: string
         name: work_mode
--- a/custom_components/tuya_local/devices/thermoval-aquati-waterheater.yaml
+++ b/custom_components/tuya_local/devices/thermoval-aquati-waterheater.yaml
@@ -1,0 +1,90 @@
+name: Thermoval Aqua Ti Water Heater
+products:
+  - id: jxma5t0bc0sagddh
+    manufacturer: Thermoval
+    model_id: "WS-42S-2"
+    model: "Aqua Ti Water Heater"
+entities:
+  - entity: water_heater
+    translation_key: water_air
+    dps:
+      - id: 1
+        type: boolean
+        name: operation_mode
+        mapping:
+          - dps_val: false
+            value: Off
+          - dps_val: true
+            constraint: work_mode
+            conditions:
+              - dps_val: Manual
+                value: Manual
+              - dps_val: Eco
+                value: Eco
+              - dps_val: Turbo
+                value: Turbo
+      - id: 2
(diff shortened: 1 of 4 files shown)


szupi-ipuzs:
Hmm, I've just noticed the "off" mode does not show correctly and actually cannot be selected. Any idea what might be causing this?
I'm using the latest HA (2025.8.0)

image

custom_components/tuya_local/devices/thermoval-aquati-waterheater.yaml
```diff
@@ -0,0 +1,92 @@
+name: Thermoval Aqua Ti Water Heater
```
make-all: The top level name should be without branding, as other products may match the config. Branding belongs under products, where it will be used in future if the product id matches.

custom_components/tuya_local/devices/thermoval-aquati-waterheater.yaml
```diff
@@ -0,0 +1,92 @@
...
+      - id: 1
+        type: boolean
+        name: operation_mode
+        mapping:
+          - dps_val: false
+            value: Off
+          - dps_val: true
+            constraint: work_mode
+            conditions:
+              - dps_val: Manual
```
make-all: These are not valid operation modes. It is better to just map true to `heat_pump`, and have a separate select entity for the modes.

szupi-ipuzs: Hmm I would prefer this to shown as the heater mode in HA.
Can I just add the new "water_heater_mode" and use it instead?

custom_components/tuya_local/translations/en.json
```diff
@@ -464,6 +464,15 @@
                     "oolong_tea": "Oolong tea"
                 }
             },
+            "water_heater_mode": {
```
make-all: You cannot extend entities by defining translations for the extra attributes. My suggestion was to add a select entity for the work_mode, and make the operation_mode a simple boolean mapped to off and heat_pump (or electric if it is not a heat pump water heater)

make-all: I guess another possibility is to change the mapping to use valid operation_mode values, such as `eco`, `electric` and `performance`, maybe that would have been a better suggestion.

custom_components/tuya_local/devices/thermoval-aquati-waterheater.yaml
make-all: No, you cannot add new features to a platform like that. But see my comment below about using some of the standard strings (such as electric, eco, performance) instead, and leaving it as operation_mode.

custom_components/tuya_local/devices/thermoval-aquati-waterheater.yaml
```diff
@@ -0,0 +1,92 @@
...
+entities:
+  - entity: water_heater
+    translation_key: water_air
+    dps:
+      - id: 1
+        type: boolean
+        name: water_heater_mode
+        mapping:
+          - dps_val: false
+            value: off
```
make-all: The "off" needs to be quoted to avoid yaml interpreting it as a boolean and the parser converting it to Python `False`.

szupi-ipuzs: You were right, it helped

custom_components/tuya_local/translations/en.json
szupi-ipuzs: Ok, I ended up creating a separate select entity for this.

PR #3610, opened 2025-08-09
Adding moes star feather switches support

hussainaljamri:
Adding device config for moes star feather series switches (only 3 and 4 gang switches, I don't own the rest so can't test).

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -589,6 +589,7 @@ the best option.
 - Mirabella Genio smartplug with USB
 - Moes 3-gang switch
 - Moes 4-gang (quad) switch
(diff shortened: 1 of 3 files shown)


make-all:
It would be useful to get the dps info for the device, so that the timers and events for scene selection can be implemented by someone else.

hussainaljamri:
> It would be useful to get the dps info for the device, so that the timers and events for scene selection can be implemented by someone else.

I've been meaning to update this PR.. I did do some modifications for the countdown feature to work. I'll do another round this week and will push an update. didn't know we can consolidate the devices in a single file (I have a lot to learn here)

PR #3635, opened 2025-08-17
New Device: CurrentUSA Vita ColorPlus Aquarium Light

Yodie98:
[currentusa_vitacolorplus_aquarium_light.yaml.zip](https://github.com/user-attachments/files/21819521/currentusa_vitacolorplus_aquarium_light.yaml.zip)

New device, a nice fishtank light
CurrentUSA Vita ColorPlus


make-all:
This is the same device as the Serene ColorPlus aquarium light, which is supported already by the rgb_led_strip config.

PR #3653, opened 2025-08-22
Create conga_cecotec_1970_vacuum.yaml

malkavi:
Add config file for vacuum model Conga 1970ti from Cecotec. Tested basic configurations + directions and other dps using tuya local.

### DIFF
--- a/custom_components/tuya_local/devices/cecotec_congas1970_vacuum.yaml
+++ b/custom_components/tuya_local/devices/cecotec_congas1970_vacuum.yaml
@@ -1,8 +1,8 @@
-name: Conga vacuum
+name: Robot vacuum
 products:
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/cecotec_congas1970_vacuum.yaml
```diff
@@ -0,0 +1,274 @@
...
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    name: Mode
+    category: diagnostic
+    dps:
+      - id: 104
+        type: string
+        name: sensor
+  - entity: sensor
```
make-all: Should be a binary_sensor

PR #3659, opened 2025-08-24
Add Backward Blade Stop for Parkside PMRDA 20-Li A1

tribut:
This feature was added by a recent firmware upgrade to Parkside PMRDA 20-Li A1 (also sold as Parkside PAMRS 1000 A1).

From Tuya Developer Console / Query Data Model:

~~~~json
{
  "abilityId":121,"accessMode":"rw","code":"valuereserved01",
  "description":"功能点名称：Backward Blade Stop （后退停刀盘）。\n数据传输类型：可下发可上报\n数据类型：value\n功能点属性：数值范围：0-1\n备注：0是关闭，1是打开",
  "extensions":{"scope":""},"name":"valuereserved01",
  "typeSpec":{"type":"value","max":9999,"min":0,"scale":0,"step":1,"unit":""}
}
~~~~

English translation via DeepL:
> Function Point Name: Backward Blade Stop.
> Data Transmission Type: Can be sent down and reported up.
> Data Type: value.
> Function Point Attributes: Numerical Range: 0-1.
> Notes: 0 is off, 1 is on."

### DIFF
--- a/custom_components/tuya_local/devices/moebot_s_mower.yaml
+++ b/custom_components/tuya_local/devices/moebot_s_mower.yaml
@@ -307,9 +307,10 @@ entities:
         name: switch
         optional: true
   - entity: switch
-    name: Backward Blade Stop
+    name: Backward blade stop
     icon: "mdi:saw-blade"
     category: config
+    hidden: unavailable
     dps:
       - id: 121
         type: integer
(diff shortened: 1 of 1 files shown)


tribut:
I took the liberty of also adding the parkside mowers to the device list. If you'd rather have this in a separate PR, just let me know.

PR #3660, opened 2025-08-24
Add new device : RIYUE BOX 3 Fan Light

r3m0x:
Added device control for RIYUE BOX 3


custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
```diff
@@ -0,0 +1,170 @@
+name: RIYUE BOX 3 Fan Light
```
make-all: Top level name should be generic, in case other brands match.

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
```diff
@@ -0,0 +1,170 @@
...
+products:
+  - id: shrracpw0ve9jooc
+    manufacturer: RIYUE
+    model: RIYUE BOX 3 - Health Lighting
```
make-all: Manufacturer generally should not be repeated in the model name.

What is this device, I cannot find any trace of it on the internet?

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
```diff
@@ -0,0 +1,170 @@
...
+            value: fresh
+          - dps_val: nature
+            value: nature
+      - id: 62
+        name: speed
+        type: integer
+        range:
+          min: 1
+          max: 100
+      - id: 64
```
make-all: This should be broken out to a sensor using the translation_key: time_remaining, and class: duration.

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
```diff
@@ -0,0 +1,170 @@
...
+        range:
+          min: 10
+          max: 1000
+      - id: 23
+        name: color_temp
+        type: integer
+        range:
+          min: 0
+          max: 1000
+      - id: 26
```
make-all: This should be broken out to a time entity.

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
```diff
@@ -0,0 +1,170 @@
...
+        type: boolean
+        name: switch
+      - id: 21
+        type: string
+        name: color_mode
+        optional: true
+        mapping:
+          - dps_val: white
+            value: white
+          - dps_val: colour
```
make-all: There is no rgbhsv or named_color dp, are you sure this light supports colour?

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
```diff
@@ -0,0 +1,170 @@
...
+        type: string
+        optional: true
+  - entity: light
+    name: Night light
+    dps:
+      - id: 105
+        type: boolean
+        name: switch
+  - entity: select
+    name: Oscillation 1
```
make-all, line 84: I don't really understand how a fan can have 4 oscillation switches, and what the difference is between pause and off. If I could find the device online, it might help with understanding, but maybe there is a better way to name these that makes it clearer?

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - id: 66
+        type: boolean
+        name: switch
+  - entity: switch
+    name: General beeper
+    dps:
+      - id: 106
+        type: boolean
+        name: switch
+#  - entity: binary_sensor
```
make-all: Not sure why you commented this out

r3m0x: Add this will caused the device unable to connect to tuya-local, and the sensor not really meaningful for normal user

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
r3m0x: Not sure if u can find it, it has 4 ways / wind direction. For ur reference:
https://detail.tmall.com/item.htm?abbucket=1&id=811238408389

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
r3m0x: it only support warm to cool color, not rgb

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
r3m0x: https://detail.tmall.com/item.htm?abbucket=1&id=811238408389 for your reference. It usually sold with Mijia, but there  is Tuya variant

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
r3m0x: Will be resolved in next commit, thanks for pointing out

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
```diff
@@ -0,0 +1,147 @@
...
+    name: Do not disturb
+    dps:
+      - id: 34
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Switch gradient
+    dps:
+      - id: 35
+        type: raw
```
make-all: `raw` in tuya usually means a `base64` encoding (sometimes `utf16b64` which is UTF-16 text encoded in base64, but in this case I think binary data). So this is not a switch, but some configuration for the behaviour of the switch - maybe for fading the light in and out rather than abruptly turning on and off. You can expose it as a text entity, but hidden by default, since it will only be useful if you can understand the binary format it is using.

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
make-all: I see you just removed it. What I meant was to replace it with
````yaml
  - entity: sensor
    class: duration
    category: diagnostic
    dps:
      - id: 64
        type: integer
        name: sensor
        unit: min
  ````

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
make-all:
````yaml
  - entity: time
    translation_key: timer
    category: config
    dps:
      - id: 26
        type: integer
        name: second
        range:
          min: 0
          max: 86400
````

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
make-all: I think adding `optional: true` will resolve this. But to follow the general pattern of such sensors, use:

````yaml
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 67
        type: bitfield
        optional: true
        name: sensor
        mapping:
          - dps_val: 0
            value: false
          - dps_val: null
            value: false
          - value: true
      - id: 67
        type: bitfield
        optional: true
        name: fault_code
      - id: 67
        type: bitfield
        optional: true
        name: description
        mapping:
          - dps_val: 0
            value: ok
          - dps_val: 1
            value: motor_fault
````

custom_components/tuya_local/devices/riyue_box3_fanlight.yaml
make-all: That link just goes to a login page, unfortunately.

PR #3661, opened 2025-08-24
Add Sensor entity Id 10 to geyserwise_water_heater.yaml to provide the current temperature in class diaqgnostic

RandMental:
Add Sensor entity Id 10 to geyserwise_water_heater.yaml to provide the current temperature in class diagnostic


make-all:
Current temperature is already exposed in the water heater entity.

PR #3662, opened 2025-08-25
Add support for Libhof IM-112

fgsfds:
Support for Libhof IM-112 Ice Maker

image

image


make-all:
Unfortunately the project is not accepting devices that are exclusively made for the Russian market at this time.

PR #3670, opened 2025-08-27
Add Goldair GGK1000 Kettle device

ashabala:


custom_components/tuya_local/devices/goldair_ggk1000_kettle
```diff
@@ -0,0 +1,127 @@
...
+          max: 100
+        unit: C
+  - entity: select
+    translation_key: kettle_mode
+    dps:
+      - id: 11
+        type: string
+        name: option
+        mapping:
+          - dps_val: boiling_quick
```
make-all: It does not make sense to have a select with only one option.

custom_components/tuya_local/devices/goldair_ggk1000_kettle
```diff
@@ -0,0 +1,127 @@
...
+            value: boil
+  - entity: number
+    name: Target temperature
```
make-all: Entities should not be duplicated without good reason.

custom_components/tuya_local/devices/goldair_ggk1000_kettle
```diff
@@ -0,0 +1,127 @@
...
+          - dps_val: standby
+            value: standby
+          - dps_val: heating
+            value: heating
+          - dps_val: cooling
+            value: cooling
+          - dps_val: warm
+            value: keeping_warm
+  - entity: sensor
+    class: temperature
```
make-all: Another duplicate

custom_components/tuya_local/devices/goldair_ggk1000_kettle
```diff
@@ -0,0 +1,127 @@
...
+    dps:
+      - id: 17
+        type: integer
+        name: sensor
+        unit: min
+        range:
+          min: 0
+          max: 720
+  - entity: number
+    name: Countdown timer
```
make-all: translation_key: timer would be better

ashabala: Have added `translation_key: timer`

custom_components/tuya_local/devices/goldair_ggk1000_kettle
ashabala: Have removed

custom_components/tuya_local/devices/goldair_ggk1000_kettle
ashabala: Have removed

custom_components/tuya_local/devices/goldair_ggk1000_kettle
ashabala: I've changed this from a select box to a diagnostic sensor. I think it is worth keeping in case other options get added in future firmware updates.

PR #3673, opened 2025-08-27
refactor doel_tiplus_litterbox.yaml: reorganize YAML entities for clarity and consistency

valdenny:
- Grouped entities by type (sensors, binary_sensors, switches, lock, numbers, selects, buttons)
- Standardized entity names and icons in English
- Added missing entities: Bag Change Mode, Thin Feces Detection, Sleep Mode and Developer Mode
- Replaced Bin Full DPS id from 103 → 119 (correct mapping)
- Removed non-existing DPS 123 "Capacity Calibration"
- Added correct DPS 130 "Sand Surface Calibration"
- Simplified status sensor (id 24 + timing fields)
- Improved structure for readability and maintainability

### DIFF
--- a/custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
+++ b/custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
@@ -36,7 +36,7 @@ entities:
 
   - entity: sensor
     name: Daily Visits
-    icon: "mdi:emoticon-poop"
+    icon: "mdi:paper-roll"
     dps:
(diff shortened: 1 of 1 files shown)


make-all:
The naming changes are the opposite of standardising the names, as you have changed them from HA recommended naming convention to your own preference.
The reordering of entities hides other changes that are being made together. It is an unnecessary change that brings no value, but if you must make changes like that, you should do it as a separate change without any other changes mixed in with it.

PR #3678, opened 2025-08-28
Fix ability to change ringtone (mode)

theqz:
The ringtone (mode) needs to be separately configured for it to be honoured by the device


make-all:
I don't see a need for this. It is no different than calling the turn_on service of the siren giving only a tone argument.

theqz:
Doing that doesn't actually change the ringtone on the device for some reason. Changing the config this way does.

make-all:
Perhaps you can capture a debug log of that. It should be working that way.

PR #3679, opened 2025-08-28
This PR fixes incorrect DPS mappings and adds missing entities for the DOEL Ti+ litter box.

valdenny:
- Corrected Bin full DPS: `103 → 119`
- Removed non-existing DPS `123 Capacity calibration'
- Added correct DPS `130 Sand surface calibration` (0–6)
- Fixed schema for DPS `129` by setting `class: measurement` (unit: min)
- Aligned DP names 128/129 to device codes (`time`, `time_clear`)
- Added missing entities:
  - Developer mode (135)
  - Sleep mode (120)
  - Thin feces detection (127)
  - Bag change mode (138)
  - Bag change count (139)
- Updated icons for Daily visits and Daily visits duration to avoid duplication with Thin feces detection (127)

### DIFF
--- a/custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
+++ b/custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
@@ -224,12 +224,13 @@ entities:
     name: Developer mode
     category: config
     icon: "mdi:cog"
+    hidden: true
     dps:
       - id: 135
         type: boolean
         name: switch
   - entity: switch
-    name: Sleep mode
+    name: Sleep
     category: config
     icon: "mdi:sleep"
     dps:
@@ -245,7 +246,7 @@ entities:
         type: boolean
         name: switch
   - entity: button
-    name: Bag change mode
+    name: Bag change
     icon: "mdi:bag-personal"
     dps:
       - id: 138
--- a/custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
+++ b/custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
@@ -29,7 +29,7 @@ entities:
           - scale: 10
   - entity: sensor
     name: Daily visits
-    icon: "mdi:emoticon-poop"
+    icon: "mdi:paper-roll"
     dps:
       - id: 7
         type: integer
@@ -37,7 +37,7 @@ entities:
         unit: visits
   - entity: sensor
     name: Daily visits duration
-    icon: "mdi:paper-roll"
+    icon: "mdi:timer-sand"
     class: duration
     dps:
       - id: 8
@@ -77,7 +77,7 @@ entities:
     name: Bin full
     icon: "mdi:trash-can"
     dps:
-      - id: 103
+      - id: 119
         type: boolean
         name: sensor
   - entity: binary_sensor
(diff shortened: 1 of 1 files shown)


valdenny:
Controls
Sensors
Configuration
State - Attributes

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -227,3 +220,42 @@ entities:
             value: "Mineral"
           - dps_val: "mixed_cat_litter"
             value: "Mixed"
+  - entity: switch
```
make-all, line 223: Should probably be hidden: true, since generally this option should not be exposed to the user.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -227,3 +220,42 @@ entities:
...
+    name: Developer mode
+    category: config
+    icon: "mdi:cog"
+    dps:
+      - id: 135
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Sleep mode
```
make-all: "mode" not needed

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
```diff
@@ -227,3 +220,42 @@ entities:
...
+  - entity: switch
+    name: Thin feces detection
+    category: config
+    icon: "mdi:emoticon-poop"
+    dps:
+      - id: 127
+        type: boolean
+        name: switch
+  - entity: button
+    name: Bag change mode
```
make-all: "mode" not needed

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
valdenny: You are right, I agree with you.
In the Tuya app I saw that the “Developer Mode” switch only enables the Detection sensitivity (DPS 124), which is already exposed as a separate entity.
I’m not sure if it also unlocks other hidden functions, but in principle I agree that it should not be exposed directly to the user.
Actually, I had the same doubt and left it visible thinking that, by seeing it, other users after me who own the same litter box might have investigated the concept further and shed more light on it.
In any case, I will correct the file and mark it as hidden: true.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
valdenny: Ok, I will fix it.

custom_components/tuya_local/devices/doel_tiplus_litterbox.yaml
valdenny: Ok, I will fix it.

valdenny:
Ok, I fixed it.
I have marked Developer mode as hidden: true and renamed "Sleep mode" to "Sleep", and "Bag change mode" to "Bag change".
Before merging, I would like to ask if it’s ok with you if I reorder the entities in the YAML.
The idea is simply to group them by type (sensors, binary_sensors, switches, etc.) to improve readability and future maintenance.
This change would not affect any DPS mappings or functionality, only the ordering.
Thanks a lot for your review and support! 🙏

make-all:
The order is not really important. From a point of view of reviewing the config against Tuya documentation, it is more useful to have the entities in order of the lowest numbered dp. In any case it will be reordered to alphabetical order in the HA UI.

PR #3683, opened 2025-08-29
New Device: Cat Litter Box (multiple brands)

robosphinx:
At time of purchase (late 2023), brand was MEEGEEM with a few others available.

Current brands are BCHARYA and LubadiPaw, though I haven't purchased either to verify product IDs.

All are physically identical as far as I can tell.

Worth renaming to something generic like `smart_cat_litter_box.yaml` or `generic_cat_litter_box.yaml`, what ever the typical convention is.  Couldn't really figure that out. 

This configuration has been running in my instance for over a year.  Finally got some time to polish it up for contribution.

Weight is in kg but I couldn't figure out how to make `unit: kg` play nicely with `class: enum`.

image

### DIFF
--- a/custom_components/tuya_local/devices/meegeem_selfcleaningcatlitterbox.yaml
+++ b/custom_components/tuya_local/devices/meegeem_selfcleaningcatlitterbox.yaml
@@ -8,7 +8,6 @@ products:
 # LubadiPaw
 # Previous brands:
 # MEEGEEM
-#primary_entity:
 entities:
   - entity: sensor
     name: Action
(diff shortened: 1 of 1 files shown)


make-all:
You can't assume that because two products look the same, they have the same software. A generic name would be confusing because there are many other litter boxes already supported that obviously are not compatible if a new config is needed.

robosphinx:
That's a fair assertion.  I don't need to know badly enough to purchase one of the current doppelgangers to find out.  Happy to keep this as brand-specific ("MEEGEEM").

PR #3686, opened 2025-08-30
Add support for FUKUMARU (AF01-W) PET feeder

Epictek:
From what I can gather the fukumaru feeders all advertise themselves as "FUKUMARU-W" through the Tuya API (I only have the AF01-W so can't test this theory) so I've added the model id to the filename and model_id

Model IDs can be seen on this page
https://fukumarupet.com/pages/support

### DIFF
--- a/custom_components/tuya_local/devices/fukumaru_af01w_pet_feeder.yaml
+++ b/custom_components/tuya_local/devices/fukumaru_af01w_pet_feeder.yaml
@@ -1,9 +1,8 @@
-name: pet feeder
+name: Pet feeder
 products:
   - id: jttvwskiffndnb1q
-    manufacturer: FUKUMARU
-    model: FUKUMARU-W
-    model_id: AF01-W
+    manufacturer: Fukumaru
+    model: AF01-W
 entities:
   - entity: button
     name: Quick feed
@@ -12,9 +11,6 @@ entities:
       - id: 2
         type: boolean
         name: button
-        mapping:
-          - dps_val: true
-            value: true
   - entity: number
     name: Manual feed
     icon: "mdi:food-drumstick"


custom_components/tuya_local/devices/fukumaru_af01w_pet_feeder.yaml
```diff
@@ -0,0 +1,93 @@
...
+    dps:
+      - id: 16
+        type: integer
+        name: value
+        unit: times
+        range:
+          min: 0
+          max: 10
+  - entity: button
+    name: Factory reset
```
make-all: use `translation_key: factory_reset' instead of the explicit name and icon

custom_components/tuya_local/devices/fukumaru_af01w_pet_feeder.yaml
```diff
@@ -0,0 +1,93 @@
...
+    icon: "mdi:restore"
+    category: config
+    dps:
+      - id: 20
+        type: boolean
+        name: button
+        mapping:
```
make-all: The mapping is unnecessary

custom_components/tuya_local/devices/fukumaru_af01w_pet_feeder.yaml
```diff
@@ -0,0 +1,93 @@
...
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 10
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: number
+    name: Voice playback count
```
make-all: "count" implies a sensor, but this is a number entity, settable by the user. What is the purpose of this?

custom_components/tuya_local/devices/fukumaru_af01w_pet_feeder.yaml
```diff
@@ -0,0 +1,93 @@
...
+    name: Feed report
+    icon: "mdi:food-drumstick"
+    category: diagnostic
+    dps:
+      - id: 5
+        type: integer
+        name: sensor
+        unit: portions
+  - entity: sensor
+    name: Battery
```
make-all: name should be omitted, as the class defines a default which is translated.

PR #3688, opened 2025-08-31
Add Princess Smart Air Cooler

OriginalArkus:
This is a straightforward implementation for the Princess Smart Air Cooler.

Tuya lists DP 12 and calls it "light", but this actually puts the machine into a silence mode whereby pressing the controls doesn't cause a beep to be emitted.


custom_components/tuya_local/devices/princess_smart_air_cooler.yaml
```diff
@@ -0,0 +1,59 @@
+name: Princess Smart Air Cooler
```
make-all, line 1: This name should be generic (just "Air cooler")

PR #3706, opened 2025-09-07
Add pioneer mini split config

JeremiahChurch:
I've been running this config for more than a year now on two units - they've worked flawlessly.

I'd love to not have to re-add this file everytime this plugin gets updated :)

All credit goes to the post in this forum thread: https://community.home-assistant.io/t/pioneer-wyt-mini-split-wifi-integration/434616/29


make-all:
This config is outdated and internally inconsistent (for example dp 119 is listed twice with different types). It cannot be accepted like this. Based on the linked thread and previous issue that was opened for this device, it is compatible with the daizuki_heatpump, and there does not appear to be a need for a new config for it.

PR #3714, opened 2025-09-09
updated device configuration

witulu:
Hi, after using this device for a while, I made some quality of life changes to the configuration.


custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -64,6 +64,11 @@ entities:
             value: cleaning
           - dps_val: sleep
             value: idle
+      - id: 6
```
make-all, line 67: This should not be added, as Home Assistant has deprecated battery_level for vacuum entities, and will remove them soon.

custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -75,16 +80,17 @@ entities:
...
-            value: Normal
+            value: Standard
           - dps_val: strong
-            value: Strong
+            value: High
       - id: 18
         type: bitfield
         name: error
-        optional: true
+        hidden: true
```
make-all, line 89: Removing the optional from dp 18 will affect device detection, I think it is there for a reason. Hiding it is OK, since there is a separate problem sensor, and it only needs to be here to report error as the entity state.

custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -75,16 +80,17 @@ entities:
...
       - id: 104
         type: string
         name: serial_number
+        hidden: true
```
make-all, line 93: I don't see any reason to hide the serial number

custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -94,7 +100,7 @@ entities:
         unit: "%"
         class: measurement
   - entity: sensor
-    name: Edge brush life
+    name: Side brush life
```
make-all, line 103: Changing entity names after release is disruptive, so I would prefer not to.

custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -75,16 +80,17 @@ entities:
           - dps_val: gentle
             value: Silent
           - dps_val: normal
-            value: Normal
+            value: Standard
           - dps_val: strong
-            value: Strong
+            value: High
```
make-all, line 85: I don't see any point in these wording changes. They are less disruptive than entity name changes, but still could mess with someone's existing automations.

custom_components/tuya_local/devices/airrobo_p20_vacuum.yaml
```diff
@@ -141,7 +147,7 @@ entities:
         type: boolean
         optional: true
   - entity: button
-    name: Reset filter
+    translation_key: filter_reset
```
make-all, line 150: This one is a little more worthwhile, since it converts to use an existing translation, but I think the reason it was not already done when the translation was added was the disruptiveness of changing the entity id (which is derived from the name/translation_key)

PR #3715, opened 2025-09-10
Update geyserwise_water_heater.yaml - added Current Temperature to Diagnostic category 

RandMental:
Update geyserwise_water_heater.yaml to add Current Temperature to the Diagnostic category

This makes the Current temperature available outside the  water heater entity

### DIFF
--- a/custom_components/tuya_local/devices/geyserwise_water_heater.yaml
+++ b/custom_components/tuya_local/devices/geyserwise_water_heater.yaml
@@ -194,3 +194,15 @@ entities:
         optional: true
         unit: C
         class: measurement
+
+  - entity: sensor
+    name: Current temperature
+    class: temperature
+    category: diagnostic
+    dps:
+      - id: 10
+        type: integer
+        name: sensor
+        optional: true
+        unit: C
+        class: measurement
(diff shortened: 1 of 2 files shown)


make-all:
Duplicate sensors are not required.

RandMental:
The Current Temperature is an Attribute of the water.heater template.  A normal Gauge needs a sensor to display the temperature in this case.    To use the temperature attribute in a Gauge card we now need to create helper functions to define sensors from the attribute.   It is overall much simpler to add the current temperature as a sensor in the Geyserwise device file (e.g. in one place)

make-all:
It is not true that it needs to be a sensor to display in the UI. Most cards will allow you to display any attribute of an entity, though some may require manual yaml config to achieve that. If the card you are using does not allow that, it is a shortcoming of that card.

RandMental:
Thanks for the quick response.

We are using the Mushroom standard gauge card

RandMental:
Hi Jason

We are trying to use the standard Lovelace Gauge, which only accepts the sensor entity – any advice would be helpful

make-all:
The standard HA gauge card also supports monitoring an attribute of an entity. It probably needs to be manually configured, as I said.

PR #3718, opened 2025-09-10
Add Insoma Water Timer Valve

trasbd:
### Summary

Added support for **Insoma Smart Water Timer Valve**.  
Product link: [Amazon](https://a.co/d/8ybD4vM)

Although this device is similar to other water timers, it did not work correctly with the existing `ble_water_valve.yaml` configuration.

### Why a new configuration was needed

The existing `ble_water_valve.yaml` did not fully support this model:

- Switch state was unreliable (did not always turn on or reflect the actual device state)  
- Irrigation time always reset to 0 minutes  
- Time values using AM/PM were not valid for this device  

### Insoma Smart Water Timer DP Mapping

| DP ID | Code                 | Type     | Unit | Min     | Max     | Step | Options                       |
|-------|----------------------|----------|------|---------|---------|------|-------------------------------|
| 1     | `switch`             | Boolean  | –    | –       | –       | –    | –                             |
| 7     | `battery_percentage` | Integer  | %    | 0       | 100     | 1    | –                             |
| 9     | `time_use`           | Integer  | s    | 0       | 2,592,000 | 1  | –                             |
| 10    | `weather_delay`      | Enum     | –    | –       | –       | –    | `cancel`, `24h`, `48h`, `72h` |
| 11    | `countdown`          | Integer  | s    | 0       | 86,400  | 1    | –                             |
| 12    | `work_state`         | Enum     | –    | –       | –       | –    | `auto`, `manual`, `idle`      |
| 13    | `smart_weather`      | Enum     | –    | –       | –       | –    | `sunny`, `cloudy`, `rainy`    |
| 15    | `use_time_one`       | Integer  | s    | 0       | 86,400  | 1    | –                             |

### DIFF
--- a/custom_components/tuya_local/devices/insoma_water_timer_valve.yaml
+++ b/custom_components/tuya_local/devices/insoma_water_timer_valve.yaml
@@ -1,0 +1,85 @@
+name: Insoma Water Timer Valve
+products:
+  - id: nxquc5lb
+    model: SOP10
+    manufacturer: Insoma
+entities:
+  - entity: valve
+    class: water
+    dps:
+      - id: 1
+        type: boolean
+        name: valve
+      - id: 13
+        type: string
+        name: weather
+        optional: true
+  - entity: sensor
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
+        unit: "%"
+  - entity: sensor
+    name: Operation
+    category: diagnostic
+    dps:
+      - id: 12
+        type: string
+        name: sensor
+  - entity: sensor
+    name: Accumulated use time
+    category: diagnostic
+    class: duration
+    dps:
+      - id: 9
+        type: integer
+        name: sensor
+        unit: s
+  - entity: sensor
+    name: Last use time
+    category: diagnostic
+    class: duration
+    dps:
+      - id: 15
+        type: integer
+        name: sensor
+        unit: s
+        optional: true
+  - entity: select
+    name: Weather delay
+    translation_key: timer
+    icon: "mdi:weather-cloudy-clock"
+    category: config
+    dps:
+      - id: 10
+        type: string
+        name: option
+        mapping:
+          - dps_val: cancel
+            value: cancel
+          - dps_val: "24h"
+            value: "24h"
+          - dps_val: "48h"
+            value: "48h"
+          - dps_val: "72h"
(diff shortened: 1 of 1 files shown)


make-all:
There is no reason for this config. It is perfectly well covered by the ble_water_valve config, and the product id is already even listed there. Regarding the specific points:
- Switch state was unreliable: you have not changed anything related to the valve entity.
- Irrigation time always reset to 0 minutes: there is nothing in the config that influences that, it seems to be a device issue
- Time values using AM/PM were not valid for this device: I am not sure what this means. There is no AM/PM in the config, it is a 24 hour timer. But if there is an issue with that, it needs to be addressed globally, not just in one config.

trasbd:
Maybe you should investigate why the code controlling optional DPs and the 24 hour timer cause these issues. At least I was trying to bandaid the problems instead of ignore.

make-all:
I did not realise the time format could be changed like this, as mine is defaulting to 24h format which works well. But I have found the setting now and tested it out. The UI for timers is not friendly when set to 12h time format, but it does work (with the confusing proviso that times less than 1 hour have to use 12:xx AM, and times longer than 11:59 need to move into the PM range). I don't see any failures to set a time, or resetting back to 0, so this must be an issue with the device itself having unwritten requirements for the times that can be set there.

I have filed a feature request with HA for a more suitable UI for relative times, as switching back to a number is not much of an improvement over the confusing 12h display, with the need to do calculations in your head and/or scale the times from seconds into minutes so they have that ugly fractional part. Switching your time format to 24h solves the display issue.

trasbd:
> I did not realise the time format could be changed like this, as mine is defaulting to 24h format which works well. But I have found the setting now and tested it out. The UI for timers is not friendly when set to 12h time format, but it does work (with the confusing proviso that times less than 1 hour have to use 12:xx AM, and times longer than 11:59 need to move into the PM range). I don't see any failures to set a time, or resetting back to 0, so this must be an issue with the device itself having unwritten requirements for the times that can be set there.
> 
> I have filed a feature request with HA for a more suitable UI for relative times, as switching back to a number is not much of an improvement over the confusing 12h display, with the need to do calculations in your head and/or scale the times from seconds into minutes so they have that ugly fractional part. Switching your time format to 24h solves the display issue.


This is still an issue that is solved with the separate configuration.

Screenshot_20260422_180107_Home Assistant

PR #3728, opened 2025-09-13
Add device: Vibrance Electric Fireplace

fender4645:
Add support for the Modern Ember Vibrance Electric Fireplace

Amazon Link: https://a.co/d/5HeLv4L

Creates the following entity types:

- Climate (control heat temperature; turn fan to low, medium, and high)
- Flame Brightness
- Flame Color
- Flame Speed
- Heat Mode (on/off)
- Sound Level
- Timer (30m, 1h, 2h, 3h, 4h, 5h)


custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+products:
+  - id: nabdubjo2qqnrxs6
+    manufacturer: Vibrance
+    model: XH-BG-15W/30WZKWIFI-10
+entities:
+  - entity: climate
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
```
make-all: `switch` has no meaning here. There is a separate entity below. You could rename it to `available` to make the whole climate entity appear as disabled when the main switch is off, otherwise just remove it from here.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+      - id: 2
+        name: temperature
+        type: integer
+        unit: C
```
make-all: Since `temperature_unit` is defined below, the unit should not be hardcoded here.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+        range:
+          min: 5
+          max: 35
+      - id: 3
+        name: current_temperature
+        type: integer
+        unit: C
+        readonly: true
```
make-all: Unit should also not be here, and readonly is not needed, as the current_temperature is inherently read-only.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+      - id: 13
+        name: temperature_unit
+        type: string
+        mapping:
+          - dps_val: c
+            value: C
+          - dps_val: f
+            value: F
+      - id: 14
+        name: temperature
```
make-all: You cannot use duplicate names for attributes. Instead, call this `temp_set_f` and use a `value_redirect: temp_set_f` in the mapping for temperature attribute above.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+        type: integer
+        unit: F
+        range:
+          min: 41
+          max: 95
+        mapping:
```
make-all: The conditional mapping belongs on the temperature attribute above.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+          - constraint: temperature_unit
+            conditions:
+              - dps_val: F
+      - id: 15
+        name: current_temperature
```
make-all: Similarly, this should be `temp_cur_f` and `value_redrect` used in a conditional mapping on `current_temperature` above to redirect here when temperature_unit is F

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+        mapping:
+          - constraint: temperature_unit
+            conditions:
+              - dps_val: F
+      - id: 5
+        name: preset_mode
+        type: string
+        mapping:
+          - dps_val: H0
+            value: "off"
```
make-all, line 46: Maybe "fan_only" rather than "off"?

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+          - dps_val: false
+            value: heat
+          - dps_val: true
+            value: "off"
+  - entity: switch
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+  - entity: switch
```
make-all: This is duplicating the climate hvac_mode, so not needed as a separate switch.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+          - dps_val: S0
+            value: "off"
+          - dps_val: S1
+            value: low
+          - dps_val: S2
+            value: medium
+          - dps_val: S3
+            value: high
+  - entity: select
+    name: timer
```
make-all: Use translation_key instead of name, and map the values to the same strings as original dps_val to use existing translations.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -0,0 +1,180 @@
...
+      - id: 103
+        name: switch
+        type: boolean
+        mapping:
+          - dps_val: false
+            value: true
+          - dps_val: true
+            value: false
+  - entity: select
+    name: flame_color
```
make-all: A light entity should be used to combine the color (named_color) and brightness into a single entity. It could even pull in the flame speed as the "effect". Unfortunately the color names need to match standard colors on the color wheel, so you need to use the in-between colors rather than two distinct color names.
````
  - entity: light
    translation_key: flame
    dps:
      - id: 101
        type: string
        name: named_color
        mapping:
          - dps_val: "1"
            value: orange
          - dps_val: "2"
            value: red
          - dps_val: "3"
            value: green
          - dps_val: "4"
            value: blue
          - dps_val: "5"
            value: orangered
          - dps_val: "6"
            value: yellow
          - dps_val: "7"
            value: violet
      - id: 102
        type: string
        name: brightness
        mapping:
          - dps_val: F1
            value: 51
          - dps_val: F2
            value: 102
          - dps_val: F3
            value: 153
          - dps_val: F4
            value: 204
          - dps_val: F5
            value: 255
      - id: 108
        type: string
        name: effect
        mapping:
          - dps_val: "1"
            value: slow
          - dps_val: "2"
            value: medium
          - dps_val: "3"
            value: fast
````

File: custom_components/tuya_local/devices/vibrance_fireplace.yaml [hunk 11]
```diff
@@ -0,0 +1,180 @@
...
+        type: string
+        mapping:
+          - dps_val: "1"
+            value: slow
+          - dps_val: "2"
+            value: medium
+          - dps_val: "3"
+            value: fast
+  - entity: select
+    name: sound_level
```
make-all: Names of entities should be human readable. Spaces rather then underscores, and first word (only) capitalized.

make-all: You could also use the following for a more usual volume slider control:
````
  - entity: number
    name: Volume
    dps:
      - id: 109
        type: string
        name: value
        unit: "%"
        range:
          min: 0
          max: 100
        mapping:
          - dps_val: S0
            value: 0
          - dps_val: S1
            value: 33
          - dps_val: S2
            value: 66
          - dps_val: S3
            value: 100
````

File: custom_components/tuya_local/devices/vibrance_fireplace.yaml [hunk 12]
```diff
@@ -0,0 +1,180 @@
+name: Vibrance Fireplace
+products:
+  - id: nabdubjo2qqnrxs6
+    manufacturer: Vibrance
```
make-all: I think manufacturer should be `Modern Ember` and `Vibrance` should be the model. The below can be `model_id`.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Makes sense. Removing.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Removing hard code and mapping to temperature_unit

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Removed

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Changed name and used the `value_redirect` in the climate code.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Moved mapping

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Removed

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Good catch. Updated.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Good idea. Yeah, I had kept them separate because I was trying to keep the colors matching the Tuya app. But this makes sense. Updated.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Great idea. Fixed.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
fender4645: Ah, didn't know `model_id` existed. Fixed.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
make-all: Did you forget to push the changes you made?

fender4645:
No, I was doing some final testing and realized that the timer switch isn’t working. It sets the timer but within a few seconds something is reverting back to “off”. I’m going to try and see if I can fix this weekend, otherwise I’ll just remove and re-submit.

--
Chris Kast
***@***.***

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -1,26 +1,29 @@
...
           min: 5
           max: 35
+        mapping:
+          - constraint: temperature_unit
+            conditions:
+              - dps_val: C
+          - constraint: temperature_unit
+            conditions:
+              - dps_val: F
+            value_redirect: temp_set_f
```
make-all: indentation on this line is incorrect

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -1,26 +1,29 @@
...
         name: temperature
         type: integer
-        unit: C
         range:
           min: 5
           max: 35
+        mapping:
+          - constraint: temperature_unit
+            conditions:
+              - dps_val: C
```
make-all: The three lines above are conflicting with the next ones below, and do not do anything, so should just be removed.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -30,21 +33,14 @@ entities:
...
-        mapping:
-          - constraint: temperature_unit
-            conditions:
-              - dps_val: F
       - id: 15
         name: current_temperature
         type: integer
-        unit: F
-        readonly: true
         mapping:
```
make-all: This mapping should be removed, as well as renaming the dp above.

custom_components/tuya_local/devices/vibrance_fireplace.yaml
```diff
@@ -69,30 +65,15 @@ entities:
...
           - dps_val: true
             value: "off"
-  - entity: switch
+  - entity: light
+    translation_key: flame
     dps:
       - id: 1
-        name: switch
         type: boolean
-  - entity: switch
```
make-all: This seems to have been lost now.

fender4645: Can you elaborate on what you mean by "lost"?

make-all: The latest commit seems to have removed that switch entity.

make-all: Oh, I see now, it has been converted to hvac_mode in the climate entity. That makes sense.

Issue #104, opened 2022-01-03
Request to support thermostat Hysen HY08WE-2 WIFI- black


mpetcuRO:
Hello, I have the following thermostat for Floor heating - [link](http://www.cn-hysen.com/productinfo/465729.html)
I have followed the info on how to configure it with this integration and is accepting it as an "anderson_gsh_heater" no other similar device in the list.
![image](https://user-images.githubusercontent.com/73108927/147945288-ab149f0d-7aa5-4ab0-8aae-debef0b784b8.png)
After that it shows in HA but the update is slow and the temperatures is adding one "0"
Actual temp, current temp is 19, set temp is 9
![image](https://user-images.githubusercontent.com/73108927/147945506-f9e13b21-6ad1-4c83-b83a-917811c48ad8.png)
in Lovelace the thermostat entity is slow to react and showing like this.
![image](https://user-images.githubusercontent.com/73108927/147945847-bdd2b903-3f1d-45fd-9914-c3f732548f57.png)
What do I need to get from the Tuya developer account to continue troubleshooting.

mpetcuRO:
I have found the following in the Tuya platform under "Get the specifications and properties of the device"
`{
  "result": {
    "category": "wk",
    "functions": [
      {
        "code": "switch",
        "desc": "{}",
        "name": "开关",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "desc": "{\"unit\":\"℃\",\"min\":5,\"max\":1220,\"scale\":1,\"step\":1}",
        "name": "温度设置",
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":5,\"max\":1220,\"scale\":1,\"step\":1}"
      },
      {
        "code": "child_lock",
        "desc": "{}",
        "name": "童锁",
        "type": "Boolean",
        "values": "{}"
      }
    ],
    "status": [
      {
        "code": "switch",
        "name": "开关",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "name": "温度设置",
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":5,\"max\":1220,\"scale\":1,\"step\":1}"
      },
      {
        "code": "temp_current",
        "name": "当前温度",
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":0,\"max\":1580,\"scale\":1,\"step\":1}"
      },
      {
[...]

PR #3736, opened 2025-09-14
Herschel XLS T-MT Thermostat integration

RichardL64:
Herschel IR Heater integration for their mains T-MT thermostat.
Working for the last 3 months on my home configuration.

Has all functionality except the thermostats internal schedule mode.

Product link:
https://www.herschel-infrared.co.uk/product/t-mt-wifi-thermostat

Dev detail here:
https://github.com/RichardL64/HomeAssistant/tree/main

### DIFF
--- a/custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
+++ b/custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
@@ -40,14 +40,14 @@ entities:
         type: integer
         mapping:
           - scale: 10
-     - id: 104
-       name: temperature_unit
-       type: string
-       optional: true
-       mapping:
-         - dps_val: f
-           value: F
-         - value: C
+      - id: 104
+        name: temperature_unit
+        type: string
+        optional: true
+        mapping:
+          - dps_val: f
+            value: F
+          - value: C
       - id: 111
         name: hvac_action
         type: boolean
--- a/custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
+++ b/custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
@@ -12,44 +12,48 @@ entities:
         type: boolean
         mapping:
           - dps_val: true
-            value: "heat"
+            value: heat
           - dps_val: false
             value: "off"
       - id: 2
         name: temperature
         type: integer
-        constraint: temperature_unit
-        conditions:
-          - dps_val: "c"
-            range:
-              min: 160
-              max: 300
-          - dps_val: "f"
-            range:
-              min: 410
-              max: 1040
+        range:
+          min: 160
+          max: 1040
         mapping:
           - scale: 10
             step: 5
+            constraint: temperature_unit
+            conditions:
+              - dps_val: c
+                range:
+                  min: 160
+                  max: 300
+              - dps_val: f
+                step: 10
+                range:
+                  min: 410
+                  max: 1040
       - id: 3
         name: current_temperature
         type: integer
         mapping:
           - scale: 10
-#      - id: 104
-#        name: temperature_unit
-#        type: string
-#        mapping:
-#          - dps_val: "c"
-#            value: "C"
-#          - dps_val: "f"
-#            value: "F"
+     - id: 104
+       name: temperature_unit
+       type: string
+       optional: true
+       mapping:
+         - dps_val: f
+           value: F
+         - value: C
       - id: 111
         name: hvac_action
         type: boolean
         mapping:
           - dps_val: true
-            value: "heating"
+            value: heating
           - dps_val: false
             value: "off"
       - id: 4
@@ -67,9 +71,8 @@ entities:
             translation_key: program
 
   - entity: lock
-    name: Child lock
-    category: config
     translation_key: child_lock
+    category: config
     dps:
       - id: 6
         name: lock
@@ -89,13 +92,27 @@ entities:
           max: 99
 
   - entity: switch
-    name: Open window technology
+    name: Open window detection
     category: config
     icon: "mdi:window-open"
     dps:
       - id: 103
         name: switch
         type: boolean
+
+  - entity: select
+    translation_key: temperature_unit
+    category: config
+    dps:
+      - id: 104
+        type: string
+        optional: true
+        name: option
+        mapping:
+          - dps_val: c
+            value: celsius
+          - dps_val: f
+            value: fahrenheit
 
   - entity: number
     name: Temperature calibration
--- a/custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
+++ b/custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
@@ -18,9 +18,16 @@ entities:
       - id: 2
         name: temperature
         type: integer
-        range:
-          min: 0
-          max: 370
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
```diff
@@ -0,0 +1,111 @@
...
+          max: 370
+        mapping:
+          - scale: 10
+            step: 5
+      - id: 3
+        name: current_temperature
+        type: integer
+        mapping:
+          - scale: 10
+      - id: 104
```
make-all, line 43: Changing the temperature_unit to F will also require at least a change to the temperature range.

Different devices work differently, some use separate dps ids for C and F, others reuse the DP with different units, others only change the display on the device, but continue to use C over the local protocol, so it may be safer to leave this out of the climate entity if you are not sure how it will affect things.

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - id: 111
+        name: hvac_action
+        type: boolean
+        mapping:
+            - dps_val: true
+              value: "heating"
+            - dps_val: false
+              value: "off"
+
+  - entity: select
```
make-all: This should be preset_mode under the climate entity rather than a separate select entity. It can then use the existing translations for thermostats: manual, away and program

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
```diff
@@ -0,0 +1,111 @@
...
+        mapping:
+          - dps_val: manual
+            value: Manual
+          - dps_val: holiday
+            value: Holiday
+          - dps_val: auto
+            value: Auto
+        
+  - entity: switch
+    name: Child lock
```
make-all: Use a lock entity with `translation_key: child_lock` instead

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
```diff
@@ -0,0 +1,111 @@
...
+    category: config
+    dps:
+      - id: 6
+        name: switch
+        type: boolean
+        
+  - entity: number
+    name: Holiday days
+    category: config
```
make-all, line 80: `class: duration`

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
```diff
@@ -0,0 +1,111 @@
...
+    dps:
+      - id: 101
+        name: value
+        type: integer
```
make-all, line 85: `unit: d`

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
```diff
@@ -0,0 +1,111 @@
...
+  - entity: switch
+    name: Child lock
+    category: config
+    dps:
+      - id: 6
+        name: switch
+        type: boolean
+        
+  - entity: number
+    name: Holiday days
```
make-all: Holiday length might be better, as longer periods may be shown as weeks or months in some contexts.

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
```diff
@@ -0,0 +1,111 @@
...
+    dps:
+      - id: 101
+        name: value
+        type: integer
+        range:
+          min: 0
+          max: 99
+          
+  - entity: switch
+    name: Open window technology
```
make-all: Open window detection

Add `icon: "mdi:window-open"`

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
```diff
@@ -0,0 +1,111 @@
...
+    category: config
+    dps:
+      - id: 103
+        name: switch
+        type: boolean
+
+  - entity: number
+    name: Calibration
```
make-all: Temperature calibration

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
```diff
@@ -0,0 +1,111 @@
...
+    category: config
+    dps:
+      - id: 106
+        name: value
+        type: integer
```
make-all, line 120: `unit: °`

make-all: Use the degree sign alone here, as C or F will be subject to conversion in the UI, which is incorrect for relative temperatures.

RichardL64: Agreed - done

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
RichardL64: Agreed - done

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
RichardL64: Agreed - done

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
RichardL64: Agreed - done

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
RichardL64: Agreed - done

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
RichardL64: Agreed - done

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
RichardL64: Agreed - done

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
RichardL64: Agreed - done

custom_components/tuya_local/devices/herschel_t_mt_thermostat.yaml
RichardL64: Looks like it always keeps the set temperature dps #2, and the current temp in dps #3 (x10)
'c' & 'f' in 104
I guess the right way to implement that is constraints and conditions - I can see how that works referring back to #104 to modify the range in #2

The challenge is if the C/F is set on the device - it conflicts with HA which seems to set it centrally for the whole HA user environment. 

Not sure how to handle that.

RichardL64: Aside - is there a way to restart the tuya integration only, without rebooting the whole HA environment?
To reduce test cycle time. (scrub that - I just found the 'reload' option Ha!

PR #3429, opened 2025-06-28
Adding support for Adlar Castra / SolarEast heat pump

paulklinkenberg:
Here's the device config for the "Adlar Castra Aurora II heat pump".
This machine comes in 3 flavors: 4kwh, 6kwh, 10kwh, which in theory should all be supported now, except I do not have all their Tuya product IDs. The entities should be complete though.

### Tested
I tested this on my HASS installation, by:
* adding/editing the yaml file in /homeassistant/custom_components/tuya_local/devices/adlar_castra_heatpump.yaml
* removing the heat pump from tuya-local
* reloading tuya-local
* re-adding the heat pump
* after every attempt, checked the tuya-local logs for errors.

### References
* Original SolarEast service manual: https://www.climadirect.com/documenten/r32-heat-pump-for-heating-cooling-dhw-new.pdf
* Manual installation steps for Local-tuya (tuya-local much better!): https://gathering.tweakers.net/forum/view_message/77626436
* Entity IDs: https://gathering.tweakers.net/forum/view_message/80552878

### ToDos / Improvements
* Add translation labels for all fields

### Example image
adlar castra heatpump tuyalocal example

Issue #3693, opened 2025-09-01
Request support for Adlar Aurora II Air Source Heat Pump


m-hot:
### Log message

```text
Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:433
integration: Tuya Local (documentation, issues)
First occurred: 21:04:46 (2 occurrences)
Last logged: 22:52:07

Device matches manta_windy_MT0200B with quality of 22%. DPS: {"updated_at": 1756757084.7168105, "1": true, "2": "heating", "4": 33, "5": "Normal", "10": 0, "11": "OFF", "13": "L1", "14": 429, "15": 0, "16": 250, "18": 491, "20": 0, "21": 33, "22": 33, "23": 15, "24": 30, "25": 0, "26": 16, "27": true, "31": false, "33": false, "35": 14, "36": 15, "37": 24, "38": 15, "39": 14, "40": 0, "41": 18, "101": 20, "102": 562, "103": 2368, "104": 289, "105": 59971, "106": 1, "107": 23, "108": 25}
Device matches manta_windy_MT0200B with quality of 22%. DPS: {"updated_at": 1756763525.9332004, "1": true, "2": "heating", "4": 33, "5": "Normal", "10": 0, "11": "OFF", "13": "L1", "14": 429, "15": 0, "16": 250, "18": 539, "20": 0, "21": 29, "22": 28, "23": 14, "24": 22, "25": 0, "26": 15, "27": true, "31": false, "33": false, "35": 14, "36": 14, "37": 21, "38": 15, "39": 14, "40": 0, "41": 20, "101": 20, "102": 570, "103": 2395, "104": 295, "105": 60019, "106": 1, "107": 20, "108": 20}
```

### Product ID

g9m7honkxjweukvt

### Product Name

Adlar Aurora II

### DPS information

```text
{
  "result": {
    "model": "{\"modelId\":\"000004u5nz\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"默认服务\",\"properties\":[
{\"abilityId\":1,\"accessMode\":\"rw\",\"code\":\"switch\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-dp_power2\",\"attribute\":\"1152\"},\"name\":\"开关\",\"typeSpec\":{\"type\":\"bool\"}},
[...]

PR #3751, opened 2025-09-18
SolarEast/Adlar heatpump change, and small issues

paulklinkenberg:
See #3429 for the original request. The device was added, and with many improvements: Thanks!
There are a few things I experienced within HASS for this device:
1. The climate entity does not have a valid "current temperature" value;
2. The climate entity has a constantly changing "target temperature"; (should target temp. and current temp. be switched?)
3. The handy, but a bit redundant, "daily power usage" data seems to be unavailable in hass. In the yaml, it can be found at https://github.com/paulklinkenberg/tuya-local/blob/feature/adlar_heatpump_improvements/custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml#L388-L393
4. The mode "dry" was used instead of "heat". I made a code change for this in this PR. I read in the comments, you added the Dry value because of a missing distinction between "heat"  and "floor_heat" in the HASS Climate entity. I think it would be less confusing to then choose "heat_cool" for floor_heat, and "heat" for heating. 
Background: This heat pump has mode "heat" and "floor_heat", but both actually do the same: air to water heating. Only diff is the use of their own set of heat curves: HH1-8 is for "heat", HL1-8 is for "floor_heat" (with max. temp of 38*C). When I change the mode from Heat to Floor_heat or vice versa, the heat curve value will change as well.

I hope you can look into this. I am available for debugging or any other questions.

HASS card config for Climate entity, containing missing data

### DIFF
--- a/custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml
+++ b/custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml
@@ -38,7 +38,7 @@ entities:
           - dps_val: heating
             value: heat
           - dps_val: floor_heating
-            value: floor_heat
+            value: heat_cool
           - dps_val: hot_water
             value: hotwater
           - dps_val: cold_and_hot_water
@@ -63,12 +63,12 @@ entities:
                 value: cool
                 available: wth_off
               # Since there are both heating and floor_heating, model
-              # the air heating as dry, and floor_heating as heat
+              # the regular (radiator) heating as heat, and floor_heating as heat_cool
               - dps_val: heating
-                value: dry
+                value: heat
                 available: wth_off
               - dps_val: floor_heating
-                value: heat
+                value: heat_cool
                 available: wth_off
               - dps_val: hot_water
                 value: "off"
@@ -88,6 +88,7 @@ entities:
         hidden: true
       - id: 2
         type: string
+        # "wth_off" true = the heatpump does not supply hot tap water
         name: wth_off
         mapping:
           - dps_val: hot_water
@@ -101,6 +102,7 @@ entities:
           - value: true
       - id: 2
         type: string
+        # "wth_on" true = the heatpump does supply hot tap water
         name: wth_on
         mapping:
           - dps_val: hot_water
@@ -112,7 +114,7 @@ entities:
           - dps_val: floor_heating_and_hot_water
             value: true
           - value: false
-      - id: 4
+      - id: 4 # temp_set
         name: temperature
         type: integer
         unit: C
@@ -132,6 +134,7 @@ entities:
       # Inlet temperature used here, as ambient is often the outdoor unit
       # I don't see an indoor_temp, but if ambient is, then it may be more
       # suitable to use here
+      # PK 20250918: ambient is indeed outdoor temp. There is no indoor temp sensor like from a connected thermostat.
       - id: 21
         name: temperature
         type: integer
@@ -382,7 +385,7 @@ entities:
         mapping:
           - scale: 100
         class: total_increasing
-      - id: 18
+      - id: 18 # It seems this entity is unreachable from within HASS
         name: today
         type: integer
         unit: kWh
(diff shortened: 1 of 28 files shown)


custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml
```diff
@@ -38,7 +38,7 @@ entities:
           - dps_val: heating
             value: heat
           - dps_val: floor_heating
-            value: floor_heat
+            value: heat_cool
```
make-all: I don't think you should change this here. There is no restriction on what the values in a select should be, so they should be correct, even if it does not directly match the climate hvac_mode.

custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml
```diff
@@ -63,12 +63,12 @@ entities:
                 value: cool
                 available: wth_off
               # Since there are both heating and floor_heating, model
-              # the air heating as dry, and floor_heating as heat
+              # the regular (radiator) heating as heat, and floor_heating as heat_cool
               - dps_val: heating
-                value: dry
+                value: heat
```
make-all, line 69: You should also change lines 80 and 83 to match these changes for the water heating on case.

custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml
```diff
@@ -132,6 +134,7 @@ entities:
       # Inlet temperature used here, as ambient is often the outdoor unit
       # I don't see an indoor_temp, but if ambient is, then it may be more
       # suitable to use here
+      # PK 20250918: ambient is indeed outdoor temp. There is no indoor temp sensor like from a connected thermostat.
       - id: 21
         name: temperature
```
make-all: The name here should be current_temperature I think (it is the sensor being discussed in the above comment, and there is already a temperature dp above - I am not sure how this slipped through the tests, I thought there was a duplicate dp name test to catch these).

make-all:
It would also be useful to add `optional: true` to dp ids 109, 111 and 112 to improve detection as noted in #3693

custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml
m-hot: > The name here should be current_temperature I think (it is the sensor being discussed in the above comment, and there is already a temperature dp above - I am not sure how this slipped through the tests, I thought there was a duplicate dp name test to catch these).

I'd stupidly missed all your hard work and built my own YAML based off the names of everything translated from the chinese (and what was surfaced in the Smart Life app from Adlar). here is what I'd created:
  - entity: sensor
    category: diagnostic
    name: "Outlet Water Temperature"
    class: temperature
    dps:
      - id: 21
        type: integer
        name: sensor
        unit: "°C"
        readonly: true
        range:
          min: -50
          max: 150

I'll add my YAML file to this ticket - if any of it can be used, please feel free.

m-hot:
> It would also be useful to add `optional: true` to dp ids 109, 111 and 112 to improve detection as noted in #3693

Yes please to this - I can confirm my 14KW Adlar Aurora II ASHP was successfully detected when I manually changed these items.

custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml
paulklinkenberg: Reverted this. I get the idea, and agree.

custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml
paulklinkenberg: Sharp! Thx, done.

custom_components/tuya_local/devices/solareast_aurora2_heatpump.yaml
paulklinkenberg: I changed the name now!

paulklinkenberg:
> It would also be useful to add `optional: true` to dp ids 109, 111 and 112 to improve detection as noted in #3693

Did this now, at the dps level for the 3 entities. I hope that was the correct place...

paulklinkenberg:
For completeness sake, I can confirm the Climate entity is now correctly working with these changes, see img attached. I tested this by overwriting the old implementation with the new one, in my local HA instance:
afbeelding

PR #3756, opened 2025-09-19
New Devices: Royal Clima RCS-1000-P 3.0 ventilation system

zpk2006:
An attempt to provide access to all found sensors using reverse engineering of the official application. Localization has been implemented for entities found in the tuya-local repo. Most likely, the configuration should be compatible with all RCS 3.0 series devices.
Screen1
Screen2

### DIFF
--- a/custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
+++ b/custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
@@ -1,0 +1,280 @@
+name: Energy Recovery Ventilator 
+products:
+  - id: ics2husbxx7n6wz5
+    manufacturer: Royal Clima
+    model: RCS-1000-P 3.0
+    name: Supply and exhaust ventilation system
+entities:
+  - entity: fan
+    translation_key: ventilation
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 2
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: "0"
+            value: heat_recovery
+          - dps_val: "1"
+            value: fresh
+          - dps_val: "2"
+            value: auto
+          - dps_val: "3"
+            value: sleep
+  - entity: sensor
+    translation_key: status
+    class: enum
+    dps:
+      - id: 2
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: "0"
+            value: Recuperation
+            icon: "mdi:shuffle-variant"
+          - dps_val: "1"
+            value: Bypass
+            icon: "mdi:weather-windy"
+          - dps_val: "2"
+            value: auto
+            icon: "mdi:arrow-decision-auto"
+          - dps_val: "3"
+            value: sleep
+            icon: "mdi:power-sleep"
+  - entity: number
+    name: Supply fan
+    icon: "mdi:tray-arrow-down"
+    dps:
+      - id: 103
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 10
+  - entity: number
+    name: Exhaust fan
+    icon: "mdi:tray-arrow-up"
+    dps:
+      - id: 104
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 10
+  - entity: sensor
+    name: CO2
+    icon: "mdi:molecule-co2"
+    dps:
+      - id: 6
+        type: integer
+        name: sensor
+        unit: ppm
+        class: measurement
+  - entity: sensor
+    name: Humidity
+    icon: "mdi:water-percent"
+    dps:
+      - id: 8
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    name: Indoor temperature
+    class: temperature
+    icon: "mdi:home-thermometer"
+    dps:
+      - id: 9
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+  - entity: sensor
+    name: Filter life
+    icon: "mdi:air-filter"
+    dps:
+      - id: 11
+        type: integer
+        name: sensor
+        unit: "Days"
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 14
+        type: boolean
+        name: lock
+        mapping:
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+    name: CO2
+    icon: "mdi:molecule-co2"
+    dps:
+      - id: 6
+        type: integer
+        name: sensor
+        unit: ppm
+        class: measurement
+  - entity: sensor
+    name: Humidity
```
make-all, line 76: Use `class: humidity` and no explicit name or icon

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+    icon: "mdi:tray-arrow-up"
+    dps:
+      - id: 104
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 10
+  - entity: sensor
+    name: CO2
```
make-all, line 67: Use `class: carbon_dioxide` instead of explicit name and icon

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+      - id: 118
+        type: integer
+        name: value
+        unit: C
+        class: measurement
+        range:
+          min: 2
+          max: 15
+  - entity: sensor
+    name: Status text
```
make-all, line 255: This can be an attribute on the status sensor above

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+  - entity: number
+    name: Bypass open range
+    icon: "mdi:thermometer-lines"
+    category: config
+    class: temperature
+    dps:
+      - id: 118
+        type: integer
+        name: value
+        unit: C
```
make-all, line 249: If this is a relative or delta temperature, then use `°` to avoid conversion when the user preferences are set to Fahrenheit.

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+      - id: 112
+        type: boolean
+        name: switch
+        mapping:
+          - dps_val: false
+            icon: "mdi:hand-pointing-up"
+          - dps_val: true
+            icon: "mdi:auto-mode"
+  - entity: sensor
+    name: Bypass
```
make-all, line 198: This should be a binary_sensor, without mapping of boolean values to strings. A `class: opening` is probably appropriate.

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+        range:
+          min: -9
+          max: 5
+  - entity: sensor
+    name: Heater configured
+    category: diagnostic
+    dps:
+      - id: 111
+        type: boolean
+        name: sensor
```
make-all, line 184: This should probably be used elsewhere as an availability flag, rather than a standalone sensor.

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+    dps:
+      - id: 109
+        type: integer
+        name: value
+        unit: ppm
+        range:
+          min: 800
+          max: 2000
+        mapping:
+          - scale: 1
```
make-all, line 162: `scale: 1` is redundant

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+        name: sensor
+        unit: "Days"
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 14
+        type: boolean
+        name: lock
+        mapping:
```
make-all, line 109: There is no need to override the icons here.

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+        unit: C
+        class: measurement
+  - entity: sensor
+    name: Filter life
+    icon: "mdi:air-filter"
+    dps:
+      - id: 11
+        type: integer
+        name: sensor
+        unit: "Days"
```
make-all, line 101: Unit should be `d`, and the sensor should have a `class: duration`.

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+    class: temperature
+    icon: "mdi:home-thermometer-outline"
+    dps:
+      - id: 22
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+  - entity: number
+    name: Humidity threshold to turn on
```
make-all, line 125: The name is a bit wordy. Just "Humidity threshold" should be enough.

`class: humidity` should also be added in place of the icon.

custom_components/tuya_local/devices/royalclima_rcs1000p_ventilation.yaml
```diff
@@ -0,0 +1,279 @@
...
+        name: button
+  - entity: button
+    translation_key: filter_reset
+    category: config
+    dps:
+      - id: 108
+        type: boolean
+        name: button
+  - entity: number
+    name: CO2 threshold to turn on
```
make-all, line 150: The name is a bit wordy, just CO2 threshold should be enough.

`class: carbon_dioxide` should also be added in place of the icon.

Issue #507, opened 2023-03-07
Request support for Ceres Plus smart pet feeder with camera


nyok92:
# Log Message

```
2023-03-07 14:41:23.118 DEBUG (MainThread) [custom_components.tuya_local.device] Refreshing device state for Test
2023-03-07 14:41:23.119 INFO (MainThread) [custom_components.tuya_local.device] Setting protocol version for Test to 3.3
2023-03-07 14:41:23.124 DEBUG (SyncWorker_2) [tinytuya.core] status() entry (dev_type is default)
2023-03-07 14:41:23.124 DEBUG (SyncWorker_2) [tinytuya.core] building command 10 payload=b'{"gwId":"bf0a4a3a4eb3586d1czint","devId":"bf0a4a3a4eb3586d1czint","uid":"bf0a4a3a4eb3586d1czint","t":"1678196483"}'
2023-03-07 14:41:23.136 DEBUG (SyncWorker_2) [tinytuya.core] sending payload
2023-03-07 14:41:23.137 DEBUG (SyncWorker_2) [tinytuya.core] payload encrypted=b'000055aa000000010000000a0000008803b59a76f9870b071fed34c6e3d69163309557ff089ec0b5b23090c607b2455f4156c34e7e404a5b1f6a667457b91611461325400e9d8942419a1f500e7bf51dc99a8822e73739ae233cf09d95d4f>
2023-03-07 14:41:23.149 DEBUG (SyncWorker_2) [tinytuya.core] received data=b'000055aa000000010000000a0000002c00000001663b112c52e9b13db8b682ad720888dfe5add1416bdeca84d639f6cd667823b86507f2700000aa55'
[...]

PR #3767, opened 2025-09-25
Update ceres_plus_petfeeder.yaml

nyok92:
Update of the Ceres plus petfeeder config


make-all:
This is quite a major change. Some of it appears to just be reordering of entities, which adds to the noise and makes it difficult to find the actual changes amongst it.

At least one dp has had its type changed, dp 246 has switched from string (as documented in #507 and most likely correct) to integer, although #507 is a bit confused about the purpose of the dp.

Changes that affect detection like that will need some evidence with them to convince me that it is a good change. Also other changes that affect entity naming need good justification, as that will affect other users who have the config installed also, so just "I like this name better" is not likely to be enough.

make-all:
Closed due to no feedback in 2 months

PR #3774, opened 2025-09-27
Feature: add an option to specify port

bbronisz:
This PR adds support to specify custom port (optional) on which the device is available.
I needed this for my own setup, probably not needed in near future, but it worked so maybe somebody else will find it useful.

I am not Python developer, so it was: reading, copying, tweaking, testing. I had issue with `voluptuous.Optional` not liking that the optional value was `int`. Fixed it with `vol.Maybe(int)` but I don't know if that's correct way.

Translations were generate by VS Code, with the exception of English and Polish.


make-all:
I don't really understand under what conditions this is needed. The ports used by Tuya devices are fixed as far as I know. If this is for some industrial use case, then I don't think it is appropriate for this integration.

bbronisz:
I have Home Assistant installed on Raspberry Pi connected to ISP provided router that has weak WiFi which cannot be extended, doesn't support any mesh technologies. So I have TP-Link Deco M5 that allow me to have better WiFi coverage.
But this way I don't have direct access from Home Assistant to devices connected to Deco network. Fortunately Deco allows port forwarding and TinyTuya also supports specifying ports for devices so I've managed to have it working.
I wouldn't call it industrial use. It would expose option available in TinyTuya to Tuya Local users.

make-all:
It seems like an overcomplex installation scenario which is not worth adding additional configuration complexity for all users to accommodate. It also potentially encourages dangerous practices such as exposing remote devices via NAT rather than the safer option of using a VPN that will not require such a workaround.

So I am going to reject this change.

bbronisz:
OK.

PR #3777, opened 2025-09-28
Add Apricus Heat Pump Water Heater

admackin:
Device: [Apricus Heat Pump Water Heater](https://apricus.com.au/all-in-one-heat-pump/)

Have verified this locally: 
image

### DIFF
--- a/custom_components/tuya_local/devices/apricus_heat_pump_water_heater.yaml
+++ b/custom_components/tuya_local/devices/apricus_heat_pump_water_heater.yaml
@@ -1,15 +1,24 @@
-name: Apricus Water Heater
+name: Apricus Heat Pump Water Heater
 entities:
-  - entity: climate
+  - entity: water_heater
     dps:
       - id: 1
-        name: hvac_mode
+        name: operation_mode
         type: boolean
         mapping:
-          - dps_val: true
-            value: heat
           - dps_val: false
             value: "off"
+          - dps_val: true
+            constraint: work_mode
+            conditions:
+              - dps_val: AUTO
+                value: Auto
+              - dps_val: ECO
+                value: Eco
+              - dps_val: BOOST
+                value: Boost
+              - dps_val: ELE
+                value: Element            
       - id: 2
         name: temperature
         type: integer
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/apricus_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,48 @@
+name: Apricus Water Heater
+entities:
+  - entity: climate
```
make-all: water_heater would be better for a heating-only device

custom_components/tuya_local/devices/apricus_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,48 @@
...
+          - dps_val: AUTO
+            value: Auto
+          - dps_val: ECO
+            value: Eco
+          - dps_val: BOOST
+            value: Boost
+          - dps_val: ELE
+            value: Element
+  - entity: sensor
+    name: Power Consumption
```
make-all: name not needed here

custom_components/tuya_local/devices/apricus_heat_pump_water_heater.yaml
```diff
@@ -1,15 +1,24 @@
...
         mapping:
-          - dps_val: true
-            value: heat
           - dps_val: false
             value: "off"
+          - dps_val: true
+            constraint: work_mode
+            conditions:
+              - dps_val: AUTO
+                value: Auto
```
make-all: The modes for operation_mode are not free-form (the user level documentation at https://www.home-assistant.io/integrations/water_heater/ makes them sound like the integration can define them as needed, but developer docs at https://developers.home-assistant.io/docs/core/entity/water-heater/ makes the statement *The allowed operation modes are the states specified in the base component and implementations of the water_heater component cannot differ.*). I would map these to `heatpump`, `eco`, `performance`, `electric` respectively.

admackin: (well it works *ok* with free-form values, but you don't get icons). Anyway I've updated, although `high_demand` seemed a better fit that `performance`

make-all (approved):

PR #3782, opened 2025-09-30
Create neopower_heat_pump_water_heater.yaml

banemonster:
Modified from the Aquatech RAPID/X6 device which partially worked, however had some invalid modes like HYB1/Boost. DP ID's changed to suit the neopower heat pump. Successfully working in Home Assistant.
[https://neopower.com.au/all-in-one-heat-pump/](https://neopower.com.au/all-in-one-heat-pump/)

Additional info like ele_heating_state & compressor_state are listed as attributes in the Problem area, but are still exposed to home assistant so can be used for automations. I'm not sure how to make these appear as sensors
Screenshot 2025-09-30 144652

brave_i5ModPvMZR

Full list of DP ID's are here
[https://github.com/banemonster/neopower_heatpump_tuya/blob/main/neopower_DP_IDs](https://github.com/banemonster/neopower_heatpump_tuya/blob/main/neopower_DP_IDs)

### DIFF
--- a/custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
+++ b/custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
@@ -1,4 +1,8 @@
-name: Neopower All in One Heat Pump
+name: Water heat pump
+products:
+  - id: opugd8qgpbqbcizu
+    manufacturer: Neopower
+    model: Black Diamond all-in-one
 entities:
   - entity: water_heater
     dps:
@@ -30,12 +34,16 @@ entities:
         range:
           min: 15
           max: 75
-      - id: 7
-        type: boolean
-        name: defrosting
       - id: 16
         type: integer
         name: current_temperature
+  - entity: binary_sensor
+    translation_key: defrost
+    category: diagnostic
+    dps:
+      - id: 7
+        type: boolean
+        name: sensor
   - entity: binary_sensor
     class: problem
     category: diagnostic
@@ -50,47 +58,63 @@ entities:
       - id: 15
         type: bitfield
         name: fault_code
+  - entity: binary_sensor
+    name: Element
+    class: running
+    category: diagnostic
+    icon: "mdi:heating-coil"
+    dps:
       - id: 32
         type: boolean
-        name: ele_heating_state
-        mapping:
-          - dps_val: false
-            value: "off"
-          - dps_val: true
-            value: "on"
+        name: sensor
+  - entity: binary_sensor
+    name: Compressor
+    class: running
+    category: diagnostic
+    dps:
       - id: 27
         type: boolean
-        name: compressor_state
-        mapping:
-          - dps_val: false
-            value: "off"
-          - dps_val: true
-            value: "on"
-      - id: 7
-        type: boolean
-        name: defrost_state
-        mapping:
-          - dps_val: false
-            value: "off"
-          - dps_val: true
-            value: "on"
+        name: sensor
+  - entity: sensor
+    name: Vent temperature
+    class: temperature
+    category: diagnostic
+    dps:
       - id: 24
         type: integer
-        name: venting_temp
+        name: sensor
+        unit: C
+  - entity: sensor
+    class: power
+    category: diagnostic
+    dps:
       - id: 18
         type: integer
-        name: power_consumption
+        name: sensor
+        unit: W
+  - entity: sensor
+    name: Top temperature
+    class: temperature
+    category: diagnostic
+    dps:
       - id: 21
         type: integer
-        name: temp_top
+        name: sensor
+        unit: C
+  - entity: sensor
+    name: Bottom temperature
+    class: temperature
+    category: diagnostic
+    dps:
       - id: 22
         type: integer
-        name: temp_bottom
+        name: sensor
+        unit: C
+  - entity: binary_sensor
+    class: opening
+    name: Four-way valve
+    category: diagnostic
+    dps:
       - id: 28
         type: boolean
-        name: four_valve_state
-        mapping:
-          - dps_val: false
-            value: "off"
-          - dps_val: true
-            value: "on"
+        name: sensor


custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,96 @@
...
+      - id: 4
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 15
+          max: 75
+      - id: 7
+        type: boolean
+        name: defrosting
```
make-all: Defrosting is not a standard part of the water_heater API, so adding it here will only show the current value under "Additional attributes".  From the manual, it seems like this should be a switch that is controllable by the user (in addition to automatically kicking in when needed).

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,96 @@
...
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: "on"
+      - id: 27
+        type: boolean
+        name: compressor_state
+        mapping:
+          - dps_val: false
+            value: "off"
```
make-all: It is better to leave the values as boolean.

This can be made into a binary_sensor with `class: running`

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,96 @@
...
+          - dps_val: true
+            value: "on"
+      - id: 7
```
make-all: This is duplicated from above

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,96 @@
...
+        type: boolean
+        name: defrost_state
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: "on"
+      - id: 24
```
make-all, line 83: Can be a sensor with `class: temperature`

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,96 @@
...
+        type: integer
+        name: venting_temp
+      - id: 18
```
make-all, line 91: Can be a sensor with `class: power` if this is a dynamic measurement. If it is a fixed value telling the max output of the device, then it may be better as an attribute on the water_heater rather than on the problem sensor. Since this is the only power sensor the device has, the sensor should be unnamed to let the default translation for power sensors be used.

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,96 @@
...
+        type: integer
+        name: power_consumption
+      - id: 21
```
make-all, line 100: can be a sensor with `class: temperature`

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,96 @@
...
+        type: integer
+        name: temp_top
+      - id: 22
```
make-all, line 109: Can be a sensor with `class: temperature`

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,96 @@
...
+        type: integer
+        name: temp_bottom
+      - id: 28
+        type: boolean
+        name: four_valve_state
```
make-all: Can be a binary_sensor with `class: opening`

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
```diff
@@ -0,0 +1,96 @@
+name: Neopower All in One Heat Pump
```
make-all: Top level name should be without branding. If you can find the product_id for this device (from the same platform.tuya.com API calls that return the local key, or from a local discovery scan by tinytuya etc), then a products section can be added with id, manufacturer and model details.

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
banemonster: Defrosting can be triggered via the control panel on the unit, however there does not appear to by anyway to control this from the Tuya app or Tuya IOT portal.

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
banemonster: This is what is returned for product_id in the same section I obtained the local_key


      "category": "znrb",

      "local_key": REDACTED

      "model": "",
      "name": "Heat Pump",
      "product_id": "opugd8qgpbqbcizu",

      "product_name": "家用热水机（四模式）",


"**家用热水机（四模式）**"translates to "**Household water heater (four modes)**"

make-all: I would make the top level name "Water heat pump" (or even "Water heater") - short names are preferred because in some places in the UI the device name gets added to the front of every entity, but it is best to keep "Water" in there to distinguish from air conditioning heat pumps. The naming that Tuya gives it is not as important as being clear to end users.

custom_components/tuya_local/devices/neopower_heat_pump_water_heater.yaml
make-all: So this should probably be a binary_sensor then, maybe with class: cold

PR #3785, opened 2025-09-30
feat (devices): add ELEGRP DTR10 Single Pole 2.4 GHz WiFi Dimmable Switch

davidmurdoch:
- Complete device configuration
- Timer system
- Physical interface settings (long press, indicator brightness)
- Tests

Supported features:
- Dimmable light control (DPS 1,2)
- Configurable minimum brightness with app auto-management (DPS 3,114)
- Bulb type selection for optimal dimming (DPS 4)
- Customizable fade on/off speeds (DPS 103,104)
- Long press brightness presets (DPS 108)
- LED indicator brightness control (DPS 101)
- Timer functionality with action selection (DPS 110,111,6)
- Firmware version reporting (DPS 113)

### DIFF
--- a/custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
+++ b/custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
@@ -98,14 +98,12 @@ entities:
             value: Slow (15s)
   # Indicator LED converted to light (brightness only)
   - entity: light
-    name: Night indicator light
+    translation_key: indicator
     category: config
-    icon: "mdi:led-on"
     dps:
       - id: 101
         type: integer
         name: brightness
-        unit: "%"
         range:
           min: 0
           max: 100
@@ -157,9 +155,8 @@ entities:
 
   # Timer
   - entity: time
-    name: Timer duration
+    translation_key: timer
     category: config
-    icon: "mdi:timer"
     dps:
       - id: 111
         type: integer
@@ -169,7 +166,7 @@ entities:
           max: 86399
         optional: true
   - entity: select
-    name: Timer auto on mode
+    name: Timer mode
     category: config
     icon: "mdi:timer-cog"
     dps:
@@ -183,8 +180,9 @@ entities:
           - dps_val: true
             value: "On"
   - entity: sensor
-    name: Time remaining
-    icon: "mdi:timer-sand"
+    translation_key: time_remaining
+    class: duration
+    category: diagnostic
     dps:
       - id: 6
         type: integer


davidmurdoch:
i'll try to fix the lint and tests tomorrow

make-all:
Delete the tests, there should be no need for device specific tests if you are not adding Python code level features.

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
```diff
@@ -0,0 +1,213 @@
+name: Dimmable switch
+products:
+  - id: elegrp_dtr10
```
make-all: This is not a product id

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
```diff
@@ -0,0 +1,213 @@
...
+        mapping:
+          - dps_val: 1000
+            value: "Max Brightness"
+          - dps_val: 258
+            value: "Nap 25%"
+          - dps_val: 505
+            value: "Cuddle 50%"
+          - dps_val: 753
+            value: "Leisure 75%"
+          - dps_val: 1000
```
make-all, line 61: There are two entries for 1000

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
```diff
@@ -0,0 +1,213 @@
...
+          - dps_val: "Immediate"
+            value: Immediate
+          - dps_val: "Fast"
+            value: Fast (1s)
+          - dps_val: "Medium"
+            value: Medium (5s)
+          - dps_val: "Slow"
+            value: Slow (15s)
+
+  - entity: number
```
make-all: make it a light entity with just brightness

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
```diff
@@ -0,0 +1,213 @@
...
+        type: integer
+        name: value
+        unit: "s"
+        range:
+          min: 0
+          max: 86400
+        optional: true
+      - id: 6
+        type: integer
+        name: value
```
make-all: you cannot have two "value" dps. In any case, a time entity would be better

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
```diff
@@ -0,0 +1,213 @@
...
+  - entity: sensor
+    name: Timer Remaining
+    icon: "mdi:timer-sand"
+    dps:
+      - id: 6
+        type: integer
+        name: sensor
+        unit: "s"
+        optional: true
+        readonly: true
```
make-all: readonly has no meaning on a sensor, sensors are always readonly

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
```diff
@@ -0,0 +1,213 @@
...
+
+  # Diagnostics
+  - entity: sensor
+    name: Firmware Version
```
make-all: does not need to be a sensor entity. It should never change, so can just be an attribute on another entity.

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
```diff
@@ -0,0 +1,213 @@
...
+          - dps_val: "Led"
+            value: LED
+          - dps_val: "Incandescent"
+            value: Incandescent
+          - dps_val: "Halogen"
+            value: Halogen
+
+  # Settings
+  - entity: select
+    name: 'Long Press "On"'
```
make-all: HA naming convention is to only capitalise the first word.

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
davidmurdoch: where might I find the id?

make-all: 1. In the tuya portal device info (same functions you can find local key)
2. By running `python -m tinytuya scan` on any PC on the same local network as the device.

davidmurdoch: A scan on the same network returns 0 devices (I have 9 devices working in home assistant using this elegrp config already).

I don't have a tuya portal set up. If I modify `diagnostics.py` in my HA is there a value it collects I can log to return the product id for use here?

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
davidmurdoch: resolved. Most of these decisions were an effort to copy the ELEGRP app's design language, which actually did implemented two entries for 1000.

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
davidmurdoch: It seems to have worked, but i've gone ahead and split them.

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
make-all: Unfortunately the product id cannot be queried after connection, it is only sent in discovery broadcasts. Some devices do not send discovery broadcasts when they are already connected, and it seems like this may be one. If you temporarily disable a device in HA, it might start responding to the scan.

custom_components/tuya_local/devices/elegrp_dtr10_dimmer_switch.yaml
```diff
@@ -0,0 +1,193 @@
...
+  # - entity: switch
+  #   name: Minimum brightness enabled
+  #   category: config
+  #   icon: "mdi:toggle-switch"
+  #   dps:
+  #     - id: 114
+  #       type: boolean
+  #       name: switch
+  #       optional: true
+  # Value can't be changed. It seems like it is hardcoded to 1000 (100%) by the
```
make-all, line 138: When value cannot be changed directly, sometimes these are affected by the light type setting (mainly minimum brightness). But since HA does not handle minimum brightness, it will not make any difference to include these even as read-only.

PR #3793, opened 2025-10-01
Add support for Mi-Heat TH11-WF thermostat

flow-c:
* Add config for new device
* Add translation_keys for "holiday_mode" and "eco_mode"
* Add default icons for said translation_keys

Basic functions work locally (without connection to the Tuya cloud): 
+ Setting target temperature
+ Setting a Boost timer
+ Switching between manual and program modes
+ Toggling of child-lock and anti-frost, eco, and holiday modes

Reading sensor data requires a cloud connection (as it does in the "Tuya - Smart  Life, Smart Living" app). Hence the following are unavailable locally:
- Current temperature and humidity
- Target temperature*
- Power / total energy readings
- Window open/closed state
- Status of any switches*

*)if changed on the device itself or through another client

### DIFF
--- a/custom_components/tuya_local/devices/miheat_th11wf_thermostat.yaml
+++ b/custom_components/tuya_local/devices/miheat_th11wf_thermostat.yaml
@@ -64,7 +64,7 @@ entities:
         name: sensor
         type: integer
         unit: "%"
-        class: meassurement
+        class: measurement
   - entity: lock
     translation_key: child_lock
     category: config


custom_components/tuya_local/translations/en.json
```diff
@@ -825,6 +825,12 @@
             "anti_frost": {
                 "name": "Anti-frost"
             },
+            "holiday_mode": {
+                "name": "Holiday mode"
+            },
+            "eco_mode": {
+                "name": "ECO mode"
+            },
```
make-all: These can be covered by the preset_mode of the climate entity, which is currently unused. HA already defines `away` and `eco` as standard preset modes.

custom_components/tuya_local/devices/mi-heat_th11-wf_thermostat.yaml
```diff
@@ -0,0 +1,188 @@
...
+          - scale: 10
+  - entity: sensor
+    name: Status
+    class: enum
+    category: diagnostic
+    dps:
+      - id: 11
+        type: string
+        name: sensor
+        class: measurement
```
make-all: A string sensor cannot be a measurement

custom_components/tuya_local/devices/mi-heat_th11-wf_thermostat.yaml
```diff
@@ -0,0 +1,188 @@
...
+    dps:
+      - id: 10
+        name: sensor
+        type: integer
+        unit: W
+        class: measurement
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: Status
```
make-all: There is `translation_key: status` that can be used instead

custom_components/tuya_local/devices/miheat_th11wf_thermostat.yaml
```diff
@@ -0,0 +1,188 @@
...
+    name: Status
+    class: enum
+    category: diagnostic
+    dps:
+      - id: 11
+        type: string
+        name: sensor
+        class: measurement
+        mapping:
+          - dps_val: 0
```
make-all, line 93: Type is defined as string, but your dps_val is an integer.

class: enum requires all possible values to be listed

custom_components/tuya_local/translations/en.json
flow-c: It would be nice to also cover Mode (manual/program) and Boost (e.g. 20min as the default) via preset_mode.

But I don't know how to implement this as the device does not have a dp for presets.

If there was such a thing as a dummy dp id so that when set in HA only the constraint dp is affected, then it would be something like this?

```
[…]
      - id: [dummy-id]
        name: preset_mode
        type: string
        mapping:
          - dps_val: [irrelevant]
            constraint: eco_mode
            conditions:
              - dps_val: true
                value: eco
          - dps_val: [irrelevant]
            constraint: holiday_mode
            conditions:
              - dps_val: true
                value: away
[…]

```

custom_components/tuya_local/devices/mi-heat_th11-wf_thermostat.yaml
flow-c: Understood: <s>class: measurement</s>

custom_components/tuya_local/devices/miheat_th11wf_thermostat.yaml
flow-c: Noted. I've changed it to `type: bitfield`

custom_components/tuya_local/devices/mi-heat_th11-wf_thermostat.yaml
flow-c: Thanks. Done.

custom_components/tuya_local/translations/en.json
make-all: It may be possible to combine two of them, but the logic for combinations can be hard to get right if they are more than simple booleans, or one boolean and a short list of enum (string) values. But combining 3 or 4 different dps complicates the logic too much, so I have not implemented support for that (in some cases there can be workarounds using value_redirect). If all the other options are only valid when manual/program is set to manual for example, that could simplify things enough to get an extra dp covered with value_redirect.

make-all: Easiest is to pick one to use for preset_mode, and do the others as separate entities.

PR #3796, opened 2025-10-02
Allow rescanning IP addresses

albertoxamin:
This PR allows going to the config and retrigger a scan of the IP address. It also does it automatically if the device is not reachable.

## Why?
In my flat I have a bad stock router which doesn't allow me to set me fixed ips for my tuya devices.

## Examples in my instance

telegram-cloud-photo-size-4-5971851979128948822-y
image


albertoxamin:
test failures are because i added a new field on the config flow and the lint I guess i need to sort imports

make-all:
This can already be achieved by setting the IP address to "Auto".

albertoxamin:
ok can you add it to the docs/readme?

make-all:
It used to be documented, but if the option is widely advertised then users tend to overuse it, and this results in bug reports for the drawbacks of using dynamic addresses (much slower reconnection times, failure to reconnect to some devices altogether if the app is open or if the user later installs localtuya) so I prefer to leave the information buried.

felipecrs:
@make-all have you considered the Home Assistant DHCP discovery for this? It "notifies" the integration on whether some IP Address for some MAC address has changed. Some official integrations do that.

Here is one reference (removing such feature, not adding):

https://github.com/home-assistant/core/pull/153373/commits/a348199377868aca818b80b7b3506e7773f62ec1

make-all:
I don't see an official API for this, but anyway looking at the DHCP browser info in my installation, it does not appear to offer any advantage over Tuya's own device id based discovery (most of my Tuya devices are on a different subnet and do not appear, which is the main drawback with Tuya discovery since it is broadcast based, not multicast).

felipecrs:
It definitely is official and several official integrations do the same thing.

This is even one of the quality gates:

https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/discovery-update-info/

And that's the way this can be done, through DHCP discovery.

I'm not sure there is any improvement over your approach, but it's probably worth trying.

In my integration, I was surprised that DHCP discovery was very fast.

Maybe it also runs lighter than a custom code, because Home Assistant processes IP/Mac changes for all integrations instead of multiple integrations scanning for the same thing.

make-all:
This would require setup to start from a MAC address if I understand correctly. I think the idea behind DHCP discovery is that the integration listens for devices with the manufacturer's MAC prefix to discover devices, and sets them up from there. But Tuya does not use a single device prefix. There are ranges registered to Tuya, but they also use ranges registered to Espressif or other chip manufacturers, and private unassigned ranges. Using Tuya's own broadcast discovery works better, and also gives some hints about the protocol, since 3.1 and 3.5 devices use a different port than 3.2/3.3/3.4 devices.

I know there is an official API for registering for DHCP discovery, but your suggestion was a bit different, it was to update IP addresses from DHCP. Unless the discovery was via DHCP in the first place, I do not see an API that allows the initial lookup for MAC address of a known IP address to work.

felipecrs:
@make-all, please check this:

- https://github.com/home-assistant/core/pull/154788

This uses DHCP discovery to update IP addresses **only**. DHCP discovery of new devices is not even possible in such integration.

> Using Tuya's own broadcast discovery works better, and also gives some hints about the protocol, since 3.1 and 3.5 devices use a different port than 3.2/3.3/3.4 devices.

None of that is necessary if the goal is to simply update the IP address.

Issue #3664, opened 2025-08-25
Request support for FancyLEDS 3 HDMI 2.1 Fancy Sync Box


LordShadowen:
### Log message

```text
[custom_components.tuya_local.config_flow] Device matches lytmi_hdmisync_backlight with quality of 82%. DPS: {"updated_at": 1756121708.136649, "20": false, "22": 1000, "23": 10, "24": "000000000386", "25": "82000000000000000000000000000000", "101": 1, "105": 0, "107": true, "108": 10, "109": false, "113": 100}
```

### Product ID

9cgdcqpe9iqg9tjn

### Product Name

FancyLEDS 3 HDMI 2.1 Fancy Sync Box

### DPS information

```text
{
  "result": {
    "model": "{\"modelId\":\"e1kj5m84\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"\",\"properties\":[
{\"abilityId\":20,\"accessMode\":\"rw\",\"code\":\"switch_led\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-dp_power\",\"attribute\":\"1027\"},\"name\":\"开关\",\"typeSpec\":{\"type\":\"bool\"}},
{\"abilityId\":21,\"accessMode\":\"rw\",\"code\":\"work_mode\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-dp_mode\",\"attribute\":\"1024\"},\"name\":\"模式\",\"typeSpec\":{\"type\":\"enum\",\"range\":[\"white\",\"colour\",\"scene\",\"music\"]}},
{\"abilityId\":22,\"accessMode\":\"rw\",\"code\":\"bright_value\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-liangdu1\",\"attribute\":\"1026\"},\"name\":\"亮度值\",\"typeSpec\":{\"type\":\"value\",\"max\":1000,\"min\":10,\"scale\":0,\"step\":1}},
{\"abilityId\":23,\"accessMode\":\"rw\",\"code\":\"temp_value\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-dp_half\",\"attribute\":\"1154\"},\"name\":\"冷暖值\",\"typeSpec\":{\"type\":\"value\",\"max\":1000,\"min\":0,\"scale\":0,\"step\":1}},
[...]

PR #3797, opened 2025-10-02
Add  FancyLEDS 3 HDMI 2.1 Fancy Sync Box

LordShadowen:
As described in #3664


make-all:
As far as I can see, all the changes here that are not just naming changes have been incorporated in the Lytmi config already. Since this config has errors I will close it, please try the latest Lytmi file and give clear feedback on what is not working (not just names that don't match the Tuya app, as replicating the Tuya app is not my goal).

PR #3802, opened 2025-10-03
Fine tune Lawn Mower devices and moebot/parkside mower

cociweb:
- Parkside mowers have a cancel feature to interrupt the ongoing task before a new command is issued. (Pause only has transiton to start_mowing/resume.)
- fixed multiple typos/logics
- Added missing Return to dock button
- Implemented spot/fixed mowing functionality by extending the LawnMoverEntityFeature core class
- Add raw text attribute to problem sensor


make-all:
It looks like you are trying to extend the lawn_mower platform with new activities and commands.  Extending the standard HA platforms is not an aim of this project, rather I prefer to work within the current limitations of HA, and extend through select entities, buttons etc as is done already in the moebot configuration.

PR #3809, opened 2025-10-06
Add support for Madimack Eco Heatpump 9 kW

mjens118:
Adds support for the Madimack Eco Heatpump 9 kW (product ID 3z98f9dmfwtieuc7).
Tested with Home Assistant 2025.10.1 and Tuya Local v2025.9.1 — fully functional via local control.

Features include:
	•	Accurate current-temperature reporting (DPS 102)
	•	Full HVAC control (heat / cool / auto / off)
	•	Preset modes: Boost, Sleep (Silence), Power
	•	Temperature range 18–40 °C

File added:
custom_components/tuya_local/devices/madimack_eco_heatpump_9kW.yaml

### DIFF
--- a/custom_components/tuya_local/devices/madimack_eco_heatpump_9kW.yaml
+++ b/custom_components/tuya_local/devices/madimack_eco_heatpump_9kW.yaml
@@ -1,0 +1,130 @@
+name: Madimack Eco Heatpump 9kW
+products:
+  - id: 3z98f9dmfwtieuc7
+    manufacturer: Madimack
+    model: Eco 9kW
+
+entities:
+  - entity: climate
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: "heat"
+      - id: 2
+        type: string
+        name: mode
+        mapping:
+          - dps_val: heating
+            value: heat
+          - dps_val: cold
+            value: cool
+          - dps_val: auto
+            value: heat_cool
+      - id: 4
+        type: integer
+        name: temperature
+        range:
+          min: 18
+          max: 40
+      - id: 6
+        type: string
+        name: temperature_unit
+        mapping:
+          - dps_val: c
+            value: C
+          - dps_val: f
+            value: F
+      - id: 102
+        type: integer
+        name: current_temperature
+      - id: 5
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: boost
+            value: boost
+          - dps_val: silence
+            value: sleep
+          - dps_val: power
+            value: boost
+
+  # Fault code as a SENSOR (no binary_sensor) – name must be 'sensor'
+  - entity: sensor
+    name: Fault code
+    category: diagnostic
+    dps:
+      - id: 15
+        type: bitfield
+        name: sensor
+
+  - entity: sensor
+    name: Compressor strength
+    category: diagnostic
+    dps:
+      - id: 20
+        type: integer
+        name: sensor
+        unit: "%"
+
+  - entity: sensor
+    name: Max temperature
+    category: diagnostic
+    dps:
+      - id: 21
+        type: integer
+        name: sensor
+        unit: "°C"
+
+  - entity: sensor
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/madimack_eco_heatpump_9kW.yaml
```diff
@@ -0,0 +1,130 @@
...
+entities:
+  - entity: climate
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
```
make-all, line 18: Below mode seems to support other modes. But it is not taken into account, because "mode" is not a standard part of climate entity.

custom_components/tuya_local/devices/madimack_eco_heatpump_9kW.yaml
```diff
@@ -0,0 +1,130 @@
+name: Madimack Eco Heatpump 9kW
```
make-all: Top level name should be unbranded. If this is a pool heat pump (like other supported madimack devices), then the top level name should probably mention that, and also on the climate entity below `translation_key: pool_heatpump` would be useful.

custom_components/tuya_local/devices/madimack_eco_heatpump_9kW.yaml
```diff
@@ -0,0 +1,130 @@
...
+            value: heat_cool
+      - id: 4
+        type: integer
+        name: temperature
+        range:
+          min: 18
+          max: 40
+      - id: 6
+        type: string
+        name: temperature_unit
```
make-all, line 52: Selecting a temperature unit of F should have some effect on the temperature range above (or perhaps require redirecting to a different dp for the temperature)

custom_components/tuya_local/devices/madimack_eco_heatpump_9kW.yaml
```diff
@@ -0,0 +1,130 @@
...
+        name: preset_mode
+        mapping:
+          - dps_val: boost
+            value: boost
+          - dps_val: silence
+            value: sleep
+          - dps_val: power
+            value: boost
+
+  # Fault code as a SENSOR (no binary_sensor) – name must be 'sensor'
```
make-all: This comment is incorrect. The convention for this integration is to turn fault sensors into a binary sensor to indicate whether there is a fault or not, with the fault code as an attribute.

A sensor with the raw fault code is not very useful, as the user still needs to decode the fault code to get any meaning from it.

Most devices have an example of this.

custom_components/tuya_local/devices/madimack_eco_heatpump_9kW.yaml
```diff
@@ -0,0 +1,130 @@
...
+  - entity: sensor
+    name: Compressor strength
+    category: diagnostic
+    dps:
+      - id: 20
+        type: integer
+        name: sensor
+        unit: "%"
+
+  - entity: sensor
```
make-all, line 103: Sensor, or number?

Is this supposed to be adjustable by the user, or is it indicating a variable maximum that depends on mode setting, or is it recording the maximum temperature reached in the past 24 hours/week/month/ever?

custom_components/tuya_local/devices/madimack_eco_heatpump_9kW.yaml
```diff
@@ -0,0 +1,130 @@
...
+    name: Max temperature
+    category: diagnostic
+    dps:
+      - id: 21
+        type: integer
+        name: sensor
+        unit: "°C"
+
+  - entity: sensor
+    name: Min temperature
```
make-all: Same questions as for max.

mjens118:
Thanks for the detailed feedback!
I’ve now updated the YAML to:
– Use an unbranded top-level name + translation_key: pool_heatpump
– Add a hidden mode constraint for proper hvac mapping
– Make the temperature range unit-aware
– Implement binary fault sensor with fault_code attribute
– Keep max/min/coil sensors as diagnostics

Let me know if there’s anything else I can tweak before merge.

PR #3812, opened 2025-10-06
Improvements to Axen heat pump configuration

amantia:
- fix warnings
- change text to match what the official app uses and what makes more sense to me
- add AC Voltage/Current diagnostics (hidden by default)


make-all:
Sorry, I made my own change that overlapped with this. The values without underscores were incorrect, so rather than adding as additional, I just changed them.

Renaming entities after release is quite disruptive, so I need a good case for making that change. I also added the sensors for voltage and current, but since they are not showing in the log on the original issue, I'm not sure how well they will work.

andraskdab:
Thanks, I will test the current version. The voltage entities work for me. 
Screenshot_20251006_154548
Screenshot_20251006_154617

PR #3817, opened 2025-10-07
Create vivo_fd55_motorized_tv_mount.yaml

tr1ckz:
Added Vivo Motorized TV Ceiling Mount support
(only up and down works)

### DIFF
--- a/custom_components/tuya_local/devices/vivo_fd55_tvmount.yaml
+++ b/custom_components/tuya_local/devices/vivo_fd55_tvmount.yaml
@@ -1,7 +1,7 @@
-name: "VIVO FD55 Motorized TV Mount"
+name: TV mount
 products:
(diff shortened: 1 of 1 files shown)


tr1ckz:
Tried adding the Stop button, but it didn't do anything. So only up and down works, if it's currently moving pressing either button will stop the mount.

tr1ckz:
open and close were backwards

make-all:
This is not really a cover device, I think it will be better to use buttons instead of a cover entity for the functions, as the device is not providing any feedback on current state or position that can be used to provide a richer interface than just open, close and stop buttons. For the limited functions supported, it seems compatible with the speaka_sptv510_tvmount config. You could just copy the relevant entities from there. And if there are other functions available that you have not yet covered in your config, you may find the same function in that config that can be copied, and if necessary modified.

PR #3819, opened 2025-10-08
Added dehumidifier's humidity as a sensor additionally

belleDESiRE:
Instead of only being available in controls, the humidity data can now also be used as a sensor


make-all:
This is an unnecessary change. The current_humidity from the humidifier entity can be used directly.

belleDESiRE:
probably my noobness in HA, but I can't assign it to an area's related sensors can I?

make-all:
You can add the humidifier entity to an area. Whether that will report the humidity in whatever context you are expecting it, I don't know, I do not have a humidifier device any longer, but the climate device does appear to have its current temperature considered by the built in Area card as far as I can tell.
But even if not, there are template helpers for special cases like this.

PR #3823, opened 2025-10-08
Adding support for Beok TGM50-WIFI-WPB-WF / M5-G2 Thermostat

diegaccio:
Added support for BEOK Smart Room Thermostat, M5-G2. The label on the box says "TGM50-WIFI-WPB-WF"

Here the home assistant log:

`Adding M5-G2 device with product id 3vkudynaojqad3al`

`Cloud device spec: [
{"id": 1, "name": "switch", "type": "Boolean", "format": "{}", "enumMap": {}}, 
{"id": 2, "name": "temp_set", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":50,\"max\":950,\"scale\":1,\"step\":5}", "enumMap": {}}, 
{"id": 3, "name": "temp_current", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":0,\"max\":999,\"scale\":1,\"step\":1}", "enumMap": {}}, 
{"id": 4, "name": "mode", "type": "Enum", "format": "{\"range\":[\"auto\"]}", "enumMap": {}}, 
{"id": 9, "name": "child_lock", "type": "Boolean", "format": "{}", "enumMap": {}}, 
{"id": 11, "name": "fault", "type": "Bitmap", "format": "{\"label\":[\"fault\"]}", "enumMap": {}}, 
{"id": 15, "name": "upper_temp", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":15,\"max\":95,\"scale\":0,\"step\":1}", "enumMap": {}}, 
{"id": 19, "name": "temp_correction", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":-99,\"max\":99,\"scale\":1,\"step\":1}", "enumMap": {}}, 
{"id": 103, "name": "frost", "type": "Boolean", "format": "{}", "enumMap": {}}, 
{"id": 109, "name": "sound", "type": "Boolean", "format": "{}", "enumMap": {}}]`


make-all:
The info from the cloud that you provided above is often only partial. Check your device against the config beok_tg50wp_thermostat, which was added recently, as it may be the same.
Comparing the configs there are only additional dps (5 for hvac_action, 101 for temperature hysteresis, 102 for external temperature limit, 104 for factory reset, 106 for backlight brightness, 107 for away mode temperature, 108 for inverting output. If some of these are missing from your device, it may be possible to make them optional and hidden when not available rather than adding a new config that is just a subset of that.

diegaccio:
Ok thanks for the clarification, I actually needed the hvac_action for an automation with the heating water pump. I added the section to my yaml file and now the hvac_action  is working like a charm. 
Just one thing I did not understand: my yaml is quite similar to the original beok_tgm50_thermostat.yaml that is present in my folder, it has more fields, but it's not picked up when I try to add the device.  It has something to do with the product id? 
Thanks

make-all:
Your model number has wp in it, so my assumption is that it may have the same incompatibility with the original tgm50 that saw the tgm50wp being created a couple of weeks ago.

diegaccio:
I checked the code in the function match_quality and I guess the product id is the one thas is checked before dpis. If there is a match there the yaml is picked up. 
So I think that adding this:

`  - id: bfd1d6e2efcf60a661gjas
    manufacturer: Beok / Avatto
    model: TGM50-WIFI-WPB-WF / M5-G2` 

in either of the two files, in the products list, will do the job, since they both have more dpis then mine, and I don't know wich one is better. The only difference is "Away mode temperature" (newly added yaml file), that's not configurable in the tuya app for my device, and the "schedule" in the original file, but mine has only the daily option. 

Let me know if that's right. Thanks

make-all:
First the config needs to match without the product id. If it does not match, then it would not be correct to add a product id.
If neither config matches, then gathering debug logs might also show that the match is partial and can be matched if some dps in the config are marked optional.  BTW, the product id is the one in the original log message you posted `3vkudynaojqad3al`. The bf... one appears to be a device id, which is unique to your specific device.

diegaccio:
Hello, last update removed the file I added and broke my configuration, I didn't have the time to fix this before, now I was forced to :) . So I added the correct product id (thanks for pointing my in the righr direction) to beok_tg50wp_thermostat, and it now works with that config file. I also removed the new file I added in this PR. Let me know If it's ok for you, thanks

PR #3825, opened 2025-10-09
Add Smart Motorized Pop-up Socket

3ek:
Added support for generic motorized power socket:
tuya_smart_motorized_pop_up_socket

I wanted to implement it as `cover` entity but was unable to do it properly - commands for each direction are in separate DPS. So I added simple buttons instead. Data model is a mess anyway, none of the descripions match actual device behavior.

I had to use Tuya Developer platform and change `Control Instruction Mode` to `DP Instruction` before receieving Device Data Model since not all commands were visible in tuya-local logs in HA.

obraz

Device Information:
>**Product Name**
智能升降插座
**Device ID**
bfa3f089992fcf874820qj
**Product Category**
clkg

Device Data Model:
```
{
   "modelId":"0000020aop",
   "services":[
      {
         "actions":[
         ],
         "code":"",
         "description":"",
         "events":[
         ],
         "name":"默认服务",
         "properties":[
            {
               "abilityId":1,
               "accessMode":"rw",
               "code":"control",
               "description":"",
               "extensions":{
                  "iconName":"icon-a_power"
               },
               "name":"灯光",
               "typeSpec":{
                  "type":"enum",
                  "range":[
                     "S3",
                     "S2",
                     "S1",
                     "OFF"
                  ]
               }
            },
            {
               "abilityId":3,
               "accessMode":"rw",
               "code":"cur_calibration",
               "description":"",
               "extensions":{
                  "iconName":"icon-dp_loop"
               },
               "name":"电源",
               "typeSpec":{
                  "type":"enum",
                  "range":[
                     "start",
                     "off"
                  ]
               }
            },
            {
               "abilityId":6,
               "accessMode":"rw",
               "code":"cur_calibration_2",
               "description":"",
               "extensions":{
                  "iconName":"icon-dp_loop"
               },
               "name":"上升",
               "typeSpec":{
                  "type":"enum",
                  "range":[
                     "open",
                     "end"
                  ]
               }
            },
            {
               "abilityId":7,
               "accessMode":"rw",
               "code":"switch_backlight",
               "description":"",
               "extensions":{
                  "iconName":"icon-dp_power3"
               },
               "name":"插座",
               "typeSpec":{
                  "type":"bool"
               }
            },
            {
               "abilityId":8,
               "accessMode":"rw",
               "code":"control_back",
               "description":"",
               "extensions":{
                  "iconName":"icon-dp_loop"
               },
               "name":"下降",
               "typeSpec":{
                  "type":"enum",
                  "range":[
                     "open",
                     "end"
                  ]
               }
            },
            {
               "abilityId":101,
               "accessMode":"ro",
               "code":"ZT",
               "description":"状态",
               "name":"状态",
               "typeSpec":{
                  "type":"enum",
                  "range":[
                     "a1",
                     "a2",
                     "a3",
                     "a4",
                     "a5",
                     "a6",
                     "a7",
                     "as8",
                     "a9",
                     "a10",
                     "a11",
                     "a12",
                     "a13",
                     "a14",
                     "a15",
                     "a16",
                     "a17",
                     "a18",
                     "a19",
                     "a20"
                  ]
               }
            }
         ]
      }
   ]
}
```

### DIFF
--- a/custom_components/tuya_local/devices/smart_motorized_pop_up_socket.yaml
+++ b/custom_components/tuya_local/devices/smart_motorized_pop_up_socket.yaml
@@ -101,4 +101,3 @@ entities:
             value: "Closing Interrupted"
           - dps_val: "a9"
             value: "Closed"
-


custom_components/tuya_local/devices/smart_motorized_pop_up_socket.yaml
```diff
@@ -0,0 +1,96 @@
+name: "Smart Motorized Pop-up Socket"
+products:
+  - id: hfeuatvaoedf2mqo
+    manufacturer: Generic
```
make-all: Just leave the manufacturer out (it will default to "Tuya") if you do not know.

custom_components/tuya_local/devices/smart_motorized_pop_up_socket.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - id: hfeuatvaoedf2mqo
+    manufacturer: Generic
+    name: "Smart Motorized Pop-up Socket"
+
+entities:
+  - entity: select
+    name: Light mode
+    icon: "mdi:lightbulb"
+    dps:
+      - id: 1
```
make-all, line 9: I think this can be a light entity, with the dp mapped to brightness (range is 0-255, with 0 being off, 255 full brightness)

custom_components/tuya_local/devices/smart_motorized_pop_up_socket.yaml
```diff
@@ -0,0 +1,96 @@
...
+    name: Open
+    icon: "mdi:arrow-up-thick"
+    dps:
+      - id: 6
+        type: string
+        name: button
+        mapping:
+          - dps_val: "open"
+            value: true
+          - dps_val: "end"
```
make-all: The false mapping is meaningless for a button. If you want to be able to stop it part way by sending end, then you will need to define another button that maps "end" to true.

custom_components/tuya_local/devices/smart_motorized_pop_up_socket.yaml
3ek: Removed.

custom_components/tuya_local/devices/smart_motorized_pop_up_socket.yaml
3ek: Since device does not support integer type brightness and there is only predefined light levels. I changed it to custom mapping to support light entity.

custom_components/tuya_local/devices/smart_motorized_pop_up_socket.yaml
3ek: There is no option to stop operation manually, there is only anti-pinch feature that performs emergency stop automatically. It works like a latch - it holds `open` when set and resets after device operation is finished.

I managed to add it as `cover` entity with help of `value_redirect` to different dp, so button mapping no longer relevant anyway.

custom_components/tuya_local/devices/smart_motorized_pop_up_socket.yaml
```diff
@@ -34,36 +32,32 @@ entities:
...
           - dps_val: "open"
-            value: true
-          - dps_val: "end"
-            value: false
-  - entity: button
-    name: Close
-    icon: "mdi:arrow-down-thick"
-    dps:
+            value: "open"
+          - dps_val: "close"
```
make-all: Should this be "close"? or "end"?

3ek: "close" is actually correct, since:
- "close" value for dp id: 6 - lifts the module
- "close" value for dp id: 8 - drops the module

I changed mapping for dp id 6 by removing `dps_val: close` entry (it has no use here) and renamed redirection to `close_redirection` to be not confused with `close` value

PR #3826, opened 2025-10-09
feat(climate): add schedule decoding and temperature retrieval

Watt25:
KKMoon knob termostats have a design flaw: it reports the manually set target temperature even in AUTO mode. (Which is schedule based of course). This modification is a workaround to decode the schedule (DP71) and in AUTO mode report the target temperature from the schedule instead of the temperature the device provides. In case the device do not provide a schedule or it is in a different format, the code falls back to report the temperature which the device provided.

### DIFF
--- a/custom_components/tuya_local/climate.py
+++ b/custom_components/tuya_local/climate.py
@@ -4,7 +4,7 @@ Setup for different kinds of Tuya climate devices
 
 import logging
 import base64
-from typing import List, Tuple
+from typing import list, tuple
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


make-all:
This does not belong inside the climate entity. I would suggest making a separate integration for this that fits over the top of the text entities exposed by this integration, similarly as was done with Pet feeder meal plans. I do not wish to expand this integration with device specific complex logic, as it is more work than I can manage already maintaining the integration and adding new devices.

PR #3831, opened 2025-10-10
Added support for Outdoor inc zip blinds

justinwyq:


custom_components/tuya_local/devices/outsmart_switch_zs3lcbcio.yaml
```diff
@@ -0,0 +1,46 @@
+name: Master Bedroom Lights
```
make-all: This config is a duplicate of the existing simple_dual_switch_timer.yaml.A new config is not required.

custom_components/tuya_local/devices/outsmart_switch_water_heater_a_zb.yaml
```diff
@@ -0,0 +1,35 @@
+name: Water Heater Switch
```
make-all: This config is already covered by somgom_single_switch.yaml.  There is an extra entity there for initial state but it is marked optional and you can hide it in your entity list if it is not supported by your device.

custom_components/tuya_local/devices/miju_smart_curtain_cm82.yaml
```diff
@@ -0,0 +1,79 @@
+name: Miju Smart Curtain (CM82)
```
make-all: This config seems to duplicate the existing loonas_curtain.yaml config. A new config is not needed.

custom_components/tuya_local/devices/outdoor_inc_zipblind.yaml
```diff
@@ -0,0 +1,56 @@
+name: Outdoor Zip Blind
```
make-all: Top level name should be without branding

custom_components/tuya_local/devices/outdoor_inc_zipblind.yaml
```diff
@@ -0,0 +1,56 @@
...
+products:
+  - id: f91miyhj
+    manufacturer: Outdoor Inc
+    model: DC1545BE
+entities:
+  - entity: cover
+    class: blind
+    current_position_dp: 8
+    set_position_dp: 9
```
make-all: These two are not valid syntax. They are defined properly in the dps list below, so can be removed

custom_components/tuya_local/devices/outdoor_inc_zipblind.yaml
```diff
@@ -0,0 +1,56 @@
...
+          min: 0
+          max: 100
+      - id: 8
+        name: current_position
+        type: integer
+        range:
+          min: 0
+          max: 100
+      - id: 3
+        name: state
```
make-all: Avoid the name "state" for attributes, as it clashes with the HA concept of entity state. Usually tuya uses "work_state", so you can use that.

PR #3833, opened 2025-10-10
Improvements on HY Multifunction Heatpump V1.2

goncalogui1233:
While using this device, I notice the following points could be improved/added:

1. Missing the Outlet Temperature, since we already have the Inlet Temperature and Tank Temperature
2. Climate current temperature was not displaying the correct temperature because it was not using the correct dps id.

### DIFF
--- a/custom_components/tuya_local/devices/hy_combo_heatpump.yaml
+++ b/custom_components/tuya_local/devices/hy_combo_heatpump.yaml
@@ -63,7 +63,7 @@ entities:
           - dps_val: null
             value: true
           - value: false
-      - id: 21
+      - id: 24
         type: integer
         name: current_temperature
         mapping:
@@ -330,6 +330,18 @@ entities:
     category: diagnostic
     dps:
       - id: 23
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: Outlet temperature
+    class: temperature
+    category: diagnostic
+    dps:
+      - id: 24
         type: integer
         name: sensor
         unit: C


custom_components/tuya_local/devices/hy_combo_heatpump.yaml
```diff
@@ -63,7 +63,7 @@ entities:
           - dps_val: null
             value: true
           - value: false
-      - id: 21
+      - id: 24
```
make-all, line 66: If 24 should be the air temperature for the climate entity, then what should the water temperature be for the water_heater?

custom_components/tuya_local/devices/hy_combo_heatpump.yaml
```diff
@@ -336,6 +336,18 @@ entities:
         class: measurement
         mapping:
           - scale: 10
+  - entity: sensor
```
make-all, line 339: This is already output as the water_heater current_temperature, and after your change also by the climate current_temperature. It does not need a 3rd representation.

custom_components/tuya_local/devices/hy_combo_heatpump.yaml
goncalogui1233: The id 21 is the air temperature. I suggested this change to align the current temperature value between the water_heater entity and climate entity.
But, for the climate entity, should the displayed the air temperature instead of the water temperature?

custom_components/tuya_local/devices/hy_combo_heatpump.yaml
goncalogui1233: Yep, I noticed it. But the idea of creating this separate sensor was to making more easy the creation of automations where we have to monitor constantly the temperature, or to create graphics to display the temperature in certain times.

custom_components/tuya_local/devices/hy_combo_heatpump.yaml
make-all: This is a multifunction heat pump, so the climate and water_heater entity are for different functions. They are not supposed to reflect the same info, water_heater is for the domestic hot water function, climate is for the heating function (I am not sure if it is via radiators for this device, or direct heat pump indoor unit). So the air temperature makes sense for the climate entity.

custom_components/tuya_local/devices/hy_combo_heatpump.yaml
make-all: climate.current_temperature and water_heater.current_temperature can be accessed in automations, and at least climate.current_temperature shows up in the history graph for climate entity (I am not sure about water_heater, because I do not have one of those in my installation).  If a user really needs a sensor, they can use a template helper, but it does not need to pollute the entity namespace of all users just for the few who need that.

PR #3839, opened 2025-10-12
Changes to Klarta Stor 2 and Klarta Forste 4 with appropriate adjustments to polish translation

kjpg99:
Hello,

Here I suggest making couple of changes to two devices – Klarta Forste 4 and Klarta Stor 2. In Klarta Forste 4 improvements include support for all the fan speeds and adding a selector for mode in which auto mode operates (as it already has been in case of Stor 2). In Stor 2 I suggest covering all the reported levels of air quality. Finally, in both devices I suggest using native preset modes – it requires certain adjustments in the polish translation file, which I also added together with some other small improvements.

The details are described in every commit made to particular files. It also should be worth mentioning that I've been using the integration with so modified devices files for already some time ans I haven't had any issues. The only difference is that now I mapped Stor 2's hazard air quality level to new "danger" level included recently in pl.json.

I hope that the changes will appear helpful. Should there be any questions, I'll try to answer as soon as possible. Thanks for a great integration!


custom_components/tuya_local/translations/pl.json
```diff
@@ -353,18 +353,21 @@
...
                             "smart": "Inteligentny",
+                            "auto": "Automatyczny",
+                            "manual": "Ręczny",
                             "strong": "Silny",
                             "custom": "Niestandardowy",
                             "high": "Wysoki",
                             "medium": "Średni",
                             "low": "Niski",
                             "displayoff": "Wyświetlacz wyłączony",
+                            "screenoff": "Wyłączone podświetlanie",
```
make-all: There is no need to introduce this as a new translation. displayoff has the same meaning, so just use that.

custom_components/tuya_local/devices/klarta_forste4_purifier.yaml
```diff
@@ -16,25 +16,33 @@ entities:
         name: preset_mode
         mapping:
           - dps_val: auto
-            value: smart
+            value: auto
```
make-all: I don't see a point to adding new translations where the existing ones are close enough in meaning.

Issue #1983, opened 2024-06-06
Parkside PMRDA 20-Li A1 reports wrong (or not exhaustive) statuses and incorrect problems


cociweb:
**Describe the bug**
Parkside PMRDA 20-Li A1 lawn mower ("type": "moebot_s_mower") sometimes has incorrect mappings in specific cases, and the fault reports always return "Tilted"

**To Reproduce**
Steps to reproduce the behaviour including the device the issue was observed with:
use and test the device.

**Expected behavior**
mapping the states and faults correctly (see below)

**Additional context**
correct info is returned from the raw_action attribute of the status but no such attribute for the problems, so individual remapping is not possible. (it would be nice if it is attached as well, but this FR is not the scope of the current ProblemRiport.) Digging deeper, the problem state (103) always return MOWING_LEAN

Only protocol 3.3 is supported.

If the bug involves a device, then please include device diagnostics from
[mower.txt](https://github.com/user-attachments/files/15612571/mower.txt)

Please check if there are messages from Tuya Local in the Home
```
2024-06-06 11:18:27.269 DEBUG (MainThread) [custom_components.tuya_local.device] Fűnyíró received {"6": 94, "101": "ERROR", "102": 1048576, "103": "MOWER_LEAN", "104": false, "105": 5, "106": **PIN should be REDACTED**, "116": false, "full_poll": true}
2024-06-06 11:18:27.271 DEBUG (MainThread) [custom_components.tuya_local.helpers.device_config] No mapping for dpid 6 (sensor), unable to determine valid values
2024-06-06 11:18:27.272 DEBUG (MainThread) [custom_components.tuya_local.helpers.device_config] sensor values: ['Tilted', 'Steep', 'Raining', 'Low battery', 'Weak signal', 'Close top cover', 'Docked', 'Undocked', 'Manually dock', 'Finished fixed mowing', 'Disconnected', 'Paused charging', 'Interrupted', 'Interrupted fixed mowing', 'Turn on button', 'Press start', 'Set 30 minute timer', 'Invalid timer', 'Charging current low', 'Caught in rain', 'Data upload failure', 'Continue', 'Stopped', 'UI locked']
2024-06-06 11:18:27.272 DEBUG (MainThread) [custom_components.tuya_local.helpers.device_config] Lawnmower: Mapped dps 103 value from MOWER_LEAN to Tilted
[...]

PR #3840, opened 2025-10-13
Parkside Lawn Mower

cociweb:
- Parkside mowers have a cancel feature to interrupt the ongoing task before a new command is issued. (The current Pause function has only transitioned to start_mowing/resume. Cancel stops and clear the other activities: mowing and paused mowing, fixed mowing, and returning, so the current (general HA) implementation is insufficient for these devices.) by extending the LawnMoverEntityFeature class
- Implemented spot/fixed mowing functionality by extending the LawnMoverEntityFeature and Activity core class
- fixed multiple typos/logics
- Added the missing Return to dock button entity
- Added two more possible warning messages
- Add raw text attribute to problem sensor
- Keeping backward compatibility with the already implemented lawn mowers (as of now, only Moebot) as forking.
- Tests are created for the device and modified on the device type level
- remove Unsupported bladestop and workmode function

this PR makes possible to add proper support PARKSIDE PMRDA A1, PARKSIDE PMRDA B1, PARKSIDE PRDA A1, PARKSIDE PAMRS A1, PARKSIDE PAMR 500 A1 (Unluckily, with the same device id)

Unfortunately, the core LawnMoverEntityFeature and LawnMoverActivity are insufficient for this type of TUYA device. (Already suggested multiple times to the HA ADR architecture board to extend the basic features, but they always rejected and proposed to implement it on the device level, so other integrations do the same, eg Mammotion, etc.

The PR is tested, and it works as intended on PARKSIDE PMRDA A1.

### DIFF
--- a/custom_components/tuya_local/devices/parkside_p_mower.yaml
+++ b/custom_components/tuya_local/devices/parkside_p_mower.yaml
@@ -49,7 +49,6 @@ entities:
             value: cancel
           - dps_val: ContinueWork
             value: resume
-            hidden: true
           - dps_val: StartReturnStation
             value: dock
       - id: 101


make-all:
This integration is not here to implement architectural changes rejected by HA developers

cociweb:
@make-all ,It is not a change, it is an extension.... 
In this case, please remove the Parkside support, as a non-suported device, because the current implementation has only 1 working feature out of the default 3 of HA for this device.
Should I make the PR for the removal?

make-all:
The Parkside is working with the Moebot config according to other users. I will not be taking the drastic measure of removing it because it is not perfect.

cociweb:
> The Parkside is working with the Moebot config according to other users. I will not be taking the drastic measure of removing it because it is not perfect.

@make-all  Yep, this knowledge/feedback is based on my opinion/tickets as well in the past [here](https://github.com/make-all/tuya-local/issues/1983) and [here](https://github.com/make-all/tuya-local/discussions/758#discussioncomment-9677041). It is working. (somehow, with hacks and workarounds).... but not correctly as it should. - I've also mentioned that time there are a couple of things that are missing, Ofc, with the old implementation, where the moebot used a vacuum platform, it is ok. (As a workaround, think many of us still stuck to the initial vacuum platform implementation - as I was....)

This extension is not an abstract/unreal extension. It just fits the TUYA dev platform and maps it correctly:

image

image


But you know it, because it is decoded in the moebot config file, but it is not available for the current implementation:
https://github.com/make-all/tuya-local/blob/851a6f07f7cc394d9264358f9380b94164f99cc0/custom_components/tuya_local/devices/moebot_s_mower.yaml#L47C5-L57
Let's take another example for the workarounds:
Also, we can take a closer look at the button entities. (The lawn mowers are controlled by them with this repo.) The back-to-home button is not implemented in the case of Moebot as well. As a workaround, I've used the work hour timer of the device to return in the automations... 
Take another one:  the cancel button was mapped to the pause feature: paused state has transition only to start_mowing or the resume/continue (which was also not implemented. It was mapped back to start_mowing.) so nobody noticed that cancel is not working, since it was a pause. :D -> its working somehow, but not correctly.

another one: the fixed_mowing was totally missed, because HA not support it, but it would be usable (especially with starting point, which is still not part of this PR, maybe later: pymoebot supports this in case of Parkside/moebot machines: it requires some additional conversion.) All in all,a resume of paused fixed mode mowing (then intending to resume) -according to the current implementation - will start a random mowing instead of resuming the fixed mowing. - The solution was to install the pymoebot integration alongside the tuya-local. but WHY is it forcefully required instead of merging this PR? :D
So as you can see, it was working somehow, but the current working logic is not proper.

So my PR is truly based on usability and real-life use cases. (ofc, there are still needs additional scripts in HA to make the device work as simply as it should: but those are not TUYA platform relevant, those are device specific and worth creating another device integration, but in this case your repo will miss the support of these machines, if you really stick to the "moebot-only" concept...)
Additionally, the other Tuya-based lawn mowers: 
JIUDU DM2
AYI DRM3 600i
Hyundai HYRM1000
Gtech RLM50
Sømløs G1/G1s
iHunt RoboMower APP 2500m2 PRO

PR #3875, opened 2025-10-21
addad "Smart P01" socket

luckyTamme:
Found my newly bought (super cheap) smart socket not being supported so I followed the guidelines and created a profile. 

The device is called Smart P01 in the tuya app, it features a switch, timer and electrical usage date (voltage, current, power, accumulated power). test_bit and deprecated timer fields are not included. Tested and working in my setup.

Device costs between 3-6€ on aliexpress.

example product:
https://de.aliexpress.com/item/1005007880121648.html

3b9343c96f781ef01ed3341d5952f1d4

fetched data from tinytuya:
```
{
  "name": "Smart P01",
  "id": "XXX",
  "key": "XXX",
  "mac": "XXX",
  "uuid": "XXX",
  "sn": "XXX",
  "category": "tdq",
  "product_name": "Smart P01",
  "product_id": "ocazi2dacnofqidn",
  "biz_type": 0,
  "model": "",
  "sub": false,
  "icon": "https://images.tuyaeu.com/smart/icon/ay15532588194335cLlF/3b9343c96f781ef01ed3341d5952f1d4.jpg",
  "mapping": {
    "1": {
      "code": "switch_1",
      "type": "Boolean",
      "values": {}
    },
    "9": {
      "code": "countdown_1",
      "type": "Integer",
      "values": {
        "unit": "s",
        "min": 0,
        "max": 86400,
        "scale": 0,
        "step": 1
      }
    },
    "20": {
      "code": "add_ele",
      "type": "Integer",
      "values": {
        "unit": "kW\u00b7h",
        "min": 0,
        "max": 50000,
        "scale": 3,
        "step": 100
      }
    },
    "21": {
      "code": "cur_current",
      "type": "Integer",
      "values": {
        "unit": "mA",
        "min": 0,
        "max": 30000,
        "scale": 0,
        "step": 1
      }
    },
    "22": {
      "code": "cur_power",
      "type": "Integer",
      "values": {
        "unit": "W",
        "min": 0,
        "max": 50000,
        "scale": 1,
        "step": 1
      }
    },
    "23": {
      "code": "cur_voltage",
      "type": "Integer",
      "values": {
        "unit": "V",
        "min": 0,
        "max": 5000,
        "scale": 1,
        "step": 1
      }
    },
    "24": {
      "code": "test_bit",
      "type": "Integer",
      "values": {
        "min": 0,
        "max": 5,
        "scale": 0,
        "step": 1
      }
    },
    "29": {
      "code": "fault",
      "type": "Bitmap",
      "values": {
        "label": ["ov_cr", "ov_vol", "ov_pwr", "ls_cr", "ls_vol", "ls_pow"]
      }
    },
    "38": {
      "code": "relay_status",
      "type": "Enum",
      "values": {
        "range": ["0", "1", "2"]
      }
    },
    "42": {
      "code": "random_time",
      "type": "String",
      "values": "{\"maxlen\":255}"
    },
    "43": {
      "code": "cycle_time",
      "type": "String",
      "values": "{\"maxlen\":255}"
    }
  },
  "ip": "192.168.0.X",
  "version": "3.5"
}
```

### DIFF
--- a/custom_components/tuya_local/devices/ledvance_smart_ceilingfan.yaml
+++ b/custom_components/tuya_local/devices/ledvance_smart_ceilingfan.yaml
@@ -69,7 +69,7 @@ entities:
           min: 0
           max: 86399
   - entity: time
-    name:  Fan timer
+    name: Fan timer
     translation_key: timer
     category: config
     dps:


custom_components/tuya_local/devices/smart_p01_socket.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 20
+        type: integer
+        name: add_ele
+        optional: true
+      - id: 42
+        type: string
+        name: random_schedule
```
make-all: This should be a text entity, so that users can use it if they can decipher the format. Usually the type will be base64 (if you have some sample data you should be able to confirm that).

custom_components/tuya_local/devices/smart_p01_socket.yaml
```diff
@@ -0,0 +1,110 @@
...
+        optional: true
+      - id: 43
+        type: string
+        name: cycle_schedule
```
make-all: Likewise

custom_components/tuya_local/devices/smart_p01_socket.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - id: 9
+        type: integer
+        name: second
+        range:
+          min: 0
+          max: 86400
+  - entity: sensor
+    class: energy
+    dps:
+      - id: 20
```
make-all: Above this is under the switch entity as the extra attribute add_ele. Usually Tuya dps called add_ele are not usable as an energy sensor, as the dp only indicates usage since the last report, so it is resetting too frequently for HA to track it.

There is a possibility that this device is different, because it is using dp 20, instead of the usual 17 for this, but one of the two should be removed. If this sensor is providing usable data, then keep the sensor and delete the attribute above, if the data is unusable, keep it in the add_ele attribute above (or on the power sensor might be better), and remove the useless sensor.

custom_components/tuya_local/devices/smart_p01_socket.yaml
```diff
@@ -0,0 +1,110 @@
+name: Smart P01 Socket
```
make-all: This top level name should be without branding, quite likely this same device is sold in different markets or by different sellers on aliexpress under different brands.

luckyTamme: The devices I bought look "brand-less" and just being labeled "Smart Socket", it looks like the same model is sold by different vendors. The device shows up as P01 smart socket in the tuya app, so I thought probably all devices with this hardware and feature-set will likely be named P01 smart socket internally.

Should I use something more generic like "Smart Socket"?

make-all: Usual terminology is smartplug

custom_components/tuya_local/devices/smart_p01_socket.yaml
```diff
@@ -0,0 +1,110 @@
+name: Smart P01 Socket
+products:
+  - id: ocazi2dacnofqidn
+    manufacturer: Smart
```
make-all: So based on your previous comment, is the manufacturer's brand "Smart", or is it generic? In the case it is generic, manufacturer should be left out (and will appear as "Tuya")

luckyTamme:
Thank you for taking over the PR and your great work!

PR #3880, opened 2025-10-22
feat(emos_ip300_camera): add EMOS IP-300 camera with floodlight

jkolo:
## Summary
Add support for EMOS IP-300 outdoor camera with integrated floodlight.

**Device Details:**
- Product ID: `wauxpj9p977uxhvb`
- Manufacturer: EMOS
- Model: IP-300
- Protocol: 3.2
- Total DPs: 34 (1 primary camera + 33 configurable options)

## Features
**Primary Entities:**
- Camera entity with motion detection and recording
- Floodlight (brightness 20-100%)

**Configuration Entities:**
- Indicator light control
- Image flip and timestamp overlay
- Motion detection sensitivity and area limiting
- Night vision modes (Auto/On/Off)
- SD card management and recording modes
- PTZ controls (Left/Right/Stop)
- PIR sensor with adjustable sensitivity
- Siren with duration and volume control
- Light warning with duration
- ONVIF stream support (with IP and password management)
- Scene linkage (6 scenes)

**Diagnostic Entities:**
- SD card status and capacity
- Alarm and initiative message events

## Technical Notes
1. **Entity Initialization**: Device sends only 1 DP (101 or 103) at startup, so null mappings with default values added for 17+ DPs to ensure entities show correct initial state instead of 'unknown'.

2. **Brightness Mapping**: Home Assistant auto-scales 0-255 to device's 20-100 range. No manual mapping needed.

3. **ONVIF Configuration**: Consolidated 3 DPs (253, 254, 255) into single switch entity following nedis_outdoor_camera pattern. Uses `sensitive: true` for IP and password fields.

4. **DP Query Behavior**: Most DPs cannot be queried individually - device only sends updates when values change via app or device triggers.

## Testing
- ✅ All 13 unit tests passing
- ✅ Protocol version 3.2 verified
- ✅ Brightness scaling verified (0-255 → 20-100)
- ✅ Entity initialization with null mappings tested
- ✅ All 34 DPs mapped and verified against Tuya Device Data Model

### DIFF
--- a/custom_components/tuya_local/devices/emos_ip300_camera.yaml
+++ b/custom_components/tuya_local/devices/emos_ip300_camera.yaml
@@ -27,10 +27,6 @@ entities:
         optional: true
         sensitive: true
         name: snapshot
-      - id: 185
-        type: base64
-        name: log_message
-        optional: true
   - entity: light
     translation_key: indicator
     category: config
@@ -230,18 +226,32 @@ entities:
             value: Continuous
           - dps_val: null
             value: null
-  - entity: switch
-    name: Siren
-    icon: "mdi:bullhorn"
+  - entity: siren
     category: config
     dps:
       - id: 159
         type: boolean
         name: switch
         optional: true
+      - id: 194
+        type: integer
+        name: duration
+        optional: true
+        unit: s
+        range:
+          min: 10
+          max: 600
         mapping:
-          - dps_val: null
-            value: null
+          - step: 10
+      - id: 195
+        type: integer
+        name: volume_level
+        optional: true
+        range:
+          min: 10
+          max: 100
+        mapping:
+          - scale: 100
   - entity: switch
     name: Motion tracking
     icon: "mdi:motion-sensor"
@@ -300,34 +310,6 @@ entities:
         range:
           min: 5
           max: 600
-  - entity: number
-    name: Siren duration
-    icon: "mdi:timer-outline"
-    category: config
-    dps:
-      - id: 194
-        type: integer
-        name: value
-        optional: true
-        unit: s
-        range:
-          min: 10
-          max: 600
-        mapping:
-          - step: 10
-  - entity: number
-    name: Siren volume
-    icon: "mdi:volume-high"
-    category: config
-    dps:
-      - id: 195
-        type: integer
-        name: value
-        optional: true
-        unit: "%"
-        range:
-          min: 10
-          max: 100
   - entity: event
     name: Alarm
     dps:
@@ -340,11 +322,11 @@ entities:
             value: null
           - value: alarm
       - id: 185
-        type: base64
+        type: utf16b64
         name: alarm_message
         optional: true
   - entity: event
-    name: Initiative message
+    name: Notification
     dps:
       - id: 212
         type: base64
@@ -355,7 +337,7 @@ entities:
             value: null
           - value: message
       - id: 212
-        type: base64
+        type: utf16b64
         name: message
         optional: true
   - entity: switch


custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,442 @@
...
+    dps:
+      - id: 194
+        type: integer
+        name: value
+        optional: true
+        unit: s
+        range:
+          min: 10
+          max: 600
+        step: 10
```
make-all: Step must be inside a mapping (the reason it was designed this way is that a step can change input values by mapping them to the nearest step)

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,442 @@
...
+        name: record
+        optional: true
+        mapping:
+          - dps_val: null
+            value: false
+            hidden: true
+      - id: 115
+        type: base64
+        optional: true
+        persist: false
```
make-all: I don't think persist: false is desirable here, as it may make the snapshot flash up for a short time only. That is if it is a decodable image, only a few Tuya cameras send actual snapshots (and they are very low resolution), most just give lookup keys for cloud storage.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,442 @@
...
+        sensitive: true
+        name: snapshot
+      - id: 109
+        type: string
+        name: memory_card_capacity
```
make-all: Probably better under the SD status than the camera.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,442 @@
...
+        type: string
+        name: option
+        optional: true
+        mapping:
+          - dps_val: "1"
+            value: Event
+          - dps_val: "2"
+            value: Continuous
+          - dps_val: null
+            value: None
```
make-all: I would just leave this as null (Unknown), since the actual mode will be one of the above, the device has just not told us which.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,442 @@
...
+    icon: "mdi:bullhorn"
+    category: config
+    dps:
+      - id: 159
+        type: boolean
+        name: switch
+        optional: true
+        mapping:
+          - dps_val: null
+            value: false
```
make-all: Is it correct that the siren is always off when the device has not informed the state? I think it is better to leave it as null (Unknown)

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,442 @@
...
+          min: 10
+          max: 100
+  - entity: event
+    name: Alarm
+    dps:
+      - id: 185
+        type: base64
+        name: event
+        optional: true
+        persist: false
```
make-all: events are non-persistent by definition, so the `persist: false` is not needed.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,442 @@
...
+        mapping:
+          - dps_val: null
+            value: null
+          - value: message
+      - id: 212
+        type: base64
+        name: message
+        optional: true
+        persist: false
+  - entity: switch
```
make-all, line 343: Is it a switch, or a binary_sensor? There is already a motion_enable on dp 134.

tests/devices/test_emos_ip300_camera.py
```diff
@@ -0,0 +1,107 @@
+"""Tests for the EMOS IP-300 camera"""
```
make-all: Tests are not needed for devices that do not introduce new features. Given the number of devices supported now, the tests would overload the free tier on Github if we had them for every device.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
jkolo: This is a switch that activates the PIR sensor for floodlight.motion_enable is for motion tracking by the camera.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,434 @@
...
+        mapping:
+          - dps_val: null
+            value: false
+            hidden: true
+      - id: 115
+        type: base64
+        optional: true
+        sensitive: true
+        name: snapshot
+      - id: 185
```
make-all: Since it is exposed as an event below, it is not needed as an attribute here.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,434 @@
...
+            value: missing
+          - dps_val: null
+            value: normal
+      - id: 109
+        type: string
+        name: capacity
+        optional: true
+  - entity: button
+    name: Format SD card
+    icon: "mdi:sd"
```
make-all, line 130: This can also use `translation_key: status`, and be mapped to translated values: `formatting`, `unformatted`, `missing` and `fault`

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,434 @@
...
+        optional: true
+        mapping:
+          - dps_val: "1"
+            value: Event
+          - dps_val: "2"
+            value: Continuous
+          - dps_val: null
+            value: null
+  - entity: switch
+    name: Siren
```
make-all: This, combined with volume and duration below can be a siren entity.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,434 @@
...
+        mapping:
+          - dps_val: null
+            value: null
+          - value: alarm
+      - id: 185
+        type: base64
+        name: alarm_message
+        optional: true
+  - entity: event
+    name: Initiative message
```
make-all: I think Notification is a better name here.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,434 @@
...
+      - id: 212
+        type: base64
+        name: event
+        optional: true
+        mapping:
+          - dps_val: null
+            value: null
+          - value: message
+      - id: 212
+        type: base64
```
make-all: Likely the type is utf16b64 (base64 encoded UTF-16). Marking the type as this will recode it to UTF-8 so it displays in HA as text.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
```diff
@@ -0,0 +1,434 @@
...
+      - id: 185
+        type: base64
+        name: event
+        optional: true
+        mapping:
+          - dps_val: null
+            value: null
+          - value: alarm
+      - id: 185
+        type: base64
```
make-all: Likely the type is utf16b64 (base64 encoded UTF-16). Using that as the type will reencode as UTF-8 so it displays as text in the HA UI.

custom_components/tuya_local/devices/emos_ip300_camera.yaml
make-all: Forger that, this is not a sensor, the status is only provided as an attribute, which is OK, because this info is only interesting when performing format operations, which is rare.

PR #3889, opened 2025-10-25
Adds support for Catit Pixi Smart Feeder (Model: 43752)

colin-nolan:
- Catit branded pet feeder was not previously matching due to model missing a non-optional dps.
- Adds additional sensors seen on Catit feeder.

### DIFF
--- a/custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
+++ b/custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
@@ -7,11 +7,10 @@ products:
     manufacturer: GiotoHun
   - id: ojqmjnoxqygqldsp
     manufacturer: Puppy Kitty
-    name: Automatic pet feeder
   - id: s3rvixmeqx62vud5
     manufacturer: Catit
-    model: "43752"
-    name: Pixi Smart Feeder
+    model: Pixi Smart Feeder
+    model_id: "43752"
 entities:
   - entity: button
     name: Quick feed
@@ -34,18 +33,14 @@ entities:
         optional: true
   - entity: sensor
     icon: "mdi:paw"
-    name: Last meal size
+    name: Feed report
     category: diagnostic
     dps:
       - id: 15
         name: sensor
         type: integer
-        range:
-          min: 1
-          max: 12
         # Only present after feeding on Catit 43752
         optional: true
-        readonly: true
   - entity: button
     translation_key: factory_reset
     category: config
@@ -56,7 +51,7 @@ entities:
         optional: true
   - entity: number
     icon: "mdi:paw"
-    name: Feed meal
+    name: Manual feed
     dps:
       - id: 3
         name: value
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
```diff
@@ -22,18 +27,25 @@ entities:
...
     dps:
+      # Not thought to be used with Catit 43752
       - id: 1
         type: base64
         name: value
         optional: true
   - entity: sensor
     icon: "mdi:paw"
-    name: Feed report
+    name: Last meal size
```
make-all: Renaming entities after release causes problems for existing users, who may have automations referencing this.

custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
```diff
@@ -22,18 +27,25 @@ entities:
...
     category: diagnostic
     dps:
       - id: 15
         name: sensor
         type: integer
+        range:
```
make-all: Range is not necessary for a sensor.

custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
```diff
@@ -22,18 +27,25 @@ entities:
...
+          min: 1
+          max: 12
+        # Only present after feeding on Catit 43752
+        optional: true
+        readonly: true
```
make-all: Readonly is implied by the fact the entity is a sensor, so adding this is not necessary.

custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
```diff
@@ -44,15 +56,16 @@ entities:
         optional: true
   - entity: number
     icon: "mdi:paw"
-    name: Manual feed
-    category: config
+    name: Feed meal
```
make-all: Renaming after release causes problems

custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
```diff
@@ -66,6 +79,7 @@ entities:
           - dps_val: 2
             value: true
           - value: false
+        readonly: true
```
make-all: Readonly is implied by the fact this is a binary_sensor

custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
```diff
@@ -79,12 +93,14 @@ entities:
           - dps_val: 4
             value: true
           - value: false
+        readonly: true
```
make-all: Implied

colin-nolan:
@make-all Many thanks for reviewing, and making necessary changes.

I did wonder whether the name changes were backwards compatible. I reasoned that the config was probably only used on setup, and hence wouldn't break existing (i.e. sensors setup with an older version would maintain their existing IDs). Sounds like I was wrong there - apologies for that. My motivation was to make the names a little more accurate.

PR #3894, opened 2025-10-26
Nexsmart Devices (Watch 2, Air 2, Air 3)

creoludifico:
Camera Device from Nexsmart. USB connected camera. No battery and therefor always on. This integration does NOT support Camera directly from Tuya - For that you will use RTSP This only expose all sensors
- I plan on importing the other types of cameras from nexsmart

### DIFF
--- a/custom_components/tuya_local/devices/nexsmart_air3_camera.yaml
+++ b/custom_components/tuya_local/devices/nexsmart_air3_camera.yaml
@@ -309,4 +309,3 @@ entities:
             value: Card error
           - dps_val: null
             value: Formatted
-


custom_components/tuya_local/devices/nexsmart_watch_2_camera.yaml
```diff
@@ -40,15 +40,21 @@ entities:
   # ------------------------------------
   # 2. Basic Configuration Switches/Lights (101, 103, 104)
   # ------------------------------------
-  - entity: light
-    translation_key: indicator
+  - entity: switch
```
make-all: Why make this change?

creoludifico: I just thought it was more intuitive with a switch instead of the "lightbulb" buttons, because it doesnt light up anything (only the small blue indicator light).

I'm currently also configuring nexsmart_air_3 and that has a "floodlight" where it's actually light, so that's why i changed it, to not get confused between actually lights vs a blue dot on the camera :)

make-all: I think the way to differentiate lights like floodlights from small indicator LEDs is to mark the latter as config (as it is more a configuration preference than a user control). Ultimately it is still a light, and you even gave it a lightbulb as an icon (the indicator translation_key uses an LED as an icon to better fit an indicator LED).

creoludifico: Ok ill make some changes.

Do you prefer i add all the 3 cams config in this PR? Or make 3 separate PRs? They are all from nexsmart

make-all: Separate PRs is better, otherwise issues with any one config will delay all of them.

creoludifico:
All 3 works! And all DP's for all devices works too! So dont mind the commit comment ^^
My mistake was to add them under v3.2 instead of the intended v3.3 :)

custom_components/tuya_local/devices/nexsmart_air_2.yaml
make-all: If it matches wireless_awake, then most likely this is useless as a local sensor, and you should instead move it to an attribute of some other entity. If you insist on keeping it, then connectivity may be a better fit for class.

make-all: This may be better as a binary_sensor with `class: battery_charging` or `plug`

custom_components/tuya_local/devices/nexsmart_air_2.yaml
```diff
@@ -283,3 +276,4 @@ entities:
         type: raw
```
make-all: Tuya type raw usually corresponds to a tuya-local `type: base64`. But that is not directly useful as a sensor, so if you can't decode it further, maybe move it to an attribute of another entity.

Messages are often the more specialized `type: utf16b64`, which is a UTF-16BE string encoded within a base64 string. These often act like events rather than sensors, with the message coming as an event occurs, then disappearing.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
```diff
@@ -62,13 +62,6 @@ entities:
...
-        mapping:
-          - dps_val: true
-            value: "on"
-            value_on: Awake
-          - dps_val: false
-            value: "off"
-            value_off: Sleeping
 
   - entity: number
     name: Low Battery Alarm Threshold
```
make-all: HA naming convention says only the first word in an entity name should be capitalized.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
make-all: Top level name should be without branding. Branding should only be under products.

make-all: Remove the name, the `class: battery` comes with translated entity names already.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
```diff
@@ -0,0 +1,279 @@
...
+    dps:
+      - id: 147
+        type: integer
+        name: value
+        unit: "%"
+        # Range is 10-50 based on device dump
+        range:
+          min: 10
+          max: 50
+          step: 1
```
make-all: step: 1 id not needed, but if included, step does not go under range, but inside a mapping.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
```diff
@@ -0,0 +1,279 @@
...
+
+  - entity: sensor
+    name: Battery report capability
```
make-all: Probably just put this as an attribute under the battery sensor. It's value as an integer is not user facing.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
creoludifico: Actually this is very usefull for doing AI simulations of the images. Because on the v2 cameras only "wireless_awake" change when it detects movement, so ill keep it and change it to "connectivity"

make-all: It seems it is misnamed then. We do not need to follow the mistakes of the manufacturer, if it fits a `class: motion` better in practice, then we should use that.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
```diff
@@ -0,0 +1,237 @@
+name: AIR 2 Camera (Battery)
```
make-all: Top level names should be generic, without any branding. Also avoid punctuation characters in device (and entity) names, as they get used as ids in places.

custom_components/tuya_local/devices/nexsmart_air2_camera.yaml
```diff
@@ -0,0 +1,237 @@
...
+    manufacturer: Nexsmart
+    model: AIR 2
+entities:
+  # ------------------------------------
+  # 1. Main Camera & Snapshot
+  # ------------------------------------
+  # USE RTSP
+
+  - entity: sensor
+    name: Latest snapshot
```
make-all: I think it would be better to use a camera for this if it is an actual snapshot, or an event if it is a descriptor of where to download the image.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
```diff
@@ -0,0 +1,237 @@
...
+    class: battery
+    category: diagnostic
+    # Battery Level Sensor (matches wireless_electricity: 145)
+    dps:
+      - id: 145
+        type: integer
+        name: sensor
+
+  - entity: binary_sensor
+    name: Power source
```
make-all: This should not need a name or icon. The values in the mapping below need to be true and false.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
```diff
@@ -0,0 +1,237 @@
...
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: "0"
+            value: Battery
+          - dps_val: "1"
+            value: Plugged in / Charging
+
+  - entity: binary_sensor
+    name: Device state
```
make-all: This should not need a name or icon (the icon provided is counterintuitive, as it would apply to the false state.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
```diff
@@ -0,0 +1,237 @@
...
+    category: diagnostic
+    icon: "mdi:sleep"
+    # Awake/Sleep Status (matches wireless_awake: 149)
+    dps:
+      - id: 149
+        type: boolean
+        name: sensor
+
+  - entity: number
+    name: Low Battery Alarm Threshold
```
make-all: Follow HA naming conventions. Only first word should be capitalised.

custom_components/tuya_local/devices/nexsmart_air_2.yaml
```diff
@@ -0,0 +1,237 @@
...
+    icon: "mdi:flip-horizontal"
+    # Flip Image Switch (matches basic_flip: 103)
+    dps:
+      - id: 103
+        type: boolean
+        name: switch
+        optional: true
+
+  - entity: switch
+    name: Watermark (OSD)
```
make-all: Again, avoid brackets. "Watermark" alone is enough (though "Timestamp" slightly beats it for number of occurences in other configs).

custom_components/tuya_local/devices/nexsmart_air_3.yaml
```diff
@@ -0,0 +1,319 @@
...
+    icon: "mdi:bell-alert"
+    category: config
+    # Audible Alarm Switch (matches ipc_audible_alarm: 242)
+    dps:
+      - id: 242
+        type: boolean
+        name: switch
+
+  - entity: switch
+    name: Siren
```
make-all: switch and volume can be combined into a siren entity.

custom_components/tuya_local/devices/nexsmart_air2_camera.yaml
creoludifico: There is sadly no way to get the stream from the local device. I've tried multiple things, and only thing i found out was to get the stream through RTSP or the Cloud. Ill remove the sensor because it's actually created from the cloud (115)

PR #3898, opened 2025-10-27
Add device EARU K6RH3A-WiFi smart thermostat

MiguelDomingues:
This device is a slightly different from the existing one at [custom_components/tuya_local/devices/earu_k6h3aw_thermostat.yaml](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/earu_k6h3aw_thermostat.yaml)

Below is a screenshot of the device after adding it in the integration:
image


MiguelDomingues:
@make-all I've removed the duplicated entity 'Working week' can you trigger the workflows again please

make-all:
I think this can be merged with the existing config

MiguelDomingues:
Don't think it can be merged without breaking the existing one.

Besides several differences in the range values (min/max), there is this one that, in my opinion, could break the existing device: id 1 (mode) in existing device has program/manul/eco, while this new device has auto/manual/eco

make-all:
Did you try the updated config?

The preset_mode is actually the same, auto is mapped to program to take advantage of existing translations.
The only real differences are:
- slight differences in additional entities, taken care of by putting them all in as optional and making them hidden if unavailable.
- different capitalisation of ALL/all in sensor selection enum.
- difference in upper temperature range. Initially I ignored this because I am not sure anyone really needs to set their thermostat between 90 and 95 degrees F, but since the logic to detect which model it is already exists for the ALL/all handling, I just changed it to reuse that logic for the temperature range as well.

PR #3902, opened 2025-10-28
Fix (time): update hour before updating minute for masked dps with the same id

SirLyoshyk:
**Issue:** If a time entity uses the same DPS ID with a masked value for hour and minute or minute and second, you are not able to update the hour or minute value, because the following minute or second value overrides the updated value

**Reason:** The updated hour in masked value is overridden during the minute value processing.

**Config example:**
```
  - entity: time
    category: config
    name: "Start time"
    dps:
      - id: 106
        optional: true
        type: base64
        mask: "00FF00"
        name: hour
        range:
          min: 0
          max: 23
      - id: 106
        optional: true
        type: base64
        mask: "0000FF"
        name: minute
        range:
          min: 0
          max: 59
```

**Proposed solution:** in case if we have such configuration, set hour value to device before processing minute value

### DIFF
--- a/custom_components/tuya_local/time.py
+++ b/custom_components/tuya_local/time.py
@@ -105,11 +105,31 @@ class TuyaLocalTime(TuyaLocalEntity, TimeEntity):
         seconds = value.second
         if self._hour_dps:
             settings.update(self._hour_dps.get_values_to_set(self._device, hours))
+
+            # If hour dps is using the same masked ID as minute dps, it's important to update hours before processing minute
+            if (
+                self._minute_dps
+                and self._hour_dps.id == self._minute_dps.id
+                and self._hour_dps.mask
+                and self._minute_dps.mask
+            ):
+                await self._device.async_set_properties(settings)
+                settings = {}
         else:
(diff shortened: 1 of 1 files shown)


make-all:
I think this is a general issue with masked dps that should be fixed at a lower level. For now I will leave this open as a reminder, but I will look at fixing this a different way that does not require splitting into multiple commands.

SirLyoshyk:
Yes, I thought about handling it during schedule changes in the queue (before taking the value from the device, check for an updated value in the queue), but I'm not very familiar with Python.
So far, I'm using this fix as a workaround and looking forward to a more elegant solution from you.

make-all:
If you could try the latest main branch, it would help confirm that the more general fix is working as intended before the next release. The changes were in helpers/device_config.py and the entity classes (time.py in the case of your specific use case) so just taking those files on top of 2025.10.1 would also work for testing it.

SirLyoshyk:
Thank you!
I can confirm the fix works for my case. Time settings are updated correctly.

PR #3904, opened 2025-10-28
Add support Humidifier for Deerma DEM-F995W

SandDeath:
Tested features:

Power control
Target humidity
Auto/manual mode
Spray level control (fan entity)
Sleep mode switch
LED and ionizer switches

worked device in HA


SandDeath:
It's very sad to see you've decided to drag politics into an open source project for some personal gain. It's incredibly sad, and it's unclear what you're trying to prove, and most importantly, who you're trying to prove it to. However, I still commend you as a developer for providing a tool like Tuya Local. You're doing excellent work for the benefit of society. However, I can't share your position.

make-all:
There is no personal gain here for me. Open Source is inherently political as a concept, so it is strange to face demands to keep politics out of Open Source.

I can understand you being upset because this adversely affects you personally, but the policy is not aimed at you as an individual.

PR #3907, opened 2025-10-30
Add support for Pro Breeze D-23 Dehumidifier

8none1:
image

image

### DIFF
--- a/custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
@@ -39,7 +39,7 @@ entities:
         name: sensor
         unit: C
   - entity: select
-    name: Fan Speed
+    name: Fan speed
     icon: "mdi:fan"
     category: config
     dps:
@@ -52,7 +52,7 @@ entities:
           - dps_val: "3"
             value: "High"
   - entity: switch
-    name: Ventilation Mode
+    name: Ventilation mode
     icon: "mdi:fan"
     category: config
     dps:


custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -0,0 +1,111 @@
...
+    class: dehumidifier
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 2
+        type: string
+        name: mode
+        mapping:
+          - dps_val: 0
```
make-all: These need to be quoted if they are strings, otherwise the yaml parser will convert them to integers.

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -0,0 +1,111 @@
...
+          max: 80
+      - id: 104
+        type: integer
+        name: current_humidity
+      - id: 103
+        type: integer
+        name: current_temperature
+
+  - entity: sensor
+    name: Temperature
```
make-all: Do not use an explicit name when there is a class of the same name, as it will prevent the HA translations from being used.

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -0,0 +1,111 @@
...
+        type: boolean
+        name: switch
+      - id: 2
+        type: string
+        name: mode
+        mapping:
+          - dps_val: 0
+            value: auto
+          - dps_val: 1
+            value: max
```
make-all: Use `boost`, as that has a translation in the default HA humidifier strings.

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -0,0 +1,111 @@
...
+        name: humidity
+        range:
+          min: 30
+          max: 80
+      - id: 104
+        type: integer
+        name: current_humidity
+      - id: 103
+        type: integer
+        name: current_temperature
```
make-all: Not needed here, as there is a sensor below

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -0,0 +1,111 @@
...
+    name: Ventilation Mode
+    icon: "mdi:fan"
+    category: config
+    dps:
+      - id: 5
+        type: boolean
+        name: switch
+
+  - entity: switch
+    name: LED Display
```
make-all: Use a light entity, with `translation_key: display` instead of name and icon

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -0,0 +1,111 @@
...
+    icon: "mdi:led-on"
+    category: config
+    dps:
+      - id: 102
+        type: boolean
+        name: switch
+
+  - entity: switch
+    name: Child Lock
```
make-all: Use a lock entity with `translation_key: child_lock`

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -0,0 +1,111 @@
...
+    icon: "mdi:lock"
+    category: config
+    dps:
+      - id: 7
+        type: boolean
+        name: switch
+
+  - entity: binary_sensor
+    name: Defrost Mode
```
make-all: Use `translation_key: defrost`, class is not needed as icon and strings will be provided by the translation.

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -0,0 +1,111 @@
...
+    class: cold
+    category: diagnostic
+    dps:
+      - id: 105
+        type: boolean
+        name: sensor
+
+  - entity: binary_sensor
+    name: Tank Problem
```
make-all: Use `translation_key: tank_full`, and add another unnamed `class: problem` binary_sensor to catch all the other non-zero values. The general problem sensor should also expose the original value as `fault_code` (search for tank_full in the other dehumidifier configs to find an example to follow).

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -0,0 +1,111 @@
+name: Pro Breeze D-23 Dehumidifier
```
make-all: Top level name should be without branding (all branding should be under products).

8none1:
Thank you for the review, I've incorporate all the requested changes (I think).

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -67,45 +59,57 @@ entities:
...
     dps:
       - id: 11
         type: bitfield
         name: sensor
+        optional: true
         mapping:
+          - dps_val: 0
+            value: false
           - dps_val: 1
             value: false
```
make-all, line 106: 1 is not captured anywhere else. It is intentionally excluded from the problem sensor? Or a duplicate of the separate defrost sensor (which is the other "normal operation" flag that dehumidifiers might have in here)

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
```diff
@@ -12,17 +11,17 @@ entities:
         type: boolean
         name: switch
       - id: 2
-        type: string
+        type: integer
```
make-all: Most commonly these would be strings (numbers but surrounded by quotes in the tuya protocol documented as type enum). It is not unheard of for tuya devices to use actual integers (unquoted, documented as type value) for discrete values, but it is not common.

Currently this integration is lenient about the difference between these cases, as in the early days before tuya developer portal existed, people would use various means to gather info, some of which stripped the quotes from strings, but if in future it becomes necessary to become strict about types, this device might no longer be detected if the type is incorrect.

8none1: Thank you for this info.  I've written a little script to dump the raw data and indeed, it is a string.  Fixing now...

custom_components/tuya_local/devices/probreeze_pbd23w_dehumidifier.yaml
8none1: I'm not sure what to do here.  `1` seems to be the default state and means "everything is fine".  I haven't seen it be a `0`, but looking at other devices suggests that normally `0` is normal.  So I captured both. 

I suppose `0` should be removed if anything.  

What do you think?

PR #3916, opened 2025-10-31
Added the Flinq Smart Multicolor Light

tomvansambeek:


tomvansambeek:
This is a series of light bulbs on a string which have a on/off, brightness and scene setting. The color of the led's is defined by the selected scene.

PR #3922, opened 2025-11-01
Add support for Cat litter MINI-B-Public

amotruk:
add support for cat litter https://www.aliexpress.com/i/1005009671893724.html

### DIFF
--- a/custom_components/tuya_local/devices/minib_public_litterbox.yaml
+++ b/custom_components/tuya_local/devices/minib_public_litterbox.yaml
@@ -2,7 +2,6 @@ name: Litter box
 
 products:
   - id: "93at0k76rfvteseb"
-    manufacturer: Tuya
     model: MINI-B-Public
 
 entities:


custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
+name: MINI-B-Public
```
make-all: Top level name should be unbranded, and clearly identify the device type (Litter box)

custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
...
+    dps:
+      - id: 2
+        type: string
+        name: option
+        mapping:
+          - dps_val: auto_clean
+            value: Auto clean
+          - dps_val: manual_clean
+            value: Manual clean
+  - entity: switch
```
make-all: Should probably be a button, since you can't undo this operation

custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
...
+    name: Empty
+    icon: "mdi:trash-can-outline"
+    dps:
+      - id: 3
+        type: boolean
+        name: switch
+  - entity: number
+    name: Cleaning delay time
```
make-all: Try to keep the entity names short, time is redundant after delay.

custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
...
+    icon: "mdi:timer"
```
make-all: `translation_key: timer` and `class: duration` would be better than setting an explicit icon

custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - id: 5
+        type: integer
+        name: value
+        unit: min
+        range:
+          min: 1
+          max: 60
+  - entity: sensor
+    name: Cat weight
+    icon: "mdi:scale"
```
make-all: `class: weight` would be better than the explicit icon. The name may even be considered redundant, as there is no other weight sensor to confuse it with. The advantage would be the use of HA built in translations, at the expense of not specifically mentioning Cat.

custom_components/tuya_local/devices/minib_public_litterbox.yaml
```diff
@@ -0,0 +1,133 @@
...
+        optional: true
+  - entity: switch
+    name: Clean
+    icon: "mdi:broom"
+    dps:
+      - id: 9
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Reset
```
make-all, line 92: Another one that would be better as a button.

custom_components/tuya_local/devices/minib_public_litterbox.yaml
```diff
@@ -0,0 +1,133 @@
...
+    name: Toilet duration
+    icon: "mdi:timer-outline"
+    dps:
+      - id: 8
+        type: integer
+        name: sensor
+        unit: s
+        optional: true
+  - entity: switch
+    name: Clean
```
make-all, line 84: Better as a button

custom_components/tuya_local/devices/minib_public_litterbox.yaml
```diff
@@ -0,0 +1,133 @@
...
+    name: Times today
+    icon: "mdi:counter"
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
+        optional: true
+  - entity: sensor
+    name: Toilet duration
+    icon: "mdi:timer-outline"
```
make-all: `class: duration` would be better than an explicit icon.

custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
...
+        type: boolean
+        name: switch
+  - entity: sensor
+    name: Cleaning times
+    icon: "mdi:counter"
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+  - entity: number
```
make-all: This would be better as a light, with `translation_key: backlight`, and the dp name `brightness`

custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
...
+    dps:
+      - id: 106
+        type: integer
+        name: value
+        unit: "%"
+        range:
+          min: 0
+          max: 100
+  - entity: sensor
+    name: Message
```
make-all: Probsbly better as an event. Does it not need to be marked optional? Normally such dps are only present when there is a relevant event to trigger the message.

custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
...
+    icon: "mdi:message-alert-outline"
+    category: diagnostic
+    dps:
+      - id: 108
+        type: string
+        name: sensor
+  - entity: sensor
+    name: Device status
```
make-all: It would be better to list all the possible values, use `class: enum` and `translation_key: status` instead of the explicit name.

custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
...
+    icon: "mdi:state-machine"
+    dps:
+      - id: 110
+        type: string
+        name: sensor
+  - entity: select
+    name: Unit setup
```
make-all: `Weight unit` might be a better name

custom_components/tuya_local/devices/mini_b_public.yaml
```diff
@@ -0,0 +1,133 @@
...
+entities:
+  - entity: switch
+    name: Power
+    icon: "mdi:power"
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+  - entity: select
+    name: Work mode
```
make-all: Just *Mode* would be better. The naming does not need to match exactly what Tuya has used, which is sometimes not quite right.

custom_components/tuya_local/devices/minib_public_litterbox.yaml
make-all, line 1: filename should follow brand_model_type pattern, so minib_public_litterbox.yaml

custom_components/tuya_local/devices/minib_public_litterbox.yaml
```diff
@@ -42,17 +48,18 @@ entities:
...
   - entity: sensor
-    name: Cat weight
-    icon: "mdi:scale"
+    class: weight
     dps:
       - id: 6
         type: integer
         name: sensor
-        unit: kg
+        unit: KG
```
make-all: Lower case was correct

amotruk: fixed

custom_components/tuya_local/devices/minib_public_litterbox.yaml
```diff
@@ -0,0 +1,170 @@
+name: Litter box
+
+products:
+  - id: "93at0k76rfvteseb"
+    manufacturer: Tuya
```
make-all: Leave the manufacturer out if you do not know. Tuya does not manufacture end products, just the WiFi modules.

make-all (approved):

PR #3924, opened 2025-11-01
Add support for Devola Designer Glass Panel Heater.

CCCowley:
Adds support for the Devola Designer Glass Panel Heater.

This device is almost identical to the Devola Intelligent Heater, which already has an existing config and shares a product ID, with some small changes:

- preset_mode value for antifrost mode is now af
- temperature is constrained to 5 degrees in antitrust mode
- temperature range is extended

All changes have been extracted from and tested in Tuya developer portal.

**Additional details:**

- Product webpage: [https://www.devola.co.uk/products/devola-designer-2kw-smart-glass-panel-heater-with-timer-black-dvpw2000b](https://www.devola.co.uk/products/devola-designer-2kw-smart-glass-panel-heater-with-timer-black-dvpw2000b)
- Product manual: [https://www.devola.co.uk/cdn/shop/files/dvpw-instruction-manual.pdf?v=17064640317587403520](https://www.devola.co.uk/cdn/shop/files/dvpw-instruction-manual.pdf?v=17064640317587403520)

### DIFF
--- a/custom_components/tuya_local/devices/devola_designer_glass_panel_heater.yaml
+++ b/custom_components/tuya_local/devices/devola_designer_glass_panel_heater.yaml
@@ -1,0 +1,61 @@
+name: Heater
+products:
+  - id: laagtqcj40q1uwg8
+    manufacturer: Devola
+    model: Designer Glass Panel Heater
+entities:
+  - entity: climate
+    translation_key: heater
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: true
+            value: heat
+          - dps_val: false
+            value: "off"
+      - id: 2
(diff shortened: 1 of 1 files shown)


make-all:
This config is identical except for the ridiculous temperature range to the andersson_gsh_heater, so I don't think the config is necessary. It is especially concerning that the manufacturer is recycling product ids for not quite compatible products.

CCCowley:
Thanks for the feedback.

Agree with all your points, I'm guessing they're both different brands selling the same white labelled product. 

The constraint on the antifrost preset is missing from the andersson_gsh_heater, which is fairly important as it is restricted in Tuya/on the device. Would you prefer the Andersson config be edited to include that? Obviously I am unable to test with that specific device.

Issue #2289, opened 2024-09-10
Unstable connections to sub devices since 2024.9.0


josiasmontag:
Since version 2024.9.0 the connection to my RainPoint water timers (#2180) is unstable. The actual water timers are sub devices of a hub.

Directly after restarting HA all entities did show as unavailable. After reloading the Tuya Local integration, the entities of one device started working but the other device was still unavailable. Reloading the integration multiple times resulted in random results, sometimes one device or the other device worked.

The debug log only shows:

```
2024-09-09 18:16:31.506 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 18:26:54.824 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 18:52:17.908 ERROR (MainThread) [custom_components.tuya_local.device] Failed to fetch device status for Bewässerung Hecke
2024-09-09 18:59:41.062 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 19:00:38.378 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 19:01:23.875 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 19:05:00.892 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 19:06:33.270 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 19:18:30.521 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 19:22:21.634 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 19:57:52.454 WARNING (MainThread) [custom_components.tuya_local.device] Bewässerung Beete error reading: Network Error: Unable to Connect
2024-09-09 20:05:52.219 ERROR (MainThread) [custom_components.tuya_local.device] Failed to fetch device status for Bewässerung Beete
[...]

PR #3929, opened 2025-11-03
Synchronize accesses to sub-devices

darchons:
When multiple sub-devices are under the same parent device, we need to make sure to access only one sub-device at a time, because the underlying tinytuya device for the parent cannot handle multiple acceses at the same time. This bug often manifests as "No local key for device!" in the tinytuya logs.

This commit adds a lock that is associated with the parent device, and acquires the lock whenever we access a sub-device or the parent device itself.

Issue #2289

### DIFF
--- a/custom_components/tuya_local/device.py
+++ b/custom_components/tuya_local/device.py
@@ -78,22 +78,33 @@ class TuyaLocalDevice(object):
             if dev_cid:
                 if hass.data[DOMAIN].get(dev_id) and name != "Test":
                     parent = hass.data[DOMAIN][dev_id]["tuyadevice"]
+                    parent_lock = hass.data[DOMAIN][dev_id]["tuyadevicelock"]
                 else:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/device.py
```diff
@@ -380,6 +392,8 @@ async def async_receive(self):
                 if self._api.parent:
                     self._api.parent.set_socketPersistent(False)
                 await asyncio.sleep(5)
+            finally:
```
make-all, line 395: I think the lock should not be held while sleeping, otherwise it is held most of the time, only the brief period when the loop is restarting can another device with the same parent get a message. This means up to a 5 second delay for each subdevice connected to the hub before a device can react.

darchons: Thanks! I updated the commit to call `release` before the `sleep` calls (for simplicity I moved the last `sleep` call outside of the `try` block)

PR #3937, opened 2025-11-04
Incude Current Temp in Water Heater Entity and Map Fault Codes

Waz-Cpt:
Inclusion of the current temperature in the water heate entity so that the value can be displayed as it is an attribute and is needed on some cards.

Added the fault mapping. Fault mapping was gathered and translated from the JSON output gathered from Tuya IoT Dev Platform under the API Explorer/Device Control/Query Things Data Model

[EMS Waterheater.json](https://github.com/user-attachments/files/23341236/EMS.Waterheater.json)

### DIFF
--- a/custom_components/tuya_local/devices/ems_waterheater.yaml
+++ b/custom_components/tuya_local/devices/ems_waterheater.yaml
@@ -25,10 +25,10 @@ entities:
       - id: 3
         type: integer
         name: current_temperature
-        unit: C
       - id: 2
         type: integer
         name: temperature
+        unit: C
         range:
           min: 10
           max: 60
@@ -39,7 +39,6 @@ entities:
   - entity: sensor
     # 2025-11-05
     deprecated: water_heater.current_temperature
-    name: Current Temperature
     class: temperature
     dps:
       - id: 3
@@ -61,14 +60,9 @@ entities:
       - id: 9
         type: bitfield
         name: fault_code
-  - entity: sensor
-    name: Fault Description
-    class: enum
-    category: diagnostic
-    dps:
       - id: 9
         type: bitfield
-        name: sensor
+        name: description
         mapping:
           - dps_val: 0
             value: ok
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/ems_waterheater.yaml
```diff
@@ -51,7 +56,48 @@ entities:
         mapping:
           - dps_val: 0
             value: false
+          - dps_val: null
+            value: false
           - value: true
       - id: 9
         type: bitfield
         name: fault_code
+  - entity: sensor
```
make-all: Just make this an attribute "description" on the sensor above.

If you needed to add a null mapping, you will also need to mark all occurences of id: 9 as `optional: true`

custom_components/tuya_local/devices/ems_waterheater.yaml
```diff
@@ -34,6 +38,7 @@ entities:
         name: mode
         hidden: true
   - entity: sensor
+    name: Current Temperature
```
make-all: Do not rename this entity. Firstly it will prevent translations from being used. Second it will interfere with any existing users setups.

Since the temperature reading is moving to the water_heater, mark this as `deprecated: water_heater.current_temperature` with a comment noting the date

custom_components/tuya_local/devices/ems_waterheater.yaml
```diff
@@ -22,6 +22,10 @@ entities:
                 value: high_demand
               - dps_val: electric
                 value: electric
+      - id: 3
+        type: integer
+        name: current_temperature
+        unit: C
```
make-all: water_heater only has one `temoerature_unit` attribute, so there is no need to state it on both temperature and current_temperature.

Waz-Cpt: Will do, is there a preference on which temp element should have hte unit?

custom_components/tuya_local/devices/ems_waterheater.yaml
Waz-Cpt: Hopefully I understand. I have created an updated draft.

custom_components/tuya_local/devices/ems_waterheater.yaml
Waz-Cpt: Done

PR #3938, opened 2025-11-04
Add Support for Rohnson R9920 Dehumidifier

AverageJoe89:
Add support for Rohnson R-9920 Genius 20 Ion & UV-C - https://www.rohnson.cz/r-9920/

Most of features works correctly.

**Known issues:**
Timer does not work (ID 12 for time value, ID 101 should be boolean for on/off timer).
Problem reporting - only know error code 8 is for empty/ejected water tray, rest I did not catch.
Unknown binary sensor on ID 105 - accroding to this source https://www.cena-vykon.cz/recenze-rohnson-r-9920/

### DIFF
--- a/custom_components/tuya_local/devices/rohnson_r9920_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/rohnson_r9920_dehumidifier.yaml
@@ -1,0 +1,190 @@
+name: Dehumidifier
+products:
+  - id: jsrcliohsha5y9kg
+    manufacturer: Rohnson
+    model: R-9920 Genius 20
+    model_id: R-9920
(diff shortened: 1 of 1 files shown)


make-all:
This seems identical to probreeze_pbd23w_dehumidifier.yaml which was added in #3907and merged just a few days ago.

Comparing the configs, only the timer is missing from that, but since you say it is not working, I don't think that is something that needs to be added.

AverageJoe89:
Product id is different, what's correct approach? Add multiple product ids under `probreeze_pbd23w_dehumidifier.yaml`? Or something else?

make-all:
I did that already

PR #3939, opened 2025-11-05
Update catit_pixi_smart_feeder.yaml

fbesserp:
Add outlet power status (id: 105) to the catit pixi smart feeder. This bit corresponds to the used power source:
- 0: Power outlet (USB-C).
- 1: Batteries.

The condition was negated, as the monitor point name is Outlet power.

### DIFF
--- a/custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
+++ b/custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
@@ -150,7 +150,6 @@ entities:
         # Present at all times on Catit 43752 but unsure about other products
         optional: true
   - entity: binary_sensor
-    name: Plug
     category: diagnostic
     class: plug
     dps:


custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
```diff
@@ -149,6 +149,19 @@ entities:
         type: string
         # Present at all times on Catit 43752 but unsure about other products
         optional: true
+  - entity: binary_sensor
+    name: Outlet Power
```
make-all: A `class: plug` (or perhaps `battery_charging`) would be better than this explicit name, to use HA built in translations for both entity name and state.

fbesserp: Agree with your suggestion. Monitoring point class was set to plug, as the device uses D size alkaline batteries.

custom_components/tuya_local/devices/catit_pixi_smart_feeder.yaml
```diff
@@ -150,8 +150,9 @@ entities:
         # Present at all times on Catit 43752 but unsure about other products
         optional: true
   - entity: binary_sensor
-    name: Outlet Power
+    name: Plug
```
make-all: Remove the name to let the class translations shine through.

Issue #3941, opened 2025-11-05
Request support for Thermostat Mini Series-Mini04 Plus


dtaylor7:
### Log message

```text
Adding Mini04 Plus device with product id qnkkma3qcqzmchx4

DPS:
{
  "updated_at": 1762380426.7773983,
  "4": "Standby",
  "7": false,
  "101": 45,
  "102": 21,
  "103": 21,
  "104": 2,
  "105": 120,
  "106": false
}
```

### Product ID

qnkkma3qcqzmchx4

### Product Name

Thermostat Mini Series-Mini04 Plus

### DPS information

```json
{
  "modelId": "000004smod",
  "services": [
    {
      "actions": [],
      "code": "",
      "description": "",
      "events": [],
      "name": "默认服务",
      "properties": [
        {
          "abilityId": 4,
          "accessMode": "rw",
          "code": "mode",
          "description": "None = 0 为枚举0 模式未使用",
          "name": "工作模式",
          "typeSpec": {
            "type": "enum",
            "range": ["None", "Standby", "Manual", "LST", "Setback", "Comfort", "Program", "RBT"]
          }
        },
        {
          "abilityId": 7,
          "accessMode": "rw",
          "code": "child_lock",
          "description": "",
          "name": "童锁",
          "typeSpec": { "type": "bool" }
        },
        {
          "abilityId": 23,
          "accessMode": "rw",
          "code": "week_program_mode",
          "description": "RAW数据格式：\n每⼀⼩时可设置⼀个模式，可设置的模式根据dp4 mode ⼯作模式的枚举值来显示（仅包含Standby、Manual、LST、Setback、Comfort），每位都是⼀个⼗六进制的值，以第⼀天的数据为例，100000000000000000000000，代表星期⼀第⼀个⼩时有模式设置，设置模式显示是根据dp 4 mode的枚举值来的，如图所示Standby = 1; Manual = 2; Lst = 3; Setback = 4; Comfort = 5",
          "name": "编程模式设置",
          "typeSpec": { "type": "raw", "maxlen": 128 }
        },
        {
          "abilityId": 101,
          "accessMode": "rw",
          "code": "manual_temperature",
          "description": "",
          "name": "手动模式温度设置",
          "typeSpec": { "type": "value", "max": 70, "min": 45, "scale": 0, "step": 5, "unit": "℃" }
        },
        {
          "abilityId": 102,
          "accessMode": "rw",
          "code": "comfort_temperature",
          "description": "",
[...]

PR #3942, opened 2025-11-06
Add support for Thermostat Mini Series-Mini04 Plus

dtaylor7:
Solves #3941

Add support for Thermostat Mini Series-Mini04 Plus - https://etowelthermostats.com/products/thermostat-mini-series-1?variant=45082151715042

### DIFF
--- a/custom_components/tuya_local/devices/etowel_mini04_thermostat.yaml
+++ b/custom_components/tuya_local/devices/etowel_mini04_thermostat.yaml
@@ -196,6 +196,7 @@ entities:
                 value: false
           - value: true
   - entity: binary_sensor
+    name: Window 3
     class: window
     hidden: unavailable
     dps:


custom_components/tuya_local/devices/etowel_mini04_thermostat.yaml
```diff
@@ -0,0 +1,45 @@
...
+    manufacturer: eTowel
+    model: Mini04 Plus
+entities:
+  - entity: climate
+    translation_only_key: thermostat
+    dps:
+      - id: 4
+        name: hvac_mode
+        type: string
+        mapping:
```
make-all, line 13: The full range for this dp is ["None", "Standby", "Manual", "LST", "Setback", "Comfort", "Program", "RBT"]

These all need to be covered (a default mapping here should be enough for hvac_mode, with an additional preset_mode mapped to the same dp giving access to the full range)
 ```
    - value: heat
      hidden: true
```

File: custom_components/tuya_local/devices/etowel_mini04_thermostat.yaml [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+          - dps_val: 'Standby'
+            value: 'off'
+          - dps_val: 'Manual'
+            value: heat
+      - id: 101
+        name: temperature
```
make-all, line 57: This is documented as `manual_temperature`, so only valid for the Manual preset mode? There will need to be logic to redirect to `comfort_temperature` when appropriate.

make-all (commented):
Config is missing dps 102, 104, 105 that are showing in the log, and 23, 107, 108. 109, 110, 111, 112 that are not showing in the log, but may be useful as optional entities nonetheless.

dtaylor7:
@make-all, thanks for sorting this. I was getting there... slowly... how did you manage to do this without testing on a real device?

PR #3958, opened 2025-11-09
Improvements to EARU K6RH3A-WiFi

MiguelDomingues:
- Fixes Hysteresis step
- ~~Add binary sensor for valve~~

### DIFF
--- a/custom_components/tuya_local/devices/earu_k6h3aw_thermostat.yaml
+++ b/custom_components/tuya_local/devices/earu_k6h3aw_thermostat.yaml
@@ -300,16 +300,3 @@ entities:
           - dps_val: null
             value: false
           - value: true
-  - entity: binary_sensor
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/earu_k6h3aw_thermostat.yaml
```diff
@@ -299,3 +300,16 @@ entities:
           - dps_val: null
             value: false
           - value: true
+  - entity: binary_sensor
```
make-all: This info is already available from the hvac_action on the climate entity, so this sensor is not needed.

MiguelDomingues: Missed that it was there as an attribute. Will remove it

Issue #3166, opened 2025-04-20
Request support for Weather station


Zylazy:
### Log message

```text
2025-04-20 07:44:11.862 WARNING (MainThread) [custom_components.tuya_local.config_flow] Adding Weather stations device with product id wkbc51nj8mqiws0n
2025-04-20 07:44:12.097 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device specification:
[...]

PR #3963, opened 2025-11-11
Add missing/broken entities to weather station

veista:
Adding missing/broken entities to weather station like wind direction and outdoor battery. Also cleaned the weather station entities, so that entities are in correct categories.

### DIFF
--- a/custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
+++ b/custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
@@ -439,6 +439,9 @@ entities:
         mask: "00FF000000000000000000000000000000000000000000000000000000000000"
         unit: "%"
         name: sensor
+      - id: 113
+        type: base64
+        name: raw_data
   - entity: text
     name: Backlight
     category: config
@@ -488,6 +491,9 @@ entities:
         name: sensor
         mask: "0000000000FFFF0000"
         unit: °
+      - id: 134
+        type: base64
+        name: raw_data
   - entity: sensor
     class: illuminance
     dps:


custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
```diff
@@ -433,13 +431,15 @@ entities:
...
+    name: Battery
+    class: battery
     category: diagnostic
-    hidden: true
     dps:
       - id: 113
         type: base64
-        optional: true
+        mask: "FFFF"
+        cutoff_mask: 2
```
make-all: What is this?

custom_components/tuya_local/helpers/device_config.py
```diff
@@ -520,6 +534,17 @@ def decode_value(self, v, device):
                     self.name,
                 )
                 return None
+        elif self.rawtype == "latin1b64" and isinstance(v, str):
```
make-all: I don't really want to bloat the integration with one-off exceptions for every weird decision that a device manufacturer makes to hide data in obscure encodings. What was the source data that led you to create this?

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
```diff
@@ -29,6 +32,8 @@ entities:
         class: measurement
   - entity: sensor
     class: battery
+    name: Indoor battery
```
make-all: It is disruptive to existing users to rename entities after the config is already released. It is maybe better to keep the default class naming here and rename the other to "Outdoor battery".

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
```diff
@@ -433,13 +431,15 @@ entities:
           - dps_val: angle
             value: Compass angle
   - entity: sensor
-    name: Power status
+    name: Battery
```
make-all: As above, it is better not to rename. In this case it seems the meaning is also changed, so it can be an exception, but use "Outdoor battery".

Is there also other info encoded in the same binary field? In that case, retaining the raw value so it can be extracted would also be useful, but since raw binary data is not really useful as a sensor, making it an attribute of the newly renamed sensor would be OK.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
```diff
@@ -433,13 +431,15 @@ entities:
...
+    class: battery
     category: diagnostic
-    hidden: true
     dps:
       - id: 113
         type: base64
-        optional: true
+        mask: "FFFF"
+        cutoff_mask: 2
+        unit: "%"
```
make-all, line 440: If it is % with no scaling, then a single byte is enough. Are you sure the mask is correct and you are not pulling in another flag that under some conditions might make the battery appear over 256%?

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
```diff
@@ -475,20 +474,22 @@ entities:
...
+    name: Wind direction
+    icon: "mdi:compass-rose"
+    dps:
       - id: 134
-        type: base64
-        optional: true
-        name: direction_data
+        type: latin1b64
+        name: sensor
+        cutoff_mask: 5
```
make-all: Again, what is this for?

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
veista: This is true, but also it would be inconsistent with the other entities, since they are Indoor, and other entities are missing Outdoor in the start. This way the naming would be consistent.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
veista: As anwsered above. To my understanding the rest is just garbage. At least my device does not return anything usefull.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
veista: Two bytes are reserved, but technically a single byte would suffice I guess.

custom_components/tuya_local/helpers/device_config.py
veista: Okay, let me start with this one. My weather station is missing a crucial sensor: Wind direction. I would very much like to add this to the repo, since re-adding it on every update would be a pain.

The wind direction entity is a pain, since it supports both angles and directions depending on the wind direction display select entity dps109. The response encoded in an obscure iso 8859-1 ("latin1") encoding, since it contains the degree symbol °. So depending on the setting it returns for example 1° or N.

For some reason the original message contains a lot of garbage at the end, this also affects the Battery entity.

The message is for example:
ADE0OLAAlAAA (Base64, latin1)
'\x00148°\x00\x94\x00\x00'

Then the garbage end is cut off.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
veista: The battery data is in base64:
'AGR/f3////////////////////////////////////8='
b'\x00d\x7f\x7f\x7f\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff'

I do not understand what the rest of the data is/could be. Maybe it is just reserved.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
veista: P.S. The naming of the original entity was wrong according to that logic.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
make-all: This can be handled already by using a 32 byte mask to match the underlying data, there is no need to introduce new syntax for this.

custom_components/tuya_local/helpers/device_config.py
make-all: The rest is not garbage, it is a numeric representation in little-endian format, which is probably more useful to extract.

The format seems to use null characters as separators though, so I don't know if this will be fixed width fields or variable width without more data collection including 2 digit and 1 digit wind directions.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
veista: I figured, but for some reason I could not make it work. I will try again.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
make-all: I'm not sure where you are getting the info that 2 bytes are reserved, but the description on #3166 says "Wireless status (battery level, disconnection)", which makes me think there is a flag in there as well, possibly in the first byte. Also, if the device supports multiple external sensors, then that may be what the "garbage" is for.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
make-all: The original naming was correct at the time, as it was the only battery class sensor that was exposed. Now that the other has been added, the most important constraint is to avoid breaking other users' configurations.

custom_components/tuya_local/helpers/device_config.py
make-all: Also, including collecting data with the Wind direction type setting to "name" rather than "angle".
If these are variable width fields as I suspect, then they will have to be supported by additional templace / integration on top of this one, as I do not wish to complicate this integration by adding that support directly.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
veista: I see.

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
veista: Other indoor sensors are with Indoor at the start. According to that logic it is wrong even if there is just one sensor. I thought it was outdoor until I removed the batteries on the indoor unit and to my disappointment realized it was for the indoor battery.

custom_components/tuya_local/helpers/device_config.py
veista: I am sorry, but I do not follow. I have tested this solution with all angles and directions.

I will give you more samples if you have a more elegant solution in mind.

make-all: Samples of smaller degree angles, and samples after changing "Wind direction display" to "Direction name" would complete the picture.

veista: AAAAMrAAAvAT = b'\x00\x00\x002\xb0\x00\x02\xf0\x13' = 2°ð
ADMzNLABTvAT = b'\x00334\xb0\x01N\xf0\x13' = 334°Nð
AAAyMrAAFvAX = b'\x00\x0022\xb0\x00\x16\xf0\x17' = 22°ð
AFdOVwABLPAD = b'\x00WNW\x00\x01,\xf0\x03' = WNW,ð
AABXAAABGPAD = b'\x00\x00W\x00\x00\x01\x18\xf0\x03' = Wð
AE5XAAABM/AD = b'\x00NW\x00\x00\x013\xf0\x03' = NW3ð

custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml
veista: Yes, it works. Must have type-o-d something last time.

custom_components/tuya_local/helpers/device_config.py
make-all: I think a mask of "0000000000FFFF0000" will extract the angle as an integer without anything else needed. The text portion in the first 5 bytes seems to be right justified when display is set to "angle", and centered when it is set to "name", and padded with NULL bytes, which will be difficult to cleanly handle.

Issue #3644, opened 2025-08-19
Request support for Geyserwise MWS


Waz-Cpt:
### Log message

```text
This error originated from a custom integration.

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:404
integration: Tuya Local (documentation, issues)
First occurred: 5:17:54 PM (2 occurrences)
Last logged: 5:18:43 PM

Adding Geyserwise MWS device with product id qwbyia2yr5ip9y4v

This error originated from a custom integration.

Logger: custom_components.tuya_local.helpers.device_config
Source: custom_components/tuya_local/helpers/device_config.py:1097
integration: Tuya Local (documentation, issues)
First occurred: 5:06:09 PM (3 occurrences)
Last logged: 5:18:42 PM

Parse error in geyserwise_mws_water_heater.yaml

This error originated from a custom integration.

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:422
integration: Tuya Local (documentation, issues)
First occurred: 5:17:55 PM (2 occurrences)
Last logged: 5:18:43 PM

Cloud device spec: [{"id": 1, "name": "switch", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 10, "name": "temp_current", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":0,\"max\":125,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 20, "name": "fault", "type": "bitmap", "format": "{\"label\":[\"Over_current\",\"Dry_burn\",\"Wt_sensor_fault\",\"Heating_loss\",\"Over_temp\",\"Water_leak\",\"Comm_failure\",\"Ct_probe_fault\",\"Pump_failure\"]}", "enumMap": {}}]

This error originated from a custom integration.

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:428
integration: Tuya Local (documentation, issues)
First occurred: 5:06:09 PM (3 occurrences)
Last logged: 5:18:43 PM

Device matches rectec_rt700_grill with quality of 29%. DPS: {"updated_at": 1755615968.455908, "1": true, "2": "Timer", "10": 75, "13": "On", "20": 0, "101": false, "102": 75, "103": 55, "104": 65, "105": 75, "106": 58, "107": 18, "108": 0, "110": "Solar", "111": 917, "112": 23, "113": 2209}
[...]

PR #3971, opened 2025-11-12
New Device Config - Geyserwise MWS #3644

Waz-Cpt:
Created a new device config and tested with two units.

Request support for Geyserwise MWS #3644

### DIFF
--- a/custom_components/tuya_local/devices/geyserwise_mws_water_heater.yaml
+++ b/custom_components/tuya_local/devices/geyserwise_mws_water_heater.yaml
@@ -146,7 +146,7 @@ entities:
     dps:
       - id: 107
         type: integer
-        name: value
+        name: sensor
         unit: h
   - entity: sensor
     name: Power source


Waz-Cpt:
@make-all would you mind giving some guidance on what I need to rework before the merge? I am still learning the ropes of all this.

Also, played around with some of the information that is in the new device docs.  Either I am really not understanding things correctly or there are some potential shortfalls with the way water_heaters are handled.

The unit is question is a strange one and might not follow the normal behaviours of a typical water_heater.

make-all:
Your branch needs to be rebased on the upstream repository. Currently the history of this PR is showing commits related to ems_waterheater that are already merged upstream.

Waz-Cpt:
> Your branch needs to be rebased on the upstream repository. Currently the history of this PR is showing commits related to ems_waterheater that are already merged upstream.

I am quite a newby with this. How do I go about doing that?

make-all:
```
git remote add upstream https://github.com/make-all/tuya-local
git fetch upstream
git rebase upstream/main
git push --force-with-lease origin main
```
You can use `-f` instead of `--force-with-lease` if you know you are the only one pushing to your version of the repository. The --force-with-lease just does some extra checks to make sure you are overwriting what you think you are overwriting on the server.

Normally the rebase will automatically connect your old commits to what was merged. But since lately I have been doing squash merges, it may find conflicts due to multiple old commits being squashed into a single commit. If that happens, you may need to do an interactive rebase `git rebase -i upstream/main`, and manually mark all your old commits that are not part of this PR as `skip`.

custom_components/tuya_local/devices/geyserwise_mws_water_heater.yaml
```diff
@@ -0,0 +1,217 @@
+name: Geyserwise MWS
```
make-all: Top level name should be generic, see geyser_water_heater.yaml for example.

custom_components/tuya_local/devices/geyserwise_mws_water_heater.yaml
```diff
@@ -0,0 +1,217 @@
...
+    name: Solar MPPT Geyser
+entities:
+  - entity: water_heater
+    dps:
+      - id: 1
+        type: boolean
+        name: operation_mode
+        mapping:
+          - dps_val: false
+            value: "Off"
```
make-all: Values should be valid operation modes. Lower case "off", electric, heat_pump etc (see https://developers.home-assistant.io/docs/core/entity/water-heater/#states for full list)

custom_components/tuya_local/devices/geyserwise_mws_water_heater.yaml
```diff
@@ -0,0 +1,217 @@
...
+        name: away_mode
+        mapping:
+          - dps_val: "Timer"
+            value: false
+          - dps_val: "Holiday"
+            value: true
+      - id: 10
+        type: integer
+        name: current_temperature
+        readonly: true
```
make-all: Readonly is only needed in exceptional cases where the HA entity model expects a value to be read-write but the device does not support that and there is a strong reason not to change the entity mapping. Here, current_temperature is always expected to be read-only.

custom_components/tuya_local/devices/geyserwise_mws_water_heater.yaml
```diff
@@ -0,0 +1,217 @@
...
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 20
+        type: bitfield
+        name: fault_code
+  - entity: sensor
```
make-all: Instead of a separate sensor, this should be an attribute of the binary sensor above.

custom_components/tuya_local/devices/geyserwise_mws_water_heater.yaml
```diff
@@ -0,0 +1,217 @@
...
+          - dps_val: 32
+            value: Water leak
+          - dps_val: 64
+            value: Comm failure
+          - dps_val: 128
+            value: Ct probe fault
+          - dps_val: 256
+            value: Pump failure
+  - entity: switch
+    name: AC Boost
```
make-all: HA naming convention is to only use capitalisation on the first word of an entity name.

custom_components/tuya_local/devices/geyserwise_mws_water_heater.yaml
```diff
@@ -0,0 +1,217 @@
...
+        name: value
+        unit: C
+        range:
+          min: 30
+          max: 75
+  - entity: sensor
+    name: Element AC Runtime
+    category: diagnostic
+    class: duration
+    readonly: true
```
make-all: readonly is not valid at entity level, and sensors are inherently readonly anyway.

Waz-Cpt:
Thank you so much for the assistance.

PR #3975, opened 2025-11-13
Change dps_val from HYB to HYB1

banemonster:
dps_val: HYB is not valid for this device and must be changed to HYB1 for the performance mode to function correctly.


banemonster:
Image showing the device working in performance mode with **dps_val: HYB1**

image

PR #3946, opened 2025-11-06
Add support for WarmeHaus Thermostatic Towel Rail Electric Heating Element

dtaylor7:
Add support for WarmeHaus Thermostatic Towel Rail Electric Heating Element - https://www.manomano.co.uk/p/warmehaus-thermostatic-towel-rail-electric-heating-element-with-timer-lcd-display-child-lock-function-dual-fuel-for-bathroom-heated-towel-rails-and-radiators-79338138?model_id=83772065

### DPS information
```
{
  "modelId": "000004yg0s",
  "services": [
    {
      "actions": [],
      "code": "",
      "description": "",
      "events": [],
      "name": "默认服务",
      "properties": [
        {
          "abilityId": 1,
          "accessMode": "rw",
          "code": "switch",
          "description": "",
          "extensions": { "iconName": "icon-dp_power3" },
          "name": "开关",
          "typeSpec": { "type": "bool" }
        },
        {
          "abilityId": 2,
          "accessMode": "rw",
          "code": "mode",
          "description": "",
          "extensions": { "attribute": "128" },
          "name": "工作模式",
          "typeSpec": {
            "type": "enum",
            "range": ["hot", "eco", "cold", "auto"]
          }
        },
        {
          "abilityId": 16,
          "accessMode": "rw",
          "code": "temp_set",
          "description": "",
          "extensions": { "iconName": "icon-dp_temp" },
          "name": "目标温度",
          "typeSpec": {
            "type": "value",
            "max": 700,
            "min": 300,
            "scale": 1,
            "step": 10,
            "unit": "℃"
          }
        },
        {
          "abilityId": 24,
          "accessMode": "ro",
          "code": "temp_current",
          "description": "",
          "extensions": { "iconName": "icon-wendu" },
          "name": "当前温度",
          "typeSpec": {
            "type": "value",
            "max": 800,
            "min": 0,
            "scale": 1,
            "step": 10,
            "unit": "℃"
          }
        },
        {
          "abilityId": 27,
          "accessMode": "rw",
          "code": "temp_correction",
          "description": "",
          "extensions": { "iconName": "icon-wendu1" },
          "name": "温度校准",
          "typeSpec": {
            "type": "value",
            "max": 5,
            "min": -5,
            "scale": 0,
            "step": 1,
            "unit": "℃"
          }
        },
        {
          "abilityId": 40,
          "accessMode": "rw",
          "code": "child_lock",
          "description": "",
          "extensions": { "iconName": "icon-dp_lock" },
[...]

PR #3990, opened 2025-11-14
Correct preset mode and temporature scale for WarmeHaus Towel Rail

dtaylor7:
- "cold" is a fixed low temperature - More aligned with "comfort"
- "hot" allows the user to define any temperature
- Missed scale for "current_temperature"

Fix's for https://github.com/make-all/tuya-local/pull/3946

### DIFF
--- a/custom_components/tuya_local/devices/warmehaus_afd02tj_thermostat_c.yaml
+++ b/custom_components/tuya_local/devices/warmehaus_afd02tj_thermostat_c.yaml
@@ -20,9 +20,9 @@ entities:
         name: preset_mode
         mapping:
           - dps_val: hot
-            value: comfort
-          - dps_val: cold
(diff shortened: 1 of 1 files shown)


make-all:
I was thinking comfort because it keeps it to the comfortable temperature that you set, and manual because you need to manually turn it off, but it is a matter of how you interpret those I guess. eco might be better than comfort for a low temperature though.

dtaylor7:
> I was thinking comfort because it keeps it to the comfortable temperature that you set, and manual because you need to manually turn it off, but it is a matter of how you interpret those I guess. eco might be better than comfort for a low temperature though.

Totally down to how you interpret. Im just going by other devices configuration.... Manual, seems to give most control on other devices. Eco and comfort is anyone's guess!

PR #3994, opened 2025-11-16
Add support for Convection Heater - Jocel JCT007452

DavidRomao:
While setting up my Convection Heater, the integration suggested the device **lehmann_lhohc2013c_radiator**. 

All controls worked as expected. However, the power levels didn't exactly match my device. 

So I copied the file for lehmann_lhohc2013c_radiator, fixed the mapping of the levels to match my device and updated the product data.

Tested locally with my device, it automatically matched it during setup.


make-all:
It is better to make the preset names more generic rather than creating a new config over minor naming differences

PR #4006, opened 2025-11-18
Add alternative dps_val conditions for modes

tkaefer:
There are ballu aura devices out there that 
* use different dps_val (upper vs lowecase) 
* have a heating mode enabled 

Unfortunately there seem to be a bit of inconsistency in the naming of the dps_val values.
I tried to make this more robust, by adding upper and lower case dps_val values.

As for my device (sorry only in German - https://www.costway.de/costway-mobile-klimaanlage-fernbedienung-tragbarer-entfeuchter-16000-btu-16-32-44-5-x-39-x-77-cm-weiss.html) there were two additional modes "auto" and "hot" (not "heat" strangely). Both values were lowercase, as for the "COOL", "DRY" and "FAN" remain uppercase.
So it's a bit of a mix up.

### DIFF
--- a/custom_components/tuya_local/devices/ballu_aura.yaml
+++ b/custom_components/tuya_local/devices/ballu_aura.yaml
@@ -15,35 +15,35 @@ entities:
           - dps_val: true
             constraint: mode
             conditions:
-             # cool
-             - dps_val: "COOL"
-               value: cool
-             - dps_val: "cool"
-               value: cool
-             # dry
-             - dps_val: "DRY"
-               value: dry
-             - dps_val: "dry"
-               value: dry
-             # fan
-             - dps_val: "FAN"
-               value: fan_only
-             - dps_val: "fan"
-               value: fan_only
-             # heat
-             - dps_val: "HEAT"
-               value: heat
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/ballu_aura.yaml
```diff
@@ -15,12 +15,35 @@ entities:
...
-              - dps_val: "COOL"
-                value: cool
-              - dps_val: "DRY"
-                value: dry
-              - dps_val: "FAN"
-                value: fan_only
+             # cool
+             - dps_val: "COOL"
+               value: cool
+             - dps_val: "cool"
```
make-all: You cannot simply duplicate these. There needs to be some way for the integration to know which one to use when selecting a mode.

make-all:
I think I misread your original submission. I don't know why you concluded that there are variants that use lower case for the existing features without evidence of that being the case, but that was what confused me. It seems it is not possible to differentiate between the models unless they have an additional model_code or feature_flags dp that is not currently supported by the integration. So your model will need a separate config.

PR #4010, opened 2025-11-18
added Afyeev 11kW(3x16A) EV charger

gec75:
This is a variant of existing Afyeev EV chargers; the other ones only partially matched.

### DIFF
--- a/custom_components/tuya_local/devices/afyeev_11kw_evcharger.yaml
+++ b/custom_components/tuya_local/devices/afyeev_11kw_evcharger.yaml
@@ -1,0 +1,355 @@
+name: EV charger
+products:
+  - id: zma7jiuomnao0lk4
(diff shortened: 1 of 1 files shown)


make-all:
This has been merged to the existing Afyeev 16A config, as there are only new entities added for the additional 2 phases and some prepay related entities, which are not really relevant to Home installations, and all these can be made optional.

gec75:
I didn't want to modify existing entries, but now I see how that works.
Not very important, but how could several dps work for the same entity?? Firmware Version (dps 23) and Identity (dps 22) below status entity? Important is that status works :) (dps 3).

make-all:
The id and firmware version are static info that is not really useful as a sensor, so I put them as attributes of the status sensor instead.

Issue #399, opened 2023-02-13
Request for new device - Fanco Eco Silent Deluxe Ceiling Fan with LED Light CFFCESD3LWHSMRABS


dezinc57:
Requesting device support for Fanco Eco Silent Deluxe DC Smart Ceiling Fan with LED Light

Fan switch is currently discoverable in Tuya Local as simple_switch only

# **Information about DPS mappings**

{
  "result": {
    "category": "fs",
    "functions": [
      {
        "code": "switch",
        "dp_id": 1,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "fan_speed_percent",
        "dp_id": 3,
        "type": "Integer",
        "values": "{\"min\":\"1\",\"max\":\"6\",\"scale\":\"0\",\"step\":\"1\"}"
      },
      {
        "code": "light",
        "dp_id": 15,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "bright_value",
        "dp_id": 16,
        "type": "Integer",
        "values": "{\"min\":\"10\",\"max\":\"1000\",\"scale\":\"0\",\"step\":\"1\"}"
      },
      {
        "code": "temp_value",
        "dp_id": 17,
        "type": "Integer",
        "values": "{\"min\":\"0\",\"max\":\"1000\",\"scale\":\"0\",\"step\":\"1\"}"
      }
    ],
    "status": [
      {
        "code": "switch",
        "dp_id": 1,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "fan_speed_percent",
        "dp_id": 3,
        "type": "Integer",
        "values": "{\"min\":\"1\",\"max\":\"6\",\"scale\":\"0\",\"step\":\"1\"}"
      },
      {
        "code": "light",
        "dp_id": 15,
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "bright_value",
        "dp_id": 16,
        "type": "Integer",
        "values": "{\"min\":\"10\",\"max\":\"1000\",\"scale\":\"0\",\"step\":\"1\"}"
      },
      {
        "code": "temp_value",
        "dp_id": 17,
        "type": "Integer",
        "values": "{\"min\":\"0\",\"max\":\"1000\",\"scale\":\"0\",\"step\":\"1\"}"
      }
    ]
  },
  "success": true,
  "t": 1676261554171,
  "tid": "a4586334ab5411ed9a310aeeb3bc8fba"
}

# Product ID

{
[...]

PR #4012, opened 2025-11-19
Added configuration for Fanco Infinity-ID fan

brendanlees:
Similar to fanco_ecosilentdeluxe, but with differing DP IDs to better suit this model.

Have verified YAML loads locally and entities appear and work as expected.


make-all:
The information on #399 was unreliable, so given the identical product id, I have merged these changes into the original Fanco Eco Silent config instead.

PR #4021, opened 2025-11-23
Add Boost countdown to Mi-Heat TH11 WF thermostat

flow-c:
- **Add support for Mi-Heat TH11-WF thermostat**
- **Fixed lint and type errors**
- **cleanup (miheat_th11wf_thermostat): changes from review**
- **Fix typo in measurement class for thermostat**
- **Add Boost countdown to Mi-Heat TH11 WF thermostat**

### DIFF
--- a/custom_components/tuya_local/devices/miheat_th11wf_thermostat.yaml
+++ b/custom_components/tuya_local/devices/miheat_th11wf_thermostat.yaml
@@ -105,6 +105,8 @@ entities:
         precision: 0
         unit: kWh
         class: total_increasing
+        mapping:
+          - scale: 100
   - entity: binary_sensor
     class: window
     dps:


custom_components/tuya_local/devices/miheat_th11wf_thermostat.yaml
```diff
@@ -130,6 +130,21 @@ entities:
...
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 119
+        type: integer
+        name: sensor
+        unit: min
+        mapping:
+          - dps_val: 0
+            constraint: Boost
```
make-all: constraint needs to reference a dp on the same entity. This is referencing a different entity without specifying the dp name.

But is this really required? No other devices have this, they just show the data the device is sending.

flow-c: You are absolutely right, the Boost entity does reset itself automatically when the time is up. It only fails when the device can't reach the Tuya cloud – but so do many other features of this gadget :/
I must have first tested this when the device was offline.

flow-c:
I'm sorry to lump in other (minor) changes here. @make-all, if you prefer these split up, I'll create a separate PR for each.

PR #4022, opened 2025-11-23
Remove deprecated items

gkwok1:


make-all:
It is premature to remove these less than 3 weeks after they were deprecated. If you disable the entities you can get rid of the warnings in your own install.

PR #4031, opened 2025-11-24
Added support for Orbegozo RRW heater

H4irBear:
Created custom_components/tuya_local/devices/orbegozo_rrw_emisor_termico.y

Tested all config, controls and sensors.

Controls:
- Climate (added modes for confort, economy, antifrost and programmed)

Config:
- Child lock
- Open window detection
- Temperature correction / calibration

Sensors:
- Problem status

### DIFF
--- a/custom_components/tuya_local/devices/orbegozo_rrw_emisor_termico.yaml
+++ b/custom_components/tuya_local/devices/orbegozo_rrw_emisor_termico.yaml
@@ -1,6 +1,6 @@
 name: Heater
 products:
-  - id: bf49669b1da42ca7f34faa
+  - id: xnljzcjrdqus2tka
     manufacturer: Orbegozo
     model: RRW Emisor Termico
 entities:
@@ -43,13 +43,13 @@ entities:
         type: string
         mapping:
           - dps_val: COMFORT
-            value: Comfort
+            value: comfort
           - dps_val: ECONOMY
-            value: Economy
+            value: eco
           - dps_val: ANTIFROST
-            value: Antifrost
+            value: away
           - dps_val: PROGRAMME
-            value: Programming
+            value: home
   - entity: lock
     translation_key: child_lock
     category: config
@@ -66,9 +66,8 @@ entities:
         type: boolean
         name: switch
   - entity: number
-    name: Temperature correction
+    translation_key: temperature_calibration
     category: config
-    icon: "mdi:pan-vertical"
     dps:
       - id: 102
         type: integer


custom_components/tuya_local/devices/orbegozo_rrw_emisor_termico.yaml
```diff
@@ -0,0 +1,96 @@
+name: Heater
+products:
+  - id: bf49669b1da42ca7f34faa
```
make-all: not a valid product id
it looks like a device id

custom_components/tuya_local/devices/orbegozo_rrw_emisor_termico.yaml
```diff
@@ -0,0 +1,96 @@
...
+                  max: 32
+      - id: 3
+        name: current_temperature
+        type: integer
+      - id: 4
+        name: preset_mode
+        type: string
+        mapping:
+          - dps_val: COMFORT
+            value: Comfort
```
make-all: Use values that match standard HA presets where possible  (all lower case, as they are translation keys): `comfort`, `eco`, `away`. 

Program does not have a direct equivalent, but I think `home` is often used for such settings.

custom_components/tuya_local/devices/orbegozo_rrw_emisor_termico.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - entity: switch
+    name: Open window detection
+    category: config
+    icon: "mdi:window-closed-variant"
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+  - entity: number
+    name: Temperature correction
```
make-all: use `translation_key: temperature_calibration` (was added in 2025.11.2) instead of explicit name and icon

custom_components/tuya_local/devices/orbegozo_rrw_emisor_termico.yaml
H4irBear: Thanks, have updated.  Wasn't aware this was different and doesn't appear to be exposed in the Tuya Dev debugging.  Found it in the json in the end.

custom_components/tuya_local/devices/orbegozo_rrw_emisor_termico.yaml
H4irBear: Have updated.

custom_components/tuya_local/devices/orbegozo_rrw_emisor_termico.yaml
H4irBear: Didn't find this key when I was making this.  Maybe my version was out of date.  Have updated to use the key.

PR #4036, opened 2025-11-24
Update electriq_cd12pro_dehumidifier.yaml

steaura:
Added child lock

### DIFF
--- a/custom_components/tuya_local/devices/electriq_cd12pro_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/electriq_cd12pro_dehumidifier.yaml
@@ -129,3 +129,4 @@ entities:
       - id: 16
         type: boolean
         name: lock
+        optional: true


custom_components/tuya_local/devices/electriq_cd12pro_dehumidifier.yaml
```diff
@@ -122,3 +122,10 @@ entities:
       - id: 19
         type: bitfield
         name: fault_code
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 16
+        type: boolean
```
make-all, line 130: Since it was not reported before, it may be better to add `optional: true` here.

steaura: Thanks. I've now done this.

PR #4039, opened 2025-11-25
Add configuration for Plikc Ario Wifi Air Vent

christianglodt:
I created this basic configuration this [device](https://plikc.com/products/ario-w-aspiratore-wifi-da-parete-ad-aspirazione-assiale-con-sensore-di-umidita-e-timer). It is a ceiling- or wall-mounted extractor fan for bathrooms.

Please let me know if there's anything else I need to do (translations?).

### DIFF
--- a/custom_components/tuya_local/devices/plikc_ariowifi_airvent.yaml
+++ b/custom_components/tuya_local/devices/plikc_ariowifi_airvent.yaml
@@ -1,58 +1,46 @@
-name: Air Vent
+name: Air vent
 products:
   - id: lc29ihcc61xcyhyj
     manufacturer: Plikc
     model: Ario Wifi
     model_id: PLKARW100
 
-# DPs that are not yet implemented:
-# -  45 "Fault Alarm":
-#          {
-#            "id": 45, "name": "fault", "type": "Bitmap",
-#            "format": "{\"label\":[\"e1\",\"e2\",\"e3\"]}",
-#            "enumMap": {}
-#          }
-# - 110: "schedule_7day" "valueRange": "{}", "valueType": "Raw"
-# - 111: "schedule_52" "valueRange": "{}", "valueType": "Raw"
-# - 112: "schedule_24h" "valueRange": "{}", "valueType": "Raw"
-# - 116: "led" ?
-
 entities:
+  - entity: switch
+    icon: "mdi:power"
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
   - entity: humidifier
     class: dehumidifier
     dps:
       # Global on/off
       - id: 1
+        type: boolean
+        name: available
+      - id: 104
         name: switch
         type: boolean
       - id: 101
         name: current_humidity
         type: integer
-        unit: "%"
-        range:
-          min: 0
-          max: 100
       # Humidity set point
       - id: 102
         name: humidity
         type: integer
-        unit: "%"
         range:
           min: 22
           max: 99
   - entity: sensor
     class: temperature
-    name: Current temperature
     dps:
       - id: 24
         name: sensor
         type: integer
-        unit: "C"
+        unit: C
         mapping:
           - scale: 10
-        range:
-          min: 0
-          max: 400
   - entity: button
     translation_key: factory_reset
     category: config
@@ -61,14 +49,7 @@ entities:
         type: boolean
         name: button
         optional: true
-  - entity: switch
-    name: Timer mode
-    dps:
-      - id: 103
-        name: switch
-        type: boolean
   - entity: lock
-    name: Child lock
     translation_key: child_lock
     category: config
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/plikc_ario_wifi_air_vent.yaml
```diff
@@ -0,0 +1,173 @@
...
+        name: option
+        mapping:
+          - dps_val: "on"
+            value: "on"
+          - dps_val: "off"
+            value: "off"
+          - dps_val: "using_on"
+            value: "using_on"
+  - entity: number
+    name: Timer time
```
make-all: Use `translation_key: timer` rather than this name

custom_components/tuya_local/devices/plikc_ario_wifi_air_vent.yaml
```diff
@@ -0,0 +1,173 @@
...
+      - id: 113
+        type: string
+        name: option
+        mapping:
+          - dps_val: "on"
+            value: "on"
+          - dps_val: "off"
+            value: "off"
+          - dps_val: "using_on"
+            value: "using_on"
```
make-all: The value here should be understandable by the end user, as it is what appears in the UI. This actually seems like a good match for `translation_key: light_mode`, where the values are "off", "on", "state" (for a light that follows state), or "locator" (for a light that is the inverse of state - common for light switches to find them in the dark)

custom_components/tuya_local/devices/plikc_ario_wifi_air_vent.yaml
```diff
@@ -0,0 +1,173 @@
...
+          - dps_val: null
+            value: false
+  - entity: switch
+    name: Humidity mode
+    dps:
+      - id: 104
+        name: switch
+        type: boolean
+  - entity: switch
+    name: Program mode
```
make-all: Program mode seems mutually exclusive with Manual mode. Would these be better as buttons rather than switches?

custom_components/tuya_local/devices/plikc_ario_wifi_air_vent.yaml
```diff
@@ -0,0 +1,173 @@
...
+      - id: 102
+        name: humidity
+        type: integer
+        unit: "%"
+        range:
+          min: 22
+          max: 99
+  - entity: sensor
+    class: temperature
+    name: Current temperature
```
make-all: name not required, it is better to let the class translation be used.

custom_components/tuya_local/devices/plikc_ario_wifi_air_vent.yaml
```diff
@@ -0,0 +1,173 @@
...
+        name: button
+        optional: true
+  - entity: switch
+    name: Timer mode
+    dps:
+      - id: 103
+        name: switch
+        type: boolean
+  - entity: lock
+    name: Child lock
```
make-all: name not required, let the translation_key be used instead

custom_components/tuya_local/devices/plikc_ario_wifi_air_vent.yaml
```diff
@@ -0,0 +1,173 @@
...
+        type: boolean
+        optional: true
+        name: lock
+        mapping:
+          - dps_val: null
+            value: false
+  - entity: switch
+    name: Humidity mode
+    dps:
+      - id: 104
```
make-all: It might be better to swap this with the switch under humidifier entity. An available dp on the humidifier entity can ensure that the main switch is on first before the humidifier can be used.

custom_components/tuya_local/devices/plikc_ario_wifi_air_vent.yaml
```diff
@@ -0,0 +1,173 @@
...
+        type: boolean
+  - entity: switch
+    name: Manual mode
+    dps:
+      - id: 106
+        name: switch
+        type: boolean
+  - entity: binary_sensor
+    name: Fan Status
+    icon: mdi:fan
```
make-all: class: running might be appropriate here.

christianglodt:
Thank you so much for the review and for doing the changes!

I was going to try to implement the changes myself this weekend, but you beat me to it :)

I'll test the new configuration and if necessary, I'll make a new pull request.

loop2:
I don't see the humidity value in the entities on HA. The humidity sensor is the ID 101

christianglodt:
The current humidity value should be available as a state attribute named "current_humidity" on the humidifier entity state. You should be able to see it if you check the state in the HA developer tools dashboard. You could also extract it to a separate entity using a template sensor, if necessary.

loop2:
yes but only if the fan is working. Instead is useful if I can read always the humidity in order to use it on HA.
I have just add this so I have always the humidity value:
`  - entity: sensor
    class: humidity
    dps:
      - id: 101
        name: sensor
        type: integer
        unit: '%'`

christianglodt:
There is a switch entity that controls if the device is "on", and indeed, when it is "off", the humidifier entity is unavailable.

However this switch does not control whether the fan actually spins.

Whether the fan spins is controlled by the "on"/"off" state of the humidifier entity in combination with the current/set humidity, or by the manual mode switch (or the timer and the schedule, which are not well tested).

I just leave the switch entity on all the time, and thus the humidifier entity is always available.

loop2:
Thanks i understand. However obtain the humidity value is simply and in some configurations can be useful.

christianglodt:
Of course, I understand it can be easier to have the current humidity as a separate entity. I would recommend leaving the humidifer "on" and creating a template sensor to extract it, with a template like this:
```
{{ state_attr('humidifier.pool_dressing_ventilator', 'current_humidity') }}
```
This way you don't have to keep your own version of the configuration for this device.

PR #4040, opened 2025-11-25
Added support for Pilot Wire Thermostat

rickbergmans:
### Device this implementation is for:
https://aliexpress.com/item/1005007319412940.html

---

### Climate
- Added HVAC modes:
  - **off**
  - **heat**
  - **auto**
- Added preset mode support:
  - **comfort**
  - **comfort 1**
  - **comfort 2**
  - **eco**
  - **anti-frost (mapped as away)**
  - **manual**
  - **program**

---

## **Config**
- Child lock (secondary lock entity)
- Target temperature handling (0.5°C resolution via scale 10)
- Current temperature sensor (scaled)
- Minimal reliable DP set implemented

---

## **Sensors**
- Current temperature
- Target temperature
- Basic status reporting


custom_components/tuya_local/devices/pilot_rp5_bath_thermostat_min.yaml
```diff
@@ -3,8 +3,8 @@ products:
   - id: c03zek9b5daz7omr
     name: Pilot Thermostat RP5
 legacy_type: climate
```
make-all: legacy_type is for migration purposes only and should not be used. Icons also are not valid at device level

custom_components/tuya_local/devices/pilot_rp5_bath_thermostat_min.yaml
```diff
@@ -0,0 +1,82 @@
+name: Pilot RP5 Thermostat (minimal+presets)
```
make-all: Device name should be generic, just `Thermostat` is OK

custom_components/tuya_local/devices/pilot_rp5_bath_thermostat_min.yaml
```diff
@@ -0,0 +1,82 @@
...
+products:
+  - id: c03zek9b5daz7omr
+    name: Pilot Thermostat RP5
```
make-all: This should be split to `manufacturer` and `model`.

custom_components/tuya_local/devices/pilot_rp5_bath_thermostat_min.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+entities:
+  - entity: climate
+    name: Bathroom Thermostat
```
make-all: It may be better to use `translation_key: thermostat` in place of the name and icon

custom_components/tuya_local/devices/pilot_rp5_bath_thermostat_min.yaml
```diff
@@ -0,0 +1,82 @@
...
+        mapping:
+          - dps_val: "Standby"
+            value: "off"
+          - dps_val: "Programming"
+            value: "auto"
+          # All heating modes -> heat
+          - dps_val: "Comfort"
+            value: "heat"
+          - dps_val: "Comfort_1"
+            value: "heat"
```
make-all: To ensure the behaviour is not random, all but one of the mappings to `heat` should be marked with `hidden: true`.
This could then be abbreviated by using a default mapping:
```
mapping:
  - dps_val: Standby
    value: "off"
  - dps_val: Programming
    value: auto
  - dps_val: Comfort
    value: heat
  - value: heat
    hidden: true
```

custom_components/tuya_local/devices/pilot_rp5_bath_thermostat_min.yaml
```diff
@@ -0,0 +1,82 @@
...
+      # --- Current temperature (°C) ---
+      - id: 116
+        type: integer
+        name: current_temperature
+        range:
+          min: -300
+          max: 900
+        mapping:
+          - scale: 10
+        unit: "°C"
```
make-all: Unit is only needed on one of the temperature dps, since climate entity only has one temperature_unit attribute so they have to be the same unit.

custom_components/tuya_local/devices/pilot_rp5_bath_thermostat_min.yaml
```diff
@@ -0,0 +1,82 @@
...
+            value: "comfort_1"
+          - dps_val: "Comfort_2"
+            value: "comfort_2"
+          - dps_val: "Thermostat"
+            value: "manual"
+          - dps_val: "Programming"
+            value: "program"
+
+  - entity: lock
+    name: Child Lock
```
make-all: use `translation_key: child_lock` instead of an explicit name and icon.

PR #4051, opened 2025-11-26
add support for TCL Airmax Inverter mini-split air conditioner model AWTLSE18-C2

jaredhobbs:
```json
{
    "name": "Air conditioner",
    "id": "10736132ac0bfbfccf46",
    "key": "a`6!_VRd*SbE7qZh",
    "mac": "ac:0b:fb:fc:cf:46",
    "uuid": "10736132ac0bfbfccf46",
    "sn": "10009737700B70",
    "category": "kt",
    "product_name": "Air conditioner",
    "product_id": "hw50w7qvxluhslkk",
    "biz_type": 18,
    "model": "",
    "sub": false,
    "icon": "https://images.tuyaus.com/smart/product_icon2/kt_1.png",
    "mapping": {
        "1": {
            "code": "switch",
            "type": "Boolean",
            "values": {}
        },
        "2": {
            "code": "temp_set",
            "type": "Integer",
            "values": {
                "unit": "\u00b0C",
                "min": 160,
                "max": 880,
                "scale": 1,
                "step": 5
            }
        },
        "3": {
            "code": "temp_current",
            "type": "Integer",
            "values": {
                "unit": "\u00b0C",
                "min": -20,
                "max": 100,
                "scale": 0,
                "step": 1
            }
        },
        "4": {
            "code": "mode",
            "type": "Enum",
            "values": {
                "range": [
                    "cold",
                    "hot",
                    "wet",
                    "wind",
                    "auto"
                ]
            }
        },
        "18": {
            "code": "humidity_current",
            "type": "Integer",
            "values": {
                "unit": "%",
                "min": 0,
                "max": 100,
                "scale": 0,
                "step": 1
            }
        },
        "136": {
            "code": "temp_set_f",
            "type": "Integer",
            "values": {
                "unit": "\u00b0F",
                "min": 61,
                "max": 88,
                "scale": 0,
                "step": 1
            }
        }
    }
}
```

### DIFF
--- a/custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
@@ -98,6 +98,9 @@ entities:
             value: quiet
           - dps_val: "auto"
             value: auto
+      - id: 18
+        type: integer
+        name: broken_current_humidity
       - id: 110
         type: bitfield
         name: flags


make-all:
Is there no temperature_unit switch in the full dps list?
Why is dp 2 used to set temperature in F when dp 136 appears to exist for this purpose?

custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
```diff
@@ -0,0 +1,305 @@
...
+        name: flags
+      - id: 113
+        type: string
+        name: swing_mode
+        mapping:
+          - dps_val: "0"
+            value: "off"
+          - dps_val: "1"
+            value: "on"
+          - hidden: true
```
make-all: I think this can cover the full range of vertical swing modes so the select entity below is not needed. Unlike hvac_mode, swing_mode is not limited to predefined modes, so you can use  Upper and Lower here (and change "on" to Full if you think it would be clearer)

custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
```diff
@@ -0,0 +1,305 @@
...
+            value: "on"
+      - id: 114
+        type: string
+        name: swing_horizontal_mode
```
make-all, line 121: Likewise

custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
jaredhobbs: thanks, fixed

custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
jaredhobbs: thanks, fixed

jaredhobbs:
> Is there no temperature_unit switch in the full dps list? Why is dp 2 used to set temperature in F when dp 136 appears to exist for this purpose?

There's no temperature_unit switch that I could find. I tried dp 136 but it wasn't working for me 🤷

custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
```diff
@@ -109,21 +109,27 @@ entities:
         name: swing_mode
         mapping:
           - dps_val: "0"
-            value: "off"
+            value: "Off"
```
make-all: "off" should be left in lower case, as that is a standard option for swing_mode

custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
```diff
@@ -98,9 +98,6 @@ entities:
             value: quiet
           - dps_val: "auto"
             value: auto
-      - id: 18
-        type: integer
-        name: current_humidity
```
jaredhobbs: the humidity value isn't changing from 0 on my unit so I removed this; is that the correct solution or is there another way to remove humidity from HA while keeping the dp?

custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
jaredhobbs: fixed

custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
make-all: If you leave it but rename to something else (eg, broken_humidity), then it will still be visible under "Attributes" in the more info dialog, but will not be used in the main UI

jaredhobbs: done, thanks!

custom_components/tuya_local/devices/tcl_airmax_airconditioner.yaml
```diff
@@ -0,0 +1,275 @@
...
+        range:
+          min: 630
+          max: 880
+        mapping:
+          - scale: 10
+            step: 10
+      - id: 3
+        type: integer
+        name: current_temperature
+        mapping:
```
make-all, line 42: It may be cleaner to do the conversion by specifying range and target_range, and a precision of 0.
```
    - id: 3
      type: integer
      name: current_temperature
      precision: 0
      range:
        min: 16
        max: 32
      mapping:
        - target_range:
            min: 61
            max: 90
```

make-all:
This appears to be very similar to daizuki_heatpump, I would like to try to merge with that.
Actually if the flags (110) and flags2 (123) fields could be fully decoded, it can probably also be merged with starlight_heatpump that daizuki was originally forked out from for the temperature unit difference, and with other similar configs that have minor feature differences.

make-all:
Sorry to not notice the similarity earlier and making you go through several iterations of modifying the file only to make the modifications in daizuki_heatpump to support this model.

jaredhobbs:
> Sorry to not notice the similarity earlier and making you go through several iterations of modifying the file only to make the modifications in daizuki_heatpump to support this model.

No worries! Thanks for adding the support!

PR #4054, opened 2025-11-27
Add Zedar K1200 Cat Litterbox device configuration

MarcoKaldenbach:
Add support for the Zedar K1200 Cat Litterbox. For the mapping of some data points I used [this git repo as reference](https://github.com/baukeposthuma/ZEDAR-K1200-Data-Points).

image
</br>
image

### DIFF
--- a/ACKNOWLEDGEMENTS.md
+++ b/ACKNOWLEDGEMENTS.md
@@ -1132,3 +1132,11 @@ Further device support has been made with the assistance of users. Please consid
 - [Junikiewicz](https://github.com/Junikiewicz) for contributing improvements to Advwin 6L pet feeder.
 - [H4irBear](https://github.com/H4irBear) for contributing support for Orbegozo RRW heater.
 - [steaura](https://github.com/steaura) for contributing improvements to Electriq CD12 Pro dehumidifier.
+- [rickbergmans](https://github.com/rickbergmans) for contributing support for Pilot Wire RP5 bath thermostat.
+- [baracici](https://github.com/baracici) for assisting with support for ElectriQ CD20PRO-LE-V4 dehumidifier.
+- [christianglodt](https://github.com/christianglodt) for contributing support for Plikc Ario WiFi air vent.
+- [george-oniceag](https://github.com/george-oniceag) for contributing support for AlecoAir PU55 Humino humidifier.
+- [yurgh](https://github.com/yurgh) for contributing improvements to Eeese Carl dehumidifier.
+- [fredppm](https://github.com/fredppm) for contributing support for Newpet pet feeder.
+- [politician](https://github.com/politician) for assisting with support for Kerui 120dB siren alarm.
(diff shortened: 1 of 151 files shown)


custom_components/tuya_local/devices/zedar_k1200_litterbox.yaml
```diff
@@ -0,0 +1,218 @@
...
+          max: 600
+  - entity: button
+    name: Manual clean
+    icon: "mdi:autorenew"
+    dps:
+      - id: 9
+        type: boolean
+        name: button
+  - entity: light
+    name: Light
```
make-all: name not needed when it is the same as the entity type, the default can be used instead

custom_components/tuya_local/devices/zedar_k1200_litterbox.yaml
```diff
@@ -0,0 +1,218 @@
...
+        name: switch
+  - entity: switch
+    name: Smooth sand
+    icon: "mdi:auto-fix"
+    dps:
+      - id: 102
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Child lock
```
make-all: Use the standard pattern from other devices for child lock:
```yaml
  - entity: lock
    translation_key: child_lock
    category: config
    dps:
      - id: 103
        type: boolean
        name: lock
```

custom_components/tuya_local/devices/zedar_k1200_litterbox.yaml
```diff
@@ -0,0 +1,218 @@
...
+        type: boolean
+        name: button
+  - entity: sensor
+    name: Remaining deodorant days
+    class: duration
+    dps:
+      - id: 109
+        type: integer
+        name: sensor
+        unit: days
```
make-all: unit: d

custom_components/tuya_local/devices/zedar_k1200_litterbox.yaml
```diff
@@ -0,0 +1,218 @@
...
+    translation_key: status
+    class: enum
+    category: diagnostic
+    dps:
+      - id: 24
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: cat_in
+            value: Cat enters
```
make-all: Try to use values from the translations for status:
- occupied
- starting (or monitor)
- available
- cleaning
- full
- sleep
- maintenance_required
- standby
- sterilizing

MarcoKaldenbach: I changed the values to translations that made sense to the state, for smoothing I added a new the translation

ricardodolnl:
I can't get the weight sensor to work. It doesn't show any data and just stays at 0g.

JBosbeer:
I have the Zedar 1200 + and it only supports the Zedar 1200 and not the + one. Is it possible to add the + one so I can add this one to home assistant. And also for me I can't get the weight sensor to work. It doesn't show any data and just stays at 0g.

PR #4069, opened 2025-11-30
Add support for EZValo Smart Desk Lamp

aminorjourney:
Added support for https://www.ezvaloledlight.com/product-category/smart-desk-lamp/, including: 

• Power On/Off of the light
• Dimmer
• Color Temperature
• "Auto" Color/Temperature mode
• Timer (for shutdown) - limited to 4 hours (240 minutes)

### DIFF
--- a/custom_components/tuya_local/devices/ezvalo_smart_desklamp.yaml
+++ b/custom_components/tuya_local/devices/ezvalo_smart_desklamp.yaml
@@ -1,9 +1,8 @@
-name: Ezvalo Smart Desk Lamp
-legacy_type: smart lamp
-products:
-  - id: eb8b877f027a017980djai
-    manufacturer: Ezvalo
-    model: Smart Desk Lamp  
+name: Desk lamp
+# products:
+#   - id: UNKNOWN
+#     manufacturer: Ezvalo
+#     model: Smart Desk Lamp  
 entities:
   - entity: light
     dps:
@@ -28,7 +27,7 @@ entities:
               max: 6500
               
   - entity: switch
-    name: Auto Mode
+    name: Auto mode
     dps:
       - id: 102
         type: boolean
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/ezvalo_smart_desklamp.yaml
```diff
@@ -0,0 +1,48 @@
+name: Desk lamp
+# products:
+#   - id: UNKNOWN
```
make-all, line 3: I had to comment this out, because you used a device id rather than product id here. If you can find the product id in Tuya portal (same API function that returns the local key), or by `python -m tinytuya scan` on a local PC with tinytuya installed (should be a matter of running `pip install tinytuya` if you have a full Python installation), then it can be uncommented.

custom_components/tuya_local/devices/ezvalo_smart_desklamp.yaml
```diff
@@ -0,0 +1,48 @@
...
+        range:
+          min: 0
+          max: 1000
+        mapping:
+          - target_range:
+              min: 2700 
+              max: 6500
+              
+  - entity: switch
+    name: Auto mode
```
make-all: Is this some light sensing or sunrise/sunset tracking mode that changes color temperature and brightness according to time of day? Or something else? It may be better to choose a name that more precisely describes the feature (without getting too wordy)

aminorjourney: Sorry for the delay. 

The auto function controls brightness and color temperature based on the ambient light in the room.  Do you have any specific request for naming that feature?  

I would perhaps call it "Auto Ambient Light"?  I will look into the ID shortly!

mapeol:
Hi!
This is excellent work, but I can't get HA to recognise the colour temperature.
Can you help me?

PR #3891, opened 2025-10-26
Add Gosund SP211 Dual Smart Plug

smeegoan:
Adds support for the Gosund SP211 dual outlet smart plug (product ID: p2tjkvgdhvuadnlj).

The device was matching `blitzwolf_bwshp6_smartplug` at 89% but only exposed one switch. I combined the dual-outlet structure from `denver_shp200mk2_dualsmartplug` with the full feature set from the Blitzwolf template.

**What works:**
- Both outlets (DP 1, 2)
- Individual countdown timers for each outlet (DP 9, 10)
- Energy monitoring: current, power, voltage, cumulative energy (DP 17-20)
- Initial state, light mode, child lock settings (DP 38-40)
- Schedule/inching support (DP 41-43)

Tested via cloud integration, all DPs reporting correctly. Log data included with device discovery.

```
[...]

PR #4070, opened 2025-11-30
Expose energy measurement in Gosund SP211

smeegoan:
Hi, thank you for adjusting and merging my previous [PR](https://github.com/make-all/tuya-local/pull/3891/) but I noticed the energy measurements do not work with the merged changes (no energy entity is exposed). 

Can you review if these changes are ok to be merged? I tested them locally and they work.

Thanks

### DIFF
--- a/custom_components/tuya_local/devices/gosund_sp211_dualsmartplug.yaml
+++ b/custom_components/tuya_local/devices/gosund_sp211_dualsmartplug.yaml
@@ -74,15 +74,20 @@ entities:
         type: integer
         name: calibration
         optional: true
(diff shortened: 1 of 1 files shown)


make-all:
As commented on the previous PR, dp 17 on Tuya smart switches is generally labelled "add_ele", and reports only increments since the last report. It resets too frequently to be useful as an energy sensor, so exposing it as one will only cause confusion.

smeegoan:
Can you point me to a device that exposes a similar attribute? I am interested in accessing its contents via an alternative strategy if possible

PR #4074, opened 2025-12-01
Add support for Funai Fuji ERW-150

o-mega:
**Device Information:**
* **Name:** Funai Fuji ERW-150 Ultimate
* **Category:** Ventilation / Air Purifier
* **Product ID:** `9xaunhonta8om3zr`
* **Config file:** `funai_erw150_ventilation.yaml`

**Functionality added:**
* **Primary Entity:** Fan (Ventilator) with 9 speed levels (mapped to percentage) and presets (Auto, Sleep, Manual).
* **Sensors:** PM2.5, CO2, TVOC, Indoor Temperature, Air Quality Text.
* **Configuration:** Damper control (Manual/Auto), Fan speed level (number), Damper position.
* **Switches:** Anion, UV Light.
* **Diagnostics:** Filter reset, Cleaning reminder, Fault codes.

**Notes for Reviewer:**
* **DP 4 (Display):** The mapping for this DP is intentionally inverted (`true` -> `false`, `false` -> `true`). Testing on the physical device confirmed that the boolean logic is reversed compared to standard Tuya behavior (sending `true` turns the light off).

Tested on real hardware.

### DIFF
--- a/custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
+++ b/custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
@@ -1,7 +1,8 @@
-name: Ventilation Breezer
+name: Ventilation fan
 products:
   - id: 9xaunhonta8om3zr
-    name: Funai Fuji ERW-150 Ventilation
+    manufacturer: Funai
+    model: Fuji ERW-150
 entities:
   - entity: fan
     translation_key: ventilation


custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
+name: Funai Fuji ERW-150 Ultimate
+products:
+  - id: 9xaunhonta8om3zr
+    name: Funai Fuji ERW-150
```
make-all: The name under products should be split to manufacturer and model, and the name above should be made generic in case other products also match (eg "Ventilation fan")

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+          - dps_val: auto
+            value: auto
+          - dps_val: sleep
+            value: sleep
+          - dps_val: manual
+            value: manual
+      - id: 101
+        type: integer
+        name: speed
+        mapping:
```
make-all: This mapping is unnecessary. Just use
```yaml
    range:
      min: 1
      max: 9
```

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
+name: Funai Fuji ERW-150 Ultimate
+products:
+  - id: 9xaunhonta8om3zr
+    name: Funai Fuji ERW-150
+entities:
+  - entity: fan
+    dps:
```
make-all, line 9: add `translation_key: ventilation`

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+        name: option
+        mapping:
+          - dps_val: "Manual"
+            value: "Manual"
+          - dps_val: "Auto_1"
+            value: "Auto 1"
+          - dps_val: "Auto_2"
+            value: "Auto 2"
+
+  - entity: number
```
make-all: It is unnecessary to duplicate this

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+    dps:
+      - id: 104
+        type: integer
+        name: value
+        range:
+          min: 1
+          max: 5
+
+  - entity: switch
+    name: UV light
```
make-all: Use `translation_key: uv_sterilization` instead of explicit name and icon.

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+        type: boolean
+        name: switch
+        mapping:
+          - dps_val: true
+            value: false
+          - dps_val: false
+            value: true
+
+  - entity: switch
+    name: Anion
```
make-all: Use `translation_key: ionizer` instead of explicit name and icon.

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+    icon: "mdi:air-filter"
+    category: config
+    dps:
+      - id: 5
+        type: boolean
+        name: switch
+
+  - entity: button
+    name: Filter reset
```
make-all: Use `translation_key: filter_reset` instead of explicit name and icon.

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+    icon: "mdi:filter-remove"
+    category: config
+    dps:
+      - id: 13
+        type: boolean
+        name: button
+
+  - entity: sensor
+    name: PM2.5
```
make-all: Remove this name and let the class translation be used.

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+    class: pm25
+    dps:
+      - id: 3
+        type: integer
+        name: sensor
+        unit: "µg/m³"
+
+  - entity: sensor
+    name: CO2
```
make-all: remove this name and let the class translation be used

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+    class: carbon_dioxide
+    dps:
+      - id: 6
+        type: integer
+        name: sensor
+        unit: "ppm"
+
+  - entity: sensor
+    name: TVOC
```
make-all: Remove this name and let the class translation be used

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+    class: volatile_organic_compounds
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
+        unit: "ppb"
+
+  - entity: sensor
+    name: Indoor temperature
```
make-all: Remove this name and let the class translation be used

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+    class: temperature
+    dps:
+      - id: 9
+        type: integer
+        name: sensor
+        unit: "C"
+
+  - entity: sensor
+    name: Air quality text
```
make-all: Use `translation_key: air_quality` instead of explicit name and icon. Preferably make it `class: enum` also, and list all the options under a mapping, with values mapped to values from [excellent, good, moderate, poor, severe, danger]

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
```diff
@@ -0,0 +1,177 @@
...
+    name: Cleaning reminder
+    class: problem
+    icon: "mdi:broom"
+    category: diagnostic
+    dps:
+      - id: 103
+        type: boolean
+        name: sensor
+
+  - entity: sensor
```
make-all: Make this a binary sensor to follow the common pattern for such sensors:

```yaml
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 18
        type: bitfield
        name: sensor
        mapping:
          - dps_val: 0
            value: false
          - value: true
      - id: 18
        type: bitfield
        name: fault_code
```

custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
o-mega: my mistake: this is not UV, its just a display light on\off state

make-all: Then use `translation_key: display`

PR #4075, opened 2025-12-01
Add support for EARU 3-Phase Breaker (EAMPDW-TY-63)

o-mega:
**Device Information:**
* **Name:** EARU 3-Phase Breaker Meter
* **Model:** EAMPDW-TY-63
* **Product ID:** `nopfbtfxnwqcrg5y`
* **Config file:** `earu_3phase_breaker.yaml`

**Functionality:**
* **Control:** Main Switch (on/off).
* **Metering:** Total Energy, Energy Today, Energy This Month (working natively).
* **Settings:** Over/Under Voltage, Over Current, Leakage protection.

**Note on Phase Measurements (Complex Data):**
The device transmits real-time metrics (Voltage, Current, Power) for each phase via DPs 6, 7, and 8 encoded as a **Base64 string**.
Currently, I have exposed these DPs as `diagnostic` string sensors (`Phase A raw`, etc.).

**Decoding Logic for Maintainers:**
I have reverse-engineered the Base64 format. It decodes into an 8-byte array (Big Endian):
* **Voltage:** Bytes 0-1 (divide by 10)
* **Current:** Bytes 2-4 (divide by 1000)
* **Power:** Bytes 5-7 (divide by 1)

*Example Python decoding logic:*
```python
# Assuming 'decoded' is the byte array from base64
volts = (decoded[0] << 8 | decoded[1]) / 10.0
amps = (decoded[2] << 16 | decoded[3] << 8 | decoded[4]) / 1000.0
watts = (decoded[5] << 16 | decoded[6] << 8 | decoded[7])
```

I'm submitting the YAML with raw sensors so users can at least control the device and use template sensors for decoding until native base64_struct support is added to the integration.

### DIFF
--- a/custom_components/tuya_local/devices/earu_3phase_breaker.yaml
+++ b/custom_components/tuya_local/devices/earu_3phase_breaker.yaml
@@ -1,4 +1,4 @@
-name: Breaker Meter
+name: Breaker meter
 products:
   - id: nopfbtfxnwqcrg5y
     manufacturer: EARU
@@ -63,8 +63,6 @@ entities:
         optional: true
         unit: W
         mask: "0000000000FFFFFF0000"
-        mapping:
-          - scale: 1
 
   - entity: sensor
     translation_key: voltage_x
@@ -111,8 +109,6 @@ entities:
         optional: true
         unit: W
         mask: "0000000000FFFFFF0000"
-        mapping:
-          - scale: 1
 
   - entity: sensor
     translation_key: voltage_x
@@ -159,8 +155,6 @@ entities:
         optional: true
         unit: W
         mask: "0000000000FFFFFF0000"
-        mapping:
-          - scale: 1
 
   - entity: number
     translation_key: sensor_threshold


custom_components/tuya_local/devices/earu_3phase_breaker.yaml
```diff
@@ -0,0 +1,128 @@
...
+    class: outlet
+    dps:
+      - id: 16
+        type: boolean
+        name: switch
+
+  - entity: sensor
+    name: Total energy
+    class: energy
+    state_class: total_increasing
```
make-all: state_class should be put as `class: total_increasing` under the dps below

custom_components/tuya_local/devices/earu_3phase_breaker.yaml
```diff
@@ -0,0 +1,128 @@
...
+entities:
+  - entity: switch
+    class: outlet
+    dps:
+      - id: 16
+        type: boolean
+        name: switch
+
+  - entity: sensor
+    name: Total energy
```
make-all: Leave out this name to let the default HA translations for energy class be used

custom_components/tuya_local/devices/earu_3phase_breaker.yaml
```diff
@@ -0,0 +1,128 @@
...
+    dps:
+      - id: 10001
+        type: integer
+        name: sensor
+        unit: "kWh"
+        mapping:
+          - scale: 100
+
+  # --- Raw Phase Data ---
+  - entity: sensor
```
make-all, line 24: See tompd_63lw_breaker.yaml for example of decoding these phase data sensors

custom_components/tuya_local/devices/earu_3phase_breaker.yaml
```diff
@@ -0,0 +1,199 @@
+name: EARU 3-Phase Breaker Meter
+products:
+  - id: nopfbtfxnwqcrg5y
+    name: EARU 3-Phase Breaker Meter
```
make-all: This should be broken into `manufacturer` and `model`

custom_components/tuya_local/devices/earu_3phase_breaker.yaml
```diff
@@ -0,0 +1,199 @@
+name: EARU 3-Phase Breaker Meter
```
make-all: This should be generic, all branding belongs under products.

Example "Breaker meter"

custom_components/tuya_local/devices/earu_3phase_breaker.yaml
```diff
@@ -0,0 +1,199 @@
...
+        name: sensor
+        unit: "kWh"
+        class: total_increasing
+        mapping:
+          - scale: 100
+
+  - entity: sensor
+    name: Phase A voltage
+    class: voltage
+    hidden: unavailable
```
make-all: For `hidden: unavailable` to work, there must be a listing under dps with a name `available`.
```yaml
      - id: 6
        type: base64
        optional: true
        name: available
        mapping:
          - dps_val: null
            value: false
          - value: true
```

custom_components/tuya_local/devices/earu_3phase_breaker.yaml
```diff
@@ -0,0 +1,199 @@
...
+      - id: 1
+        type: integer
+        name: sensor
+        unit: "kWh"
+        class: total_increasing
+        mapping:
+          - scale: 100
+
+  - entity: sensor
+    name: Phase A voltage
```
make-all: Instead of name, use
```yaml
    translation_key: voltage_x
    translation_placeholders:
      x: A
```

There are similar keys for current and power.

custom_components/tuya_local/devices/earu_3phase_breaker.yaml
```diff
@@ -0,0 +1,199 @@
...
+        type: base64
+        name: sensor
+        optional: true
+        unit: W
+        mask: "0000000000FFFFFF0000"
+        mapping:
+          - scale: 1
+
+  - entity: number
+    name: Over voltage protection
```
make-all: Instead of this name, use:
```yaml
    translation_key: sensor_threshold
    translation_placeholders:
      sensor: Over voltage
```

Similarly for the rest of the number entities below.

PR #4094, opened 2025-12-05
Fix preset_mode mapping for Vonroc GPH-XA-HEMAN heater (Tuya firmware change)

danyrd92:
This PR fixes the preset_mode mapping for the Vonroc GPH-XA-HEMAN heater in the Tuya Local device definition.

Background:
The device previously reported the preset_mode (DP 4) using string values: -AF
-low
-high

The current YAML maps these to:
- id: 4 name: preset_mode type: string optional: true mapping:
    - dps_val: AF value: "away"
    - dps_val: low value: "eco"
    - dps_val: high value: "comfort"

After a Tuya firmware/product template update, the same DP now reports stringified enum values instead: -"0" → Comfort
-"1" → Eco
-"2" → Anti-freeze (Away)

As a result, the existing mapping no longer matches the raw DP values, and preset_mode stops working correctly in Home Assistant.

Changes:
This PR updates the mapping for DP 4 to align with the new enum values exposed by Tuya:
- id: 4 name: preset_mode type: string optional: true mapping:
    - dps_val: "2"
      value: "away"      # Anti-freeze
    - dps_val: "1"
      value: "eco"
    - dps_val: "0"
      value: "comfort"

The entity-facing preset names (away, eco, comfort) are preserved, so existing automations and front-end usage can remain unchanged.

Testing:
*Verified the DP values using Tuya IoT Platform “Device Debugging”: -mode = "0" → Comfort
-mode = "1" → Eco
-mode = "2" → Anti-freeze
*Applied the updated YAML to Tuya Local.
*Restarted Home Assistant.

Confirmed that:
*The correct preset mode is reported in the climate entity. *Changing presets in Home Assistant successfully switches the heater to the expected mode (comfort/eco/anti-freeze).

Notes:
This change reflects a firmware/product template update on Tuya’s side and brings the device definition back in sync with the actual DP values.

No breaking changes are expected for users already relying on away/eco/comfort preset names in Home Assistant; only the underlying DP mapping is updated.

### DIFF
--- a/custom_components/tuya_local/devices/frankever_bv05_watervalve.yaml
+++ b/custom_components/tuya_local/devices/frankever_bv05_watervalve.yaml
@@ -1,8 +1,8 @@
 name: Water valve
-#products:
-#  - id: UNKNOWN
-#    manufacturer: FrankEver
-#    model: FK-BV05
+# products:
+#   - id: UNKNOWN
+#     manufacturer: FrankEver
+#     model: FK-BV05
 entities:
   - entity: switch
     icon: "mdi:pipe-valve"


danyrd92:
It looks like the CI failure is not related to the changes in this PR.
The failing test (TestDeviceConfig.test_config_files_parse) is complaining about:

custom_components/tuya_local/devices/frankever_bv05_watervalve.yaml
Validation error: products: expected a list

As far as I can see, this file is unrelated to the device definition I modified, and I did not change it in this branch. The test suite appears to validate all device configs in the repository, so this existing issue is surfaced when running the tests for my PR.

I am not sure whether this validation error will prevent this PR from being merged into the main branch, or if you prefer to handle that faulty config separately. Please let me know if you would like me to adjust anything in this PR to move it forward.

make-all:
Unfortunately doing a change like this in such a simplistic way is likely to cause problems for other users. Even if the manufacturer is pushing the firmware update out to existing devices, they are unlikely to all get updated at the same time. There may also be other devices matching this config that are not being updated.
But with these changes, the values are mutually exclusive, so should be able to be handled by either a conditional mapping, or availability flags.

danyrd92:
I understand what you mean. In my specific case, since it is a heater, it had not been used since last winter, so I do not know whether the update comes from the device or from the server.

You are absolutely right in saying that I should have modified the code so that both commands (new and old) remain valid.

Thank you very much for your work and your support.

PR #4103, opened 2025-12-06
Update petree_cube_litter_box.yaml

PurpleFries:
Add name to LED enable/disable switch, Add 'is Cleaning' diagnostic, Add Dump Sand button

### DIFF
--- a/custom_components/tuya_local/devices/petree_cube_litter_box.yaml
+++ b/custom_components/tuya_local/devices/petree_cube_litter_box.yaml
@@ -20,7 +20,6 @@ entities:
           min: 1
           max: 60
   - entity: light
-    name: LED
     translation_key: indicator
     category: config
     dps:


custom_components/tuya_local/devices/petree_cube_litter_box.yaml
```diff
@@ -20,6 +20,7 @@ entities:
           min: 1
           max: 60
   - entity: switch
+    name: LED
```
make-all: The correct change here would be to change the entity type to light so the indicator translation key will be used as originally intended.

custom_components/tuya_local/devices/petree_cube_litter_box.yaml
```diff
@@ -143,3 +144,22 @@ entities:
...
         name: button
+  - entity: button
+    name: Dump Sand
+    category: config
+    dps:
+      - id: 118
+        type: boolean
+        name: button
+  - entity: binary_sensor
+    name: Cleaning
```
make-all, line 154: Maybe `class: running` will improve the state appearance in the UI

PR #4104, opened 2025-12-07
Add HYS-U1S-13A. Add Tank full and Defrost for Electriq PD45E

charlesomer:
Adds smart plug HYS-U1S-13A.

Fixes an issue with Electriq PD45E where laundry and purify are the wrong way round. Also adds Tank full and Defrost sensors.

Let me know if there's anything wrong here :)

### DIFF
--- a/custom_components/tuya_local/devices/smartplugv1.yaml
+++ b/custom_components/tuya_local/devices/smartplugv1.yaml
@@ -94,4 +94,3 @@ entities:
         mapping:
           - scale: 60
             step: 60
-


custom_components/tuya_local/devices/electriq_pd45e_dehumidifier.yaml
```diff
@@ -70,10 +70,37 @@ entities:
...
-    category: diagnostic
     dps:
       - id: 110
         type: integer
         name: sensor
         unit: C
         class: measurement
+  - entity: binary_sensor
+    class: problem
+    name: Tank full
```
make-all: Use `translation_key: tank_full` here (no need for a class or icon either because name, icon and states are all overridden by the translation key). Also the icon override in the mapping is not needed.

custom_components/tuya_local/devices/electriq_pd45e_dehumidifier.yaml
```diff
@@ -70,10 +70,37 @@ entities:
...
+      - id: 11
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 2
+            value: true
+            icon: "mdi:cup-water"
+          - value: false
+  - entity: binary_sensor
+    name: Defrost
```
make-all: Use `translation_key: defrost` instead of the name, icon and class here, also mapping is not needed.

custom_components/tuya_local/devices/smartplugv1.yaml
```diff
@@ -6,6 +6,9 @@ products:
     model: EP2
   - id: 37mnhia3pojleqfh
     model: SP111-A
+  - id: kOKEJBFoF9VNB5FR
+    model: HYS-U1S-13A
+    manufacturer: Tuya
```
make-all: Manufacturer is not Tuya, they only make the Wifi module, if you do not know the manufacturer, leave it out.

charlesomer:
Apologies for the delay in getting back to you, thank you for completing this :)

PR #4108, opened 2025-12-08
Add model to device info

FredrikM97:
This allow to partly fix https://github.com/make-all/tuya-local/discussions/3474 and expose the manufacturer + model for other integrations. 

This give a little more info on the device too:
image

### DIFF
--- a/custom_components/tuya_local/config_flow.py
+++ b/custom_components/tuya_local/config_flow.py
@@ -533,6 +533,7 @@ def create_test_device(hass: HomeAssistant, config: dict):
     subdevice_id = config.get(CONF_DEVICE_CID)
     device = TuyaLocalDevice(
         "Test",
+        "Test Model",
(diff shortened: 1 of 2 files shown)


make-all:
This is not exposing the model, it is incorrectly tying the config file to device constructor, and using a fake for the conection test, which is when such information would be most useful.
I prefer to leave this until the connection rework when the change can be done correctly.

FredrikM97:
I see, this was the closest I could come up with for any indication of the model. Any time line for the rework or anything yoy need help with?

make-all:
Currently I am waiting for tinytuya to release 2.0.0 to begin the work, as that will anyway require a major rewrite.

FredrikM97:
I see, looks like they just started with 2.0.0?

make-all:
I'm going to close this, as bringing config info into the device creates circular references based on past experience, it is important to keep these separate.

PR #4109, opened 2025-12-08
fix(ilife_a30_pro_vacuum): Use pause DP for activation, map correct status

MIKOLAJW197:
1. **Tuya DPS structure:**
   - `switch_go` (dp_id: 1, type: bool)
   - `pause` (dp_id: 2, type: bool)
   - `switch_charge` (dp_id: 3, type: bool)

2. Previously, using the **`activate`** DPS caused incorrect behavior:
   - `start` was incorrectly triggering **pause = true**
   - `pause` was incorrectly triggering **pause = false**

3. Updated the logic to use the **`pause` DPS** directly for proper start/pause handling.

4. Adjusted status mapping so that the Tuya `"smart"` mode is now correctly recognized as **`cleaning`**.

 **Tuya DPS structure:**
   ```json
   {
     "code": "switch_go",
     "custom_name": "",
     "dp_id": 1,
     "time": 1764709913492,
     "type": "bool",
     "value": false
   },
   {
     "code": "pause",
     "custom_name": "",
     "dp_id": 2,
     "time": 1764665210045,
     "type": "bool",
     "value": false
   },
   {
     "code": "switch_charge",
     "custom_name": "",
     "dp_id": 3,
     "time": 1764710024252,
     "type": "bool",
     "value": false
   }
 ```

### DIFF
--- a/custom_components/tuya_local/devices/ilife_a30_pro_vacuum.yaml
+++ b/custom_components/tuya_local/devices/ilife_a30_pro_vacuum.yaml
@@ -11,7 +11,7 @@ entities:
         name: power
       - id: 2
         type: boolean
-        name: activate
+        name: pause
       - id: 4
         type: string
         name: command
@@ -39,7 +39,7 @@ entities:
           - dps_val: standby
             value: standby
           - dps_val: smart
-            value: smart
+            value: cleaning
           - dps_val: zone_clean
             value: zone
           - dps_val: part_clean
--- a/custom_components/tuya_local/vacuum.py
+++ b/custom_components/tuya_local/vacuum.py
@@ -45,6 +45,7 @@ class TuyaLocalVacuum(TuyaLocalEntity, StateVacuumEntity):
         self._locate_dps = dps_map.get("locate")
         self._power_dps = dps_map.get("power")
         self._activate_dps = dps_map.get("activate")
+        self._pause_dps = dps_map.get("pause")
         self._direction_dps = dps_map.get("direction_control")
         self._error_dps = dps_map.get("error")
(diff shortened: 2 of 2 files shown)


make-all:
It should not be necessary to add new handling of ficticious features to vacuums to get this working. Other vacuums are working OK without it. It seems that you have anyway mostly just duplicated the activate logic for a dp named pause instead.

Regarding the change from smart to cleaning. That should also not be needed. It is usually better to differentiate the different cleaning modes in status, but for the vacuum state any unrecognised statuses are converted to "cleaning" automatically.

MIKOLAJW197:
Hi @make-all 

1. Without this change, **`activate` works incorrectly**:  
   - Pressing **pause** on an entity makes it **start**, and clicking **play** makes it **pause**.  
   - This happens because the **wrong Boolean parameter is passed**:  
     - `Start` → `active = True`  
     - In the case of DP **pause**, we are sending `pause = True`.

2. Regarding the mapping:  
   - When the vacuum is started in **'smart' mode**, the entity still shows **'paused'**.

Edit: It seems that we are based on **active** with False it it catched here: (which in fact is **paused**)

```
elif self._activate_dps and self._activate_dps.get_value(self._device) is False:
            return VacuumActivity.PAUSED

```

If you have other idea to fix this just leave me a comment and I will implement it and test on my local device ;)

make-all:
It sounds like the value just needs to be inverted, no need to introduce more complexity in the python code.
```yaml
      - id: 2
        type: boolean
        name: activate
        mapping:
          - dps_val: true
            value: false
          - dps_val: false
            value: true
```

PR #4112, opened 2025-12-09
[feat] add frizzlife_lp365p_watermonitorandshutdoff

nalditopr:
led by community work

https://community.home-assistant.io/t/frizzlife-smart-water-monitor/875335

added support for [Frizzlife Smart Water Monitor](https://www.frizzlife.com/products/lp365 )

### DIFF
--- a/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
+++ b/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
@@ -56,7 +56,7 @@ entities:
         name: sensor
         class: measurement
         unit: L/h
-  
+
   # Current flow rate
   - entity: sensor
     class: volume_flow_rate
@@ -105,7 +105,7 @@ entities:
         mapping:
           - dps_val: "AUTO"
             value: "Auto"
-          - dps-val: "ADVANCED_SETTING"
+          - dps_val: "ADVANCED_SETTING"
             value: "Advanced"
 
   # AUTO Mode #


custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
```diff
@@ -0,0 +1,540 @@
...
+entities:
+  # Valve
+  - entity: valve
+    class: water
+    dps:
+      - id: 1
+        type: boolean
+        name: valve
+
+  # Temperature sensor (converted to F)
```
make-all: Leave unit conversions to HA, don't try to do them in a config file.

The range and units should be the device's native representation

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
```diff
@@ -0,0 +1,540 @@
...
+    class: temperature
+    dps:
+      - id: 12
+        type: integer
+        name: sensor
+        unit: °F
+        class: measurement
+        mapping:
+          - scale: 10.0
+            invert: false
```
make-all: There is no need to specify invert: false, that is the normal case

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
nalditopr: Thanks for the feedback, I will refactor and commit new PR as soon as possible. Thanks for your efforts.

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
```diff
@@ -0,0 +1,538 @@
...
+    name: Volume Today
+    class: water
+    dps:
+      - id: 103
+        type: integer
+        optional: true
+        name: sensor
+        class: total_increasing
+        unit: gal
+        mapping:
```
make-all, line 45: Again, leave the unit conversion to HA (leave the unit as L, and remove any transformation).

In this case, there is no user preference, so the unit conversion will not be automatic like it is for temperature, but you should be able to select the units for display in the entity properties in the UI.

PR #4113, opened 2025-12-09
Add SIME ECOMAXI VB 200 heat pump water heater

m4rt0s:
Added new device config custom_components/tuya_local/devices/sime_ecomaxi_vb200_heatpump.yaml for SIME ECOMAXI VB 200 heat pump water heater: climate (on/off, presets auto/eco/turbo, setpoint 35–65 °C, current temp) + switches (power, anti-legionella, electric anode) + diagnostic sensors/binary_sensors (evaporator/air outlet/ambient/solar collector temps, EEV opening, board temps, pump/compressor/fan, defrost, electric heater, error code). Names/icons inline; no new translation keys or global icons.
Screenshot 2025-12-11 at 10 26 17

Tests: pytest tests/test_device_config.py; yamllint
 custom_components/tuya_local/devices/sime_ecomaxi_vb200_heatpump.yaml

Product details:
https://doc.sime.it/SimeWSPortale/DownloadPdf?docId=3296f1ce-860e-4115-a285-827b28e28a96&fileName=6333542G%20Ecomaxi%20VB%20200-300%20(IT-EN-ES-DE-EL-FR-NL).pdf&lingua=EN

### DIFF
--- a/custom_components/tuya_local/devices/sime_ecomaxi_vb200_waterheater.yaml
+++ b/custom_components/tuya_local/devices/sime_ecomaxi_vb200_waterheater.yaml
@@ -1,8 +1,8 @@
-name: Heat pump water heater
+name: Water heater
 products:
   - id: ukqqxsk7vfz8co9z
-    manufacturer: SIME
-    model: ECOMAXI VB 200 WiFi
+    manufacturer: Sime
+    model: EcoMaxi VB 200 WiFi
 entities:
   - entity: water_heater
     dps:
@@ -125,7 +125,7 @@ entities:
   - entity: binary_sensor
     name: Water pump
     icon: "mdi:pump"
-    class: power
+    class: running
     category: diagnostic
     dps:
       - id: 28
@@ -134,7 +134,7 @@ entities:
   - entity: binary_sensor
     name: Compressor
     icon: "mdi:engine"
-    class: power
+    class: running
     category: diagnostic
     dps:
       - id: 29
@@ -143,7 +143,7 @@ entities:
   - entity: binary_sensor
     name: Fan
     icon: "mdi:fan"
-    class: power
+    class: running
     category: diagnostic
     dps:
       - id: 27
@@ -159,7 +159,6 @@ entities:
   - entity: binary_sensor
     name: Electric heater
     icon: "mdi:lightning-bolt"
-    class: power
     category: diagnostic
     dps:
       - id: 33


custom_components/tuya_local/devices/sime_ecomaxi_vb200_heatpump.yaml
```diff
@@ -0,0 +1,180 @@
+name: Heat pump water heater
+products:
+  - id: ukqqxsk7vfz8co9z
+    manufacturer: SIME
+    model: ECOMAXI VB 200 WiFi
+entities:
+  - entity: climate
```
make-all: water_heater has its own entity type

custom_components/tuya_local/devices/sime_ecomaxi_vb200_heatpump.yaml
```diff
@@ -0,0 +1,180 @@
...
+        unit: C
+        range:
+          min: 35
+          max: 65
+      - id: 16
+        name: current_temperature
+        type: integer
+        unit: C
+  - entity: switch
+    name: Power
```
make-all: This is just duplicating the control above

custom_components/tuya_local/devices/sime_ecomaxi_vb200_heatpump.yaml
```diff
@@ -0,0 +1,180 @@
...
+    name: Fan
+    icon: "mdi:fan"
+    class: running
+    category: diagnostic
+    dps:
+      - id: 27
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
+    name: Defrost
```
make-all: Use `translation_key: defrost` instead of the explicit name, icon and class.

custom_components/tuya_local/devices/sime_ecomaxi_vb200_heatpump.yaml
```diff
@@ -0,0 +1,180 @@
...
+  - entity: binary_sensor
+    name: Electric heater
+    icon: "mdi:lightning-bolt"
+    class: heat
+    category: diagnostic
+    dps:
+      - id: 33
+        type: boolean
+        name: sensor
+  - entity: sensor
```
make-all: I think it is better to follow the standard pattern used in other configs of a binary_sensor with `class: problem`, the sensor mapped to false for 0, true for any other value, and a fault_code attribute with the raw value. Generally this type of field is a bitfield (bitmap in Tuya docs) rather than an integer, but the difference only matters when you start mapping non-zero values.

custom_components/tuya_local/devices/sime_ecomaxi_vb200_heatpump.yaml
```diff
@@ -0,0 +1,180 @@
...
+        type: boolean
+        mapping:
+          - dps_val: false
+            value: "off"
+            icon: "mdi:water-boiler-off"
+          - dps_val: true
+            value: heat
+            icon: "mdi:water-boiler"
+      - id: 2
+        name: preset_mode
```
make-all: After changing to water_heater, these should combine with the above to map to `operation_mode`, as `heat_pump`. `eco`, `performance`

```yaml
    - id: 1
      type: boolean
      name: operation_mode
      mapping:
        - dps_val: false
          value: "off"
         - dps_val: true
           constraint: mode
           conditions:
             - dps_val: "0x00"
               value: heat_pump
             - dps_val: "0x02"
                value: eco
              - dps_val: "0x03"
                value: performance
    - id: 2
      type: string
      name: mode
      hidden: true
```

m4rt0s:
Thanks for the review! I’ve applied the requested changes:

- Switched the primary entity to water_heater and mapped DP1+DP2 to operation_mode with off/heat_pump/eco/performance (DP2 hidden as mode).

- Removed the duplicate DP1 power switch.

- Used translation_key: defrost for the defrost binary sensor.

- Modeled DP15 as a problem binary sensor with 0→false, others→true, plus fault_code bitfield.

m4rt0s:
I separated the modes into a dedicated select because the device drops back to off whenever HA writes DP1 and DP2 together. With on/off isolated on DP1 and the mode on its own select (DP2), Home Assistant no longer sends a combined write that turns the heater off when changing modes. This mirrors the pattern used in other Tuya water heaters where the mode DPS can be set independently while the power DPS stays stable.

custom_components/tuya_local/devices/sime_ecomaxi_vb200_waterheater.yaml
```diff
@@ -0,0 +1,181 @@
...
+    dps:
+      - id: 2
+        name: option
+        type: string
+        mapping:
+          - dps_val: "0x00"
+            value: auto
+            icon: "mdi:autorenew"
+          - dps_val: "0x02"
+            value: "heat pump"
```
make-all, line 42: Since this no longer needs to correspond to the HA water_heater supported operation_modes, you could go back to the original names if that is more obvious - in line with manual and Tuya app.

m4rt0s:
Thanks for the thorough code review. I believe everything is now in place, really appreciate the work you do to keep this project in shape.

custom_components/tuya_local/devices/sime_ecomaxi_vb200_waterheater.yaml
```diff
@@ -125,7 +125,7 @@ entities:
   - entity: binary_sensor
     name: Water pump
     icon: "mdi:pump"
-    class: running
+    class: power
```
make-all: I'm not sure there is any point changing the class to power, according to the docs at https://www.home-assistant.io/integrations/binary_sensor/ it just shows the states "On" and "Off", which is the default, and you have already overridden the name and icon. I would leave this class for actual power sensors only. Running at least changes the state display, so does make some sense at least for Water pump, Compressor and Fan.

PR #4119, opened 2025-12-10
Add support for SRNE SE Series lithium battery storage

barijaona:


custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
```diff
@@ -0,0 +1,310 @@
+name: LiFP battery
+products:
+  - id: l2ak4zqufuarm0j3
+    manufacturer : SRNE
+    model: SR-SE05B
+entities:
+  - entity: sensor
+    class: battery
+    translation_key: Charge
```
make-all: No translation_key by this name exists. There is no reason to override the default HA translations from the battery class, so just leave it out

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
```diff
@@ -0,0 +1,310 @@
...
+    dps:
+      - id: 101
+        type: integer
+        name: sensor
+        unit: A
+        class: measurement
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Cycle number
```
make-all: Is this frequency?

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
```diff
@@ -0,0 +1,310 @@
...
+  - entity: sensor
+    category: diagnostic
+    class: temperature
+    name: Environment temperature
+    dps:
+      - id: 119
+        type: integer
+        optional: true
+        name: sensor
+        translation_key: Environment temperature
```
make-all: There is no translation key by this name, and anyway translation keys do not belong under the dps config.

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
```diff
@@ -0,0 +1,310 @@
...
+    class: voltage
+    name: Cell voltage 1
+    dps:
+      - id: 120
+        type: integer
+        optional: true
+        name: sensor
+        unit: V
+        class: measurement
+        precision: 3
```
make-all: Does this really need overriding? 3 should be the default with the given scale factor.

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
```diff
@@ -0,0 +1,310 @@
...
+        name: sensor
+        translation_key: Environment temperature
+        unit: C
+        class: measurement
+        mapping:
+          - scale: 10
+  - entity: sensor
+    category: diagnostic
+    class: voltage
+    name: Cell voltage 1
```
make-all: I would suggest to use:
```
translation_key: voltage_x
translation_placeholders:
  x: Cell 1
```
This will allow the word "Voltage" to be translated and will do reordering if needed according to language.
In English, "Voltage Cell 1" probably makes more sense, because the "1" is referring to the Cell number, not a different voltage reading of a single cell.

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
```diff
@@ -0,0 +1,310 @@
...
+    translation_key: Charge
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    class: voltage
+    translation_key: Battery voltage
```
make-all: No translation key by this name exists, make it a name, or even just leave it out and use the default "voltage" class translation, since it is the main voltage sensor this device has so it is not ambiguous (it will be obvious from the value that it is a battery voltage, not a mains voltage).

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
```diff
@@ -0,0 +1,310 @@
...
+        type: integer
+        name: sensor
+        unit: V
+        class: measurement
+        precision: 2
+        mapping:
+          - scale: 100
+  - entity: sensor
+    class: current
+    name: Charge current
```
make-all: Debatable whether the name is needed here. It does add some clarification so I will allow it, but removing it will have the advantage of allowing translations to be used.

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
```diff
@@ -0,0 +1,310 @@
...
+        class: measurement
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Cycle number
+    dps:
+      - id: 108
+        type: integer
+        name: sensor
+        unit: cyc
```
make-all: I don't think this is a valid unit in any system, either Hz if this is frequency, or spell it out in full as `cycles`

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
barijaona: No, it's the number of cycles the battery has been through, reflecting its age

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
barijaona: If I don't include `precision: 3`, the value displayed is truncated to `3 V`, which is not really useful…

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
barijaona: Removed.

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
barijaona: Removed

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
barijaona: Changed it to `cycles`

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
barijaona: Removed

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
barijaona: I applied your advice on the 16 cells

custom_components/tuya_local/devices/srne_energy_storage_battery.yaml
make-all: I see this is a recent behaviour change in the HA UI. My voltages are also showing with no decimals, where before they had 1 decimal place (scale: 10). I don't really want to have to set the precision explicitly in 1300+ config files, so I guess I will have to make a change to send a precision based on scale in the default case instead of letting HA handle defaults.

You can remove the precision from here.

PR #4120, opened 2025-12-10
Add Henden Essential Heat Pump device

marcins:
Adds config for the Henden Essential Pool Heat Pump: https://www.hendenpools.com.au/products/p/henden-essential-inverter-heat-pump-21kw

I worked out a lot of the DPs manually / with Tuya debugging output, but then found the `fairland_x20_poolheatpump.yaml` was pretty close so copied a few things from there.

I had to leave off the availability check for the power sensor (DPS 113), as it reports `False` even though the power reported on DPS 112 appears to be correct (based on comparing with the consumption monitor in my pool equipment switchboard).

For reference, here's the output from accessing the device directly with TinyTuya when the heat pump is running:
```
Device status: {'dps': {'101': True, '102': 'smart', '103': 27, '104': False, '105': 100, '106': 'warm', '107': 28, '108': 18, '109': 40, '110': 0, '111': 0, '112': 3360, '113': False, '114': '1', '115': '0'}}
```

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,6 @@
 fuzzywuzzy
 levenshtein
+pycares~=4.11.0
 pytest-homeassistant-custom-component==0.13.254
 pytest
 pytest-asyncio


custom_components/tuya_local/devices/henden_essential_pool_heatpump.yaml
```diff
@@ -0,0 +1,146 @@
...
+          min: 18 # dps 108 is target_temp_low - can this value be driven by that?
+          max: 40 # dps 109 is target_temp_high
+      - id: 108
+        type: integer
+        name: min_temperature
+      - id: 109
+        type: integer
+        name: max_temperature
+      - id: 114
+        type: boolean # actually enum?
```
make-all: So it should be a `string`.  The type is used to match the data coming from the device, not the HA value after mapping.

custom_components/tuya_local/devices/henden_essential_pool_heatpump.yaml
```diff
@@ -0,0 +1,146 @@
...
+      - id: 106
+        name: mode
+        type: string
+        hidden: true
+      - id: 107
+        name: temperature
+        type: integer
+        unit: C
+        range:
+          min: 18 # dps 108 is target_temp_low - can this value be driven by that?
```
make-all: min and max temperature will be passed to HA as the range if available, to constrain the UI. But the widest range supported should still be specified here as a backup, and will be used to prevent out of range values being sent to the device.

ps-jay:
This is so good!  Thank you @marcins !
I installed my heat pump just last week and powered it on today. 

I've setup an integral helper for kWh energy based on the power consumption, to add it to the energy dashboard too!

PR #4123, opened 2025-12-11
fix(vacuum): Mark status 'docked' as entity state 'docked'

MIKOLAJW197:
When DP value is mapped to **docked** as status state of the device is shown as **idle**

image


make-all:
Yes this seems to be an oversight, maybe because older vacuums did not have as many distinct states as the iLife A30 does, and always reported charging or charged when docked.

PR #4126, opened 2025-12-12
Feature/add-device-config Casdon's Inline Water Dispenser

GavinIves:
Added a new configuration file for Casdon's Inline Water Dispenser. It has been tested by me and has been running well for some time.

### DIFF
--- a/custom_components/tuya_local/translations/pt-PT.json
+++ b/custom_components/tuya_local/translations/pt-PT.json
@@ -942,7 +942,7 @@
                     "drain_cold_water": "Drenar água fria",
                     "drain_hot_water": "Drenar água quente",
                     "drain_normal_water": "Drenar água normal",
-                    "high_temperature_emptying": "Esvaziamento de alta temperatura"
+                    "high_temperature_emptying": "Esvaziamento de alta temperatura",
                     "high_temperature_emptied": "Esvaziado em alta temperatura"
                 }
             },


GavinIves:
I'm a beginner, this is my first time using Git, please contact me if you have any questions. Since this device is only sold in mainland China, I can replace the Chinese characters in the files with English if needed. This configuration file has been running stably on my HA for a month without any problems. I still have a question: if I write in English, where should I write the translation file so that it displays Chinese instead of English in the Chinese-language HA?

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
```diff
@@ -0,0 +1,199 @@
+name: Inline Water Dispenser
+products:
+  - id: szx3eawxigxumxi5
+    name: KG1
+    manufacturer: Casdon
+    model: Y-GX1200-KG1
+
+entities:
+  - entity: switch
```
make-all: Consider using a valve entity, with `class: water`
No name or icon should be necessary, and default HA translations icons for water valve will be used.

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
```diff
@@ -0,0 +1,199 @@
...
+    translation_key: status
+    class: enum
+    icon: mdi:cup-water
+    dps:
+      - id: 5
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: standby
+            value: 待机
```
make-all: Use the already defined translation keys from https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/zh-Hans.json#L853 where possible.

If any extra are needed, we can consider adding them.

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
```diff
@@ -0,0 +1,199 @@
...
+            value: 出冷水
+          - dps_val: high_temp_emptying
+            value: 高温排水中
+          - dps_val: high_temp_emptyed
+            value: 高温排水完成
+          - dps_val: water_shortage
+            value: 缺水
+
+  - entity: lock
+    name: 童锁
```
make-all: Use `translation_key: child_lock` instead of this name and icon.

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
```diff
@@ -0,0 +1,199 @@
...
+
+  - entity: switch
+    name: 制冷开关
+    icon: mdi:snowflake
+    dps:
+      - id: 102
+        type: boolean
+        name: switch
+
+  - entity: switch
```
make-all, line 73: Use `entity: light`, depending on the type of light this is, it may also be useful to use `translation_key: indicator`, `translation_key: display`, `translation_key: backlight`

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
```diff
@@ -0,0 +1,199 @@
...
+    name: 照明开关
+    icon: mdi:lightbulb
```
make-all: icon should not be needed if you use a light entity. Name should also not be needed.

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
```diff
@@ -0,0 +1,199 @@
...
+    dps:
+      - id: 103
+        type: boolean
+        name: switch
+
+  - entity: number
+    name: 制冷温度
```
make-all: I think you could consider using a climate entity to combine this and the refrigeration switch above and temperature sensor below.

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
```diff
@@ -0,0 +1,199 @@
...
+
+  - entity: number
+    name: 出水量设置
+    icon: mdi:water
+    class: volume
+    dps:
+      - id: 101
+        type: integer
+        name: value
+        unit: ml
```
make-all: Unit should be mL

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
```diff
@@ -0,0 +1,199 @@
...
+    class: volume
+    dps:
+      - id: 26
+        type: integer
+        name: sensor
+        unit: ml
+        class: measurement
+
+  - entity: binary_sensor
+    name: 故障状态
```
make-all: Leave out the name and let the default HA translation for "problem" class be used.

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
```diff
@@ -0,0 +1,199 @@
...
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 6
```
make-all, line 144: Normally it is useful to also add the fault_code as a attribute so automations can still check the code:
```
- id: 6
  type: bitfield
  name: sensor
  mapping:
    - dps_val: 0
      value: false
    - value: true
- id: 6
  type: bitfield
  name: fault_code
```

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
GavinIves: ok

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
GavinIves: This switch remains on because it's not a light source, but rather a light-on mode. Turning it on indicates turning on the light to fill with water, not that the light is currently on.

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
GavinIves: It's too complicated. It's just a refrigeration unit, which makes it easier to use.

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
GavinIves: ok

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
GavinIves: ok

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
GavinIves: ok

GavinIves (commented):
fix and translation

custom_components/tuya_local/devices/water_dispenser_kg1.yaml
GavinIves: ok

custom_components/tuya_local/devices/casdon_kg1_waterdispenser.yaml
```diff
@@ -1,185 +1,185 @@
...
-      - bit: 9
-        name: E10:Cooling_fault
-      - name: Normal
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 6
+        type: bitfield
+        name: sensor
```
make-all, line 146: and this should be named `description`, since it is not boolean it cannot be the sensor of a binary_sensor

GavinIves: yes

GavinIves:
Hello, I'd like to know how to set a step for my number so that it can only increment or modify a fixed value each time, like a slider that allows you to set the number by sliding instead of typing it in. Can this be done in our project, or does it need to be configured in HA? Thank you very much.

make-all:
```
mapping:
  - step: 10
 ```
under the dp definition will give a step size of 10 (when you change it will snap to nearest)

GavinIves:
Please do not change the temperature control like this. It has two temperature controls, one for hot water and one for chilled water. You have combined them into one. It's fine that the cooling temperature can only be adjusted when the cooling function is on, but the hot water temperature cannot be adjusted during cooling. This machine can dispense hot water while cooling.

Another problem is that this water outlet button doesn't translate into a valve.
image

make-all:
The latter would seem to be a bug in HA, since the translation file definitely defines a name for water class valves.

https://github.com/home-assistant/core/blob/dev/homeassistant/components/valve/strings.json

But this could also be complicated by the fact that you already installed it using your config, so the entity name changes in later versions of the config are not being fully applied.

GavinIves:
Okay, please display the hot water temperature and cooling temperature separately, not combined. Otherwise, I can't control the cooling temperature and the hot water temperature at the same time.
image

PR #4137, opened 2025-12-14
feat: Add support for Tion Breezer Bio X (9mqdhwklpvnnvb7t)

levikinva:
Device: Tion Breezer Bio X
Product ID: 9mqdhwklpvnnvb7t
Link to product: https://tion.kz/product/breezer-tion-bio-x/

t1
t2

### DIFF
--- a/custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
+++ b/custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
@@ -1,6 +1,8 @@
-name: Tion Breezer Bio X
+name: Ventilator with heater
 products:
   - id: 9mqdhwklpvnnvb7t
+    manufacturer: Tion
+    model: Breezer Bio X
 entities:
   # Main entities
   - entity: switch
@@ -32,7 +34,7 @@ entities:
           min: 0
           max: 30
   - entity: sensor
-    name: "Indoor Temperature"
+    name: "Indoor temperature"
     class: temperature
     dps:
       - id: 108
@@ -40,7 +42,7 @@ entities:
         type: integer
         unit: "°C"
   - entity: sensor
-    name: "Outdoor Temperature"
+    name: "Outdoor temperature"
     class: temperature
     dps:
       - id: 112
@@ -48,7 +50,7 @@ entities:
         type: integer
         unit: "°C"
   - entity: sensor
-    name: "Filter Life"
+    name: "Filter life"
     icon: "mdi:air-filter"
     dps:
       - id: 115
@@ -56,62 +58,83 @@ entities:
         type: integer
         unit: "%"
   - entity: sensor
-    name: "CO2"
     class: carbon_dioxide
     dps:
       - id: 106
         name: sensor
         type: integer
         unit: "ppm"
+      - id: 122
+        name: available
+        type: string
+        mapping:
+          - dps_val: "absent"
+            value: false
+          - dps_val: null
+            value: false
+          - value: true
   - entity: sensor
-    name: "PM2.5"
     class: pm25
     dps:
       - id: 116
         name: sensor
         type: integer
         unit: "µg/m³"
+      - id: 123
+        name: available
+        type: string
+        mapping:
+          - dps_val: "absent"
+            value: false
+          - dps_val: null
+            value: false
+          - value: true
   - entity: sensor
-    name: "Humidity"
     class: "humidity"
     dps:
       - id: 119
         name: sensor
         type: integer
         unit: "%"
+      - id: 124
+        name: available
+        type: string
+        mapping:
+          - dps_val: "absent"
+            value: false
+          - dps_val: null
+            value: false
+          - value: true
 
   # Configuration entities
   - entity: switch
-    name: "Night Mode"
+    translation_key: sleep
     category: config
-    icon: "mdi:weather-night"
     dps:
       - id: 132
         name: switch
         type: boolean
-  - entity: switch
-    name: "Child Lock"
+  - entity: lock
+    translation_key: child_lock
     category: config
-    icon: "mdi:hand-back-right-off"
     dps:
       - id: 111
-        name: switch
+        name: lock
         type: boolean
-  - entity: select
-    name: "Light Mode"
+  - entity: light
+    name: "Indicator light"
     category: config
-    icon: "mdi:lightbulb-on-outline"
     dps:
       - id: 104
-        name: option
+        name: brightness
         type: string
         mapping:
           - dps_val: "0"
-            value: "Off"
+            value: 0
           - dps_val: "50"
-            value: "Dim"
+            value: 128
           - dps_val: "100"
-            value: "On"
+            value: 255
   - entity: switch
     name: "Sound"
     category: config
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+    icon: "mdi:information-outline"
+    dps:
+      - id: 118
+        name: sensor
+        type: string
+  - entity: sensor
+    name: "Total Work Time"
+    category: diagnostic
+    class: duration
+    disabled: true
```
make-all: If the intention is to disable this entity by default, then `hidden: true` will do that.

levikinva: Thanks for the review! Fixed.

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
+name: Tion Breezer Bio X
```
make-all: Branding should be in the products section (as manufacturer and model tags).
Top level name should be generic in case of other devices matching the same config.

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+products:
+  - id: 9mqdhwklpvnnvb7t
+entities:
+  # Main entities
+  - entity: switch
```
make-all, line 8: I think this device should be a higher level type of device, maybe climate since it has temperature control also.

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+  - entity: sensor
+    name: "Filter Life"
+    icon: "mdi:air-filter"
+    dps:
+      - id: 115
+        name: sensor
+        type: integer
+        unit: "%"
+  - entity: sensor
+    name: "CO2"
```
make-all: sensors with classes should not need names that are essentially the same as the class.

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+  - entity: switch
+    name: "Night Mode"
+    category: config
+    icon: "mdi:weather-night"
+    dps:
+      - id: 132
+        name: switch
+        type: boolean
+  - entity: switch
+    name: "Child Lock"
```
make-all: Standard pattern is to use a lock entity for this:

```
- entity: lock
  translation_key: child_lock
  category: config
  dps:
    - id: 111
      type: boolean
      name: lock
```

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+    class: "humidity"
+    dps:
+      - id: 119
+        name: sensor
+        type: integer
+        unit: "%"
+
+  # Configuration entities
+  - entity: switch
+    name: "Night Mode"
```
make-all: I think `translaton_key: sleep` can be used here instead of this name and icon

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+        type: boolean
+  - entity: switch
+    name: "Child Lock"
+    category: config
+    icon: "mdi:hand-back-right-off"
+    dps:
+      - id: 111
+        name: switch
+        type: boolean
+  - entity: select
```
make-all: This should be a light entity, with dp name brightness, and values mapped to 0, 128, 255

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+  - entity: switch
+    name: "Sound"
+    category: config
+    icon: "mdi:volume-high"
+    dps:
+      - id: 103
+        name: switch
+        type: boolean
+  - entity: switch
+    name: "Filter Reset"
```
make-all: Should be a `button` entity, with `translation_key: filter_reset`

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+  - entity: sensor
+    name: "Indoor Temperature"
+    class: temperature
+    dps:
+      - id: 108
+        name: sensor
+        type: integer
+        unit: "°C"
+  - entity: sensor
+    name: "Outdoor Temperature"
```
make-all: HA naming convention is to only capitalise the first word of entity and device names (unless abbreviations or branding is involved)

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+        unit: "days"
+        range:
+          min: 30
+          max: 365
+        mapping:
+          - step: 1.0
+
+  # Diagnostic entities
+  - entity: sensor
+    name: "Heater Type"
```
make-all: This will never change, so probably not useful as a sensor. May be an attribute on another entity, especially if the value can be used to selectively enable features.

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
```diff
@@ -0,0 +1,237 @@
...
+  - entity: sensor
+    name: "Error Code"
+    category: diagnostic
+    icon: "mdi:alert-circle-outline"
+    dps:
+      - id: 127
+        name: sensor
+        type: string
+  - entity: sensor
+    name: "CO2 Sensor Status"
```
make-all: should maybe just be an attribute of the CO2 sensor. Especially if it is just indicating whether the sensor is fitted or not, in which case it can be the `available` attribute with a mapping to true/false depending on value.

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: I tried the climate implementation, but found it misleading. It implies the device can control the room's temperature, which it cannot; it only heats the incoming air. This is a secondary function, as the device's main purpose is ventilation.

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: This is diagnostic information about the device's heater type. I couldn't find a suitable way to display it as a static label. It's included with `hidden: true` for completeness. This value is purely informational and cannot be used to enable or disable any device function, as all device configurations come with a heater.

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: fix

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: fix

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: fix

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: fix

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: fix

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: fix

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: fix

custom_components/tuya_local/devices/tion_breezer_bio_x.yaml
levikinva: fix

levikinva:
@make-all Please review

eXist-FraGGer:
Hello, when this functionality will be added and why it is closed?

PR #4140, opened 2025-12-14
Added Create Warm Clear Heater device

carpabg:
Added Create Warm Clear Heater device

### DIFF
--- a/custom_components/tuya_local/devices/create_warm_clear_heater.yaml
+++ b/custom_components/tuya_local/devices/create_warm_clear_heater.yaml
@@ -52,7 +52,7 @@ entities:
           - dps_val: working
             value: heating
       - id: 26
-        type: text
+        type: string
         optional: true
         name: custom_program
   - entity: select


custom_components/tuya_local/devices/create_warm_clear_heater.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - id: 4
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: CF
+            value: comfort
+          - dps_val: EC
+            value: eco
+          - dps_val: AF
+            value: antifrost
```
make-all: `away` is the standard HA preset name for AF function

custom_components/tuya_local/devices/create_warm_clear_heater.yaml
```diff
@@ -0,0 +1,159 @@
...
+            value: scheduled
+      - id: 11
+        type: string
+        name: hvac_action
+        mapping:
+          - dps_val: standby
+            value: idle
+          - dps_val: working
+            value: heating
+      - id: 26
```
make-all, line 54: Make it a `text` entity, so it can be edited. If the data is base64 encoded, then make the type base64 to get some input checking.
Mark the entity as `hidden: true`, because most users will not know how to use it.

custom_components/tuya_local/devices/create_warm_clear_heater.yaml
```diff
@@ -0,0 +1,159 @@
...
+    translation_key: time_remaining
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 20
+        type: integer
+        name: sensor
+        unit: min
+  - entity: switch
+    name: Night mode
```
make-all: `translation_key: sleep` seems to fit closely enough.

custom_components/tuya_local/devices/create_warm_clear_heater.yaml
```diff
@@ -0,0 +1,159 @@
+name: CREATE WARM CLEAR Heater
+products:
+  - id: ggz9d3bputnvntlx
+    manufacturer: Create
+    model: CREATE - WARM CLEAR
```
make-all: model does not need to repeat the manufacturer name, and preferably not in all caps.

custom_components/tuya_local/devices/create_warm_clear_heater.yaml
```diff
@@ -0,0 +1,159 @@
+name: CREATE WARM CLEAR Heater
```
make-all: Top level name should be without branding (just "Heater" for example)

custom_components/tuya_local/devices/create_warm_clear_heater.yaml
carpabg: Done

custom_components/tuya_local/devices/create_warm_clear_heater.yaml
carpabg: done

custom_components/tuya_local/devices/create_warm_clear_heater.yaml
carpabg: done

custom_components/tuya_local/devices/create_warm_clear_heater.yaml
carpabg: done

custom_components/tuya_local/devices/create_warm_clear_heater.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - id: 11
+        type: string
+        name: hvac_action
+        mapping:
+          - dps_val: standby
+            value: idle
+          - dps_val: working
+            value: heating
+      - id: 26
+        type: text
```
make-all: Type should be string, base64 or hex (usually if it is documented as "raw" in Tuya cloud metadata, it is base64, otherwise if it is documented as text or string, then it can be arbitrary readable string data, or hex encoded binary and you need to judge by observing the data, if it is documented as "enum" then there should also be a fixed set of values documented). There are also cases where utf16b64 would be appropriate, but for a "custom_program" I expect it will be binary data encoded in either hex or base64.

make-all: In this context, there won't be any real difference between string, hex and base64, since the latter 2 only get decoded if there is further processing (like mask or format tags) but it might help future improvements if the type is tagged correctly.

make-all (changes requested):
You will need to fix the error yourself, as I do not have write access to make the final changes before merge.

carpabg:
Done

PR #4149, opened 2025-12-16
Add support fo oLight Spheres

met67:
This device file adds support for oLight Sphere (https://www.olight.com/store/sphere-ambient-light-with-app-control). It's a BLE Mesh device that needs a BLE gateway and doesn't support Tuya Cloud method for obtaining the private keys (see this: https://community.home-assistant.io/t/integrating-olight-spheres-in-ha/907929/2).

### DIFF
--- a/custom_components/tuya_local/devices/olight_sphere_light.yaml
+++ b/custom_components/tuya_local/devices/olight_sphere_light.yaml
@@ -1,4 +1,4 @@
-name: oLight Sphere
+name: Ambient light
 products:
   - id: d0dpga0h
     manufacturer: oLight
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/olight_sphere.yaml
```diff
@@ -0,0 +1,168 @@
...
+    category: config
+    hidden: unavailable
+    dps:
+      - id: 7
+        type: integer
+        optional: true
+        name: second
+        range:
+          min: 0
+          max: 43200
```
make-all: Time entity can only be used if the timer is 24h, since the input cannot be limited.

custom_components/tuya_local/devices/olight_sphere.yaml
```diff
@@ -0,0 +1,168 @@
...
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        unit: "%"
+  - entity: switch
+    dps:
+      - id: 104
+        type: boolean
+        name: sleep
```
make-all: name of the dp must be switch. This name belongs at entity level (and there is a translation_key for sleep, so that should be used instead of name).

met67:
I modified timer and sleep as suggested (for sleep I took example from another device, actually it's an optional DP that isn't always returned but is definitely supported)

custom_components/tuya_local/devices/olight_sphere.yaml
met67: Ok, I changed timer to 24h...

custom_components/tuya_local/devices/olight_sphere.yaml
```diff
@@ -178,3 +179,4 @@ entities:
           - dps_val: null
             value: false
           - value: true
+
```
make-all: Did you make the change via GitHub web UI? It seems to add an extra newline every time you open it - if you use the web UI to fix the highlighted issue, you will probably find you need to remove 2 lines, because it will add another one next time you open it.

met67: Ach, sorry! Yes, I use the web UI for simplicity, I'll pay attention to it...

custom_components/tuya_local/devices/olight_sphere.yaml
```diff
@@ -0,0 +1,181 @@
...
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        unit: "%"
+        force: true
+  - entity: switch
+    translation_key: sleep
+    category: config
+    hidden: unavailable
```
make-all: `hidden: unavailable` is for handling multiple variations of a device some which have and some which do not have this entity. If as you say, the device sometimes does not report the state of this dp, but it definitely has it, you should leave out the availability logic, and just let it report "Unknown" when it has not been reported by the device.

custom_components/tuya_local/devices/olight_sphere_light.yaml
```diff
@@ -0,0 +1,181 @@
...
+            value: false
+          - value: true
+  - entity: sensor
+    class: battery
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        unit: "%"
+        force: true
```
make-all, line 156: If you need to force this dp to be reported, then it must be optional also.

PR #4169, opened 2025-12-19
Create advwin_humidifier.yaml

baronorder:
Added new device config for Advwin 13L Humidifier

### DIFF
--- a/custom_components/tuya_local/devices/advwin_humidifier.yaml
+++ b/custom_components/tuya_local/devices/advwin_humidifier.yaml
@@ -15,9 +15,9 @@ entities:
         type: string
         mapping:
           - dps_val: AUTO
-            value: Auto
+            value: auto
           - dps_val: MANUAL
-            value: Normal
+            value: normal
       - id: 13
         name: humidity
         type: integer
@@ -53,32 +53,17 @@ entities:
         name: switch
         type: boolean
         optional: true
-  - entity: select
+  - entity: fan
     name: Spray level
-    icon: mdi:heat-wave
+    icon: mdi:spray
     category: config
     dps:
       - id: 23
-        name: option
+        name: speed
         type: integer
         range:
           min: 0
           max: 6
-        mapping:
-          - dps_val: 0
-            value: Off
-          - dps_val: 1
-            value: "1"
-          - dps_val: 2
-            value: "2"
-          - dps_val: 3
-            value: "3"
-          - dps_val: 4
-            value: "4"
-          - dps_val: 5
-            value: "5"
-          - dps_val: 6
-            value: "6"
   - entity: sensor
     class: temperature
     dps:


custom_components/tuya_local/devices/advwin_humidifier.yaml
```diff
@@ -0,0 +1,135 @@
...
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+      - id: 24
+        name: mode
+        type: string
+        mapping:
+          - dps_val: AUTO
+            value: Automatic
```
make-all: Predefined modes "auto" and "normal" would be better for these mappings, to make use of HA translations.

custom_components/tuya_local/devices/advwin_humidifier.yaml
```diff
@@ -0,0 +1,135 @@
...
+            value: Manual
+      - id: 13
+        name: humidity
+        type: integer
+        optional: true
+        range:
+          min: 30
+          max: 95
+        mapping:
+          - step: 1
```
make-all: step: 1 is unnecessary, remove the mapping.

custom_components/tuya_local/devices/advwin_humidifier.yaml
```diff
@@ -0,0 +1,135 @@
...
+    category: diagnostic
+    dps:
+      - id: 22
+        type: bitfield
+        optional: true
+        name: sensor
+        mapping:
+          - dps_val: 2
+            value: true     # Tank is empty
+          - dps_val: 0
```
make-all, line 125: All these false cases can be combined with a default that also catches any other unexpected values.
```
   - value: false
```

A general `class: problem` binary sensor would also be useful here to signal when there is a fault (other status) and retain the raw fault code.
```
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 22
        type: bitfield
        optional: true
        name: sensor
        mapping:
          - dps_val: 0
            value: false
          - dps_val: null
            value: false
          - dps_val: 2
            value: false
          - value: true
       - id: 22
         type: bitfield
         optional: true
         name: fault_code
```

custom_components/tuya_local/devices/advwin_humidifier.yaml
baronorder: Fixed in new commit

custom_components/tuya_local/devices/advwin_humidifier.yaml
baronorder: Fixed in new commit

custom_components/tuya_local/devices/advwin_humidifier.yaml
baronorder: Fixed in new commit

PR #4172, opened 2025-12-19
Create auchsiag_fireplace_750w_1500w.yaml

aureli169:
Add support for AUCHSIAG Fireplace. Works fine on my Fireplace to launch the heating and to control all the lights of the electrical fireplace.

### DIFF
--- a/custom_components/tuya_local/devices/auchsiag_fireplace.yaml
+++ b/custom_components/tuya_local/devices/auchsiag_fireplace.yaml
@@ -155,4 +155,3 @@ entities:
         type: integer
         name: sensor
         unit: min
-


aureli169:
I have removed the extra keys ('step' and 'readonly') that were causing validation errors. Ready for review!

custom_components/tuya_local/devices/auchsiag_fireplace_750w_1500w.yaml
```diff
@@ -0,0 +1,161 @@
+name: AUCHSIAG Fireplace
+products:
+  - id: "00000442xl"
```
make-all: product ids are generally longer than this for WiFi devices. Are you sure this is the product id, and not the model id or some other identifier that Tuya uses in the cloud?

make-all:
I suggest to look at the other fireplace configs for ideas how to use light entities for the flame and embers, and translation keys for other common entities like the timer and time_remaining

aureli169:
Thank you for the guidance! I have updated the configuration based on your suggestions.
I converted the flame and ember controls into light entities (using the effect list for colors) and implemented the standard translation_keys for the common entities.
I think everything should be in order now, ready for a new review!

aureli169:
I have implemented the translation keys for the common entities (timer, power, time remaining) as requested.

However, regarding the light entities: I tested it, and I must revert to select entities for the Flame and Ember control.

Technical reason: This device shares the same DP (id: 1) for the main power and the "light" power. It is impossible to turn off the flame/ember light without turning off the heater itself. Using a light entity in Home Assistant is misleading here, because clicking the "turn off" button on the light entity shuts down the entire fireplace (including heat), which is bad UX.

Since the colors are fixed presets and cannot be toggled independently of the heating unit, select entities are the most accurate representation of the hardware capabilities.

custom_components/tuya_local/devices/auchsiag_fireplace_750w_1500w.yaml
aureli169: You are absolutely right, I apologize for the mistake! I double-checked my logs and indeed 00000442xl was the Model ID.

I have found the correct Product ID (cokyjkkw8gqwteww) and updated the configuration accordingly

aureli169:
You are absolutely right, I apologize for the mistake! I double-checked my logs and indeed 00000442xl was the Model ID.

I have found the correct Product ID (cokyjkkw8gqwteww) and updated the configuration accordingly

PR #4175, opened 2025-12-20
Add support for Martec Viper DC WiFi Ceiling Fans

benhopperau:
## Summary
Adds support for Martec Viper DC WiFi Ceiling Fans with and without light.

## Devices Added
- **martec_viper_fan.yaml** - Martec Viper DC WiFi Ceiling Fan (fan only model)
- **martec_viper_fan_light.yaml** - Martec Viper DC WiFi Ceiling Fan with Light

## Product Information
- Product ID: `lb9da3os0ihifq7u`
- Manufacturer: Martec
- Protocol: Fan only uses 3.3, Fan with light uses 3.5
- Product pages:
  - With light: https://www.martecaustralia.com.au/product/viper-dc-ceiling-fan-with-light/
  - Fan only: https://www.martecaustralia.com.au/product/viper-dc-ceiling-fan/

### Fan
- Power switch (DP 1)
- Preset modes: nature, sleep, fresh, smart, custom (DP 2)
- Speed control: 6 speeds, auto-scaled to percentage (DP 3)
- Direction control: forward/reverse (DP 8)
- Timer: cancel, 1h-8h (DP 22)

### Light (with light model only)
- Power switch (DP 15)
- Brightness: 1-100% (DP 16)
- Color temperature: continuous 2700K-6000K (DP 17) or fixed presets (DP 19)
- Fixed color temperature presets: Night, Warm, White, Cool

## Testing
Tested with models MVDC134W (fan only) and MVDC1343W (with light). All functions working as expected.

All Viper DC variants should be compatible if they share the same WiFi module and only differ in physical attributes (color, size, number of blades).

### DIFF
--- a/custom_components/tuya_local/devices/martec_viper_fan_light.yaml
+++ b/custom_components/tuya_local/devices/martec_viper_fan_light.yaml
@@ -13,6 +13,7 @@ entities:
       - id: 2
         type: string
         name: preset_mode
+        optional: true
         mapping:
           - dps_val: nature
             value: nature
@@ -27,6 +28,7 @@ entities:
       - id: 3
         name: speed
         type: integer
+        optional: true
         range:
           min: 1
           max: 6
@@ -38,9 +40,11 @@ entities:
       - id: 15
         name: switch
         type: boolean
+        optional: true
       - id: 16
         name: brightness
         type: integer
+        optional: true
         range:
           min: 1
           max: 100
@@ -60,6 +64,7 @@ entities:
       - id: 19
         type: string
         name: fixed_color_temp
+        optional: true
         hidden: true
         mapping:
           - dps_val: night


custom_components/tuya_local/devices/martec_viper_fan.yaml
```diff
@@ -0,0 +1,64 @@
+name: Martec Viper Ceiling Fan
+products:
+  - id: lb9da3os0ihifq7u
```
make-all: This device is already supported as  templeandwebster_alina_fan.yaml

custom_components/tuya_local/devices/martec_viper_fan_light.yaml
```diff
@@ -0,0 +1,118 @@
+name: Martec Viper Ceiling Fan with Light
+products:
+  - id: lb9da3os0ihifq7u
```
make-all, line 3: Is the same product id really reused here?

custom_components/tuya_local/devices/martec_viper_fan_light.yaml
```diff
@@ -0,0 +1,118 @@
+name: Martec Viper Ceiling Fan with Light
```
make-all: Top level name should be generic, as other brands are likely to match (as the above file demonstrates)

custom_components/tuya_local/devices/martec_viper_fan_light.yaml
```diff
@@ -0,0 +1,118 @@
...
+      - id: 3
+        name: speed
+        type: integer
+        range:
+          min: 1
+          max: 6
+      - id: 8
+        name: direction
+        type: string
+        mapping:
```
make-all: This mapping is not needed, as HA has a fixed set of values for direction, and they translate directly to the same strings that Tuya uses.

custom_components/tuya_local/devices/martec_viper_fan_light.yaml
```diff
@@ -0,0 +1,118 @@
...
+          - dps_val: night
+            value: 2700
+          - dps_val: warm
+            value: 4000
+          - dps_val: white
+            value: 5000
+          - dps_val: cold
+            value: 6000
+  - entity: select
+    name: Fixed color temperature
```
make-all: If the above is working, there should be no need for this separate control.

custom_components/tuya_local/devices/martec_viper_fan.yaml
benhopperau: Reason for the PR is that while the seem to share a product ID the temple and webster does not have a light config.

Is it preferred to add the light features to that?

custom_components/tuya_local/devices/martec_viper_fan_light.yaml
benhopperau: Yep, that what has made this a little confusing to build out. The only difference I can spot is the protocol version. Fan only uses 3.3, Fan with light uses 3.5.

benhopperau:
Keen to understand if I should update the templeandwebster_alina_fan.yaml as it shares the same device id or spin this into another version.

make-all:
I think it would be better to merge the configs into the existing temple and webster config, making the light entity `hidden: unavailable`; with an `available` dp that maps to true when preset (non-null).

make-all:
I have merged this into the templeandwebster_alina_fan config

PR #4188, opened 2025-12-22
Add support for Junsun EV Charger

bartowl:
This EV Charger supports charging current adjustment mid-charging without disconnecting.
Tested with single phase 230V 32A Type2 Junsun charger from Aliexpress.

### DIFF
--- a/custom_components/tuya_local/devices/junsun_ev_charger.yaml
+++ b/custom_components/tuya_local/devices/junsun_ev_charger.yaml
@@ -39,7 +39,6 @@ entities:
         name: mode_set
   - entity: sensor
     class: energy
-    category: diagnostic
     dps:
       - id: 1
         type: integer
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/junsun_ev_charger.yaml
```diff
@@ -0,0 +1,234 @@
...
+    category: config
+    icon: "mdi:ev-plug-type2"
+    dps:
+      - id: 4
+        type: integer
+        name: value
+        unit: A
+        range:
+          min: 6
+          max: 20
```
make-all: If this is a 32A charger, why is it limited here to 20A?

bartowl: Sorry, it really supports 32A I have just limited it locally to 20A since my wiring does not allow for more. Should be fixed now

PR #4189, opened 2025-12-22
Expose more temperature configurations to HA

flow-c:


custom_components/tuya_local/devices/miheat_th11wf_thermostat.yaml
```diff
@@ -200,3 +251,37 @@ entities:
...
+    category: config
+    mode: box
+    dps:
+      - id: 166
+        type: integer
+        name: value
+        mapping:
+          - scale: 10
+            step: 5
+            range:
```
make-all, line 263: range and unit should be outside of the mapping. default is for a different purpose (reverse mapping for actions where an optional parameter is not provided), but you can map `dps_val: null` to a value if you need to.

PR #4193, opened 2025-12-23
Create kogan_KASMCDSKTLA_kettle.yaml

baronorder:
Device config support for Kogan KASMCDSKTLA smart kettle

### DIFF
--- a/custom_components/tuya_local/devices/kogan_kasmcdsktla_kettle.yaml
+++ b/custom_components/tuya_local/devices/kogan_kasmcdsktla_kettle.yaml
@@ -27,10 +27,10 @@ entities:
           min: 40
           max: 100
 
- - entity: sensor
-   translation_key: status
-   class: enum
-   dps:
+  - entity: sensor
+    translation_key: status
+    class: enum
+    dps:
       - id: 8
         name: sensor
         type: string


custom_components/tuya_local/devices/kogan_KASMCDSKTLA_kettle.yaml
```diff
@@ -0,0 +1,65 @@
+name: Kogan 1.7L Double Wall Glass Smart Kettle
```
make-all: Top level name should be generic, as the same basic device is likely sold in different markets with different branding. Branding should remain under products where it is linked to a specific id.

custom_components/tuya_local/devices/kogan_KASMCDSKTLA_kettle.yaml
```diff
@@ -0,0 +1,65 @@
...
+products:
+  - id: 7occ66lr74vfznwt
+    manufacturer: Kogan
+    model: KASMCDSKTLA
+entities:
+  - entity: switch
```
make-all: I think following the water_heater pattern of other kettles will give a better UI for the temperature setting.

custom_components/tuya_local/devices/kogan_KASMCDSKTLA_kettle.yaml
```diff
@@ -0,0 +1,65 @@
...
+    dps:
+      - id: 4
+        name: value
+        type: integer
+        unit: °C
+        range:
+          min: 40
+          max: 100
+  - entity: select
+    name: Status
```
make-all: use `translation_key: status`, and the values under option should be all lowercase to match the translation keys under that.

custom_components/tuya_local/devices/kogan_KASMCDSKTLA_kettle.yaml
```diff
@@ -0,0 +1,65 @@
...
+        type: string
+        mapping:
+          - dps_val: standby
+            value: Standby
+          - dps_val: heating
+            value: Heating
+          - dps_val: cooling
+            value: Cooling
+  - entity: select
+    name: Preset
```
make-all: Use `translation_key: kettle_mode` instead of this name, and the options: `water`, `custom`, `coffee`, `green_tea`, `honey_water` and `infant_formula`

baronorder: fixed in new commit

custom_components/tuya_local/devices/kogan_KASMCDSKTLA_kettle.yaml
make-all: Needs quotes to avoid yaml parser converting it to boolean

baronorder: Ahhh that would be why, testing it now

custom_components/tuya_local/devices/kogan_KASMCDSKTLA_kettle.yaml
baronorder: Fixed in newcommit

custom_components/tuya_local/devices/kogan_KASMCDSKTLA_kettle.yaml
baronorder: Converted to water heater entity

custom_components/tuya_local/devices/kogan_KASMCDSKTLA_kettle.yaml
baronorder: Fixed with generic name

custom_components/tuya_local/devices/kogan_KASMCDSKTLA_kettle.yaml
baronorder: Thanks for that, I was trying to figure out why it was failing!

PR #4195, opened 2025-12-23
EVONP0 Pool Controller custom device configuration

AntoineWaes9359:


AntoineWaes9359:
EVONP0 Pool Controller device class, Pool Inverter device Aqualyzer

custom_components/tuya_local/devices/solar_inverter_SPD800.yaml
```diff
@@ -0,0 +1,117 @@
+name: Micro Inverter Stepuptech SPD-800
```
make-all: This device already exists in the config file list, there is no need for this new file

make-all:
Due to high volume of PRs I do not have time to deal with PRs that need major overhaul to update them to modern config format.

PR #4198, opened 2025-12-24
Add device support for Stadler Form Lukas Dehumidifier

RigoWeb:
Added support for Stadler Form Lukas Dehumidifier, I've tested all entities except untestable errors like 'Refrigerant Leak'...

### DIFF
--- a/custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
@@ -1,4 +1,4 @@
-name: Stadler Form Lukas
+name: Dehumidifier
 products:
   - id: 6ny5ejcir57trwao
     manufacturer: Stadler Form
@@ -36,29 +36,26 @@ entities:
         name: current_humidity
 
   # Fan speed (1–4)
-  - entity: select
-    name: "Fan Speed"
+  - entity: fan
     dps:
-      - id: 4
-        type: string
-        name: option
-        mapping:
-          - dps_val: "1"
-            value: "1"
-          - dps_val: "2"
-            value: "2"
-          - dps_val: "3"
-            value: "3"
-          - dps_val: "4"
-            value: "4"
-
-  # Swing
-  - entity: switch
-    name: Swing
-    dps:
-      - id: 8
+      - id: 1
         type: boolean
         name: switch
+      - id: 4
+        type: string
+        name: speed
+        mapping:
+          - dps_val: "1"
+            value: 25"
+          - dps_val: "2"
+            value: 50
+          - dps_val: "3"
+            value: 75
+          - dps_val: "4"
+            value: 100
+      - id: 8
+        type: boolean
+        name: oscillate
 
   # Night mode
   - entity: switch
@@ -70,7 +67,7 @@ entities:
 
   # Child lock
   - entity: lock
-    name: "Child Lock"
+    translation_key: child_lock
     dps:
       - id: 16
         type: boolean


custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
```diff
@@ -0,0 +1,114 @@
+name: Stadler Form Lukas
+products:
+  - id: bf4b1d3329427223dczfve
```
make-all: This appears to be a device id, not a product id

custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - id: 1
+        type: boolean
+        name: switch
+
+      - id: 5
+        name: mode
+        type: string
+        mapping:
+          - dps_val: "DRY"
+            value: DRY
```
make-all: Modes should be mapped where possible to existing translations, using all lower case, or where no existing mode fits, capitalise the first letter of the first word only.

Predefined modes from [HA](https://github.com/home-assistant/core/blob/dev/homeassistant/components/humidifier/strings.json#L65): `auto`, `away`, `baby`, `boost`, `comfort`, `eco`, `home`, `normal`, `sleep`.

Additional modes defined by this integration for dehumidifiers: `laundry`, `purify`

Probably `normal` and `laundry` fit here.

custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
```diff
@@ -0,0 +1,114 @@
...
+          min: 25
+          max: 80
+        mapping:
+          - step: 5
+
+      - id: 6
+        type: integer
+        name: current_humidity
+
+  # Fan speed (1–4)
```
make-all, line 38: Would probably be better as a fan combined with swing (oscillate).

custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
```diff
@@ -0,0 +1,114 @@
...
+  - entity: switch
+    name: Swing
+    dps:
+      - id: 8
+        type: boolean
+        name: switch
+
+  # Sleep-Mode
+  - entity: switch
+    name: "Night Mode"
```
make-all: use `translation_key: sleep` instead of this name

custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - id: 19
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          # No error
+          - dps_val: null
+            value: false
+          # Default: some bit is set → error
```
make-all: I suggest to also add mappings to false for tank full and defrost mode, since those are expected during normal operation and not a fault with the device. Give them their own binary_sensor, there are already translation keys for `tank_full` and `defrost` to use for these.

RigoWeb: Do you mean like this?

```
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 19
        type: bitfield
        name: sensor
        mapping:
          - dps_val: 0
            value: false
          - dps_val: 2
            value: false
          - value: true
      - id: 19
        type: bitfield
        name: fault_code
  - entity: binary_sensor
    translation_key: tank_full
    dps:
      - id: 19
        type: bitfield
        name: sensor
        mapping:
          - dps_val: 2
            value: true
          - value: false
  - entity: binary_sensor
    translation_key: defrost
    dps:
      - id: 19
        type: bitfield
        name: sensor
        mapping:
          - dps_val: 4
            value: true
          - value: false
```

make-all: Yesm like that, but probably 4 also does not need to be flagged as a problem

custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
RigoWeb: fixed

custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
RigoWeb: fixed

custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
RigoWeb: fixed

custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
RigoWeb: I tried this first, but since the fan doesn't have an on/off switch, I didn't like operating it with the fan entity. In the manufacturer's app, this is also solved with a separate entity for the speed and the swing switch. I hope this is okay...

custom_components/tuya_local/devices/stadlerform_lukas_dehumidifier.yaml
RigoWeb: Implemented as you suggested, many thanks, also for the great integration :-)

PR #4200, opened 2025-12-26
Arçelik AHPH-MM.. series heat pump yaml file addition

Korkuttum:
This file is compatible with 

AHPH-MM08
AHPH-MM10
AHPH-MM12
AHPH-MM14
AHPH-MM16 

model heat pumps.

### DIFF
--- a/custom_components/tuya_local/devices/arcelik_ahphmm_comboheatpump.yaml
+++ b/custom_components/tuya_local/devices/arcelik_ahphmm_comboheatpump.yaml
@@ -1,4 +1,4 @@
-name: Arçelik heat pump
+name: Combo heat pump
 products:
   - id: l01fseshej0xmz8x
     manufacturer: Arçelik
@@ -101,9 +101,16 @@ entities:
           - dps_val: auto_dhw
             value: heat_cool
             available: water_on
-      - id: 117
+      - id: 115
         type: boolean
         name: hvac_action
+        mapping:
+          - dps_val: true
+            value: defrosting
+          - value_redirect: actual_hvac_action
+      - id: 117
+        type: boolean
+        name: actual_hvac_action
         mapping:
           - dps_val: false
             value: idle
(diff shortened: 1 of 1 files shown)


make-all:
Rather than low level number, sensor, switch and select entities, this should make use of a climate entity for the air heating/cooling functions, and a water_heater entity for the water heating. See other combo device configs for an example.

Korkuttum:
Actually, I did it that way because it is used like this in the Tuya app, but I also added the thermostats. Please take a look at the Tuya app.
<p align="center">
  
  
  
</p>

make-all:
What it looks like in the Tuya app is irrelevant.

This integration is for integrating devices to Home Assistant, not duplicating the Tuya app.

Korkuttum:
I understand you. I made the additions to the code as you requested, but users who use this integration may already have established usage habits. Also, the application already includes built-in thermostats that automatically turn on and off when the temperature reaches a certain level, so we are essentially adding the same functionality again on the Home Assistant side.

Korkuttum:
Right now, it has become far from useful for me. There is no holiday mode switch, and there is no sensor that I can use to monitor defrosting.
Thermostats are not very useful for these tasks, and the water heater’s HVAC mode is not working.

make-all:
Defrost sensor is merged with the climate hvac_action, and holiday mode was already available in the water_heater entity.

Can you explain more about the water heater's operation mode?
Does it not work when the air is on heating, cooling or auto, or only when the air is off?

Korkuttum:
[Query Properties.json](https://github.com/user-attachments/files/24369208/Query.Properties.json)
[Query Things Data Model.json](https://github.com/user-attachments/files/24369209/Query.Things.Data.Model.json)

It only works in the modes listed below:

          - dps_val: cool
            value: cool
          - dps_val: heat
            value: heat
          - dps_val: auto
            value: auto
          - dps_val: hot_water
            value: hotwater
          - dps_val: cool_hot_water
            value: hotwater_cool
          - dps_val: heat_hot_water
            value: hotwater_heat
          - dps_val: auto_dhw
            value: hotwater_auto

Korkuttum:
Wouldn’t it make more sense to track defrost as a sensor?
For me, it would be much more logical to follow the device app’s structure and treat all settings as configuration.
This didn’t turn out to be very user-friendly.

In the end, this is a device where you can only set the temperature and change the mode from the outside; it already has its own thermostat and water heater thermostat inside. This is very confusing.
If the device we were using were actually a thermostat device, then of course it should have a climate entity. But for this specific device, there’s no need to go on such an adventure.

Korkuttum:
Defrost has no connection with the HVAC mode. It’s a system that operates to melt ice when the device gets too cold and starts to frost over.
Vacation mode is also a mode in which the device does not turn on, even if you try to power it on or if a physical thermostat is connected to the device.

Korkuttum:
It’s not suitable for use at all, just so you know. For it to be user-friendly, the best approach would be to follow the app. I won’t be using it this way, just so you’re aware, because it’s very impractical.

make-all:
This type of comment is not constructive. It is impossible to know what is not working when you make such vague complaints as "it is not suitable for use at all".

Korkuttum:
> Right now, it has become far from useful for me. There is no holiday mode switch, and there is no sensor that I can use to monitor defrosting.
Thermostats are not very useful for these tasks, and the water heater’s HVAC mode is not working.

I really want to help you — I even want to prepare a YAML file for a different espresso machine I have — but you’re not taking my requests into account.

From a software perspective, what you’re doing may be correct, but it’s not usable. In fact, I think the company that sells this device and built its app has implemented it in the most accurate and device-appropriate way.

In my opinion, it would be better to expose the entities in the same structure as the device’s app. Some sensors that don’t even exist in the app could actually be even more useful for users.

There is no thermostat in the app, yet we are creating two thermostats. If a user wants a thermostat, they can already create one using a template — that feels more like a matter of personal preference.

From the perspective of someone who values and actively supports open-source projects, your approach and firm position on this topic feel a bit difficult to align with.  

Open-source development is often at its best when different viewpoints and user feedback are carefully considered, especially when it comes to usability. That’s why I believe revisiting this approach and aligning it more closely with how the device is actually designed and used could be beneficial for everyone.

PR #4206, opened 2025-12-27
Create inkbird_ivc001w_fan.yaml

baronorder:
Device is unusual in that it does not follow the normal design of either a thermostat or HVAC device and is much more similar to a fan with a bunch of extra settings. 

Also does a very non standard base64 encoded string to show current sensor readings that needs a set of template sensors to expose sadly - would be amazing to have the ability to do "virtual" DP's and base64 decoding/encoding natively. 

I left the comments in for the bits that really need explaining as the device is weird - most people using this in HA will purely be looking to easily control the fan and speed rather than using all of the capabilities of the device.

### DIFF
--- a/custom_components/tuya_local/devices/inkbird_ivc001w_fan.yaml
+++ b/custom_components/tuya_local/devices/inkbird_ivc001w_fan.yaml
@@ -59,7 +59,6 @@ entities:
         optional: true
         mask: "00FFFF00000000000000000000"
         mask_signed: true
-        endianness: big
         name: current_temperature
         mapping:
           - scale: 10
@@ -67,7 +66,6 @@ entities:
         type: base64
         optional: true
         mask: "0000000000FFFF000000000000"
-        endianness: big
         name: current_humidity
         mapping:
           - scale: 10
@@ -128,7 +126,6 @@ entities:
         type: base64
         optional: true
         mask: "000000000000000000FFFF0000"
-        endianness: big
         name: value
         unit: C
         range:
@@ -146,7 +143,6 @@ entities:
         type: base64
         optional: true
         mask: "0000000000000000000000FFFF"
-        endianness: big
         name: value
         unit: C
         range:
@@ -163,6 +159,5 @@ entities:
         type: base64
         optional: true
         mask: "00000000000000FFFF00000000"
-        endianness: big
         name: sensor
         unit: rpm


custom_components/tuya_local/devices/inkbird_ivc001w_fan.yaml
```diff
@@ -9,14 +9,14 @@ entities:
...
     dps:
       - id: 1
-        #Yes it is backwards and yes its just as dumb as you are thinking it is
+        # Yes it is backwards and yes its just as dumb as you are thinking it is
         name: switch
         type: boolean
         mapping:
           - dps_val: true
-            value: off
+            value: "off"
```
make-all: This is incorrect, a switch should have boolean values

custom_components/tuya_local/devices/inkbird_ivc001w_fan.yaml
```diff
@@ -0,0 +1,110 @@
...
+          - dps_val: timing
+            value: smart
+      - id: 103
+        name: speed
+        type: integer
+        range:
+          min: 1
+          max: 10
+      - id: 132
+        name: raw_status
```
make-all, line 36: These can be decoded into their respective attributes and sensor or number entities using  `mask`

custom_components/tuya_local/devices/inkbird_ivc001w_fan.yaml
```diff
@@ -0,0 +1,110 @@
...
+    dps:
+      - id: 23
+        name: option
+        type: string
+        mapping:
+          - dps_val: "c"
+            value: "c"
+          - dps_val: "f"
+            value: "f"
+  - entity: number
```
make-all: This would be better as a light entity, with `translation_key: display` instead of name and icon.

custom_components/tuya_local/devices/inkbird_ivc001w_fan.yaml
```diff
@@ -0,0 +1,110 @@
...
+    dps:
+      - id: 44
+        name: value
+        type: integer
+        unit: "%"
+        range:
+          min: 0
+          max: 100
+  - entity: number
+    name: Temperature Calibration
```
make-all: Use `translation_key: temperature_calibration` instead of this name and icon

make-all:
I don't know why you conclude that this does not follow the normal design of a thermostat, as the basic features of a switch, a temperature setting and temperature reading are all there.

baronorder:
Looking to rework at the moment - getting the mask extract to work has been a bit of a nightmare but I am almost there.
Having some issues with the returned values being x10 what they actually are as "scale" seems to be ignored on values returned from the base64 / mask extract? Would love some guidance on a neat way to work around that issue?

make-all:
Scale works on masked values. It does need to be inside a mapping

baronorder:
Working through it, having a hard time getting the temperature/humidity high and low points working as they are both packed values for both high and low. I just can't get the config to work right and would appreciate any guidance you have.

the values appear to function like this using the humidity one as an example:
DP 102: 32-bit Integer Structure (Big Endian)
Total Value Example: 39,322,000 (0x02580190)

      High Setpoint (60.0%)           Low Setpoint (40.0%)
┌───────────────────────────────┐┌───────────────────────────────┐
│         16 Bits (MSB)         ││         16 Bits (LSB)         │
├─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┤├─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┤
│1│1│1│1│1│1│1│1│1│1│1│1│1│1│1│1││1│1│1│1│1│1│1│1│1│1│1│1│1│1│1│1│
└─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┘└─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┘
        Hex: 0x0258                      Hex: 0x0190
      Decimal: 600                     Decimal: 400
     (Scaled / 10 = 60.0)             (Scaled / 10 = 40.0)

     Mask: FFFF0000                   Mask: 0000FFFF
     
Is there an example you have on how values like this can be dealt with? 
I have the previous base64 decode to get the sensor values out working fine now but at the very least writing to these is causing problems, the closest I have gotten to functional (after changing types to deal with errors when trying to set) is:
     
     ```
     - entity: number
    name: Humidity High Setpoint
    category: config
    icon: "mdi:water-percent"
    dps:
      - id: 102
        name: value
        type: integer
        mask: "FFFF0000"
        endianness: big
        mapping:
          - scale: 10
            step: 0.1
        unit: "%"

  - entity: number
    name: Humidity Low Setpoint
    category: config
    icon: "mdi:water"
    dps:
      - id: 102
        name: value
        type: integer
        mask: "0000FFFF"
        endianness: big
        mapping:
          - scale: 10
            step: 0.1
        unit: "%"
          ```

make-all:
I don't have any example of the raw data received in dp 102, but generally masked values will come through as either hex or base64 from the device. ~~In theory integer should be supported if it really is coming through as an integer, but is untested~~. Edit: after checking the code, the mask requires a byte array to operate on, which integer does not provide, so a code modification would be required to handle masking integer values.

baronorder:
Just to make sure I am not going insane - this is the returned feedback from the tuya API explorer for that humidity value:
```
 "code": "humd_set",
        "custom_name": "",
        "dp_id": 102,
        "time": 1766984494037,
        "type": "value",
        "value": 39322000
```

make-all:
The dp 102 decoding can be added in a subsequent PR

Issue #233, opened 2022-09-13
Unable to add C16 thermostat


PsychoMnts:
I'm trying to add the C16 thermostat, but it reports "Sorry, there is no support for this device".
Log:

Device matches None with quality of 0%. DPS: {'2': 170, '3': 'Smart', '4': 0, '5': 245, '6': 350, '7': False, '8': 200, '9': False, '10': False, '11': False, '12': '5_1_1', '14': '0', '15': 0, '17': 0, '21': False, '22': 0, '23': 4, '24': 'Standby', 'updated_at': 1663097383.9795818}

PsychoMnts:
I removed entity id number 26 and I was able to add the device.

make-all:
OK, I will mark that one as optional.

make-all:
DP 26 is now optional in 0.19.0, so your device should match now.

Issue #212, opened 2022-07-25
Remove Grid connect/arlec GPO/switch “master” dependency


JanekSMC:
Hi, you guys are absolutely changing my world! It’s fantastic having local support for easily accessible products in AU (yay Bunnings!!!)

Is there a chance we could remove the master switch dependency on the grid connect series of GPOs and switches?  I change it in the config, but every update it changes back (obviously). 

It seems to me the master switch is only of use to switch off all outlets. You cannot have master on and outlets off, and as such when both/all outlets are off, master defaults to off, and inhibits the individual switch entities. 

This requires me to switch in the master, then select the switch I need on. This is different to the behaviour of the Tuya app which I can select any switch, and master can turn them all off. 

I hope this makes sense.

make-all:
The override behaviour of the master switch was removed in the previous release.  It should now be possible to turn on individual switches when the master is off (whether the master automatically turns on will be down to device behaviour, there may be some delay if it does),

JanekSMC:
My goodness! You are amazing. I never ever expected it to be done so quickly!!! It’ll be nice not to manually edit the master lockout every update!

I’m sos sorry I haven’t checked in before now, I keep getting caught up with my toddlers and all my “fun” stuff gets put on the back burner for ages. 

Honestly, Thankyou. I wish you knew just how appreciative I am.

PR #4216, opened 2025-12-28
add Faroro TD20 Pet Treat Camera support

augs:
## Adds support for the Faroro TD20 Pet Treat Camera.

This device was previsouly sold on Amazon at https://www.amazon.com/dp/B0C6JMWTYN

Local DPS discovery of my reference device returns:
```
[
 {"id": 103, "name": "basic_flip", "type": "Boolean", "format": "{}", "enumMap": {}},
 {"id": 108, "name": "basic_nightvision", "type": "Enum", "format": "{\"range\":[\"0\",\"1\",\"2\"]}", "enumMap": {}},
 {"id": 134, "name": "motion_switch", "type": "Boolean", "format": "{}", "enumMap": {}},
 {"id": 168, "name": "motion_area_switch", "type": "Boolean", "format": "{}", "enumMap": {}},
 {"id": 169, "name": "motion_area", "type": "String", "format": "{\"maxlen\":255}", "enumMap": {}}
]
```

### Functionality not implemented

Camera audio and video is not implemented as it is not supported by tuya-local

Camera motion scheduling is also not exposed via any DPS that I can find so it is not implemented.

DPS #169 `motion_area` is only exposed as a text field.  The device expects a JSON payload defining an overlay box.  The values are most likely percentages of camera area.  Below is an example for the entire area:
`{"num":1,"region0":{"x":0,"y":0,"xlen":100,"ylen":100}}`

DPS #212 logic is not implemented as it would require some custom parsing of the payload JSON.  It triggers when motion is detected and an image is uploaded to a cloud bucket.  Below is an example:
```
{"v":"4.0","bucket":"ty-us-storage$$","cmd":"ipc_motion","type":"image","with":"resource","alarm":true,"time":1766941681,"files":[["/######-##########-#####################/unify/###########.jpeg","#################"]]}
```

DPS #233 `主动消息推送` or `proactive push notifications` does not appear to accept any value changes.

### DIFF
--- a/custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
+++ b/custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
@@ -1,8 +1,9 @@
-name: Pet Treat Camera
+name: Camera pet feeder
 products:
   - id: m6xa97cporn9eqzy
     manufacturer: Faroro
-    model: TD20
+    model: Pet Treat Camera
+    model_id: TD20
 entities:
   - entity: button
     name: Feed
@@ -20,7 +21,7 @@ entities:
         type: boolean
         name: switch
   - entity: select
-    name: Night vision mode
+    name: Night vision
     translation_key: light_mode
     category: config
     dps:
@@ -36,20 +37,23 @@ entities:
             value: "off"
   - entity: switch
     name: Motion sensing
+    icon: "mdi:motion-sensor"
     category: config
     dps:
       - id: 134
         type: boolean
         name: switch
   - entity: switch
-    name: Use motion area map
+    name: Limit motion area
+    icon: "mdi:picture-in-picture-bottom-right"
     category: config
     dps:
       - id: 168
         type: boolean
         name: switch
   - entity: text
-    name: Motion area map
+    name: Motion area
+    icon: "mdi:picture-in-picture-bottom-right"
     category: config
     hidden: true
     dps:
@@ -57,7 +61,6 @@ entities:
         type: string
         name: value
   - entity: sensor
-    name: Feed status
     translation_key: status
     icon: "mdi:paw"
     category: diagnostic


custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
```diff
@@ -0,0 +1,72 @@
+name: Faroro Pet Treat Camera
```
make-all: Top level name should be generic, as other brands are likely to be used for the same device in different markets.

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - entity: button
+    name: Feed
+    icon: "mdi:paw"
+    dps:
+      - id: 231
+        type: boolean
+        name: button
+        optional: true
+  - entity: switch
+    name: Flip Video
```
make-all: Use 'translation_key: flip_image'

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
```diff
@@ -0,0 +1,72 @@
...
+    category: config
+    dps:
+      - id: 103
+        type: boolean
+        name: switch
+  - entity: select
+    name: Night Vision
+    category: config
```
make-all, line 26: Add `translation_key: mode`, and use all lower-case for the values in the mapping below to match existing translations.

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
```diff
@@ -0,0 +1,72 @@
...
+        optional: true
+  - entity: switch
+    name: Flip Video
+    category: config
+    dps:
+      - id: 103
+        type: boolean
+        name: switch
+  - entity: select
+    name: Night Vision
```
make-all: Note HA naming convention is to only capitalise the first word in device and entity names.

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - entity: text
+    name: Motion Area Map
+    category: config
+    hidden: true
+    dps:
+      - id: 169
+        type: string
+        name: value
+  - entity: sensor
+    name: Feed status
```
make-all: Use `translation_key: status`, and match values in the mapping to existing translations: `fault` or `no_food`, and `feeding_complete`

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
```diff
@@ -0,0 +1,72 @@
...
+    icon: "mdi:paw"
+    category: diagnostic
+    class: enum
+    dps:
+      - id: 236
+        name: sensor
+        type: integer
```
make-all, line 71: Is this documented as a bitmap or enum type?
If bitmap, the type should be bitfield rather than integer (which does bitwise comparisons instead of exact matches). If enum, then it should be a string, with double-quotes around the dps_val below.

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
augs: done, should I also rename the file?

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
augs: done

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
augs: ack, done globally

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
augs: `translation_key: mode` lacks an `On` value so I elected to use `translation_key: light_mode` which has the required values and more accurately maps to what this setting actually controls: the IR emitter.

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
augs: done

custom_components/tuya_local/devices/faroro_pettreat_camera.yaml
augs: Apologies for being new to Tuya in general.  I found 3 different documentation sources for `feed_report`, they all return a number which should indicate the quantity of portions dispensed:
- https://developer.tuya.com/en/docs/iot-device-dev/pet_feeder_solution?id=Kay7r9zu4rupm#title-5-Manual%20feeding indicates type `Value` with valid values of 0-20
- https://developer.tuya.com/en/docs/iot/pet-feeder-product-function-definition?id=K9tlinn9lgmrq#title-0-Standard%20functions indicates type `Value` with valid values of 0-12
- https://developer.tuya.com/en/docs/iot/s?id=K9gf48qvg0myl#title-1-Support%20reporting%20instructions%20to%20the%20cloud%20only indicates type `Integer` with valid values of 0-12

For this particular device, it only permits dispensing 1 portion in the manual_feed DPS:
```
>>> d = tinytuya.Device(..., ..., ..., version=3.4)
>>> d.status()
{'dps': {'103': False, '108': '1', '134': False, '168': False, '169': '{"num":1,"region0":{"x":24,"y":22,"xlen":76,"ylen":78}}', '233': '0', '236': 1}}
>>> d.set_value(231, 0)
>>> d.receive()
>>> d.set_value(231, 1)
{'protocol': 4, 't': 1767448348, 'data': {'dps': {'236': 1}}, 'dps': {'236': 1}}
>>> d.set_value(231, 2)
>>> d.receive()
```

I have also been unable to create a "fault" condition, even by disconncting both motors of the 2 stage feeding process.  The strings I supplied here came from reversing the vendor specific (not smartlife) android app strings.

PR #4220, opened 2025-12-29
Update device_config.py - add support for bitwise masking on integer DPs

baronorder:
Its been a while since I was this deep in the weeds with Python but here we go...
Feel free to edit however you prefer!

This PR addresses an issue where using the mask parameter on data points of type: integer would result in correctable crashes and incorrect data parsing. This is required due to at least one device (Inkbird IVC-001W) that presents DPs in this fashion but I am sure that there are others out there.

Previously, the masking logic in device_config.py assumed that any masked value was a byte array (hex/base64). This caused two issues for packed integers (e.g., two 16-bit values packed into a 32-bit integer):

On read: The mask was ignored, returning the full raw integer instead of the specific bit range.

On write: The integration crashed with cannot convert 'float' object to bytes because it attempted to pass a numeric value into int.from_bytes.

Changes:

Updated get_value to check if the raw value is an integer. If so, it applies the mask and bit-shift directly without byte conversion.

Updated get_values_to_set to implement a "Read-Modify-Write" cycle for integers. It reads the current device state, clears the masked bits, and merges the new value, ensuring the unmasked parts of the integer remain untouched.

Backwards compatibility notes: The new logic only triggers if mask is defined and the raw value is an integer; existing hex/base64 implementations are unaffected.

### DIFF
--- a/custom_components/tuya_local/helpers/device_config.py
+++ b/custom_components/tuya_local/helpers/device_config.py
@@ -478,11 +478,7 @@ class TuyaDpsConfig:
         mask = self.mask
         # Get raw value directly avoiding accidental scaling by decoded_value()
         raw_from_device = device.get_property(self.id)
-        # Decode only if strictly required (hex/base64) otherwise assume it's the raw int
-        if self.rawtype in ["hex", "base64", "utf16b64"]:
-            bytevalue = self.decode_value(raw_from_device, device)
-        else:
-            bytevalue = raw_from_device
+        bytevalue = self.decode_value(raw_from_device, device)
 
         if mask and isinstance(bytevalue, bytes):
             value = int.from_bytes(bytevalue, self.endianness)
@@ -1092,10 +1088,7 @@ class TuyaDpsConfig:
             if self.id in pending_map:
                 decoded_value = self.decode_value(pending_map[self.id], device)
             else:
-                if self.rawtype in ["hex", "base64", "utf16b64"]:
-                    decoded_value = self.decode_value(raw_current, device)
-                else:
-                    decoded_value = raw_current
+                decoded_value = self.decode_value(raw_current, device)
 
             if isinstance(decoded_value, int):
                 current_value = decoded_value


custom_components/tuya_local/helpers/device_config.py
```diff
@@ -476,7 +476,14 @@ def endianness(self):
     def get_value(self, device):
         """Return the value of the dps from the given device."""
         mask = self.mask
-        bytevalue = self.decoded_value(device)
+        # Get raw value directly avoiding accidental scaling by decoded_value()
+        raw_from_device = device.get_property(self.id)
+        # Decode only if strictly required (hex/base64) otherwise assume it's the raw int
+        if self.rawtype in ["hex", "base64", "utf16b64"]:
```
make-all: I don't think utf16b64 should be included here (it is not handled in decode_value and isn't appropriate for masking).

In fact, since all unhandled types are returned as-is from decode_value, you can get rid of the if...else... completely,, and just use the line below unconditionally.

custom_components/tuya_local/helpers/device_config.py
```diff
@@ -476,7 +476,14 @@ def endianness(self):
     def get_value(self, device):
         """Return the value of the dps from the given device."""
         mask = self.mask
-        bytevalue = self.decoded_value(device)
+        # Get raw value directly avoiding accidental scaling by decoded_value()
```
make-all, line 479: I have some reservation about not using decoded_value here, as there may be something relying on a mapping being done first. But after reviewing a sample of uses, I found several that are more likely to be fixed by this change and none that would be broken, so I think I will take this as it is and deal with any fallout if it comes.

custom_components/tuya_local/helpers/device_config.py
```diff
@@ -1072,14 +1086,28 @@ def get_values_to_set(self, device, value, pending_map={}):
...
             if self.id in pending_map:
                 decoded_value = self.decode_value(pending_map[self.id], device)
-            current_value = int.from_bytes(decoded_value, endianness)
-            result = (current_value & ~mask) | (mask & int(result * mask_scale))
-            result = self.encode_value(result.to_bytes(length, endianness))
+            else:
+                if self.rawtype in ["hex", "base64", "utf16b64"]:
+                    decoded_value = self.decode_value(raw_current, device)
+                else:
+                    decoded_value = raw_current
```
make-all: Again, the if...else can be eliminated here, since decode_value returns its argument unchanged if it is not a byte type that needs decoding

PR #4221, opened 2025-12-30
Add Ventilation RPM sensor to heatpump configuration

jarivanb:
Nothing special, just added a new sensor: added Ventilation RPM sensor with diagnostic category.

### DIFF
--- a/custom_components/tuya_local/devices/steinbach_solid_4.3_heatpump.yaml
+++ b/custom_components/tuya_local/devices/steinbach_solid_4.3_heatpump.yaml
@@ -73,7 +73,7 @@ entities:
         type: bitfield
         name: fault_code
   - entity: sensor
-    name: Fan
+    name: Fan speed
     category: diagnostic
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/steinbach_solid_4.3_heatpump.yaml
```diff
@@ -72,3 +72,14 @@ entities:
       - id: 21
         type: bitfield
         name: fault_code
+  - entity: sensor
+    name: Ventilation RPM
+    class: speed
```
make-all: rpm is not valid units for speed. This class needs to be removed.

make-all:
I don't think "Ventilation" is the correct word here. It implies a fan that exchanges indoor air for outdoor air. Maybe just "Fan", as I assume this refers to the fan that forces air past the fins on the compressor unit.

jarivanb:
Thanks for the review! I've updated the pull request.

Issue #4167, opened 2025-12-19
Request support for MINI-F-Public


Permanently:
### Log message

```text
2025-12-18 22:38:48.256 WARNING (MainThread) [custom_components.tuya_local.config_flow] Adding MINI-F-Public device with product id y3xefowizhyexugf
[...]

PR #4229, opened 2025-12-31
Add configuration for MINI-F-Public

Permanently:
Addressing #4167; making a start on configuration for Mini-F as the existing for Mini-B is extremely similar: https://github.com/make-all/tuya-local/issues/4167#issuecomment-3672897184


make-all:
This config is identical to the Mini-B config except for the products details. There is no reason to add a new config, the product details can be added to the existing config.

Issue #4239, opened 2026-01-02
Request support for Elko/Schneider Wifi Thermostat


hally0:
### Log message

```text
Device matches zyhps01_human_presence_sensor with quality of 26%. DPS: {"updated_at": 1767391592.5205498, "101": "02.01.01.68", "102": "02.02.15.6D", "103": 1000, "104": 1750, "105": 2200, "107": 2200, "108": 0, "109": 2200, "110": 2200, "111": 200, "112": 200, "113": 1800, "114": 2400, "115": false, "116": false, "117": false, "118": 8, "121": "Heat", "122": "Unoccupied", "123": "Close", "128": 0, "129": "10K", "130": -300, "131": false, "132": 100, "133": 0, "134": 60, "139": "NotConnected", "140": "Manual", "145": "EKO07262;B;16A", "146": false, "150": 0, "151": 773999, "152": true}
```

### Product ID

fy012zc3l51doakx

### Product Name

CFMT-B

### DPS information

```text
{
  "result": {
    "model": "{\"modelId\":\"0000047443\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"默认服务\",\"properties\":[
{\"abilityId\":101,\"accessMode\":\"ro\",\"code\":\"firmware_version\",\"description\":\"Fixed 11 characters in dotted versioning format:\\nXX.YY.NN.MM\\n\",\"name\":\"Firmware Version\",\"typeSpec\":{\"type\":\"string\",\"maxlen\":255}},
{\"abilityId\":102,\"accessMode\":\"ro\",\"code\":\"hardware_version\",\"description\":\"Fixed 11 characters in dotted versioning format:\\nXX.YY.ZZ.NN\\n\",\"name\":\"Hardware Version\",\"typeSpec\":{\"type\":\"string\",\"maxlen\":255}},
{\"abilityId\":103,\"accessMode\":\"rw\",\"code\":\"power_rating\",\"description\":\"\",\"name\":\"Power Rating\",\"typeSpec\":{\"type\":\"value\",\"max\":3600,\"min\":0,\"scale\":0,\"step\":100,\"unit\":\"Watt\"}},
[...]

PR #4247, opened 2026-01-03
Add EKO07262 thermostat

hally0:
Ref. #4239 

I've tested it locally and it seems to work fine. I haven't add all the functionality that the thermostat provides though. Just enough to use it in heating mode.

image

### DIFF
--- a/custom_components/tuya_local/devices/elko_cfmtb_thermostat.yaml
+++ b/custom_components/tuya_local/devices/elko_cfmtb_thermostat.yaml
@@ -65,7 +65,7 @@ entities:
           - dps_val: "Off"
             value: none
   - entity: number
-    name: Power Rating
+    name: Power rating
     category: config
     class: power
     dps:
@@ -84,7 +84,7 @@ entities:
         name: lock
         type: boolean
   - entity: number
-    name: Screen Brightness (Wake)
+    name: Active screen brightness
     icon: "mdi:brightness-6"
     category: config
     dps:
@@ -96,7 +96,7 @@ entities:
           min: 10
           max: 100
   - entity: number
-    name: Screen Brightness (Standby)
+    name: Standby screen brightness
     icon: "mdi:brightness-4"
     category: config
     dps:
@@ -132,7 +132,7 @@ entities:
           - dps_val: "47K"
             value: "47K Ohm"
   - entity: sensor
-    name: Device presetting
+    name: Device type
     icon: "mdi:cog"
     category: diagnostic
     class: enum
@@ -140,7 +140,6 @@ entities:
       - id: 118
         type: integer
         name: sensor
-        readonly: true
         mapping:
           - dps_val: 1
             value: "Heat Pump/Oil Boiler"
@@ -161,7 +160,7 @@ entities:
           - dps_val: 9
             value: "Regulator Mode (output displayed is based on percentage)"
   - entity: sensor
-    name: Floor Temperature
+    name: Floor temperature
     category: diagnostic
     hidden: unavailable
     dps:
@@ -180,7 +179,7 @@ entities:
             value: false
           - value: true
   - entity: sensor
-    name: Air Temperature
+    name: Air temperature
     category: diagnostic
     dps:
       - id: 104


custom_components/tuya_local/devices/elko_cfmt-b_thermostat.yaml
```diff
@@ -0,0 +1,89 @@
...
+          - dps_val: Off
+            value: off
+      - id: 152
+        name: hvac_action
+        type: boolean
+        mapping:
+          - dps_val: true
+            value: heating
+          - dps_val: false
+            value: idle
```
make-all: This is a duplicate of above

custom_components/tuya_local/devices/elko_cfmt-b_thermostat.yaml
```diff
@@ -0,0 +1,89 @@
...
+        name: preset_mode
+        type: string
+        mapping:
+          - dps_val: Manual
+            value: manual
+          - dps_val: Schedule
+            value: program
+          - dps_val: Override
+            value: boost
+          - dps_val: Off
```
make-all: `off` (with any case) is a keyword in yaml, that is equivalent to `false`. To use the string "off" or "Off", you need to enclose it in double-quotes.

But it is a bit strange to have a preset for off, when the on/off state of the device is controlled from the switch that is mapped to hvac_mode above. Is this perhaps more of an `away` preset. which keeps the anti-frost protection running?

I suggest that you add `translation_key: thermostat` below `entity: climate`, which will give access to translations for preset modes `manual`, `program`, `temp_override` or `perm_override` for the other presets, and use either `away` or `none` for this one (which are both predefined presets from HA).

custom_components/tuya_local/devices/elko_cfmt-b_thermostat.yaml
```diff
@@ -0,0 +1,89 @@
...
+    category: diagnostic
+    dps:
+      - id: 105
+        name: sensor
+        type: integer
+        unit: C
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Power
```
make-all: Remove the name, as it is the same as class, so all it does is hides translations that are available.

custom_components/tuya_local/devices/elko_cfmt-b_thermostat.yaml
```diff
@@ -0,0 +1,89 @@
...
+    class: power
+    category: diagnostic
+    dps:
+      - id: 150
+        name: sensor
+        type: integer
+        unit: W
+  - entity: sensor
+    name: Energy
```
make-all: Remove this name too.

custom_components/tuya_local/devices/elko_cfmtb_thermostat.yaml
```diff
@@ -0,0 +1,89 @@
...
+        type: integer
+        unit: W
+  - entity: sensor
+    name: Energy
+    class: energy
+    category: diagnostic
+    dps:
+      - id: 151
+        name: sensor
+        type: integer
```
make-all, line 207: energy device class requires `class: total_increasing` here to set the expected state class.

The other sensors should probably have `class: measurement` on their sensor dp, to enable long-term statistic collection

custom_components/tuya_local/devices/elko_cfmt-b_thermostat.yaml
hally0: That might be the case. I'll have to do some more testing on this. I was just going off this:
```
{
    "abilityId": 140,
    "accessMode": "rw",
    "code": "runmode",
    "description": "",
    "name": "Run Mode ",
    "typeSpec": {
      "type": "enum",
      "range": [
        "Manual",
        "Schedule",
        "Override",
        "Off"
      ]
    }
}
```

I'll do some more investigation into the different modes. I'm also looking into adding more diagnostic/config entities. Should i make the PR draft in the meanwhile?

make-all: It's up to you, you can either leave this as draft until it is done, or we can go to merge with the current config and you can extend it with an update later. I'd suggest if merging now, to make the presets match translations first though.

hally0: Are the translations resolved now, or am i still missing something?

PR #4249, opened 2026-01-03
Add support for Meaco DD8L Pro dehumidifier

anthonws:
Adding support for Meaco DD8L Pro dehumidifier.

### DIFF
--- a/custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
@@ -17,7 +17,8 @@ entities:
         range:
           min: 35
           max: 80
-        step: 5
+        mapping:
+          - step: 5
       - id: 6
         name: current_humidity
         type: integer
@@ -26,11 +27,11 @@ entities:
         type: string
         mapping:
           - dps_val: dehumidify
-            value: "Smart Humidity"
+            value: normal
           - dps_val: drying
-            value: "Laundry"
+            value: laundry
           - dps_val: Sleep
-            value: "Sleep"
+            value: sleep
 
   - entity: select
     translation_key: timer
@@ -90,13 +91,14 @@ entities:
             value: 23h
           - dps_val: 24h
             value: 24h
-        # device published a time remaining, which is also visible in its control panel/screen. Adding this as an attribute instead of creating a new sensor.
+        # Added as an attribute instead of sensor since it is a string.
       - id: 102
         name: timer_remaining
         type: string
 
   - entity: select
     name: Fan mode
+    translation_key: mode
     icon: "mdi:fan"
     category: config
     dps:
@@ -105,16 +107,17 @@ entities:
         type: string
         mapping:
           - dps_val: 200w
-            value: 200W (Lowest)
+            value: low
           - dps_val: 400w
-            value: 400W
+            value: medium
           - dps_val: 600w
-            value: 600W (Highest)
+            value: high
           - dps_val: auto
-            value: Auto
+            value: auto
 
   - entity: select
     name: Swing mode
+    translation_key: mode
     icon: "mdi:swap-vertical"
     category: config
     dps:
@@ -123,35 +126,41 @@ entities:
         type: string
         mapping:
           - dps_val: b_swing
-            value: Big (Wide swing)
+            value: high
           - dps_val: m_swing
-            value: Medium
+            value: medium
           - dps_val: s_swing
-            value: Small
+            value: low
           - dps_val: n_swing
-            value: Static (Fixed)
+            value: "off"
 
   - entity: binary_sensor
     translation_key: tank_full
-    category: diagnostic
     dps:
       - id: 19
         name: sensor
-        type: integer
+        type: bitfield
         mapping:
-            # Captured when tank was full
+          # Captured when tank was full
           - dps_val: 1
             value: true
-            # Safety code for fallback
+          - value: false
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 19
+        type: bitfield
+        name: sensor
+        mapping:
           - dps_val: 0
             value: false
-            # captured when tank was empty and device was on
-          - dps_val: 8
+          - dps_val: 1
             value: false
-            # captured when tank was empty and device was off
-          - dps_val: 24
-            value: false
-
+          - value: true
+      - id: 19
+        type: bitfield
+        name: fault_code
   - entity: lock
     translation_key: child_lock
     category: config


custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
+name: Meaco DD8L Pro Dehumidifier
+products:
+  - id: meaco_dd8l_pro
```
make-all: Product ID is in an alphanumeric format without any symbols like _. It is not free form, you need to get the product id used by the actual device from the Tuya developer portal or `python -m tinytuya scan`

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
+name: Meaco DD8L Pro Dehumidifier
```
make-all: Top level name should be generic. Branding belongs under products.

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+products:
+  - id: meaco_dd8l_pro
+    name: Meaco DD8L Pro
```
make-all: This should be split into `manufacturer` and `model`

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+        name: switch
+        type: boolean
+      - id: 2
+        name: humidity
+        type: integer
+        range:
+          min: 35
+          max: 80
+        mapping:
+          - scale: 1
```
make-all: 1 is the default scale, it does not need to be specified

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+            step: 5
+      - id: 6
+        name: current_humidity
+        type: integer
+        class: measurement
```
make-all: class is not valid here. humidifiers are always treated as returning measurements by HA.

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+  - entity: sensor
+    name: Humidity
+    class: humidity
+    category: diagnostic
+    dps:
+      - id: 6
+        name: sensor
+        type: integer
+        unit: "%"
+        class: measurement
```
make-all: There is no need to duplicate the current_humidity from the humidifier here.

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+
+  - entity: sensor
+    name: Operating status
+    category: diagnostic
+    class: enum
+    icon: "mdi:state-machine"
+    dps:
+      - id: 5
```
make-all: This duplicates the humidifier mode, except it seems to have extra values that do not make sense as dehumidifier modes. I have not seen a read-write dps that has different read options than write before, so possibly one of these dps is a mistake.

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
+name: Meaco DD8L Pro Dehumidifier
+products:
+  - id: meaco_dd8l_pro
+    name: Meaco DD8L Pro
+
+entities:
+  - entity: humidifier
+    dps:
```
make-all, line 10: add `class: dehumidifier`

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+            value: Big (Wide swing)
+          - dps_val: m_swing
+            value: Medium
+          - dps_val: s_swing
+            value: Small
+          - dps_val: n_swing
+            value: Static (Fixed)
+
+  - entity: binary_sensor
+    name: Tank full
```
make-all: Use `translation_key: tank_full` instead of the name and class here.

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+        name: sensor
+        type: integer
+        mapping:
+          - dps_val: 0
+            value: false
+          - dps_val: 1
+            value: true
+
+  - entity: lock
+    name: Child lock
```
make-all: Use `translation_key: child_lock` instead of the name here

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+    category: config
+    dps:
+      - id: 16
+        name: lock
+        type: boolean
+
+  - entity: switch
+    name: Ionizer
```
make-all: Use `translation_key: ionizer` instead of the name and icon here.

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+            value: Clean Filter C1
+          - dps_val: filter_c2
+            value: Clean Filter C2
+          - dps_val: cable_tidy
+            value: Connect Cable Tidy
+          - dps_val: room_temp
+            value: Room Temperature Issue
+
+  - entity: select
+    name: Off timer
```
make-all: Since this is the only timer, use `translation_key: timer` instead of the name and icon here.

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+    icon: "mdi:timer-off"
+    category: config
+    dps:
+      - id: 17
+        name: option
+        type: string
+        mapping:
+          - dps_val: cancel
+            value: "Off"
```
make-all: Use the same values as the dps_val in this list, as they are all covered by translation_key: timer

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
```diff
@@ -0,0 +1,264 @@
...
+          - dps_val: 21h
+            value: 21 hours
+          - dps_val: 22h
+            value: 22 hours
+          - dps_val: 23h
+            value: 23 hours
+          - dps_val: 24h
+            value: 24 hours
+
+  - entity: sensor
```
make-all: In general enum values that are set will not update as the timer counts down, so it is not correct to duplicate the select as a sensor here.

If this device happens to update them, you should anyway see that reflected in the state of the select entity above.

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
anthonws: There's another timer. On Timer. But I left it out because I don't need it :S

anthonws:
@make-all Thank you so much for your kind feedback! First time contributing and this was a heck of a journey for me. Naturally that I leveraged the good help of Gemini, but learned a lot going through the DPS debug logs and testing.
Please check if the latest changes conform to your requirements.

anthonws:
No more changes now :) Everything is working on my end. Will wait for your feedback. TY.

custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
make-all: OK, so keep the name, but add the translation_key: timer for the options below and the icon (remove the explicit icon).

PR #4255, opened 2026-01-05
Add Klarstein Bornholm (version based on WBR3, with 1kW)

maricn:
I already implemented this over a year ago (see https://github.com/make-all/tuya-local/compare/main...maricn:tuya-local:main), and it has been working fine for me for a year until recently.

Then i pulled upstream two months ago and since then I can't install tuya-local with my own manifest anymore, so the integration stopped working. Now I resolved some syntax changes, rebased and i'm pushing to upstream for approval. I haven't tested the syntax changes from the last two commits locally.

### DIFF
--- a/custom_components/tuya_local/devices/klarstein_bornholm_smart_heater.yaml
+++ b/custom_components/tuya_local/devices/klarstein_bornholm_smart_heater.yaml
@@ -63,9 +63,18 @@ entities:
             value: false
           - dps_val: false
             value: true
+  - entity: light
+    translation_key: display_light
+    category: config
+    dps:
       - id: 102
         type: boolean
         name: switch
+        mapping:
+          - dps_val: true
+            value: false
+          - dps_val: false
+            value: true
   - entity: lock
     translation_key: child_lock
     category: config


custom_components/tuya_local/devices/klarstein_bornholm_smart_heater.yaml
```diff
@@ -0,0 +1,171 @@
...
+    dps:
+      - id: 5
+        type: boolean
+        name: switch
+        mapping:
+          - dps_val: true
+            value: false
+          - dps_val: false
+            value: true
+      - id: 102
```
make-all, line 70: There are two dps here on the same entity with the name switch, which is a conflict.
What is the purpose of these 2?

maricn: hm, i probably introduced this issue in this old commit https://github.com/make-all/tuya-local/pull/4255/commits/aa6d796655490452f20df4607dc93c45d521fe02 but it worked for me in Nov 2024..

i assume one is light for the current temperature on display, and the other one is "background" light for the config buttons..

i will effectively revert this commit now..

maricn: @make-all thanks for highlighting that issue.. does it look better now?

PR #4258, opened 2026-01-05
Wellio_D008A_20L_dehumidifier

laskas-p:
This PR adds support for the Wellio_D008A_20L_dehumidifier
Tested locally with Tuya-Local 2025.12.3
All DP mappings confirmed working.

### DIFF
--- a/custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
@@ -102,7 +102,7 @@ entities:
 
   # Child lock control
   - entity: lock
-    translation_key: child_lock 
+    translation_key: child_lock
     category: config
     dps:
       - id: 7


custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+      # Target humidity setpoint controlled by the dehumidifier
+      - id: 4
+        name: humidity
+        type: integer
+        unit: "%"
+        range:
+          min: 29
+          max: 91
+        step: 1
```
make-all: Step needs to be inside a `mapping`, but 1 is the default step so this can just be removed.

There are more like this below.

custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
```diff
@@ -0,0 +1,107 @@
+name: Wellio Dehumidifier
```
make-all: Use a generic name here in case other products also match this config.

custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+products:
+  - id: i3il8w2rhtn90s33
+    name: Wellio Dehumidifier
```
make-all: Split name into `manufacturer` and `model`.
But it isn't necessary to include the device type here (that should be in the name above).

custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+entities:
+  # Main dehumidifier entity exposed to Home Assistant
+  - entity: humidifier
+    dps:
```
make-all, line 12: Use `class: dehumidifier`

custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
```diff
@@ -0,0 +1,107 @@
...
+      - id: 2
+        name: mode
+        type: string
+        mapping:
+          - dps_val: "1"
+            value: dehumidify
+          - dps_val: "2"
+            value: laundry
+          - dps_val: "3"
+            value: fan
```
make-all: There is a predefined preset `purify` for fan-only cases (most dehumidifiers have air filters and the fans are weak compared to a standalone fan so the fan-only mode is more for running as an air purifier than circulating air).

custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
```diff
@@ -0,0 +1,107 @@
...
+        type: string
+        mapping:
+          - dps_val: "1"
+            value: low
+          - dps_val: "2"
+            value: high
+
+  # Internal temperature sensor
+  - entity: sensor
+    name: Temperature
```
make-all: Leave out the `name` when it is the same as `class`

custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
```diff
@@ -0,0 +1,107 @@
...
+        name: value
+        type: integer
+        range:
+          min: 0
+          max: 24
+        step: 1
+        unit: "h"
+
+  # Fault or status code reported by the device
+  - entity: sensor
```
make-all: Look at other configs to see how to define a binary_sensor with class `problem` to cover fault sensors.

Probably one of the bits is for a tank_full indicator, which should be separated from the general problems, since it is expected in normal use. Maybe also a defrost indicator that periodically comes on in normal use.

custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
```diff
@@ -0,0 +1,107 @@
...
+    name: Fault code
+    category: diagnostic
+    dps:
+      - id: 11
+        name: sensor
+        type: integer
+
+  # Child lock control
+  - entity: switch
```
make-all: Use a lock entity with `translation_key: child_lock`, for consistency with other devices.

custom_components/tuya_local/devices/wellio_d008a20l_dehumidifier.yaml
```diff
@@ -0,0 +1,107 @@
...
+        name: humidity
+        type: integer
+        unit: "%"
+        range:
+          min: 29
+          max: 91
+        step: 1
+
+  # Fan speed selector
+  - entity: select
```
make-all: Can be implemented as a fan entity, together with the swing (oscillate) below (and reusing the same switch as humidifier).

PR #4270, opened 2026-01-06
New device: Stadler Form Noah and Noah Pro

Smalle007:
Stadler Form Noah Pro Humidifier support added.

Implemented:
1. Humidifier with switch, current humidity, humidity and mode
2. Fan Speed
3. Child Lock
4. Filter Usage
5. Water Level
6. Display night mode
7. Filter Change Warning
8. Empty Tank Warning

Stadler Form Noah Pro Home Assistant

### DIFF
--- a/custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
+++ b/custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
@@ -58,7 +58,7 @@ entities:
         type: boolean
         name: lock
   - entity: sensor
-    name: Filter Usage
+    name: Filter usage
     icon: mdi:filter
     category: diagnostic
     dps:
@@ -67,7 +67,7 @@ entities:
         name: sensor
         unit: "%"
   - entity: sensor
-    name: Water Level
+    name: Water level
     icon: mdi:cup-water
     category: diagnostic
     dps:
@@ -90,7 +90,7 @@ entities:
           - dps_val: Light_off
             value: 0
   - entity: binary_sensor
-    name: Filter Change Required
+    name: Filter change
     class: problem
     category: diagnostic
     dps:


Smalle007:
Implemented:
1. Humidifier with switch, current humidity, humidity and mode
2. Fan Speed
3. Child Lock
4. Filter Usage
5. Water Level
6. Display night mode
7. Filter Change Warning
8. Empty Tank Warning

Stadler Form Noah Pro Home Assistant

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
```diff
@@ -0,0 +1,103 @@
+name: Humidifier
+products:
+  - id: bf27abad144ae931b76ypd
```
make-all: This appears to be the device id (hex, except at the end and 22 characters, unique to a single device), not the product id (alphanumeric throughout, 16 characters, common for all devices of a specific model).

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
```diff
@@ -0,0 +1,103 @@
...
+        type: boolean
+      - id: 13
+        name: current_humidity
+        type: integer
+      - id: 101
+        name: humidity
+        type: integer
+        range:
+          min: 40
+          max: 80
```
make-all, line 23: Is there any restriction on the step? Many humidifiers are only controlable in 5% or 10% steps.

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - id: 103
+        name: mode
+        type: boolean
+        mapping:
+          - dps_val: true
+            value: auto
+          - dps_val: false
+            value: manual
```
make-all: The [HA default preset list](https://github.com/home-assistant/core/blob/dev/homeassistant/components/humidifier/strings.json#L68) does not include manual, but maybe `normal` fits?

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
```diff
@@ -0,0 +1,103 @@
...
+    name: Water Level
+    icon: mdi:cup-water
+    category: diagnostic
+    dps:
+      - id: 17
+        name: sensor
+        type: integer
+        unit: "%"
+  - entity: select
+    translation_key: display
```
make-all, line 79: display is not a translation_key for select entities. It would be better to use a light entity here (which does have a translation_key: display), and map the options below to `brightness`:
```yaml
  - entity: light
    translation_key: display
    category: config
    dps:
      - id: 105
        name: brightness
        type: string
        mapping:
          - dps_val: Normal
            value: 255
          - dps_val: Dimmer
            value: 128
          - dps_val: Light_off
            value: 0
```

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - id: 103
+        name: mode
+        type: boolean
+        mapping:
+          - dps_val: true
+            value: auto
+          - dps_val: false
+            value: manual
+  - entity: number
+    name: Fan Speed
```
make-all: Using a fan entity here will have some benefits for  the UI and voice interfaces.

Smalle007: @make-all  also remapped the fan speed to a fan entity. :)
  
  - entity: fan
    dps:
      - id: 1
        name: switch
        type: boolean
      - id: 102
        name: speed
        type: string
        mapping:
          - dps_val: "level_1"
            value: 20
          - dps_val: "level_2"
            value: 40
          - dps_val: "level_3"
            value: 60
          - dps_val: "level_4"
            value: 80
          - dps_val: "level_5"
            value: 100

Ugzuzg: Shouldn't `dps_val` still be a number from 1 to 5?

I have landed on the following fan entity for Stadler Form Noah (without Pro):
```yaml
   - entity: fan
     dps:
       - id: 1
         name: switch
         type: boolean
       - id: 102
         name: speed
         type: integer
         mapping:
           - dps_val: 1
             value: 20
           - dps_val: 2
             value: 40
           - dps_val: 3
             value: 60
           - dps_val: 4
             value: 80
           - dps_val: 5
             value: 100
       - id: 103
         name: preset_mode
         type: boolean
         mapping:
           - dps_val: true
             value: auto
           - dps_val: false
             value: normal
```

Not so sure about duplicating the preset_mode for a fan.

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
```diff
@@ -0,0 +1,112 @@
+name: Humidifier
+products:
+  - id: hupfwl9kw70i6hoc
```
Ugzuzg, line 3: We can add Stadler Form Noah (without Pro), too. I have one, it works the same. How do I find its product id?

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
make-all: The dps_val depends on the values used by device itself. We are mapping those values to values used in HA, which in the case of speed should be a number between 0 and 100.

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
Ugzuzg: Found it through the API. The product ID for my non-Pro is the same: hupfwl9kw70i6hoc.

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
Smalle007: @Ugzuzg hmm good point. Think I made a mistake there. :) Pushed the change.

custom_components/tuya_local/devices/stadlerform_noah_humidifier.yaml
Smalle007: @Ugzuzg also my bad, looks like I used the wrong product id. Pushed a change for this.

Smalle007:
@Ugzuzg @make-all new to this all, anyone has an idea why the workflows are not running? :) Think everything is done and working. Also added both product id's for Noah and Noah pro.

make-all:
It is a standard Github security measure that PRs from first time contributors do not result in any automated scripts running, so for this PR you will not see any results until manually approved (even my changes to your PR need approval before the workflows run).

PR #4293, opened 2026-01-09
Added Realme TechLife Robot Vacuum

exileed:
Device: TechLife Robot Vacuum
Product ID: aayuntqnrmc7jnq9
Link to product: https://www.realme.com/cz/realme-techlife-robot-vacuum

### DIFF
--- a/custom_components/tuya_local/devices/realme_techlife_vacuum.yaml
+++ b/custom_components/tuya_local/devices/realme_techlife_vacuum.yaml
@@ -193,7 +193,7 @@ entities:
     translation_key: mopping
     dps:
       - id: 110
-        name: water_mode
+        name: option
         type: string
         mapping:
           - dps_val: close


custom_components/tuya_local/devices/reamle_techlife_vacuum.yaml
```diff
@@ -0,0 +1,331 @@
+name: Realme Techlife vacuum
```
make-all: Top level name should be generic in case other brands also match.
Branding should go under products section

custom_components/tuya_local/devices/reamle_techlife_vacuum.yaml
```diff
@@ -0,0 +1,331 @@
...
+            value: mop
+          - dps_val: quiet
+            value: quiet
+          - dps_val: auto
+            value: normal
+          - dps_val: strong
+            value: turbo
+          - dps_val: max
+            value: max
+      - id: 110
```
make-all: This should be split to a separate select entity, as the HA vacuum entity does not support water_mode.

custom_components/tuya_local/devices/reamle_techlife_vacuum.yaml
```diff
@@ -0,0 +1,331 @@
...
+    category: diagnostic
+    class: duration
+    icon: "mdi:clock-outline"
+    dps:
+      - id: 120
+        name: sensor
+        type: integer
+        unit: s
+  - entity: sensor
+    name: Filter time
```
make-all: New 'translation_key: filter_life' can be used here. For consistency, the above should probably also rename "time" to "life" as any future translations to cover those are likely to use similar names to existing translations.

custom_components/tuya_local/devices/reamle_techlife_vacuum.yaml
```diff
@@ -0,0 +1,331 @@
...
+    icon: "mdi:clock-outline"
+    dps:
+      - id: 121
+        name: sensor
+        optional: true
+        type: integer
+        unit: s
+  - entity: select
+    translation_key: language
+    icon: "mdi:translate"
```
make-all: icon is not needed with the above translation_key

custom_components/tuya_local/devices/reamle_techlife_vacuum.yaml
```diff
@@ -0,0 +1,331 @@
...
+    category: config
+    dps:
+      - id: 133
+        type: integer
+        name: option
+        mapping:
+          - dps_val: 0
+            value: chinese
+          - dps_val: 1
+            value: english female
```
make-all: I suggest just english here, as that already exists as a translation.

custom_components/tuya_local/devices/reamle_techlife_vacuum.yaml
```diff
@@ -0,0 +1,331 @@
...
+          - dps_val: 2
+            value: english children
```
make-all: Suggest capitalising English here, and use "child" rather than "children", as it should be a single voice.

custom_components/tuya_local/devices/realme_techlife_vacuum.yaml
```diff
@@ -0,0 +1,331 @@
...
+        name: option
+        mapping:
+          - dps_val: 0
+            value: chinese
+          - dps_val: 1
+            value: english female
+          - dps_val: 2
+            value: english children
+  - entity: binary_sensor
+    name: Mop attached
```
make-all, line 335: This can maybe also used in some lists above to filter the options.

Issue #3801, opened 2025-10-03
Request support for Klarta Humea Grande


mwolanski:
### Log message

```text
2025-10-03 22:04:22.653 WARNING (MainThread) [custom_components.tuya_local.config_flow] Adding KLARTA HUMEA GRANDE device with product id oi4ntqm3pc2g0cod
2025-10-03 22:04:22.653 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches jh_pcmini with quality of 17%. DPS: {"updated_at": 1759521859.287485, "1": true, "6": "Cancel", "10": 22, "14": 51, "16": false, "19": "CO", "20": 0, "29": false, "101": "CO", "102": "Lack_of_water", "103": "Medium_speed", "104": 180}
```

### Product ID

oi4ntqm3pc2g0cod

### Product Name

Klarta Humea Grande

### DPS information

```text
{
  "result": {
    "active_time": 1759520809,
    "category": "jsq",
    "create_time": 1759520809,
    "icon": "smart/icon/bay1647258892363WRU8/8e108b7fe7a8cd231aa73dfa91b97649.jpg",
    "id": "bffe04dcb07035d11b3x2e",
    "is_online": true,
    "lat": "51.11",
    "lon": "16.96",
    "model": "HUMEA GRANDE",
    "name": "KLARTA HUMEA GRANDE",
    "product_id": "oi4ntqm3pc2g0cod",
    "product_name": "KLARTA HUMEA GRANDE",
    "sub": false,
    "time_zone": "+02:00",
    "update_time": 1759521253,
    "uuid": "dd6c37defe170f40"
  },
  "success": true,
  "t": 1759524103956,
  "tid": "5f605284a09911f0a25e9ed280a40f39"
}

{
  "result": {
    "properties": [
      {
        "code": "switch",
        "custom_name": "",
        "dp_id": 1,
        "time": 1759522531660,
        "type": "bool",
        "value": false
      },
      {
        "code": "work_mode",
        "custom_name": "",
        "dp_id": 6,
        "time": 1759520814454,
        "type": "enum",
        "value": "Cancel"
      },
      {
        "code": "temp_current",
        "custom_name": "",
        "dp_id": 10,
        "time": 1759520814454,
        "type": "value",
        "value": 22
      },
      {
        "code": "humidity_current",
        "custom_name": "",
        "dp_id": 14,
        "time": 1759523874358,
        "type": "value",
        "value": 50
      },
      {
[...]

PR #4297, opened 2026-01-10
Add support for Klarta Humea Grande

PawelSwaczyna:
I tested this yaml configuration for a few weeks, and it works correctly. Color names for the backlight correspond to those in the Tuya App. The water_level sensor has three states: "Water_enough", "Lack_of_water", and "Tank_removed".

This solves the issue: https://github.com/make-all/tuya-local/issues/3801

### DIFF
--- a/custom_components/tuya_local/devices/klarta_humea_grande_humidifier.yaml
+++ b/custom_components/tuya_local/devices/klarta_humea_grande_humidifier.yaml
@@ -81,44 +81,51 @@ entities:
       - id: 29
         type: boolean
         name: lock
-  - entity: number
+  - entity: fan
     name: Spray level
-    icon: "mdi:fan"
+    icon: "mdi:spray"
     category: config
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/klarta_humea_grande_humidifier.yaml
```diff
@@ -0,0 +1,163 @@
...
+            value: 45
+          - dps_val: "50RH"
+            value: 50
+          - dps_val: "55RH"
+            value: 55
+          - dps_val: "60RH"
+            value: 60
+          - dps_val: "65RH"
+            value: 65
+          - dps_val: "70RG"
```
make-all: typo?

PawelSwaczyna: Yes, it should be 70RH. I hope I fixed that.

PR #4312, opened 2026-01-12
Add Meaco Clean Air Purifier device configuration

malrogers7:
Creates a device page for Meaco Clean CA-HEPA 76x5 Air Purifier

Confirmed working on my device and can be discovered when adding the device to Tuya Local.

### DIFF
--- a/custom_components/tuya_local/devices/meaco_clean_purifier
+++ b/custom_components/tuya_local/devices/meaco_clean_purifier
@@ -1,4 +1,4 @@
-name: Air Purifier
+name: Air purifier
 products:
   - id: 4kqsufvuurljmryl
     manufacturer: Meaco
@@ -118,18 +118,17 @@ entities:
       - id: 6
         name: switch
         type: boolean
-  - entity: select
-    name: Indicator Light
+  - entity: light
     translation_key: indicator
     category: config
     dps:
       - id: 101
-        name: option
+        name: brightness
         type: string
         mapping:
           - dps_val: Close
-            value: "Off"
-          - dps_val: Standard
-            value: "Standard"
+            value: 0
           - dps_val: Soft
-            value: "Soft"
+            value: 128
+          - dps_val: Standard
+            value: 255


custom_components/tuya_local/devices/meaco_clean_purifier
```diff
@@ -0,0 +1,136 @@
+name: Meaco Air Purifier
```
make-all: Top level name should be generic in case other devices also match this config.
Branding should be under products listings only.

custom_components/tuya_local/devices/meaco_clean_purifier
```diff
@@ -0,0 +1,136 @@
...
+products:
+  - id: 4kqsufvuurljmryl
+    manufacturer: Meaco
+    model: Meaco76x5WiFi
+    name: Meaco Clean CA-HEPA 76x5 Air Purifier
```
make-all: Use model here, and model_id for the above. Do not repeat the manufacturer here, or the device type (which should be the generic name above).

custom_components/tuya_local/devices/meaco_clean_purifier
```diff
@@ -0,0 +1,136 @@
...
+    class: pm25
+    dps:
+      - id: 2
+        name: sensor
+        type: integer
+        class: measurement
+        unit: ugm3
+  - entity: sensor
+    translation_key: filter_life
+    name: Filter Remaining
```
make-all: Don't override the name, the translation_key already provides a suitable one.

custom_components/tuya_local/devices/meaco_clean_purifier
```diff
@@ -0,0 +1,136 @@
...
+          - dps_val: Cancel
+            value: cancel
+          - dps_val: "2h"
+            value: "2h"
+          - dps_val: "4h"
+            value: "4h"
+          - dps_val: "8h"
+            value: "8h"
+  - entity: sensor
+    name: Timer remaining
```
make-all: Use `translation_key: time_remaining` instead of explicit name and icon.

make-all: also use `class: duration` to allow customization of units for display in the UI

custom_components/tuya_local/devices/meaco_clean_purifier
```diff
@@ -0,0 +1,136 @@
...
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 22
+        name: sensor
+        type: bitfield
+        mapping:
+          - dps_val: 0
+            value: false
+          - dps_val: 1
```
make-all: Rather than listing the values for true, use a catch-all default:
```
        - value: true
```

Also, add the raw value as an attribute
```
      - id: 22
        name: fault_code
        type: bitfield
```

custom_components/tuya_local/devices/meaco_clean_purifier
```diff
@@ -0,0 +1,136 @@
...
+        name: lock
+        type: boolean
+  - entity: switch
+    translation_key: ionizer
+    category: config
+    dps:
+      - id: 6
+        name: switch
+        type: boolean
+  - entity: select
```
make-all: This can be a light entity:

```yaml
  - entity: light
    translation_key: indicator
    category: config
    dps:
      - id: 101
        name: brightness
        type: string
        mapping:
          - dps_val: Close
            value: 0
          - dps_val: Soft
            value: 128
          - dps_val: Standard
            value: 255
```

custom_components/tuya_local/devices/meaco_clean_purifier
malrogers7: Thanks.  Changed.

custom_components/tuya_local/devices/meaco_clean_purifier
malrogers7: Thanks.  Changed.

custom_components/tuya_local/devices/meaco_clean_purifier
malrogers7: Thanks.  Changed.

custom_components/tuya_local/devices/meaco_clean_purifier
malrogers7: Thanks.  Changed.

custom_components/tuya_local/devices/meaco_clean_purifier
malrogers7: Thanks.  Changed and added.

custom_components/tuya_local/devices/meaco_clean_purifier
malrogers7: Thanks.  I'm not sure about the light entity here.  It's a light indicator which is meant to show the air quality.  It has discrete states and is not a dimmable room light.  A slider implies you can set the brightness to 42% or 88%, but that's not correct.  "Soft" and "Standard" from the manufacturer are self-explanatory.  
Also, I'm concerned that making it a light entity will cause it to be triggered by 'Turn off all lights' service calls, which isn't desired for a device display.
This is also more of a 'mode' switch. As the light can in fact be off when in the sleep mode, even though the switch is set to standard or soft. Again, I think this would be odd if the light entity was on.

make-all: There are many such indicators in other devices, I guess it can be considered a matter of preference whether you want this type of lighting on devices to go off with the rest of your lights or not, but this can be controlled by using areas, labels or light groups. The override of the sleep does make a difference, but I think this can be handled by adding the sleep dp as the "available" dp on the light entity, so it is not available for controlling when in sleep mode.

malrogers7:
Thanks for sorting this! I was just about to make those changes, and I think your light entity is better.

PR #4317, opened 2026-01-13
Added on option saving authorization data and polling climate devices to prevent them from entering sleep mode.

Stormsidee:
Added on option saving authorization data and polling climate devices to prevent them from entering sleep mode.

### DIFF
--- a/custom_components/tuya_local/climate.py
+++ b/custom_components/tuya_local/climate.py
@@ -117,9 +117,7 @@ class TuyaLocalClimate(TuyaLocalEntity, ClimateEntity):
         self._last_poll_success = False
         self._last_data_refresh = None
 
-        # Флаг для устройств через шлюз
         self._is_gateway_device = False
-        # Проверяем, есть ли device_cid (для устройств через шлюз)
         if hasattr(self._device, 'device_cid') and self._device.device_cid:
             self._is_gateway_device = True
             _LOGGER.info("Climate device %s is connected via gateway (device_cid: %s)",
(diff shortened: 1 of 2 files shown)


make-all:
This needs more explanation of what you are trying to accomplish here.

It seems you are adding cloud functionality to try to support battery powered devices, which is not in line with the projects aims. Perhaps starting a discussion first would have been a good idea.

Stormsidee:
> This needs more explanation of what you are trying to accomplish here.
> 
> It seems you are adding cloud functionality to try to support battery powered devices, which is not in line with the projects aims. Perhaps starting a discussion first would have been a good idea.

There are several problems. The first is a problem with battery-operated devices, specifically the zigbee_zb10wt_temphumid_sensor device in your list. This device is powered by a battery, and if it does not transmit data to HASS for a couple of days (no HASS login, and no changes in temperature or humidity), the device enters sleep mode. Once the device enters sleep mode, it disconnects from your integration. It is not clear why this happens (either the local key changes when you reconnect to the network, or something else happens with the parameters), but it does happen. The second problem is the dewin_kws306wf_energymeter device. This device is powered by a 220 V (380 V) network and also falls out of integration due to going into sleep mode (or standby mode).

Also, we have scenes and automations for temperature sensors, but since it constantly falls out of integration, the automations also don't work.

Stormsidee:
@make-all 
At the moment, we turned it off for 20 minutes, and after turning it back on, the KWS306 and the gateway through which the temperature sensor is connected stopped working through integration, but the sensor is still listed in the device list.
However, with our fixes, the KWS306 immediately connected, and it also stopped connecting through integration before.

make-all:
I don't understand why you have made specific changes to the climate platform. Neither of the devices you describe having problems with has a climate entity.

make-all:
Probably something in here is fixing the issue with KWS306 connection, but I cannot tell what, since there are no changes in here related to connection handling. As I said, battery powered devices and other devices that put the WiFi into sleep mode are not supported by this integration. I already cannot keep up with the inflow of PRs and issues, and do not want to make things worse by enabling inherently unstable by design devices.

PR #4319, opened 2026-01-13
Add Engo E901WiFi Thermostat (X3TH09RW) configuration

AlexFozor:
New device configuration, not found in database before.

The work was checked on a personal HA.

### DIFF
--- a/custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
+++ b/custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
@@ -19,7 +19,7 @@ entities:
         name: hvac_action
         type: string
         mapping:
-          - dps_val: heating_on           
+          - dps_val: heating_on
             value: heating
             constraint: hvac_mode
             conditions:


custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
+name: Engo E901WIFI Thermostat
```
make-all: Top level name should be generic, in case other devices also match the config.

custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
...
+products:
+  - id: bowuem6d9ihzwymj
+    name: Engo E901WIFI Thermostat
```
make-all: Split this into `manufacturer` and `model`. Thermostat will be in the generic name above, so does not need repeating here.

custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
...
+        mapping:
+          - dps_val: hot
+            value: heat
+          - dps_val: cold
+            value: cool
+      - id: 109
+        name: hvac_action
+        type: string
+        mapping:
+          - dps_val: heating_on
```
make-all, line 22: What is output when the device is cooling? If it is the same value, then a condition based on `constraint: hvac_mode` should be used to map to the correct action. If there is a different value, then that should also be in this mapping.

custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
...
+            value: heating
+          - dps_val: heating_off
+            value: idle
+      - id: 16
+        name: temperature
+        type: integer
+        range:
```
make-all, line 34: `unit: C` is needed here.

custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
...
+        type: integer
+        mapping:
+          - scale: 10
+  - entity: number
+    name: Temperature Correction
+    category: config
+    dps:
+      - id: 27
+        name: value
+        type: integer
```
make-all, line 56: `unit: °` should be used here.

custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
...
+        range:
+          min: -35
+          max: 35
+        mapping:
+          - scale: 10
+            step: 5
+  - entity: switch
```
make-all: Standard pattern here is
```yaml
  - entity: lock
    translation_key: child_lock
    category: config
    dps:
      - id: 40
        name: lock
        type: boolean
```

custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
...
+    name: Child Lock
+    category: config
+    dps:
+      - id: 40
+        name: switch
+        type: boolean
+  - entity: sensor
+    name: Battery
```
make-all: Do not use the same name as the class name, just leave the name out so the class translations can be used.

custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
...
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 35
+        name: sensor
+        type: integer
+        unit: "%"
+  - entity: sensor
```
make-all: Standard pattern here is
```yaml
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 45
        name: sensor
        type: bitfield
        mapping:
          - dps_val: 0
            value: false
          - value: true
      - id: 45
        name: fault_code
        type: bitfield
```

custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
...
+        type: integer
+        mapping:
+          - scale: 10
+      - id: 32
+        name: holiday_temperature
+        type: integer
+        mapping:
+          - scale: 10
+  - entity: number
+    name: Temperature Correction
```
make-all: Use `translation_key: temperature_calibration`

custom_components/tuya_local/devices/engo_e901wifi_thermostat.yaml
```diff
@@ -0,0 +1,102 @@
...
+        unit: "%"
+  - entity: sensor
+    name: Fault
+    category: diagnostic
+    dps:
+      - id: 45
+        name: sensor
+        type: bitfield
+  - entity: sensor
+    name: Heating Ratio
```
make-all: HA naming convention for entity names is to only capitalize the first word

PR #4321, opened 2026-01-13
Add Fairland Mr.Pure salt chlorinator alternative product

verwilst:
My Fairland InverX Mr.Pure salt chlorinator wasn't detected as such by tuya-local. The front of the device has "Fairland InverX" instead of the Aquark logo. I looked up the product ID in the Tuya platform, added it to the yaml locally, after which it was detected correctly.

Product: https://www.fairlandgroup.com/product-18/200.html

### DIFF
--- a/custom_components/tuya_local/devices/mrpure_saltchlorinator.yaml
+++ b/custom_components/tuya_local/devices/mrpure_saltchlorinator.yaml
@@ -159,6 +159,7 @@ entities:
     dps:
       - id: 114
         type: boolean
+        optional: true
         name: sensor
         mapping:
           - dps_val: true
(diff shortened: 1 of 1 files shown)


verwilst:
image

make-all:
Blindly adding product ids to configs that fail to match is not a good idea. The log and dps info needs to be checked against the config to confirm that it is really compatible, and adjustments made to the config first so that it can match without the product id, except in extreme cases where the new device is sending almost no info on first connection.

verwilst:
First timer here for this project, I wasn't aware about this. It recommended me things like generic_ir_remote and loonas_curtain instead, so I figured it was just missing the product id.

I've been using the device through this integration for a while and it seems quite complete and fully working, so it might be a small thing that's missing?

Is this what I need to compare it with? Is there documentation that can guide me along?

Thanks!

image

PR #4322, opened 2026-01-14
Adds Xtuos pet feeder device

ben-h-a:
adds full device config for Xtuos dual pet feeder

### DIFF
--- a/.github/workflows/yamltests.yml
+++ b/.github/workflows/yamltests.yml
@@ -22,6 +22,9 @@ jobs:
       matrix:
         python-version: ['3.13']
 
+    env:
+      PYTHONPATH: .
+
     steps:
       - uses: actions/checkout@v4
         with:


custom_components/tuya_local/devices/xtuos_2x_petfeeder.yaml
```diff
@@ -0,0 +1,147 @@
...
+    name: Feed report 2
+    category: diagnostic
+    dps:
+      - id: 118
+        type: integer
+        name: sensor
+        unit: portions
+  - entity: sensor
+    name: Power state
+    class: enum
```
make-all, line 108: enum requires all the options to be listed in a mapping below.

ben-h-a: Added missing enum mapping

PR #4330, opened 2026-01-15
Add support for Gidrolock Standard WiFi water leak sensor

nkaliev:
New device configuration, not found in database before.
Device can control water supply valve, automatically close it by leak caught on external sensor.
Guard can be disabled by enabling cleaning mode.

### DIFF
--- a/custom_components/tuya_local/devices/gidrolock_standard_wifi.yaml
+++ b/custom_components/tuya_local/devices/gidrolock_standard_wifi.yaml
@@ -12,7 +12,6 @@ entities:
         name: valve
   - entity: binary_sensor
     class: moisture
-    category: diagnostic
     dps:
       - id: 101
         type: boolean


custom_components/tuya_local/devices/gidrolock_standard_wifi.yaml
```diff
@@ -0,0 +1,26 @@
...
+entities:
+  - entity: valve
+    class: water
+    dps:
+      - id: 1
+        type: boolean
+        name: valve
+  - entity: binary_sensor
+    class: moisture
+    category: diagnostic
```
make-all: This seems like the main function of this device, so I wouldn't put it under diagnostic.

nkaliev: I agree. Changed

PR #4341, opened 2026-01-17
fix casdon_kg1_waterdispenser.yaml bug

GavinIves:
To fix 
Entity sensor.inline_water_dispenser_ti_ji (<class 'custom_components.tuya_local.sensor.TuyaLocalSensor'>) is using state class 'measurement' which is impossible considering device class ('volume') it is using; expected None or one of 'total', 'total_increasing'; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/make-all/tuya-local/issues

This entity only represents the water consumption for this specific instance; no statistics are needed, only measurements.

Entity sensor.inline_water_dispenser_ti_ji (<class 'custom_components.tuya_local.sensor.TuyaLocalSensor'>) is using native unit of measurement 'ml' which is not a valid unit for the device class ('volume') it is using; expected one of ['mL', 'MCF', 'ft³', 'CCF', 'L', 'gal', 'm³', 'fl. oz.']; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/make-all/tuya-local/issues

### DIFF
--- a/custom_components/tuya_local/devices/casdon_kg1_waterdispenser.yaml
+++ b/custom_components/tuya_local/devices/casdon_kg1_waterdispenser.yaml
@@ -130,14 +130,13 @@ entities:
         name: current_temperature
 
   - entity: sensor
-    name: water_volume
+    class: volume
     category: diagnostic
     dps:
       - id: 26
         type: integer
         name: sensor
         unit: mL
-        class: measurement
 
   - entity: binary_sensor
     class: problem


custom_components/tuya_local/devices/casdon_kg1_waterdispenser.yaml
```diff
@@ -131,12 +131,11 @@ entities:
 
   - entity: sensor
     category: diagnostic
-    class: volume
```
make-all, line 134: Removing the class it is necessary to add a name. Since this is already released, a name of "Volume" will avoid changing the entity_id.

Alternatively, the message seems to suggest removing the class: measurement is an option here, and if you do not need expect long term statistics is probably best. So you could leave the class here and remove from below instead, which will keep translations for the entity name.

GavinIves: I have fixed it.

GavinIves: Since this entity records a single water usage transaction, and I don't need long-term recording and statistics, is it unnecessary to use the class: measurement? I changed the name to water_volume for better understanding. If you feel this is inappropriate, you can modify and publish it directly.

make-all: name: ... should be human readable, so capitalised and without underscore.

Can you try it with class: volume  (no name) like before, but without `class: measurement`? This will be least disruptive if it can work.

Otherwise a `name: Volume`. But either way, the `class: measurement` should probably be removed, as it is not really a continuous measurement that should be recorded for long-term statistics.

GavinIves: Understood. I've revised it again. I originally thought that the class "measurement" required long-term statistics and recording, because that's what ChatGPT told me.

PR #4342, opened 2026-01-17
Adds support for Tuya M604 quad wall switch (PID iv8nm7yimgfz2kmx).

MrJinxed55:
Tested in Home Assistant:
- 4 relays (DP1–DP4)
- countdown timers (DP7–DP10, seconds)
- initial relay state (DP14: power_off/power_on/last)
- backlight (DP16)

Also exposes DP17/DP18/DP19 as config text entities (base64 blobs) with inline documentation of the observed encoding.

### DIFF
--- a/custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
+++ b/custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
@@ -1,7 +1,6 @@
-name: Quad switch (M604)
+name: Quad switch
 products:
   - id: iv8nm7yimgfz2kmx
-    manufacturer: Tuya
     model: M604
 entities:
   - entity: switch
@@ -40,62 +39,54 @@ entities:
         name: switch
         type: boolean
 
-  - entity: number
+  - entity: time
     category: config
-    class: duration
     translation_key: timer_x
     translation_placeholders:
       x: "1"
     dps:
       - id: 7
-        name: value
         type: integer
-        unit: s
+        name: second
         range:
           min: 0
           max: 86400
 
-  - entity: number
+  - entity: time
     category: config
-    class: duration
     translation_key: timer_x
     translation_placeholders:
       x: "2"
     dps:
       - id: 8
-        name: value
         type: integer
-        unit: s
+        name: second
         range:
           min: 0
           max: 86400
 
-  - entity: number
+  - entity: time
     category: config
-    class: duration
     translation_key: timer_x
     translation_placeholders:
       x: "3"
     dps:
       - id: 9
-        name: value
         type: integer
-        unit: s
+        name: second
         range:
           min: 0
           max: 86400
 
-  - entity: number
+  - entity: time
     category: config
-    class: duration
     translation_key: timer_x
     translation_placeholders:
       x: "4"
     dps:
       - id: 10
-        name: value
         type: integer
-        unit: s
+        name: second
         range:
           min: 0
           max: 86400
@@ -114,8 +105,6 @@ entities:
             value: "on"
           - dps_val: last
             value: memory
-        optional: true
-        force: true
 
   - entity: light
     translation_key: backlight
@@ -146,10 +135,11 @@ entities:
   - entity: text
     name: Cycle schedule
     category: config
+    hidden: true
     dps:
       - id: 17
         name: value
-        type: string
+        type: base64
         optional: true
 
   # DP18 (random_time) is a base64-encoded schedule payload.
@@ -169,10 +159,11 @@ entities:
   - entity: text
     name: Random schedule
     category: config
+    hidden: true
     dps:
       - id: 18
         name: value
-        type: string
+        type: base64
         optional: true
 
   # DP19 (switch_inching) is a base64-encoded inching configuration blob.
@@ -195,8 +186,9 @@ entities:
   - entity: text
     name: Inching config
     category: config
+    hidden: true
     dps:
       - id: 19
         name: value
-        type: string
+        type: base64
         optional: true


custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
```diff
@@ -0,0 +1,191 @@
...
+    category: config
+    translation_key: timer_x
+    translation_placeholders:
+      x: "1"
+    dps:
+      - id: 7
+        name: value
+        type: integer
+        unit: s
+        class: duration
```
make-all: This belongs in the entity definition above (same level as category)

MrJinxed55: Fixing :), fixing lint as well.

MrJinxed55: Fixed finally.

MrJinxed55:
Just a question : Is there some other way I should have solved the base64 fields that contain multiple values?

From what I could tell, mask/format works well for fixed binary layouts or extracting a single value, 
but these DPS encode variable-length lists of records (and DP19 ordering can vary), 
so YAML-only didn’t seem able to safely expose per-switch values.

If there’s an existing pattern in tuya-local for parsing multi-record base64 blobs into multiple entities, 
I’d love a pointer—otherwise happy to keep them as raw text with documentation.

make-all:
text entities is best, as the format will require external decoding and encoding due to the complexity.
But I would make the type base64 rather than a string, as that will give some checking on input.
Also entities that require advanced knowledge to make use of like these should be marked `hidden: true` so that normal users do not need to deal with them. They can be individually enabled as needed after the device is added.

But I am checking now if the whole config can be merged with novadigital_quad_switch, and will hold off on adding comments to avoid wasting your time fixing up this only to be not used.

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
```diff
@@ -0,0 +1,202 @@
+name: Quad switch (M604)
```
make-all: Remove the branding from the top level name, as other devices may also match this

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
```diff
@@ -0,0 +1,202 @@
...
+products:
+  - id: iv8nm7yimgfz2kmx
+    manufacturer: Tuya
```
make-all: Tuya does not manufacture end devices, so leave this out if unknown

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
```diff
@@ -0,0 +1,202 @@
...
+  - entity: switch
+    translation_key: switch_x
+    translation_placeholders:
+      x: "4"
+    dps:
+      - id: 4
+        name: switch
+        type: boolean
+
+  - entity: number
```
make-all: The current standard for minute or second resolution timers that extend to 24 hours is to use a time entity. Please follow that.

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
```diff
@@ -0,0 +1,202 @@
...
+        type: string
+        mapping:
+          - dps_val: power_off
+            value: "off"
+          - dps_val: power_on
+            value: "on"
+          - dps_val: last
+            value: memory
+        optional: true
+        force: true
```
make-all: force: true is unlikely to have any positive effect here, and can sometimes have negative effects. Please test if it actually makes a difference before including it.

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
```diff
@@ -0,0 +1,202 @@
...
+  # - ctrl encodes switch and enable:
+  #     sw1 enabled=0x01 disabled=0x00
+  #     sw2 enabled=0x03 disabled=0x02
+  #     sw3 enabled=0x05 disabled=0x04
+  #     sw4 enabled=0x07 disabled=0x06
+  # Example observed: sw1 enabled Sun/Tue/Thu/Sat 12:00-20:30,
+  #                   start 01:00, end 00:30 => AVUC0ATOADwAHg==
+  - entity: text
+    name: Cycle schedule
+    category: config
```
make-all, line 137: Add `hidden: true` here, as most users will not want this cluttering up the UI when they cannot make any real use of it.

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
```diff
@@ -0,0 +1,202 @@
...
+    dps:
+      - id: 17
+        name: value
+        type: string
```
make-all: Make the `type: base64` to have checking on input instead of blindly sending invalid data to the device.

make-all (commented):
After thinking about it, because this device is using the dps ids usually used for energy, power, current for the base64 schedule and inching config, it is better not to try to merge with the existing quad switch that is otherwise similar, as that could affect compatibility with other quad switches that have power monitoring.

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
MrJinxed55: I set to time and name:second

But in HA I get a time set UI, i.e. 12:00 AM. This feels strange, because I am setting how many seconds after the switch is turned on, to turn it off.

So for example setting it at 1:00 PM does not really make sense.

Is there a way to get the time component to understand that it's a duration and not a specific time?

I found a lot of devices using "class: duration" but all were number, select or sensor.

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
MrJinxed55: I have set them to hidden, but they are still visible when I am testing.
I deleted the device, updated the yaml and added the device again.
(My testing is looking at the device from integrations->tuya-local->device card.)

Tested manually setting the entity hidden in the UI, but I still see the entity on the device card and can add it to dashboards without trouble.

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
make-all: You need the time set to 24h in user settings for time entities to work for durations, unfortunately the HA developers do not give any way for the integration to force 24h display for specific entities, though it has been requested.

There is also no way to restrict a time entity to a specific range, so timers that are not 24h are still using number entities. class: duration is only valid for number and sensor entities, any select entities using it are wrong.

MrJinxed55: Correction : It's a "Countdown start timer", so I just used the wrong naming.

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
make-all: I'm not sure where you have added the `hidden: true`, as the changes are not yet published, but it needs to be at entity level (same as category, name etc) to have any effect. It is not expected to change any already existing entities, but if you delete the device, then set it up again, you should see the entities disabled by default (unless there is old config for those entities that wasn't removed). They are still accessible on the device details, but I think you need to enable them before they can be added to a card.

custom_components/tuya_local/devices/wifi_m604_quad_switch.yaml
MrJinxed55: Should I go back to a number (device_class: duration) since a time entity represents time-of-day (absolute) and can’t represent a relative duration/countdown?

MrJinxed55: I understood your point now, I am keeping time as you recommended.

PR #4346, opened 2026-01-18
Add Waterco Electroheat ECO-VS heat pump configuration

Keou0007:
Added configuration for Waterco Electroheat ECO-VS heat pump, which is a rebadged Madimack Elite V3, but requires a scaling factor of 10 on the water inlet and outlet temperatures.

### DIFF
--- a/custom_components/tuya_local/devices/waterco_electroheat_eco-vs_heatpump.yaml
+++ b/custom_components/tuya_local/devices/waterco_electroheat_eco-vs_heatpump.yaml
@@ -71,7 +71,6 @@ entities:
   - entity: sensor
     category: diagnostic
     name: Power level
-    class: power_factor
     dps:
       - id: 109
         type: integer


custom_components/tuya_local/devices/waterco_electroheat_eco-vs_heatpump.yaml
```diff
@@ -0,0 +1,228 @@
...
+        name: min_temperature
+      - id: 117
+        name: current_temperature
+        type: integer
+        mapping:
+          - scale: 10
+  - entity: sensor
+    category: diagnostic
+    name: Power level
+    class: power_factor
```
make-all: I don't think power_factor is the correct class for this

Keou0007: I have simply copied the contents of the madimack_elite_v3_heatpump_updated.yaml file, which is currently working fine for my heater except for the 10x scaling error on the water temperatures. The only changes I've made are to add the scale factors, and update the product id, manufacturer and model. If the power level sensor is incorrectly configured, then it is also incorrectly configured for the elite v3, which was not my doing. The "Power level" sensor is showing the correct value for me, matching what I see in the Tuya app. Perhaps you are confusing with the "Power" sensor, which is at id 125 with class measurement, and is also correctly showing the power usage in kW.

make-all: Sure, and I have fixed that in commit 749807b after this PR brought it to my attention. The problem is not with the value, just with the incorrect class it is marked with.

Keou0007: fair enough, I removed it.

PR #4347, opened 2026-01-18
Add support for Mercator Ceiling Fan with Light

wfsyd:
Add support for Mercator Grange Ceiling Fan with Tricolour LED


wfsyd (commented):
Corrected format

make-all:
This has been merged with the existing ovlaim_ceiling_fan_light.yaml file, which was already only differing in the strings used in the timer select (which can be accommodated because they are mutually exclusive so there is no confusion about which should be used when).

PR #4348, opened 2026-01-18
Add configuration for Eberg HUMI H03G1 humidifier

drkn:
Tested configuration for Eberg HUMI H03G1 humidifier.

### DIFF
--- a/custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
+++ b/custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
@@ -113,6 +113,8 @@ entities:
         mapping:
           - dps_val: 0
             value: false
+          - dps_val: 1
+            value: false
           - value: true
       - id: 22
         type: bitfield


custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
```diff
@@ -0,0 +1,131 @@
...
+          max: 80
+        mapping:
+          - step: 1
+      - id: 14
+        type: integer
+        name: current_humidity
+      - id: 4
+        type: string
+        name: mode
+        mapping:
```
make-all, line 26: I don't quite get the meaning of these as modes. It seems to be turning something off and on, but what is it that is being turned off and on here?

Ideally the modes should be from the predefinied translations in  https://github.com/home-assistant/core/blob/3fe91751f5e8a3e12a0c8b82cf692b2c8f6b6826/homeassistant/components/humidifier/strings.json#L67

custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
```diff
@@ -0,0 +1,131 @@
...
+          - dps_val: level_1
+            value: low
+          - dps_val: level_2
+            value: medium
+          - dps_val: level_3
+            value: high
+  - entity: switch
+    category: config
+    name: Sleep Mode
```
make-all: `translation_key: sleep` can be used here in place  of explicit name and icon

custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
```diff
@@ -0,0 +1,131 @@
...
+    icon: "mdi:sleep"
+    dps:
+      - id: 16
+        type: boolean
+        name: switch
+  - entity: switch
+    category: config
+    name: Ionizer
```
make-all: This can use `translation_key: ionizer` in place of explicit name and icon

custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
```diff
@@ -0,0 +1,131 @@
...
+          - dps_val: 1h
+            value: 1h
+          - dps_val: 2h
+            value: 2h
+          - dps_val: 4h
+            value: 4h
+          - dps_val: 8h
+            value: 8h
+  - entity: sensor
+    name: Countdown Left
```
make-all: `translation_key: time_remaining` can be used here in place of explicit name and icon.

custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
```diff
@@ -0,0 +1,131 @@
...
+  - entity: binary_sensor
+    translation_key: tank_empty
+    category: diagnostic
+    dps:
+      - id: 22
+        type: bitfield
+        optional: true
+        name: sensor
+        mapping:
+          - dps_val: 0
```
make-all: Since dp 22 has a number of other fault codes bundled into it, it would be better to identify which bit is for tank_empty and only map that bit to true, with a default of false.

custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
```diff
@@ -0,0 +1,131 @@
...
+      - id: 22
+        type: bitfield
+        name: sensor
+        optional: true
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+  - entity: sensor
+    name: fault_code
```
make-all: This should just be an attribute of the problem sensor above.

custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
drkn: Humi has "humidistat" setting on dps 4 - basically auto mode which tries to keep the humidity at certain level:

- "off" means - "dp_spray_mode_off" (from tuya app) - keep spraying all the time and do not use humidistat
- "on" means - "dp_spray_mode_on" (from tuya app) - use humidistat. 

I found it more convenient to keep it on humidistat mode setting rather than making it a switch. I can change it to "Humidistat" switch you think otherwise:

image

make-all: Mode is OK, I just want to try to match it to predefined modes that already have translations. Continuous is common for dehumidifiers too, I am not sure why HA does not provide a predefined preset for this. Only auto, away, baby, boost, comfort, eco, home, normal, sleep.

The humidistat mode could fit multiple of those (auto and normal seem most likely, but comfort is possible too), but the "manual" only really matches "boost" at a stretch. I might define a continuous for humidifier in the tuya-local specific translations, similar to what I have already done for dehumidifier, to make up for this shortcoming.

make-all: I just added `continuous` as a translation for humidifiers, so you can use that for "On", and either `normal` or `auto` for "Off".

drkn: Will do thanks. I'll commit my changes soon, but still trying to figure out dps - 13. It just ... disappeared in tinytuya (still visible in tuya iot). Marked as optional for now, but wanted to check why it happened before I commit.

custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
```diff
@@ -0,0 +1,131 @@
...
+        type: boolean
+        optional: true
+        name: switch
+      - id: 13
+        type: integer
+        name: humidity
+        range:
+          min: 30
+          max: 80
+        mapping:
```
make-all: step: 1 is the default, so this mapping can be removed

custom_components/tuya_local/devices/eberg_humi_h03g1_humidifier.yaml
make-all: Adding `optional: true` is the correct fix for this. Likely it is not sent by the device when it is not relevant (power is off, or mode is manual/continuous)

drkn: You saved my day :) Turning off humidistat hides dp 13. Mystery solved

drkn:
It seems there's problem with `unstranslated_entities.py`?

PR #4349, opened 2026-01-18
Add support for Smart FONT7-SMART-B 2.5L Water Fountain

lorek123:
- Add device configuration for Smart pet water fountain
- Product ID: cafirr70niv1hj2x (same as the 宠乐园 device)
- Includes all DPs: switch, work mode, filter/cleaning reminders, water level sensors, water remaining, intake tracking, and fault binary sensors
- Note: Switch and water level enum may not work on all devices


make-all:
Already supported as kanchou_petfountain

PR #4356, opened 2026-01-19
Add support for ANWIO Smart Filament bulbs

dpage:


make-all:
The config as-is offers no additional functionality over the standard rgbcw_lightbulb.yaml

dpage:
Yes, I started to wonder about that after I created the PR, however the integration failed to connect to those particular devices without it. I'm digging in a little more...

PR #4361, opened 2026-01-20
Clarified Mini-B/Mini-F settings

Permanently:
Tweaked and clarified a couple things, such as number graphs, switching out the "reset" switch for a button (as the reset only repositions the litter) and editing the language to be suitable for clumping litter.

### DIFF
--- a/ACKNOWLEDGEMENTS.md
+++ b/ACKNOWLEDGEMENTS.md
@@ -1223,3 +1223,4 @@ Further device support has been made with the assistance of users. Please consid
 - [AlexFozor](https://github.com/AlexFozor) for contributing support for Engo E901WiFi thermostat.
 - [nkaliev](https://github.com/nkaliev) for contributing support for Gidrolock Standard leak detection water valve.
 - [cm-mojo](https://github.com/cm-mojo) for assisting with support for Pro Breeze AirFlo 43" pedestal fan.
+- [wfsyd](https://github.com/wfsyd) for contributing support for Mercator ceiling fan with lights, which was merged with Ovlaim ceiling fan with lights.
--- a/custom_components/tuya_local/manifest.json
+++ b/custom_components/tuya_local/manifest.json
@@ -18,5 +18,5 @@
         "tinytuya==1.17.4",
         "tuya-device-sharing-sdk~=0.2.1"
     ],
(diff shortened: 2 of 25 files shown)


custom_components/tuya_local/devices/minib_public_litterbox.yaml
```diff
@@ -70,6 +70,8 @@ entities:
       - id: 7
         type: integer
         name: sensor
+        unit: times
+        precision: 0
```
make-all: precision: 0 should not be needed here.

custom_components/tuya_local/devices/minib_public_litterbox.yaml
```diff
@@ -107,6 +109,8 @@ entities:
       - id: 103
         type: integer
         name: sensor
+        unit: times
+        precision: 0
```
make-all: nor here

custom_components/tuya_local/devices/minib_public_litterbox.yaml
```diff
@@ -130,7 +134,7 @@ entities:
           - dps_val: "Standby"
             value: Standby
```
make-all, line 133: all the values in this list should be using translations from under the [status key](https://github.com/make-all/tuya-local/blob/984658391688430dcc72181a094c9227f0aa8f39/custom_components/tuya_local/translations/en.json#L927).

Permanently: Sorry, gonna have to ask for clarification- what do you mean here? Clumping/clustering isn't mentioned under the status key in the linked file, and everything else here has been left untouched

make-all: `caking` I think is equivalent

make-all: I see there is a lot here, which is maybe why I let this slip through in the original PR. You don't need to worry about this. I will look at myself it later, as several new translation strings are probably needed here.

Permanently: ~Okay dope; shall I mark as resolved for now?~ nvm lol

PR #4376, opened 2026-01-23
Improved Luko Dehumidifer by adding additional DPs and sensors.

cemilbrowne:
Added additional DPIDs to the existing Luko Dehumidifer model (I wrote the original).

### DIFF
--- a/custom_components/tuya_local/devices/luko_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/luko_dehumidifier.yaml
@@ -55,19 +55,20 @@ entities:
           - dps_val: "1"
             value: "circulate"
   - entity: number
-    name: Temperature Set
+    class: temperature
     category: config
     dps:
       - id: 45
         name: value
         type: integer
+        unit: C
         range:
           min: 50
           max: 350
         mapping:
           - scale: 10
   - entity: valve
-    name: Fresh Air Valve
+    name: Fresh air
     category: config
     dps:
       - id: 109


custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -45,8 +45,44 @@ entities:
             value: 80
           - dps_val: "5speed"
             value: 100
+  - entity: select
```
make-all: Make this the preset_mode of the fan entity. Add `translation_only_key: ventilation` (only is to avoid changing the entity id since it is already released without any translation key or name). And for the values use "fresh" and "circulate".

custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -45,8 +45,44 @@ entities:
...
+    category: config
+    dps:
+      - id: 109
+        name: value
+        type: integer
+        range:
+          min: 1
+          max: 99
   - entity: sensor
+    name: Temperature
```
make-all: Name should not be added. I don't think diagnostic category is needed either.

custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -45,8 +45,44 @@ entities:
...
+      - id: 45
+        name: value
+        type: integer
+        range:
+          min: 50
+          max: 350
+        mapping:
+          - scale: 10
+  - entity: number
+    name: Fresh Air Valve
```
make-all: You can make this a valve entity. `class: gas` might affect the icon or other UI appearance, but I am not sure if it will be appropriate where the gas in question is fresh air, probably it is meant more for flamable gases.  dp name below should also be valve in this case.

custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -55,3 +91,36 @@ entities:
...
+  - entity: sensor
+    name: Absolute Humidity
+    class: absolute_humidity
+    category: diagnostic
+    dps:
+      - id: 107
+        name: sensor
+        type: integer
+        class: measurement
+        mapping:
```
make-all, line 99: There should be a unit here. absolute_humidity should support g/m³ and mg/m³. I think without a supported unit, HA will raise a warning, in which case removing the class will be needed if the unit does not match one of those, or is unknown.

make-all: To help figure out the unit, absolute humidity of 100% relative humidity at 20 degrees C, is 17.29 g/m³, or 17290 mg/m³

custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -55,3 +91,36 @@ entities:
         class: measurement
         mapping:
           - scale: 10
+  - entity: sensor
+    name: Absolute Humidity
```
make-all: This name should be left out so the class translation is used

cemilbrowne:
I think I've made all of those changes; thank you for the feedback.

custom_components/tuya_local/devices/luko_dehumidifier.yaml
```diff
@@ -45,18 +45,15 @@ entities:
...
             value: 100
-  - entity: select
-    name: Circulation
-    category: config
-    dps:
       - id: 115
-        name: option
+        name: preset_mode
         type: string
+        translation_only_key: ventilation
```
make-all: This needs to go at entity level (where `class: ventilation` is).

cemilbrowne: Thank you - moved.

PR #4378, opened 2026-01-23
Change class key to translation_only_key for meaco dd8l dehumidifier

malrogers7:
Changes class to translation_only_key allowing modes to show with full names (e.g. 'laundry' becomes 'Dry clothes')

### DIFF
--- a/custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/meaco_dd8l_pro_dehumidifier.yaml
@@ -6,7 +6,7 @@ products:
 
 entities:
   - entity: humidifier
-    class: dehumidifier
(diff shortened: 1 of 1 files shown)


make-all:
The intention is to actually use the already existing class key from HA, which used to work when translations were first introduced.

I have made a different change to the translation files, which hopefully will resolve the issue without this change.

malrogers7:
Thanks. I have tried this (reinstating the original code and copying in your amended translation file (only enough.json)), but it doesn't seem to work - the 'laundry' mode has gone back to that, rather than 'Dry clothes'.
Thanks for your work on this integration, by the way. It's a lot to keep on top of and a great help.

make-all:
I have reverted to translation keys, but changed the name and updated all configs that extended the modes beyond what HA offers by default.

PR #4380, opened 2026-01-24
add device config for NUTROMO SH014 space heater

snicker:
bought one of these today
https://a.co/d/0SFytkC

there seem to be some very similar looking models branded differently, wouldn't surprise me if they are also functional with this config:
https://a.co/d/8Ck4Kwm
https://a.co/d/6J03wGS

tested and functional. there are a couple DPS I could not identify their purpose, they are included in the comments of the config.

### DIFF
--- a/custom_components/tuya_local/devices/nutromo_sh014_space_heater.yaml
+++ b/custom_components/tuya_local/devices/nutromo_sh014_space_heater.yaml
@@ -1,4 +1,4 @@
-name: NUTROMO SH014 Space Heater
+name: Space heater
 products:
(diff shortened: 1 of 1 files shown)


make-all:
You cannot assume something is generic unless there are multiple devices matching, as is the case for smartplugs and lightbulbs. Given how old this project is, and the number of heaters already supported it is extremely unlikely that you have stumbled on a Tuya standard dps mapping for heaters that is widespread among other heaters.

Please rename the file and update the product info. Based on the amazon page linked, manufacturer is Nutromo, and model is SH014.

snicker:
thanks @make-all  - PR is updated

PR #4385, opened 2026-01-24
Add device support eeese thor dehumidifier

burned89:


custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
```diff
@@ -0,0 +1,172 @@
+name: Thor Dehumidifier
```
make-all: Branding should only go under products, as there may be other devices that also match the config

custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
```diff
@@ -0,0 +1,172 @@
...
+products:
+  - id: kga5rstkass0sroi
+    manufacturer: eeese air care
```
make-all: Generally brand and model names should be capitalised, unless it is part of the branding

custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
```diff
@@ -0,0 +1,172 @@
...
+      - id: 2
+        name: humidity
+        type: integer
+        range:
+          min: 35
+          max: 80
+      - id: 5
+        name: mode
+        type: string
+        mapping:
```
make-all, line 22: Try to use mode names (in the value fields) that are already supported by [HA](https://github.com/home-assistant/core/blob/dev/homeassistant/components/humidifier/strings.json#L65), or add `translation_key: extended` to the entity definition to use extra modes from [this integration](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/en.json#L329)

custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
```diff
@@ -0,0 +1,172 @@
...
+          - dps_val: 21h
+            value: 21h
+          - dps_val: 22h
+            value: 22h
+          - dps_val: 23h
+            value: 23h
+          - dps_val: 24h
+            value: 24h
+  - entity: number
+    category: config
```
make-all: Based on the name, this should be a sensor I think. In that case, `translation_key: time_remaining`  should be used instead of the name, and the range is not needed.

Whether it is a sensor or number, `class: duration` should be added.

custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
```diff
@@ -0,0 +1,172 @@
...
+      - id: 19
+        type: bitfield
+        name: sensor
+        optional: true
+        mapping:
+          - dps_val: 1
+            value: true
+          - value: false
+  - entity: switch
+    name: Beep Switch
```
make-all: Just "Beep" is better. The entity is a switch, so it does not need that in the name.

custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
burned89: I changed it to sensor, but now its not working anymore. (Already deleted the device an added again). Do you see whats wrong?
image

custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
burned89: I changed it to Beep. Originally in the tuya app it's called also "Beep Switch".

custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
burned89: I changed it to the HA strings

custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
burned89: I capitalized it and removed "Thor" from the name.

custom_components/tuya_local/devices/eeese_thor_dehumidifier.yaml
make-all: category needs to be diagnostic (or removed) for a sensor

burned89: it works now - thank you :)

PR #4391, opened 2026-01-26
Add Vevor YT60307 Weather Station to devices

jvtcl:
- Tested in Home Assistant
- Wind direction is not directly available as a data point and must be fetched from the device status (DP 113) instead


custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
```diff
@@ -0,0 +1,125 @@
+name: Weather Station
+products:
+  - id: 1roznvzhxv5kqyg1
+  - manufacturer: Vevor
```
make-all: This should be a single list item.

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
```diff
@@ -0,0 +1,125 @@
...
+  - model: YT60307
+entities: 
+  - entity: sensor
+    name: Outdoor temperature
```
make-all: Since this is a weather station, outdoor temperature should be the default (unnamed so the HA default class translations are used). Same for humidity below.

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
```diff
@@ -0,0 +1,125 @@
...
+    dps:
+      - id: 102
+        type: integer
+        name: sensor
+        unit: mm
+        class: measurement
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: Atmospheric pressure
```
make-all: Name the same as class should be left out so the class translation is used.

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
```diff
@@ -0,0 +1,125 @@
...
+    dps:
+      - id: 121
+        type: integer
+        name: sensor
+        unit: hPa
+        class: measurement
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Wind speed
```
make-all: likewise

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
```diff
@@ -0,0 +1,125 @@
...
+    dps:
+      - id: 112
+        type: integer
+        name: sensor
+        unit: kph
+        class: measurement
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: Wind gust
```
make-all, line 57: Leave out the name here too, the difference between gust and speed is less important than having translations.

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
```diff
@@ -0,0 +1,125 @@
...
+    dps:
+      - id: 132
+        type: integer
+        name: sensor
+        unit: kph
+        class: measurement
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: Light intensity
```
make-all: Leave out the name, but the proper class to use is `illuminance`. It only supports `lx` as units, so you need to scale in the other direction (`scale: 0.1`)

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
```diff
@@ -0,0 +1,125 @@
...
+      - id: 115
+        type: integer
+        name: sensor
+        unit: klux
+        class: measurement
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: UV index
+    class: uv_index
```
make-all: There is no uv_index class.

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
```diff
@@ -0,0 +1,125 @@
...
+    dps:
+      - id: 116
+        type: integer
+        name: sensor
+        unit: uvi
+        class: measurement
+        mapping:
```
make-all: Scale of 1 is the default, so this mapping can be removed

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
```diff
@@ -0,0 +1,125 @@
...
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    class: wind_direction
+    dps:
+      - id: 113
+        type: base64
+        optional: true
+        name: sensor
+        mask: "000000000000000000000000000000000000000000000000000000000000000000000000000000000000FFFF0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
make-all: Mask can be split across lines by using `\` at the
end of the line without closing the quotes. Try to split on an even boundary to make following the hex digits easier.

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
jvtcl: To clarify, does this mean to remove the `name` item for this entity altogether?

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
jvtcl: Do you mean that as the first entry, it will be recognized as the default? Is the only action needed here to remove the name?

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
jvtcl: Is it ok to leave this intact after removing the `class` item? I adapted this from [another weather station](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml#L329-L337) in the devices list.

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
jvtcl: I think the name needs to stay in place here. The linter is detecting duplicate entries for the `wind_speed` class. For [other weather stations](https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/immax_neolite_weatherstation.yaml#L278-L300), the second `wind_speed` entity for gust speed is named.

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
make-all: I was saying to remove the name, but you removed the class instead. You should have left the class and removed the name, as the class comes with translations, but a hard-coded name does not.

custom_components/tuya_local/devices/vevor_yt60307_weather_station.yaml
make-all: I must not have been paying attention here. The name is required here to differentiate it from wind speed, which should be the one with the name left off.

PR #4392, opened 2026-01-26
6 Additional Devices

demitryk1:
These are configs for 6 additional devices.

Elegrp DTR10/30 Smart Dimmer Switch
Elegrp SSS10/30 Smart Sensor Switch 
Elegrp SPN10/30 Smart Switch
Minoston Timer Switch
Ohmax  Smart Switch
Breathe Air Monitor +


make-all:
It is usually better not to bundle devices together in a single PR like this, as they will all be delayed until all the configs are passing review.

As it is, 4 of the 6 configs are in the wrong location.

make-all:
Closing, as submitter is non-responsive

Issue #4396, opened 2026-01-26
Request support for Andalung Portable EV Charger


nathan-gs:
### Log message

```text
Device matches afyeev_16a_evcharger with quality of 92%. DPS: {"updated_at": 1769447795.8332334, "3": "charger_free", "4": 6, "9": 552, "10": 0, "13": "controlpi_12v", "14": "charge_now", "15": 0, "16": false, "18": false, "23": "HW:1.0, SW:1.6.3", "24": 10, "27": "online"}
```

### Product ID

cxriym8xwbnj4xdl

### Product Name

Andalung Portable EV Charger

### DPS information

```text
{
  "result": {
[...]

PR #4397, opened 2026-01-26
Andalung EV Charger - closing #4396

nathan-gs:
Closing #4396


make-all:
This device is the same as the already supported evsun_evcharger.yaml

I have merged some changes into that as improvements

PR #4398, opened 2026-01-26
feat (devices): add Klarstein WonderSky Smart 360W ACO14

er2er:
Add support for [Klarstein WonderSky Smart 360W ACO14](https://www.klarstein.pl/Wondersky-panel-na-podczerwien-360-W-4-8-m-Minimalistyczny-design-60-cm-x-60-cm-360-W.html).

### Logic explanation for Light Entity

The device uses a combined logic for the ambiance light control using a single Enum DP (`102`) instead of separate `switch` and `work_mode` DPs.

* **DP 102 (Enum):**
    * `"0"`: Light **OFF**
    * `"1"`: Light **ON** + **Cold White**
    * `"2"`: Light **ON** + **Warm White**

In the YAML configuration, this is handled by mapping DP `102` twice within the `light` entity:

1.  As a Switch:
    * Maps `"0"` to `off` (`false`).
    * Maps both `"1"` and `"2"` to `on` (`true`).

2.  As Color Temperature:
    * Maps `"1"` to 6500 Kelvin.
    * Maps `"2"` to 2700 Kelvin.
    * This allows the Home Assistant Color Temp slider to "snap" to the nearest supported mode (Cold or Warm) when adjusted.

3. Note that it's not possible to turn the light in warm color with light.turn_on unless an optional entity for setting DP 102 directly is mae available.

### Brightness Logic

* **DP 101 (Enum):**
    * The device does not use a standard Integer range for brightness.
    * It uses an Enum range from `"0"` (minimum nonzero brightness) to `"8"` (max brightness).
    * These Enum values are mapped to specific discrete values on the Home Assistant brightness scale.

----

Device Debug Infor from Tuya IoT:
```
switch: Boolean "{true,false}"
temp_set: Integer {
  "unit": "°C",
  "min": 5,
  "max": 50,
  "scale": 0,
  "step": 1
}
temp_current: Integer {
  "unit": "℃",
  "min": 0,
  "max": 50,
  "scale": 0,
  "step": 1
}
eco: Boolean "{true,false}"
work_state: Enum {
  "range": [
    "heating",
    "standby",
    "timer"
  ]
}
countdown_set: Enum {
  "range": [
    "cancel",
    "1h",
    "2h",
    "3h",
    "4h",
    "5h",
    "6h",
    "7h",
    "8h",
    "9h",
    "10h",
    "11h",
    "12h",
    "13h",
    "14h",
    "15h",
    "16h",
    "17h",
    "18h",
    "19h",
    "20h",
    "21h",
    "22h",
    "23h",
    "24h"
  ]
}
countdown_left: Integer {
  "unit": "min",
  "min": 0,
  "max": 1440,
  "scale": 0,
  "step": 1
}
dg: Enum {
  "range": [
    "0",
    "1",
    "2",
    "3",
    "4",
    "5",
    "6",
    "7",
    "8"
  ]
}
ledmode: Enum {
  "range": [
    "0",
    "1",
    "2"
  ]
}
```

DP Mapping:

| ID | Code | Type | Description |
| :--- | :--- | :--- | :--- |
| 1 | `switch` | Boolean | HVAC Power (On/Off) |
| 2 | `temp_set` | Integer | Target Temperature (5-50°C) |
| 3 | `temp_current` | Integer | Current Room Temperature |
| 6 | `eco` | Boolean | Preset Mode (False: Normal, True: Eco) |
| 11 | `work_state` | Enum | Heating Status (`heating`, `standby`, `timer`) |
| 19 | `countdown_set` | Enum | Timer Selection (`cancel`, `1h`...`24h`) |
| 20 | `countdown_left` | Integer | Time Remaining (minutes) |
| 101 | `dg` | Enum | Light Brightness (`0`=10% ... `8`=100%) |
| 102 | `ledmode` | Enum | Light Mode (`0`: Off, `1`: Cold, `2`: Warm) |

### DIFF
--- a/custom_components/tuya_local/devices/klarstein_wondersky_heater.yaml
+++ b/custom_components/tuya_local/devices/klarstein_wondersky_heater.yaml
@@ -1,4 +1,4 @@
-name: Infrared Heater with Ambient Light
+name: IR heater
 products:
   - id: canayin50j2qfgml
     manufacturer: Klarstein


custom_components/tuya_local/devices/klarstein_wondersky.yaml
```diff
@@ -0,0 +1,170 @@
+name: Klarstein Wondersky IR Heater with Ambient Light
```
make-all: Top level name should be generic, in case other products match the same config.

custom_components/tuya_local/devices/klarstein_wondersky.yaml
```diff
@@ -0,0 +1,170 @@
...
+products:
+  - id: canayin50j2qfgml
+    manufacturer: Klarstein
+    model: Wondersky Smart 360W
+    model_id: ACO14
+entities:
+  - entity: climate
+    translation_key: radiator
```
make-all: This translation key is not defined, use `heater`

custom_components/tuya_local/devices/klarstein_wondersky.yaml
```diff
@@ -0,0 +1,170 @@
...
+        name: hvac_action
+        mapping:
+          - dps_val: "standby"
+            value: "idle"
+          - dps_val: "heating"
+            value: "heating"
+          - dps_val: "timer"
+            value: "heating"
+  - entity: light
+    translation_key: light
```
make-all: This translation key is not defined. Remove it.

custom_components/tuya_local/devices/klarstein_wondersky.yaml
```diff
@@ -0,0 +1,170 @@
...
+  - entity: select
+    translation_key: timer
+    category: config
+    dps:
+      - id: 19
+        type: string
+        name: option
+        mapping:
+          - dps_val: cancel
+            value: "off"
```
make-all: cancel is the key used by the timer translation

custom_components/tuya_local/devices/klarstein_wondersky.yaml
```diff
@@ -0,0 +1,170 @@
...
+            value: "20h"
+          - dps_val: "21h"
+            value: "21h"
+          - dps_val: "22h"
+            value: "22h"
+          - dps_val: "23h"
+            value: "23h"
+          - dps_val: "24h"
+            value: "24h"
+  - entity: select
```
make-all: This is duplicating controls already available in the light so should not be needed.

custom_components/tuya_local/devices/klarstein_wondersky_heater.yaml
```diff
@@ -0,0 +1,170 @@
...
+    dps:
+      - id: 102
+        type: string
+        name: switch
+        mapping:
+          - dps_val: "0"
+            value: false
+          - dps_val: "1"
+            value: true
+          - dps_val: "2"
```
make-all, line 57: One of these should be marked `hidden: true`, otherwise turning the light on does not have well defined behaviour.

custom_components/tuya_local/devices/klarstein_wondersky.yaml
er2er: Thanks for reviewing, I've removed this entity as requested.
What I was trying to achieve with this extra entity was to make it possible to _turn the light on in warm color_.
Currently, with `light.turn_on` specifying the temperature of 2700K, the current implementation would still set DP 102 to "1" (cold).
Not sure how to achieve the same without this optional (hidden) entity, but it's fine to evolve this specification over time.

make-all: I think that can be fixed in the light implementation. I didn't think of lights using the same dp for on and off as another function at the time I originally wrote it.

PR #4423, opened 2026-01-31
added support for negative temperatures and humidity fot the probe 1

gaxkalik:
The temperetures for the sensors were unsigned and displayed wrong
Also I found the humidity mask (I have tested all the changes)

### DIFF
--- a/custom_components/tuya_local/devices/inkbird_ibs_m1s.yaml
+++ b/custom_components/tuya_local/devices/inkbird_ibs_m1s.yaml
@@ -29,6 +29,25 @@ entities:
         class: measurement
         endianness: little
         mask: "FFFF00"
+        mask_signed: true
+        mapping:
+          - scale: 10
+            constraint: type
+            conditions:
+              - dps_val: 0
+                invalid: true
+  - entity: sensor
+    name: Humidity probe 1
+    class: humidity
+    dps:
+      - id: 114
+        optional: true
+        type: base64
+        name: sensor
+        unit: "%"
+        class: measurement
+        endianness: little
+        mask: "FFFF000000"
         mapping:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/inkbird_ibs_m1s.yaml
```diff
@@ -29,6 +29,25 @@ entities:
...
         mask: "FFFF00"
+        mask_signed: true
+        mapping:
+          - scale: 10
+            constraint: type
+            conditions:
+              - dps_val: 0
+                invalid: true
+  - entity: sensor
+    name: Humidity probe 1
```
make-all, line 40: Does this also apply to other probes?
Is the humidity sensing probe optional? In which case this should be hidden if unavailable?

gaxkalik: The gateway supports multiple types of devices and not all of them can measure the humidity, so it should be optional. Also, the gateway supports connection up to 8 devices and all of them can have humidity values. The problem is that I can't test the masks for all possible devices as I don't have them but the idea is to shift the mask by 20 characters (add 20 zeros at the end of previous mask).

make-all: Do you know the value it puts in the humidity field when the probe does not support humidity? I think this can be used to hide the probes sensors that are not supported by default.

gaxkalik: I am getting this string (in raw hex format) from the gateway that has no device "ffffffffffffffff0000". So I guess it should be something like "null" but I am not sure

PR #4434, opened 2026-02-02
Adding support for Konyks eCosy 

NarfX4:
Devices are wire pilot control for heaters, quite French specific. 
The DP mapping is quite strange and I spent some time debugging it. It's been tested live at home with 5 eCosy running.


NarfX4:
What needs to be done to progress on that request?

make-all:
Sorry, I didn't notice the test failures were timeouts due to GitHub infrastructure issues, not actual errors in the submission.
Due to high volume of requests lately I have been only able to focus on PRs that are ready and skipped over the ones with errors indicated.

custom_components/tuya_local/devices/konyks_ecosy_2024.yaml
```diff
@@ -0,0 +1,28 @@
+name: eCosy
+products:
+  - id: evyy1wbhi4t7uftn
+    model: "eCosy 2024"
+entities:
+  - entity: climate
```
make-all: There is no point to use a climate entity here, as there is no temperature and current_temperature nor hvac_mode (though the below switch should probably be hvac_mode).

Is there really only 2 dps for this device?

custom_components/tuya_local/devices/konyks_ecosy_2024.yaml
```diff
@@ -0,0 +1,28 @@
+name: eCosy
```
make-all: Top level name should be generic, not a specific model name, as other devices may also match this config.
The device type should also be at the end of the filename to assist users to understand what they are selecting.

(konyks_ecosy2024_heater.yaml for example)

custom_components/tuya_local/devices/konyks_ecosy_2024.yaml
```diff
@@ -0,0 +1,28 @@
...
+products:
+  - id: evyy1wbhi4t7uftn
+    model: "eCosy 2024"
```
make-all: `manufacturer: Konyks` seems to be missing here.

custom_components/tuya_local/devices/konyks_ecosy_2024.yaml
```diff
@@ -0,0 +1,28 @@
...
+          - dps_val: hot
+            value: "Confort"
+          - dps_val: comfortable1
+            value: "Confort -1"
+          - dps_val: comfortable2
+            value: "Confort -2"
+          - dps_val: eco
+            value: "Eco"
+          - dps_val: cold
+            value: "Hors gel"
```
make-all: Is it anti-frost?

In general you should use presets that are defined in 
https://github.com/home-assistant/core/blob/dev/homeassistant/components/climate/strings.json#L151
or use a translation_key for extended presets from 
https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/translations/fr.json#L164

custom_components/tuya_local/devices/konyks_ecosy_2024.yaml
NarfX4: Actually, I tried all combinations and only the climate entity works with indeed only 2 DPs.

custom_components/tuya_local/devices/konyks_ecosy_2024.yaml
NarfX4: OK so we should use pilotwire. So far, I haven't seen any other device doing that.

custom_components/tuya_local/devices/konyks_ecosy_2024.yaml
NarfX4: Indeed, sorry for that.

PR #4436, opened 2026-02-03
Add support for Topgreener Smart Round Plug-In (TGWF115PRM)

d-allin:
https://topgreener.com/product/round-smart-wi-fi-plug-in-10a-with-energy-monitoring-white/

Definitions from tuya currently missing scale=1 for cur_power and cur_voltage (see https://github.com/home-assistant/core/issues/152295) so scaling by 10 in yaml.

`[{"id": 1, "name": "switch_on", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 2, "name": "countdown", "type": "Integer", "format": "{\"unit\":\"\",\"min\":0,\"max\":86400,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 4, "name": "cur_current", "type": "Integer", "format": "{\"unit\":\"mA\",\"min\":0,\"max\":30000,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 5, "name": "cur_power", "type": "Integer", "format": "{\"unit\":\"W\",\"min\":0,\"max\":50000,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 6, "name": "cur_voltage", "type": "Integer", "format": "{\"unit\":\"V\",\"min\":0,\"max\":2500,\"scale\":0,\"step\":1}", "enumMap": {}}]`

As this is an upstream thing being corrected in a dirty way downstream via this method, it may be preferable to leave the scaling alone and have users correct with [template sensors](https://github.com/home-assistant/core/issues/152295#issuecomment-3324154167). Otherwise I'll update the scaling when (if) Tuya ever fixes the problem on their end.


make-all:
It is OK to put correct scaling in if you are certain it is incorrect in the documentation. The actual scaling seems consistent with other smartplugs, so any "correction" in future is likely to be to the documentation rather than the firmare.

custom_components/tuya_local/devices/topgreener_smartroundplugin.yaml
```diff
@@ -0,0 +1,54 @@
+name: Smart Round Plug-In
+products:
+  - id: bWBb4e6BTZGie0Lv
```
make-all, line 3: Is this the product id? It looks too long, and maybe a device id for your specific device?

d-allin: The `id` specified (bWBb4e6BTZGie0Lv) corresponds to the `product_id` reported by apigw.tuyaus.com

```
    "product_id": "bWBb4e6BTZGie0Lv",
    "product_name": "Smart Round Plug-In",
```

This same value is also used as the `pid` throughout, so I assume this to be correct.

The id itself (that I assume is specific for my device) is longer, at twenty characters.

make-all:
This is the same as already existing smartplugv1

PR #4443, opened 2026-02-03
Add F32 Thermostat configuration

dvismans:
Adding configuration for F32 Thermostat (Product ID: enasdcyucyl7tgrz). Includes climate entity with C/F switching, child lock, and diagnostic sensors.


custom_components/tuya_local/devices/et_f32_thermostat.yaml
```diff
@@ -0,0 +1,99 @@
...
+        type: string
+        name: temperature_unit
+        mapping:
+          - dps_val: c
+            value: C
+          - dps_val: f
+            value: F
+
+  - entity: lock
+    name: Child lock
```
make-all: use `translation_key: child_lock` instead of explicit name

custom_components/tuya_local/devices/et_f32_thermostat.yaml
```diff
@@ -0,0 +1,99 @@
...
+    category: config
+    dps:
+      - id: 40
+        type: boolean
+        name: lock
+
+  - entity: select
+    name: Temperature unit
```
make-all: Use `translation_key: temperature_unit` instead of explicit name. and icon, and make options lower-case to match the corresponding keys

custom_components/tuya_local/devices/et_f32_thermostat.yaml
```diff
@@ -0,0 +1,99 @@
...
+        type: string
+        name: option
+        mapping:
+          - dps_val: c
+            value: Celsius
+          - dps_val: f
+            value: Fahrenheit
+
+  - entity: binary_sensor
+    name: Problem
```
make-all: Remove the explicit name to use HA`s class translations.

custom_components/tuya_local/devices/et_f32_thermostat.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - id: 45
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+
+  - entity: button
+    name: Factory reset
```
make-all: Use `translation_key: factory_reset` instead of this name, remove the class (not a perfect match and the translation key does most of the same UI customizations) and add `hidden: true`, as most users will not want this exposed.

custom_components/tuya_local/devices/et_f32_thermostat.yaml
```diff
@@ -0,0 +1,99 @@
+name: F32 Thermostat
```
make-all: Remove model name from top level name, as the config may also cover other devices.

custom_components/tuya_local/devices/et_f32_thermostat.yaml
```diff
@@ -0,0 +1,99 @@
...
+products:
+  - id: enasdcyucyl7tgrz
+    name: F32
```
make-all: Remove this, as it is the model (already stated below)

PR #4459, opened 2026-02-06
Add Support for Smart 3D Oscillating DC Fan (Kogan)

angus-thompson:
Added configuration for Smart 3D Oscillating DC Fan including entities for fan control, modes, oscillation, and timer settings.

This is a relatively cheap Tuya based standard pedastal fan in Australia from Kogan [link](https://www.kogan.com/au/buy/kogan-smarterhometm-premium-rechargeable-3d-oscillating-dc-fan-white-kogan/). It has both vertical and horizontal oscillation, vertical oscillation defined by DP 6: Run/Stop, and Horizontal by DP 7 as sweep angle in degrees. Speed is DP 3 as a value between 1 to 10. It also has various modes, which could be incorporated into the fan entity as preset modes, however these unfortunately do not appear in homekit well due to limitations, so for now I have them as a separate entity (and default hidden this as they aren't particularly useful). Noted however this config uses "3D Mode" (vertical and horizontal oscillation) as the oscillate attribute as I think this is more appropriate than using just vertical oscillation, or a set angled of horizontal oscillation as this binary value. Finally it has a timer function, I have included this but also default hidden it as timers in smart home products can be better triggered through automations.


custom_components/tuya_local/devices/smart_3d_oscillating_dc_fan.yaml
```diff
@@ -0,0 +1,114 @@
+name: Smart 3D Oscillating DC Fan
```
make-all: Top level name should be generic, as other models may also match

custom_components/tuya_local/devices/smart_3d_oscillating_dc_fan.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 3
+        type: integer
+        name: speed
+        range:
+          min: 1
+          max: 10
+      - id: 2
```
make-all, line 19: I don't think trying to map multiple modes into a boolean oscillate is a good idea, leave this out.

custom_components/tuya_local/devices/smart_3d_oscillating_dc_fan.yaml
```diff
@@ -0,0 +1,114 @@
...
+        type: string
+        name: oscillate
+        mapping:
+          - dps_val: "3Dmode"
+            value: true
+          - dps_val: normal
+            value: false
+
+  - entity: select
```
make-all: This can be the preset_mode of the fan (use `translation_key:: fan_with_presets` to pull in translations for most of the below)

custom_components/tuya_local/devices/smart_3d_oscillating_dc_fan.yaml
```diff
@@ -0,0 +1,114 @@
...
+          - dps_val: normal
+            value: Normal
+          - dps_val: nature
+            value: Nature
+          - dps_val: sleep
+            value: Sleep
+          - dps_val: auto
+            value: Auto
+          - dps_val: "3Dmode"
+            value: 3D Mode
```
make-all, line 42: Aside from 3Dmode, the original tuya dps_val correspond to translated preset_modes already, so can map to themselves. 3D Mode can remain to a string, as it seems like a marketing term that should not be translated, and `circulate` is only available with the ventilation translation key (which does not have the other modes).

custom_components/tuya_local/devices/smart_3d_oscillating_dc_fan.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+  - entity: select
+    name: Vertical Oscillation
```
make-all: HA naming convention is to only capitalise the first word of device and entity names.

custom_components/tuya_local/devices/smart_3d_oscillating_dc_fan.yaml
```diff
@@ -0,0 +1,114 @@
...
+            value: "60°"
+          - dps_val: "90"
+            value: "90°"
+          - dps_val: "120"
+            value: "120°"
+          - dps_val: "150"
+            value: "150°"
+
+  - entity: select
+    name: Timer
```
make-all: Use `translation_key: timer` instead of the explicit name and icon. Not sure why this is marked as hidden.

custom_components/tuya_local/devices/smart_3d_oscillating_dc_fan.yaml
```diff
@@ -0,0 +1,114 @@
...
+    icon: "mdi:timer-outline"
+    category: config
+    hidden: true
+    dps:
+      - id: 22
+        type: string
+        name: option
+        mapping:
+          - dps_val: cancel
+            value: "Off"
```
make-all: When using the translation_key, map these to the same string as original tuya values.

custom_components/tuya_local/devices/smart_3d_oscillating_dc_fan.yaml
angus-thompson: I haven't had any issues with it. Mostly using this because otherwise presets in homekit don't display well. I find if i want to oscillate this fan, I would like it to oscillate vertically and horizontally, there are no DPS that will let me do that

PR #4462, opened 2026-02-06
Custom cover state fix + light unavailable -> off

Itstommy10:


make-all:
An extensive PR like this needs an explanation

PR #4469, opened 2026-02-07
Add Airton Aircon device

FinweVI:
### Description

Adds support for the [Airton Aircon](https://airton.shop/products/climatiseur-fixe-reversible).

Tested only on the Monosplit 2500W, but I suspect they will all function the same.

### Device details

-  Manufacturer: Airton
-  Model: 409730
-  Product ID: keyquxnsj75xc8se
-  Category: HVAC

### DIFF
--- a/custom_components/tuya_local/devices/airton_aircon.yaml
+++ b/custom_components/tuya_local/devices/airton_aircon.yaml
@@ -98,25 +98,23 @@ entities:
       - id: 101
         name: use_number
         type: integer
-        hidden: true
+        optional: true
       - id: 103
         name: electricity
         type: integer
-        hidden: true
         optional: true
       - id: 104
         name: electricity_number
         type: integer
-        hidden: true
         optional: true
       - id: 112
         name: unit_type
         type: string
-        hidden: true
+        optional: true
       - id: 106
         name: horizontal_swing
         type: string
-        hidden: true
+        optional: true
       - id: 107
         name: swing_mode
         type: string
@@ -138,25 +136,23 @@ entities:
       - id: 108
         name: swing3d
         type: boolean
-        hidden: true
+        optional: true
       - id: 111
         name: clean
         type: boolean
-        hidden: true
+        optional: true
       - id: 115
         name: heat8
         type: boolean
-        hidden: true
         optional: true
       - id: 21
         name: countdown
         type: string
-        hidden: true
         optional: true
       - id: 22
         name: countdown_left
         type: integer
-        hidden: true
+        optional: true
   - entity: light
     translation_key: display
     category: config
@@ -196,26 +192,18 @@ entities:
         name: sensor
         mapping:
           - dps_val: 0
-            value: false
+            value: true
+            constraint: fault_code_2
+            conditions:
+              - dps_val: 0
+                value: false
           - value: true
       - id: 20
         type: bitfield
         name: fault_code
-  - entity: binary_sensor
-    name: Problem 2
-    class: problem
-    category: diagnostic
-    dps:
       - id: 113
         type: bitfield
-        name: sensor
-        mapping:
-          - dps_val: 0
-            value: false
-          - value: true
-      - id: 113
-        type: bitfield
-        name: fault_code
+        name: fault_code_2
   - entity: sensor
     name: Total usage time
     icon: "mdi:clock-outline"


custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
+name: Air conditioner
+products:
+  - id: keyquxnsj75xc8se
+    name: Reversible Air Conditioner 2500W (409730)
```
make-all: make this model (shorten by removing the "Air conditioner" that is already covered in the name above)

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
...
+          - scale: 10
+      - id: 5
+        name: fan_mode
+        type: string
+        optional: true
+        mapping:
+          - dps_val: auto
+            value: auto
+          - dps_val: mute
+            value: Silence
```
make-all: quiet

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
...
+          - dps_val: low
+            value: low
+          - dps_val: low_mid
+            value: low-mid
```
make-all: medlow

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
...
+          - dps_val: mid
+            value: medium
+          - dps_val: mid_high
+            value: mid-high
```
make-all: medhigh

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
...
+          - dps_val: high
+            value: high
+          - dps_val: turbo
+            value: Turbo
```
make-all: strong

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
...
+            conditions:
+              - dps_val: false
+                value: none
+              - dps_val: true
+                value: sleep
+          - dps_val: true
+            value: eco
+            constraint: hvac_mode_switch
+            conditions:
+              - dps_val: cold
```
make-all: I think it is better to leave the cold mapping out of this list, as it may cause side-effects on the hvac_mode

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
...
+                value: eco
+              - dps_val: auto
```
make-all: These can be combined as a list
```
- dps_val: [auto, heat, wet, fan]
  invalid: true
```

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
...
+      - id: 2
+        name: temperature
+        type: integer
+        range:
+          min: 160
+          max: 300
+        mapping:
+          - scale: 10
+            step: 10
+            constraint: hvac_mode_switch
```
make-all, line 43: You need to adjust the range for temperature_unit `- dps_val: f`, so you need to drop this condition so you can fit that in instead.

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
...
+        type: string
+        mapping:
+          - dps_val: c
+            value: C
+          - dps_val: f
+            value: F
+      - id: 114
+        name: active_mode
+        type: string
+        hidden: true
```
make-all: hidden: true should not be used here, as this is not accessilbe anywhere else.

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -0,0 +1,183 @@
...
+        optional: true
+      - id: 9
+        name: unknown_9
```
make-all: Downloading the Query Things Data Model result from API Explorer in the Tuya developer portal should show what these are

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Done

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Done

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Done

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Done

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Done

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Done

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Done

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: My account was out of subscription since 2023.
I applied for an extension.

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Well, the ECO mode works *only* in Cold mode.
If I try to send the DPS in another mode, the device is reporting an error.
How should I handle that?

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: I tried to comply.
But I'm not even sure how to set my AC in Fahrenheit (no possibility on the remote), I think those are sold on the EU market only (at least with this brand).

I could try sending the `f` dps I guess.
Let me know if it needs other modifications or further research.

custom_components/tuya_local/devices/airton_aircon.yaml
make-all: The default value is already eco, you only need conditions for the exceptions

custom_components/tuya_local/devices/airton_aircon.yaml
make-all: You added a select entity for setting the temperature unit.

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: [airton_aircon_api_spec_en.json](https://github.com/user-attachments/files/25242088/airton_aircon_api_spec_en.json)

Attached what I could get (I had an LLM translate it, it was in Chinese initially).
Apparently it's a switch for the dry mode?

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: I tried to add a selector for that.
But, while the setting temp seems to work (the AC shows the temp in F instead of C), the current temp doesn't seem to be anything else than C.
I'm not sure if it's worth it to try to force Fahrenheit on this device.

FinweVI: I removed the possibility to use Fahrenheit, since it's not consistent across sensors.

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Based on the API Spec, I was able to find some new sensors/settings, including this one.
Some, however, seem to be optional, or present but with no effect.
My latest commit represent that and brings only the ones that made sense for me (and were available).

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -105,61 +105,40 @@ entities:
       - id: 105
         name: temperature_unit
         type: string
-        mapping:
-          - dps_val: c
-            value: C
-          - dps_val: f
-            value: F
+        hidden: true
```
make-all: hidden only hides from the extra_attributes list, but since temperature_unit is a standard attribute of the climate entity, it will still be used. But now without mapping, which will leave invalid units.

It is better to hardcode the units on temperatre dp, and remove this if you found that unit switching does not work (or is too difficult to support)

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -105,61 +105,40 @@ entities:
...
-        hidden: true
       - id: 101
-        name: unknown_101
+        name: use_number
         type: integer
         hidden: true
-      - id: 102
-        name: unknown_102
+      - id: 103
+        name: electricity
```
make-all, line 103: Probably should be a power or energy sensor (depending on what it is actually reporting)

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -176,3 +155,111 @@ entities:
...
+        optional: true
+  - entity: switch
+    name: Self-clean
+    icon: "mdi:shimmer"
+    category: config
+    dps:
+      - id: 111
+        name: switch
+        type: boolean
+  - entity: select
```
make-all: Should be moved to climate entity `swing_horizontal_mode`

custom_components/tuya_local/devices/airton_aircon.yaml
```diff
@@ -176,3 +155,111 @@ entities:
...
+        name: option
+        mapping:
+          - dps_val: "off"
+            value: "Off"
+          - dps_val: same
+            value: Focus
+          - dps_val: opposite
+            value: Diffuse
+  - entity: select
+    name: Vertical swing
```
make-all: Should be moved to climate `swing_mode`

FinweVI: The Horizontal swing, despite being in the DPS, is not working on the device as there's no vanes for it.
I just used my vertical swing as the swing mode.

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: It's supposed to be the accumulated amount of consumed power, but it's not reporting anything when I try it, so I just moved it to hidden.

custom_components/tuya_local/devices/airton_aircon.yaml
FinweVI: Just hardcoded it to C.

Axilane:
Hello. First of all thank you so much for your work. 

In have tried the driver with my 2 Airton heat pumps. They work very well . I only regret 2 missing feature support. 
The first one is the vertical management of the air flow. It can be one of the  five vertical position from 1 to 5 or the auto travel of the blades.
The second missing is the speed of the air flow from Auto to  the slow, medium, high etc ...
These are important features regarding the comfort of use or the heat pump, which might be very nice to have, if you have a little of Time to implement in the Airton driver. 
I thank you in advance . 
Fabrice.

make-all:
These features are implemented in the config. Horizontal airflow is not, as the values are not known.

Axilane:
Hello Jason,

Thank you very much for your so prompt answer.
Sorry, you are right I can configure these features manually.
My remark was not accurate enough.
I was speaking about the individual command set in order to remoterly 
configure the heat pump from an automation or a script.

Kind regards,
Fabrice




------ Message d'origine ------
De "Jason Rumney" ***@***.***>
À "make-all/tuya-local" ***@***.***>
Cc "Axilane" ***@***.***>; "Comment" 
***@***.***>
Date 18/02/2026 23:06:11
Objet Re: [make-all/tuya-local] Add Airton Aircon device (PR #4469)

make-all:
What is missing for automations and scripts?
https://www.home-assistant.io/integrations/climate/#climate-control-actions

Axilane:
Hello Jason,

Right now I"m using the "sensei_airconditioner" driver which allow manual control but  no programable control of these 2 features also. . 

regarding the air flow control with the Airton Heat Pump I do have these options :
 "fan_modes": [
                "auto",
                "quiet",
                "low",
                "medlow",
                "medium",
                "medhigh",
                "high",
                "strong"
Regarding the vertical movement or position I do have this : 
              "options": [
                "Off",
                "Swing",
                "1",
                "2",
                "3",
                "4",
                "5"
Thanks for all your help and support. 
Kind regards, Fabrice

PR #4477, opened 2026-02-09
Add support to Ketotek KTF017802 Thermostat

marco-calautti:
This time around the yaml should have no issues, and should implement the thermostat properly, fixing #4474

### DIFF
--- a/custom_components/tuya_local/devices/pni_ct45_thermostat.yaml
+++ b/custom_components/tuya_local/devices/pni_ct45_thermostat.yaml
@@ -90,7 +90,7 @@ entities:
       - id: 1
         type: boolean
         optional: true
-        name: operating_mode
+        name: operation_mode
         mapping:
           - dps_val: false
             value: "off"
@@ -130,6 +130,7 @@ entities:
         optional: true
         name: min_temperature
   - entity: select
+    name: Output
     translation_key: mode
     category: config
     dps:
@@ -150,6 +151,17 @@ entities:
         type: boolean
         optional: true
         name: switch
+      - id: 28
+        name: brightness
+        type: string
+        optional: true
+        mapping:
+          - dps_val: low
+            value: 85
+          - dps_val: middle
+            value: 170
+          - dps_val: high
+            value: 255
   - entity: valve
     dps:
       - id: 7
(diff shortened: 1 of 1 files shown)


marco-calautti:
@make-all I guess the YAML tests failure is unrelated to my pull request, right?

marco-calautti:
I think I also found two additional DPS that contain the weekly program. That is a base64 string that when decoded goves a sequence of hex values of the form:
[HH][MM][TT] ... Meaning that from HH:MM, the temperature is set to T. How can I make the YAML do that conversion?

make-all:
It depends on the exact format whether this is feasible without external logic. If it is fixed length, and the numbers encoded as integers, not BCD or ASCII, then a time entity for hour and minute, and number entity with mask on the individual dps might work. If more complex deciding is needed, then use a text entity with base64 type value, so the decoding can be layered on top of this integration.

marco-calautti:
The format Is quite simple. Fixed length, 3 bytes each entry. One byte per value. So, it is just a matter of decoding the base64, and take the integer value of each byte.

The total payload is also of fixed length: 9 bytes, so 3 entries overall.

marco-calautti:
Something like this creates the right entities in home assistant, but they are not filled with anything, and if I try to set some value, I get an error "Cannot convert NoneType object to bytes":
```yaml
- entity: time
    category: config
    name: "Workday Part 1 Time"
    icon: "mdi:calendar"
    dps:
      - id: 119
        type: base64
        name: hour
        mask: "FF0000000000000000"
        range:
          min: 0
          max: 23
      - id: 119
        type: base64
        name: minute
        mask: "00FF00000000000000"
        range:
          min: 0
          max: 59
- entity: number
  category: config
  class: temperature
  name: "Workday Part 1 Temp."
  icon: "mdi:thermometer-plus"
  dps:
    - id: 119
      type: base64
      name: value
      unit: °
      mask: "0000FF000000000000"
      range:
        min: 5
        max: 70
```

marco-calautti:
Even the simple:
```yaml
  - entity: text
    category: config
    name: "Test"
    icon: "mdi:calendar"
    dps:
      - id: 119
        type: string
        name: value
```
does not work. I get "unknown" in the text box. But I can read this dps using the tuya platform:
```json
{
  "result": {
    "properties": [
      {
        "code": "Prog_Workday1",
        "custom_name": "",
        "dp_id": 119,
        "time": 1770709283471,
        "type": "raw",
        "value": "BgAUCAAPCx4P"
      }
    ]
  },
  "success": true,
  "t": 1770730766840,
  "tid": "eafc5624068511f18255b6b5e3717ca9"
}
```

make-all:
It seems they are cloud only then

marco-calautti:
Ok, but then I do not think there is anything else I can add to the yaml, if you want to merge it.

make-all:
Merged into Hysen HT101RF, as discussed on #4468

PR #4480, opened 2026-02-10
feat(eurom_wifi_thermostat): Add temperature sensor

filipekiss:
I recently setup this Wifi Thermostat and I wanted to use it paired with a threshold sensor to control multiple heaters at my living room, but the device does not expose the temperature as a sensor, only as a climate entity.

This exposes the `current_temperature` as a temperature sensor like the `eurom_sani_bathroom_towel_radiator` does.

CleanShot 2026-02-10 at 13 15 50@2x


make-all:
There is no need for such a change. If you want a separate sensor you can use a template sensor helper. The reason the Eurom Sani bathroom towel radiator has a separate sensor is that the climate current temperature depends on which mode the device is in.

PR #4482, opened 2026-02-10
Add GPH-DA-LX / Weber heat 2000W support

w-itidea:


make-all:
I don't understand what you've done with the presets here. What are the actual values used by the heater? You only need to cover those, not all possible spelling and case variants. Unless you have multiple heaters that are using different values?

custom_components/tuya_local/devices/gph_da_lx_panel_heater.yaml
```diff
@@ -0,0 +1,113 @@
+name: Heater
+products:
+  - id: bfd46bc528e265e3a37rkw
```
make-all, line 3: This appears to be a device id, not a product id

custom_components/tuya_local/devices/gph_da_lx_panel_heater.yaml
```diff
@@ -0,0 +1,113 @@
+name: Heater
+products:
+  - id: bfd46bc528e265e3a37rkw
+    model: GPH-DA-LX
+    name: "GPH-DA-LX panel heater"
```
make-all, line 5: name not required here.

custom_components/tuya_local/devices/gph_da_lx_panel_heater.yaml
```diff
@@ -0,0 +1,113 @@
...
+                  min: 5
+                  max: 45
+              - dps_val: [FrostProtection, FrostPro, frostpro]
+                range:
+                  min: 6
+                  max: 6
+      - id: 3
+        name: current_temperature
+        type: integer
+        unit: C
```
make-all, line 43: unit not required here. climate cannot have separate current_temperature and temperature units.

custom_components/tuya_local/devices/gph_da_lx_panel_heater.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - id: 4
+        name: preset_mode
+        type: string
+        optional: true
+        mapping:
+          - constraint: preset_mode
```
make-all, line 49: This is overly complex for a config for a single device, and I suspect these conditions are not needed. But if retained, constraint is not required when it points to the current dp.

custom_components/tuya_local/devices/gph_da_lx_panel_heater.yaml
```diff
@@ -0,0 +1,113 @@
...
+          - dps_val: "3h"
+            value: "3h"
+          - dps_val: "4h"
+            value: "4h"
+          - dps_val: "5h"
+            value: "5h"
+          - dps_val: "6h"
+            value: "6h"
+  - entity: switch
+    name: Display sleep
```
make-all, line 108: I think you can use `translation_key: sleep` here.

make-all (changes requested):

make-all:
No feedback for months, I do not have enough info to fix the issues in this config myself

PR #4486, opened 2026-02-11
Add support for Duux Beam 2

rplinzner:
# Description
Adds support for [Duux Beam 2](https://duux.com/en/product/beam-2-black/)

# Device details

- Manufacturer: Duux
- Model: Beam 2
- Product ID: `wksnqcoh6yzpoam9`
- Category: Humidifier

# Home assistant screenshot
<p align="middle">
Screenshot 2026-02-11 102903
Screenshot 2026-02-11 102958
Screenshot 2026-02-11 103018
</p>

### DIFF
--- a/custom_components/tuya_local/devices/duux_beam2_humidifier.yaml
+++ b/custom_components/tuya_local/devices/duux_beam2_humidifier.yaml
@@ -92,7 +92,6 @@ entities:
             value: 6h
   - entity: switch
     translation_key: sleep
-    icon: "mdi:weather-night"
     category: config
     dps:
       - id: 21


make-all:
Your product id appears to actually be a device id, which is unique to your specific device.
product id should come from the Tuya "Query Device Details" API call, or from `python -m tinytuya scan` (active Tuya API account not required)

rplinzner:
@make-all Correct, fixed

rplinzner:
@make-all Fixed the linting issues

custom_components/tuya_local/devices/duux_beam_2.yaml
```diff
@@ -99,7 +100,7 @@ entities:
         type: boolean
         optional: true
   - entity: switch
-    category: diagnostic # the switch does turn off/on the device, but it's impossible to change its value in the Tuya app. Use id 1 instead.
+    category: diagnostic
```
make-all: Diagnostic is not valid for a switch. But you seem to be saying this is read-only, so it should be a binary_sensor, or probably better the "action" dp of the humidifier, with values mapped to `humidifying` and `idle`

custom_components/tuya_local/devices/duux_beam_2.yaml
```diff
@@ -0,0 +1,139 @@
...
+    dps:
+      - id: 14
+        type: integer
+        optional: true
+        name: sensor
+        class: measurement
+        unit: C
+  - entity: select
+    translation_key: timer
+    icon: "mdi:timer-outline"
```
make-all: icon need not be overridden here, as the translation_kay already sets a suitable one

custom_components/tuya_local/devices/duux_beam_2.yaml
```diff
@@ -0,0 +1,139 @@
...
+          - dps_val: "3"
+            value: 3h
+          - dps_val: "4"
+            value: 4h
+          - dps_val: "5"
+            value: 5h
+          - dps_val: "6"
+            value: 6h
+  - entity: switch
+    name: Night mode
```
make-all: Maybe `translation_key: sleep` can be used here

rplinzner: clarification: night mode just makes the display dimmer

custom_components/tuya_local/devices/duux_beam_2.yaml
rplinzner: it's R/W actually. The thing is that this is just redundant way of turning off/on the device and the tuya app does not seem to use it. I can just expose it normally. This is the model from the API. Added it to the device schema for the sake of completeness 
```
{
          "abilityId": 10,
          "accessMode": "rw",
          "code": "switch",
          "description": "",
          "extensions": { "iconName": "icon-dp_power2" },
          "name": "总开关",
          "typeSpec": { "type": "bool" }
        },
```

rplinzner:
All fixed :)

custom_components/tuya_local/devices/duux_beam_2.yaml
```diff
@@ -0,0 +1,137 @@
...
+        type: integer
+      - id: 12
+        name: mode
+        type: string
+        optional: true
+        mapping:
+          - dps_val: auto
+            value: "auto"
+          - dps_val: general
+            value: "Continuous"
```
make-all: Add `translation_key: extended` to the humidifier definition above to make "continuous" available as a translated mode.

rplinzner: done

custom_components/tuya_local/devices/duux_beam_2.yaml
rplinzner: this was also corrected btw :)

PR #4489, opened 2026-02-11
feat(arida_eva_humidifier): add support for Arida Eva humidifier

benjamin-lyon:
Adds support for the [Arida Eva air humidifier](https://arida.no/produkt/arida-eva/). I've tested the integration locally and it works without issue.

### DIFF
--- a/custom_components/tuya_local/devices/arida_eva_humidifier.yaml
+++ b/custom_components/tuya_local/devices/arida_eva_humidifier.yaml
@@ -126,7 +126,6 @@ entities:
             value: true
           - value: false
   - entity: binary_sensor
-    name: Device fault
     class: problem
     category: diagnostic
     dps:
@@ -140,13 +139,9 @@ entities:
           - dps_val: 2
             value: false
           - value: true
-  - entity: sensor
-    name: Fault code
-    category: diagnostic
-    dps:
       - id: 22
-        name: sensor
-        type: integer
+        name: fault_code
+        type: bitfield
         optional: true
   - entity: lock
     translation_key: child_lock


custom_components/tuya_local/devices/arida_eva_humidifier.yaml
```diff
@@ -0,0 +1,184 @@
+name: Humidifier
+products:
+  - id: fv30bc
```
make-all: Is this a Zigbee or Bluetooth device? Product ID for WiFi devices is normally longer than this

custom_components/tuya_local/devices/arida_eva_humidifier.yaml
```diff
@@ -0,0 +1,184 @@
...
+        type: boolean
+        name: mode
+        mapping:
+          - dps_val: true
+            value: auto
+          - dps_val: false
+            value: normal
+  - entity: sensor
+    class: temperature
+    category: diagnostic
```
make-all: I think it is better to leave out the diagnostic category for most sensors, as HA now splits them into their own category.

custom_components/tuya_local/devices/arida_eva_humidifier.yaml
```diff
@@ -0,0 +1,184 @@
...
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 20
+        type: integer
+        name: sensor
+        unit: min
+  - entity: binary_sensor
+    translation_key: tank_empty
+    category: diagnostic
```
make-all: This should probably not be diagnostic, because it is expected to happen in normal usage.

custom_components/tuya_local/devices/arida_eva_humidifier.yaml
```diff
@@ -0,0 +1,184 @@
...
+    dps:
+      - id: 22
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 1
+            value: true
+          - value: false
+  - entity: binary_sensor
+    name: Communication error
```
make-all: Does this need its own sensor? Maybe a general problem sensor that catches all exceptional faults (ie not tank_empty) is enough?

custom_components/tuya_local/devices/arida_eva_humidifier.yaml
benjamin-lyon: It's WiFI. I've found the correct product ID and updated it 👍

custom_components/tuya_local/devices/arida_eva_humidifier.yaml
benjamin-lyon: Looked at how other humidifier YAMLs are configured and did that. Now there's a 'Device fault' sensor that should catch everything other than the empty tank fault, and also added a 'Fault code' sensor that displays the fault code value.

PR #4496, opened 2026-02-12
YSR-CGD-RGB Smart Bowl Lamp

pacorola:
Hello! I created this DPS that fits this particular lamp, which worked with the rgbcW_lightbulb preset, but has there the color temperature value wrong (curiously it also happens in smartlife). This file allows the light (and others with the same problem) to be set up correctly and have color temp right.

https://es.aliexpress.com/item/1005010380798747.html

### DIFF
--- a/custom_components/tuya_local/devices/rgbwc_lightbulb.yaml
+++ b/custom_components/tuya_local/devices/rgbwc_lightbulb.yaml
@@ -1,7 +1,8 @@
 name: RGBCW lightbulb
 products:
   - id: leirmht5udei1xzw
-    name: Smart LED
+    manufacturer: YSRSAI
+    model: YSR-CGD-RGB
 entities:
   - entity: light
     dps:


custom_components/tuya_local/devices/rgbcw_lightbulb_inverted.yaml
```diff
@@ -0,0 +1,192 @@
+name: RGBCW lightbulb (Inverted color temp)
+products:
+  - id: aok3caeleulg1neh
```
make-all: These products should not be listed here. Instead you should put the product details for your backwards bulb.

custom_components/tuya_local/devices/rgbwc_lightbulb.yaml
```diff
@@ -0,0 +1,192 @@
...
+          - dps_val: null
+            value: 0
+      - id: 23
+        name: color_temp
+        type: integer
+        optional: true
+        range:
+          min: 1
+          max: 1000
+        mapping:
```
make-all, line 41: This is not how you reverse a range, it is only modifying two specific values.

pacorola: Turns out you are right!!!! It only works at extremes, do you know how can I fix it please?

make-all:
Test this and make sure it works.  If it is not giving correct results, the other thing to try will be to swap the min and max under target_range, which should do the inversion at the same time as the range conversion.

pacorola:
I made a whole lot of changes! Tested and works flawlessly! Thanks for your suggestions, now I believe its perfect. Correct product ID, invert: true, and correct scenes.

PR #4504, opened 2026-02-15
Add Tuya PS580 wall socket support

lbp0:


make-all:
Can be covered by smartplug_usb.yaml, with some tweaks.

lbp0:
> Can be covered by smartplug_usb.yaml, with some tweaks.

Thanks for your answer! However, since that file contains many more DPs, auto config does not work for this device. I hope this file will make it easier to add the device

PR #4508, opened 2026-02-15
Add YAML configuration for Hokkaido air conditioner

King2zzo:
This PR adds a new YAML configuration for a Hokkaido air conditioner to enable local control through tuya-local.
DPS mapping was derived from device inspection and validated through real-world usage.


custom_components/tuya_local/devices/hokkaido_airconditioner.yaml
```diff
@@ -0,0 +1,273 @@
+name: Air conditioner
+products:
+  - id: nnlr9l5ipq5jfoxp
+    manufacturer: Hokkaido
+    model: HOKKAIDO Air Conditioner
```
make-all: Leave the model out if it is just the brand name from above and the device type

custom_components/tuya_local/devices/hokkaido_airconditioner.yaml
```diff
@@ -0,0 +1,273 @@
...
+  - entity: switch
+    name: Countdown off
+    icon: "mdi:timer-off"
+    category: config
+    dps:
+      - id: 110
+        type: boolean
+        name: switch
+  - entity: time
+    name: Countdown
```
make-all: Use `translation_key: timer` instead of explicit name and icon

custom_components/tuya_local/devices/hokkaido_airconditioner.yaml
```diff
@@ -0,0 +1,273 @@
...
+  - entity: sensor
+    name: Model
+    category: diagnostic
+    dps:
+      - id: 108
+        type: bitfield
+        name: sensor
+        optional: true
+  - entity: switch
+    name: Countdown on
```
make-all: Use Timer rather than Countdown to be consistent with translations.

PR #4511, opened 2026-02-15
Add Gosund SP6 device configuration

ivasso:
Add support for the Gosund SP6 smartplug

https://www.tradeinn.com/techinn/en/gosund-sp6-16a-usb-c-20w-tuya-smart-wifi-bluetooth-socket-smart-plug/142153224/p


make-all:
Merged into smartplug_usb.yaml, as the dps assignments are all matching generic Tuya smartplugs

PR #4530, opened 2026-02-18
Add support for Mini Wi-Fi Curtain Module (uoa3mayicscacseb)

matteoflavi:
Tested locally.
DP1 = control (open/close/stop)
DP2 = percent_control used as position (0–100)
Reverse motor supported via DP8.
Calibration supported via DP3.


make-all:
This is not required, as loratap_zigbee_curtain.yaml already defines a compatible device.

PR #4538, opened 2026-02-20
Create varin_va_e003_fan_light.yaml

Mighte1:
Creates support for Varin VA-E003 Smart Ceiling Fan with Light (sold in NL at bol.com)
Protocol 3.5 needed.

Contains:
* Main light (brighness/temperature)
* Auxiliary light (some fixed color choice options, ro select scene/music). This is limited. Use tuya app to have a lot of options here
*  Master switch
*  Fan with 6 speeds and choose direction

### DIFF
--- a/custom_components/tuya_local/devices/varin_va_e003_fan_light.yaml
+++ b/custom_components/tuya_local/devices/varin_va_e003_fan_light.yaml
@@ -1,4 +1,8 @@
-name: Varin VA-E003 Fan Light
+name: Fan light
+products:
+  - id: ufrvcfjnmhnqebge
+    manufacturer: Varin
+    model: VA-E003
 entities:
   - entity: light
     name: Main Light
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/varin_va_e003_fan_light.yaml
```diff
@@ -0,0 +1,112 @@
...
+        mapping:
+          - dps_val: colour
+            value: "off"
+          - dps_val: scene
+            value: Scene
+          - dps_val: music
+            value: Music
+
+  - entity: select
+    name: Auxiliary Color
```
make-all: It would be better to put these as a `named_color` attribute of the light.

custom_components/tuya_local/devices/varin_va_e003_fan_light.yaml
```diff
@@ -0,0 +1,112 @@
...
+        name: direction
+        optional: true
+        mapping:
+          - dps_val: forward
+            value: forward
+          - dps_val: reverse
+            value: reverse
+      - id: 106
+        type: string
+        name: preset_mode
```
make-all, line 111: `preset_mode` requires a mapping with the list of options, otherwise HA does not know what to put there.

custom_components/tuya_local/devices/varin_va_e003_fan_light.yaml
```diff
@@ -0,0 +1,112 @@
...
+
+  - entity: switch
+    name: Master Switch
+    dps:
+      - id: 20
+        type: boolean
+        name: switch
+
+  - entity: fan
+    dps:
```
make-all, line 89: Adding `translation_key: fan_with_presets` will allow the fan presets to be translated as long as the mapping values match entries in the translation files.

custom_components/tuya_local/devices/varin_va_e003_fan_light.yaml
```diff
@@ -0,0 +1,112 @@
+name: Varin VA-E003 Fan Light
```
make-all: Top level name should be generic (eg "Fan light") in case other devices match the config. Product details belong below this inside a `products` list, indexed by product_id (which you can generally find in the same place you found your local_key, or by `python -m tinytuya scan` from a PC on the same network as the device (you may need to disconnect the device from HA, and run `pip install tinytuya` first)

PR #4540, opened 2026-02-20
Add ultenic d5 vacuum support

RoBSPRR:


custom_components/tuya_local/devices/ultenic_d5_vacuum.yaml
```diff
@@ -0,0 +1,221 @@
...
+  - entity: vacuum
+    dps:
+      - id: 1
+        type: boolean
+        name: power
+        optional: true
+      - id: 11
+        type: bitfield
+        name: error
+        mapping:
```
make-all, line 16: There is no point to this mapping. It is better to keep integer error codes than to mark everything as "unknown" and lose the info that distinguishes them.

custom_components/tuya_local/devices/ultenic_d5_vacuum.yaml
```diff
@@ -0,0 +1,221 @@
+name: Ultenic D5 vacuum
```
make-all, line 1: Top level name should be generic in case other devices match the same config

custom_components/tuya_local/devices/ultenic_d5_vacuum.yaml
```diff
@@ -0,0 +1,221 @@
...
+          - dps_val: 11
+            value: unknown
+      - id: 38
+        type: string
+        name: status
+        mapping:
+          - dps_val: 0
+            value: standby
+          - dps_val: 1
+            value: cleaning
```
make-all, line 46: You can leave the cleaning states as different states to differentiate them in the status attribute, unknown statuses all get converted to cleaning for the entity state.

custom_components/tuya_local/devices/ultenic_d5_vacuum.yaml
```diff
@@ -0,0 +1,221 @@
...
+            value: charging
+          - dps_val: 6
+            value: cleaning
+      - id : 25
+        type: string
+        name: command
+        mapping:
+          - dps_val: NONE
+            value: stop
+          - dps_val: NONE
```
make-all, line 63: You cannot have duplicate dps_val in a mapping. It isn't clear what you are trying to accomplish here, but probably the solution is something else

custom_components/tuya_local/devices/ultenic_d5_vacuum.yaml
```diff
@@ -0,0 +1,221 @@
...
+    dps:
+      - id: 39
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    name: Cleaned area
+    category: diagnostic
+    icon: "mdi:floor-plan"
```
make-all, line 116: `class: area`

custom_components/tuya_local/devices/ultenic_d5_vacuum.yaml
```diff
@@ -0,0 +1,221 @@
...
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 42
+        type: integer
+        name: sensor
+        unit: min
+        class: measurement
+#  - entity: number
+#    name: Volume
```
make-all, line 133: This is commented out so maybe not important, but if you reinstate this, use `translation_key: volume` for audio volume control.

custom_components/tuya_local/devices/ultenic_d5_vacuum.yaml
```diff
@@ -0,0 +1,221 @@
...
+#    dps:
+#      - id: 43
+#        type: integer
+#        name: value
+#        unit: "%"
+#        range:
+#          min: 0
+#          max: 100
+  - entity: sensor
+    name: Sensor
```
make-all, line 145: This name is too generic, the name should explain what the sensor is measuring.

custom_components/tuya_local/devices/ultenic_d5_vacuum.yaml
```diff
@@ -0,0 +1,221 @@
...
+    name: Main brush
+    category: diagnostic
+    icon: "mdi:circle"
+    dps:
+      - id: 48
+        type: integer
+        name: sensor
+        unit: "%"
+  - entity: switch
+    name: Lamps
```
make-all, line 181: Use a light entity for lights.

custom_components/tuya_local/devices/ultenic_d5_vacuum.yaml
```diff
@@ -0,0 +1,221 @@
...
+  - entity: button
+    name: Sensor reset
+    category: config
+    class: restart
+    dps:
+      - id: 54
+        type: boolean
+        name: button
+  - entity: button
+    name: Filter reset
```
make-all, line 197: Use `translation_key: filter_reset`, and remove the name and class.

make-all (changes requested):

PR #4545, opened 2026-02-21
add support for grid connect and fan force fan and lights

kooori:
add support for grid connect and fan force fan and lights

### DIFF
--- a/custom_components/tuya_local/devices/fanforce_ceilingfanlight.yaml
+++ b/custom_components/tuya_local/devices/fanforce_ceilingfanlight.yaml
@@ -1,8 +1,8 @@
-name: Fan and Light Ceiling Fan
+name: Ceiling fan and light
 products:
   - id: lb9da3os0ihifq7u
     manufacturer: FanForce
-    model: CeilingLight
+    model: Ceiling Light
 entities:
   - entity: fan
     dps:


custom_components/tuya_local/devices/grid_connect_ceilinglight.yaml
```diff
@@ -0,0 +1,38 @@
+name: Grid Connect Ceiling Light
+products:
+  - id: gp3mfqqhyesuyder # Replace with the actual product ID if you have it
```
make-all: What does this comment mean? This is not the actual product id of this device?

In any case, this and the other light below already seem to be covered by the cct_lightbulb.yaml config.

custom_components/tuya_local/devices/fanforce_ceilingfanlight.yaml
```diff
@@ -0,0 +1,65 @@
+name: BBQ Fan
```
make-all: This name seems specific to your installation

custom_components/tuya_local/devices/fanforce_ceilingfanlight.yaml
```diff
@@ -0,0 +1,65 @@
...
+products:
+  - id: lb9da3os0ihifq7u
+    manufacturer: FanForce
+    model: CeilingLight
+entities:
+  - entity: fan
+    translation_only_key: fan_with_presets
```
make-all: There are no presets, so this is not required

custom_components/tuya_local/devices/fanforce_ceilingfanlight.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - entity: light
+    dps:
+      - id: 15
+        type: boolean
+        name: switch
+      - id: 16
+        name: brightness
+        type: integer
+        range:
+          min: 1
```
make-all: Is this range correct? Usually lights that go to 1000 have a minimum of 10 (1%) rather than 1

kooori (commented):
Thanks for the feedback. This is my first every commit on github so appreciate the patience. 

I made the changes based on your feedback

make-all:
The light config files need to be removed, or their existence justified to proceed to merge.

Issue #807, opened 2023-06-16
Request support for Daewoo Dhome DA9KWE airconditioner


TanYauRei:
# Log Message

```
2023-06-16 12:12:15.682 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches feit_dimmer with quality of 57%. DPS: {"updated_at": 1686910333.7208335, "1": false, "2": 16, "3": 27, "4": "Fan", "5": "Low", "22": 0, "101": false}
2023-06-16 12:12:15.688 WARNING (MainThread) [custom_components.tuya_local.config_flow] Report this to https://github.com/make-all/tuya-local/issues/

```

# Information about DPS mappings

```
{
  "result": {
    "category": "kt",
    "functions": [
      {
        "code": "switch",
        "dp_id": 1,
        "lang_config": {
          "false": "OFF",
          "true": "ON"
        },
        "name": "Switch",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "dp_id": 2,
        "lang_config": {
          "unit": "℃"
        },
        "name": "Set Temp",
        "type": "Integer",
        "values": "{\"unit\":\"℃\",\"min\":16,\"max\":32,\"scale\":0,\"step\":1}"
      },
      {
        "code": "mode",
        "dp_id": 4,
        "lang_config": {
          "cool": "",
          "dry": "",
          "fan": "",
          "heat": ""
        },
        "name": "Mode",
        "type": "Enum",
        "values": "{\"range\":[\"Cool\",\"Dry\",\"Fan\",\"Heat\"]}"
      },
      {
        "code": "fan_speed_enum",
        "dp_id": 5,
        "lang_config": {
          "hig": "",
          "low": ""
        },
        "name": "Fan Speed",
        "type": "Enum",
        "values": "{\"range\":[\"Low\",\"Hig\"]}"
      },
      {
        "code": "Sleep",
        "dp_id": 101,
        "lang_config": {
          "false": "",
          "true": ""
        },
        "name": "Sleep",
        "type": "Boolean",
        "values": "{}"
      }
    ],
    "status": [
      {
        "code": "switch",
        "dp_id": 1,
        "lang_config": {
          "false": "OFF",
          "true": "ON"
        },
        "name": "Switch",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "dp_id": 2,
        "lang_config": {
          "unit": "℃"
        },
[...]

PR #4556, opened 2026-02-23
Fix daewoo dhome heatpump High fan mode

diogolccsilva:
While testing my air conditioner (Beltax A011_D2) using the daewoo_dhome_heatpump device config (the one that was suggested and fit best) I found that the High setting for the fan mode didn't work. 

After looking at the debug logs I noticed the High fan mode value was being sent as "Hig":

image

And I confirmed this as well looking at the yaml configuration for the device type.

I manually changed in my Home Assistant and it solved the problem. Hoping to get this fixed for everyone.

### DIFF
--- a/custom_components/tuya_local/devices/daewoo_dhome_heatpump.yaml
+++ b/custom_components/tuya_local/devices/daewoo_dhome_heatpump.yaml
@@ -43,7 +43,7 @@ entities:
         mapping:
           - dps_val: Low
             value: low
-          - dps_val: Hig
+          - dps_val: High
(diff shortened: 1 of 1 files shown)


make-all:
Hig is the string given on #807, and there have been no complaints from owners of that model.
Without seeing full details of your model's data model, there is a possibility that other things are also not matching, or missing if you reuse the Daewoo config for a differenr product.

diogolccsilva:
@make-all sorry for that, you are right. I created a new pull request for my device #4567

PR #4560, opened 2026-02-23
Add RESTMO BT Water Meter configuration

pmyers:
Added support for the RESTMO BT Water Meter (FML026A). This device uses DPs 107 for Total Consumption and 5 for Flow Rate. Verified working on local hardware.

### DIFF
--- a/custom_components/tuya_local/devices/inkbird_itc308_thermostat.yaml
+++ b/custom_components/tuya_local/devices/inkbird_itc308_thermostat.yaml
@@ -216,6 +216,7 @@ entities:
   - entity: number
     category: config
     name: Heating hysteresis
+    class: temperature_delta
     icon: "mdi:thermometer-chevron-up"
     dps:
       - id: 117
@@ -239,6 +240,7 @@ entities:
   - entity: number
     category: config
     name: Cooling hysteresis
+    class: temperature_delta
     icon: "mdi:thermometer-chevron-down"
     dps:
       - id: 118
--- a/tests/devices/test_inkbird_itc308_thermostat.py
+++ b/tests/devices/test_inkbird_itc308_thermostat.py
@@ -90,7 +90,8 @@ class TestInkbirdITC308Thermostat(
                     "step": 0.1,
                     "min": -9.9,
                     "max": 9.9,
-                    "unit": "°",
+                    "device_class": NumberDeviceClass.TEMPERATURE_DELTA,
+                    "unit": NumberDeviceClass.CELSIUS,
                 },
                 {
                     "name": "number_compressor_delay",
@@ -125,6 +126,7 @@ class TestInkbirdITC308Thermostat(
                     "step": 0.1,
                     "min": 0.3,
                     "max": 15.0,
+                    "device_class": NumberDeviceClass.TEMPERATURE_DELTA,
                     "unit": UnitOfTemperature.CELSIUS,
                 },
                 {
@@ -134,6 +136,7 @@ class TestInkbirdITC308Thermostat(
                     "step": 0.1,
                     "min": 0.3,
                     "max": 15.0,
+                    "device_class": NumberDeviceClass.TEMPERATURE_DELTA,
                     "unit": UnitOfTemperature.CELSIUS,
                 },
             ]
--- a/tests/devices/test_inkbird_sousvide.py
+++ b/tests/devices/test_inkbird_sousvide.py
@@ -70,7 +70,8 @@ class TestInkbirdSousVideCooker(
                     "max": 9.9,
                     "scale": 10,
                     "step": 0.1,
-                    "unit": "°",
+                    "device_class": NumberDeviceClass.TEMPERATURE_DELTA,
+                    "unit": UnitOfTemperature.CELSIUS,
                 },
             ]
         )
--- a/tests/devices/test_nashone_mts700wb_thermostat.py
+++ b/tests/devices/test_nashone_mts700wb_thermostat.py
@@ -1,6 +1,7 @@
 from homeassistant.components.climate.const import ClimateEntityFeature, HVACMode
+from homeassistant.components.number import NumberDeviceClass
 from homeassistant.components.sensor import SensorDeviceClass
-from homeassistant.const import UnitOfTime
+from homeassistant.const import UnitOfTemperature, UnitOfTime
 
 from ..const import NASHONE_MTS700WB_THERMOSTAT_PAYLOAD
 from ..helpers import assert_device_properties_set
@@ -56,7 +57,8 @@ class TestNashoneMTS700WBThermostat(
             self.entities.get("number_temperature_calibration"),
             min=-5,
             max=5,
-            unit="°",
+            device_class=NumberDeviceClass.TEMPERATURE_DELTA,
+            unit=UnitOfTemperature.CELSIUS,
         )
         self.setUpBasicSelect(
             TIMER_DPS,


custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
```diff
@@ -0,0 +1,54 @@
+name: RESTMO BT Water Meter
+products:
+  - id: mqqna0px
+    name: RESTMO BT Water Meter (FML026A)
+entities:
+  - entity: sensor
+    name: Total Consumption
```
make-all: Leave out the name so the default translations for the water class show through.

custom_components/tuya_local/devices/basic_ir_remote.yaml
```diff
@@ -11,7 +11,7 @@ products:
     model: S16
   - id: x0lyfgjuguuh1vof
     manufacturer: Avatto
-    model: S16Pro  # also includes rf, but that is not supported yet
+    model: S16Pro # also includes rf, but that is not supported yet
```
make-all: Please undo this change, it is incorrect

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
```diff
@@ -0,0 +1,54 @@
+name: RESTMO BT Water Meter
```
make-all: You should follow HA naming convention here and only capitalise the first word.

Also, leave branding out of here, as other devices may also match.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
```diff
@@ -0,0 +1,54 @@
...
+products:
+  - id: mqqna0px
+    name: RESTMO BT Water Meter (FML026A)
```
make-all: Split this into manufacturer (Restmo) and model (FML026A)

The rest is either covered by the name above or irrelevant.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - id: 107
+        name: sensor
+        type: integer
+        unit: L
+        class: total_increasing
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: Flow Rate
+    class: water
```
make-all: water is incorrect class, it should be `volume_flow_rate`

Again, it is better to let the default translation show by removing the name.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
```diff
@@ -0,0 +1,54 @@
...
+    dps:
+      - id: 5
+        name: sensor
+        type: integer
+        unit: L/min
+        mapping:
```
make-all, line 25: `class: measurement`

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - scale: 10
+  - entity: binary_sensor
```
make-all: This is not necessary, it is duplicating information already available from the flow rate above.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
```diff
@@ -0,0 +1,54 @@
...
+    dps:
+      - id: 5
+        name: sensor
+        type: integer
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+  - entity: sensor
+    name: Battery
```
make-all: Remove the name

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
```diff
@@ -0,0 +1,54 @@
...
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 4
+        name: sensor
+        type: integer
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    name: Unit Mode
```
make-all: I suspect this is for setting the units, not just reading them. If the values output by the device are affected by this, then the units need to be dynamic for the sensors above.

make-all (changes requested):

pmyers:
Hi @make-all, and thank you for the corrections and best practices info. This was my first attempt at contributing anything to the HA ecosystem, so I appreciate your patience as I learn a bit. 😄 

I'll take a look at all of your comments and incorporate the changes you suggest, and will push an update to the branch later today that will hopefully be acceptable.

Thanks again!

custom_components/tuya_local/devices/basic_ir_remote.yaml
pmyers: Yep, will remove. The reason I touched that file at all in the first place was... I was getting linter errors on my branch after I rebased the branch off of main due to this file. 

The linter was complaining about a the id: line because there was an extra space after the colon. So, in order to get the linter to pass I removed the space. But then later, it looks like someone realized this was bad on main and fixed it there, obviating the need for me to fix it in this branch at all.

The line 14 change was collateral damage... For some reason my IDE (Antigravity/VSCode) would not let me save a change to that file to fix the id: line without removing that space on line 14. No idea why.

In any case, I'll revert any changes to this file. Now that the file lints properly, no need for any changes on this branch to it. 👍

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
pmyers: Do you mean leave out the name for "Total Consumption"? Just wanted to make sure.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
pmyers: Ok, so would "name: Water meter" be better then?

And if we're taking out the branding, should I change the filename as well? The filename has the brand in it too.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
pmyers: I'll need to look to see whether the device allows setting the unit or if this is read-only. The device is VERY passive... unless you wake it up explicitly, it isn't accessible and it only wakes up on its own to transmit data after the completion of a flow event. So, even if this was a writable value, you would need to first wake up the device before you could try writing to it.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
pmyers: Ah ok, based on context from your other comments below, you meant to take out "name: Total Consumption". Got it, and will do.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
make-all: The file name should be left, there are over 1000 config files, and they need to be meaningfully differentiated, but besides initial config, the filename is not used in the UI, so branding there is not as confusing when it mismatches with actual device.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
make-all: Yes, just remove the whole line,

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
make-all: If it is read/write, then use a select entity. Either way, the options should be listed in a mapping, and any mapping to HA's standard representation done.

custom_components/tuya_local/devices/restmo_fml026a_water_meter.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - entity: sensor
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 4
+        name: sensor
+        type: integer
+        unit: "%"
+        class: measurement
+  - entity: sensor
```
pmyers: @make-all I confirmed that writing a value here has no effect on the device. It's a read-only value, apparently. I also confirmed that changing the physical display on the device from L to gal does not seem to change this value at all. My conclusion is that the display units on the physical LCD does not change the fact that on the wire these values are always sent in L units. I don't think it hurts to leave this in as a sensor value, but if you would prefer that we remove it entirely I can do that to simplify things.

If we leave it in as a read-only sensor value, I can also remove the mapping for gal. I left it in only because, perhaps there are other similar devices out there that can transmit data in gallons rather than liters, and in that case having both as possible values here would be useful. But... let me know what you would prefer here.

pmyers: Actually, on further testing it turns out that I think the right path is to just remove this entirely. It doesn't provide any value since it can't be changed (so it isn't selectable) and the way I had it in the previous commit threw an error on testing because it expected an enum type. I removed this sensor in the latest commit here.

pmyers:
Thanks for the review, feedback and merge @make-all 👍

PR #4571, opened 2026-02-24
 Add distance sensor (DP 122) to ZG-205W mmWave profile

mxwmnn:
## Summary
- Add a `Distance` sensor entity to `zg205w_mmWave_presence_sensor.yaml` using DP `122`.
- Configure it as a measurement sensor with `cm` unit and `mdi:signal-distance-variant` icon.
- Keep existing entities unchanged.

## Why
Some ZG-205W devices expose live distance data on DP 122 (`distance`), but the current profile does not surface it in Home Assistant.  
This change makes that data available as a native sensor without affecting existing mappings.

image

## Validation
- `pytest tests/test_device_config.py` (passes)
- `yamllint custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml` (passes)
- `ruff check .` (passes)
- `ruff check --select I .` (passes)
- `ruff format --check .` (passes)

### DIFF
--- a/custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
+++ b/custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
@@ -73,12 +73,22 @@ entities:
   - entity: sensor
     name: Distance
     icon: mdi:signal-distance-variant
+    hidden: unavailable
     dps:
       - id: 122
         type: integer
+        optional: true
         name: sensor
         unit: cm
         class: measurement
+      - id: 122
+        type: integer
+        optional: true
+        name: available
+        mapping:
+          - dps_val: null
+            value: false
+          - value: true
   - entity: number
     name: Stationary sensitivity
     category: config


custom_components/tuya_local/devices/zg205w_mmWave_presence_sensor.yaml
```diff
@@ -70,6 +70,15 @@ entities:
         name: sensor
         unit: lx
         class: measurement
+  - entity: sensor
+    name: Distance
+    icon: mdi:signal-distance-variant
+    dps:
+      - id: 122
+        type: integer
```
make-all, line 79: If this is not supported by all models, then `optional: true` is required here.

It would also be useful to add below
```yaml
      - id: 122
        type: integer
        optional: true
        name: available
        mapping:
          - dps_val: null
            value: false
          - value: true
```

And above in the entity definition: `hidden: unavailable`

mxwmnn:
Applied your suggestion: marked DP 122 as optional, added an `available` helper for null, and hid the Distance entity when unavailable. Don’t know if every device supports it, API request to iot.tuya with query things data model returned it always for my ZG-205W models. (WiFi)

PR #4575, opened 2026-02-25
Add support for Infrared Heater Panels from Termoplaza

Schm1tz1:
I added support for Termoplaza Wifi models (https://termoplaza.de/produkt/termoplaza-1100-wifi-heizpaneel/). Values have been extracted from Tuya via API, integrated and double-checked with the Tuya app. Works smoothly for 2 heater panels in my local HA on my fork for about a week now.
Any feedback appreciated, especially on entity best practices - structure and naming ist best guess and what I have seen for similar device configurations.

### DIFF
--- a/custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
+++ b/custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
@@ -21,6 +21,7 @@ entities:
         range:
           min: 5
           max: 30
+        unit: C
         optional: true
       - id: 3
         name: current_temperature
@@ -30,11 +31,11 @@ entities:
         type: string
         optional: true
         mapping:
-          - dps_val: "ECO"
+          - dps_val: ECO
             value: eco
-          - dps_val: "Comfort"
+          - dps_val: Comfort
             value: comfort
-          - dps_val: "Night"
+          - dps_val: Night
             value: sleep
           - dps_val: "Off"
             value: none
@@ -48,6 +49,7 @@ entities:
   - entity: sensor
     name: Surface status
     class: enum
+    translation_key: status
     category: diagnostic
     dps:
       - id: 11
@@ -55,11 +57,11 @@ entities:
         name: sensor
         mapping:
           - dps_val: heating
-            value: Heating
+            value: heating
           - dps_val: warm
-            value: Warm
+            value: warm
           - dps_val: cold
-            value: Cold
+            value: cold
   - entity: sensor
     name: Duty cycle
     dps:
@@ -70,6 +72,7 @@ entities:
         unit: "%"
   - name: Turbo mode timer
     entity: select
+    translation_key: timer
     category: config
     dps:
       - id: 19
@@ -77,31 +80,31 @@ entities:
         type: string
         optional: true
         mapping:
-          - dps_val: 0
-            value: "Off"
-          - dps_val: 1
+          - dps_val: "0"
+            value: cancel
+          - dps_val: "1"
             value: 2h
-          - dps_val: 2
+          - dps_val: "2"
             value: 4h
-          - dps_val: 3
+          - dps_val: "3"
             value: 6h
-          - dps_val: 4
+          - dps_val: "4"
             value: 8h
-          - dps_val: 5
+          - dps_val: "5"
             value: 10h
-          - dps_val: 6
+          - dps_val: "6"
             value: 12h
-          - dps_val: 7
+          - dps_val: "7"
             value: 14h
-          - dps_val: 8
+          - dps_val: "8"
             value: 16h
-          - dps_val: 9
+          - dps_val: "9"
             value: 18h
-          - dps_val: 10
+          - dps_val: "10"
             value: 20h
-          - dps_val: 11
+          - dps_val: "11"
             value: 22h
-          - dps_val: 12
+          - dps_val: "12"
             value: 24h
   - name: Program mode
     entity: switch


Schm1tz1:
For reference: This is that tinytuya was able to discover:

```json
{
  "id": "REDACTED",
  "name": "TermoPlaza Heater",
  "key": "REDACTED",
  "mac": "REDACTED",
  "uuid": "REDACTED",
  "sn": "REDACTED",
  "category": "qn",
  "product_name": "TermoPlaza heater",
  "product_id": "xlln3q6hczrp31df",
  "biz_type": 0,
  "model": "TermoPlaza900",
  "sub": false,
  "icon": "https://images.tuyaeu.com/smart/icon/bay1607348090427rdBH/3a5fc0a4cb2c9ae45401ad5d5b205cdf.png",
  "mapping": {
      "1": {
          "code": "switch",
          "type": "Boolean",
          "values": {}
      },
      "2": {
          "code": "temp_set",
          "type": "Integer",
          "values": {
              "unit": "\u00b0C",
              "min": 5,
              "max": 30,
              "scale": 0,
              "step": 1
          }
      },
      "3": {
          "code": "temp_current",
          "type": "Integer",
          "values": {
              "unit": "\u00b0C",
              "min": -20,
              "max": 50,
              "scale": 0,
              "step": 1
          }
      },
      "4": {
          "code": "mode",
          "type": "Enum",
          "values": {
              "range": [
                  "Comfort",
                  "ECO",
                  "Night",
                  "Off"
              ]
          }
      },
      "7": {
          "code": "child_lock",
          "type": "Boolean",
          "values": {}
      },
      "11": {
          "code": "work_state",
          "type": "Enum",
          "values": {
              "range": [
                  "heating",
                  "warm",
                  "cold"
              ]
          }
      },
      "12": {
          "code": "work_power",
          "type": "Integer",
          "values": {
              "unit": "%",
              "min": 0,
              "max": 100,
              "scale": 1,
              "step": 5
          }
      },
      "19": {
          "code": "countdown_set",
          "type": "Enum",
          "values": {
              "range": [
                  "0",
                  "1",
                  "2",
                  "3",
                  "4",
                  "5",
                  "6",
                  "7",
                  "8",
                  "9",
                  "10",
                  "11",
                  "12"
              ]
          }
      },
      "21": {
          "code": "fault",
          "type": "Bitmap",
          "values": {
              "label": [
                  "temperature_sensor_failure"
              ],
              "maxlen": 1
          }
      },
      "23": {
          "code": "week_program_mode",
          "type": "Raw",
          "values": {
              "maxlen": 128
          }
      },
      "110": {
          "code": "program_mode",
          "type": "Boolean",
          "values": {}
      },
      "112": {
          "code": "brightness",
          "type": "Enum",
          "values": {
              "range": [
                  "0",
                  "1",
                  "2",
                  "3",
                  "4"
              ]
          }
      },
      "113": {
          "code": "set_int_temp",
          "type": "Integer",
          "values": {
              "unit": "\u00b0C",
              "min": 45,
              "max": 95,
              "scale": 0,
              "step": 1
          }
      },
      "114": {
          "code": "cur_int_temp",
          "type": "Integer",
          "values": {
              "unit": "\u00b0C",
              "min": -10,
              "max": 110,
              "scale": 0,
              "step": 1
          }
      },
      "115": {
          "code": "reset",
          "type": "Boolean",
          "values": {}
      }
  },
  "ip": "",
  "origin": "cloud",
  "ver": "3.5"
}
```

Schm1tz1:
Seems like I missed the required mappings in the initial commit. Added the enum mappings even though it is "just a string to reflect the status of the panel surface" to match the original type.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
+name: TermoPlaza heater
```
make-all: Top level name should be without branding in case other products also match.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+        mapping:
+          - dps_val: true
+            value: "heat"
+          - dps_val: false
+            value: "off"
+      - id: 2
+        name: temperature
+        type: integer
+        unit: C
+        mapping:
```
make-all: mapping is not needed here, as scale 1 and step 1 are already the default

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+        name: preset_mode
+        type: string
+        optional: true
+        mapping:
+          - dps_val: ECO
+            value: eco
+          - dps_val: Comfort
+            value: comfort
+          - dps_val: Night
+            value: away
```
make-all: `sleep` may be better here.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+            value: none
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 7
+        type: boolean
+        name: lock
+  - entity: sensor
+    translation_key: work_state
```
make-all: There is no translation_key by this name. But this entity looks like it would be useful as the hvac_action dp of the climate entity, depending on what Warm and Cold actually mean here.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+        mapping:
+          - dps_val: heating
+            value: Heating
+          - dps_val: warm
+            value: Warm
+          - dps_val: cold
+            value: Cold
+  - entity: sensor
+    translation_key: work_power
+    class: power_factor
```
make-all: power_factor is a specific measurement for inductive loads. I think this is just a duty cycle or percentage of max power, so does not meet the definition. The class should be removed.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+        name: sensor
+        mapping:
+          - dps_val: heating
+            value: Heating
+          - dps_val: warm
+            value: Warm
+          - dps_val: cold
+            value: Cold
+  - entity: sensor
+    translation_key: work_power
```
make-all: There is no translation_key by that name defined, other devices use the name "Power level" for this I think, but if it is measuring duty cycle rather than instantaneous power level, maybe "Duty cycle", or "Average power" or something is better.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+    class: power_factor
+    dps:
+      - id: 12
+        type: integer
+        optional: true
+        name: sensor
+        unit: "%"
+  - name: Turbo Mode Timer
```
make-all: HA naming convention is to only capitalise the first word of entity and device names.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+    entity: select
+    translation_key: countdown_set
```
make-all: `translation_key: timer` is the standard

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+    entity: switch
+    translation_key: timer
+    category: config
+    dps:
+      - id: 110
+        type: boolean
+        name: switch
+  - name: Brightness
+    entity: select
+    translation_key: light_mode
```
make-all: It you use this translation key, then it is better not to use a name (unless there are 2 or more entities using the same key).

But this could also be represented as a light entity with brightness dp.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+            value: Low
+          - dps_val: 1
+            value: Average
+          - dps_val: 2
+            value: Above avg
+          - dps_val: 3
+            value: Comfort
+          - dps_val: 4
+            value: High
+  - name: Surface Temperature Setpoint
```
make-all: Just "Surface temperature" is better.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+    class: temperature
+    dps:
+      - id: 113
+        type: integer
+        name: value
+        unit: C
+        range:
+          min: 45
+          max: 95
+  - name: Current Surface Temperature
```
make-all: Just "Surface temperature" is better.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,173 @@
...
+    dps:
+      - id: 114
+        class: measurement
+        type: integer
+        name: sensor
+        unit: C
+        range:
+          min: -10
+          max: 110
+  - name: Reset
```
make-all: Name should not be used here, as it prevents the translations from being used.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
Schm1tz1: Sure, understood. Fixed!

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
Schm1tz1: Sure!

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
Schm1tz1: Yes, that's what I also thought but we already have that information and state in the climate entity. It's more or less just an additional sensor information. This is not documented at all, but looks to me it's just an additional information based on the climate state and measured surface temperature - heating/warm/cold. That way it can show that there is a hot surface due to residual heat even with hvac state off. Not sure what's a suitable translation_key, in that case but can use name instead of translation key.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
Schm1tz1: Light entity sounds good.

Schm1tz1 (commented):
Ok, thanks for the feedback! Will run some local testing before pushing the changes.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
make-all: I see. There is not currently an hvac_action in the climate entity, so the information is not there, but the distinction between warm and cold would be lost, as both really map to `idle` for hvac_mode. So keeping this with a name of "Surface" in addition to adding an hvac_action attribute to climate may be best.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
make-all: The brightness steps need to be mapped to values between 0 and 255 for brightness, as that is the native range used by HA.

Schm1tz1:
Ok, I see - some fixes to be done. FYI - I reverted back from the light/brightness to a brightness level switch that is closer to the original device config in Tuya and has a better usability. Mapping 256 levels that are a brightness percentage in HA to discrete levels is quite a pain it you just have 4 or 5 discrete levels in your end-device.
Still I am looking for a good overview of the existing **translation_key**s. Is the anything available?

make-all:
The only list of translation keys is in the source: custom_components/tuya_local/translations/en.json

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,167 @@
...
+        type: integer
+        unit: C
+        range:
+          min: 5
+          max: 30
+        optional: true
+      - id: 3
+        name: current_temperature
+        type: integer
+        unit: C
```
make-all: I forgot to submit this comment previously:
unit is not needed here, because climate can only have one temperature_unit, and the priority is:
1. temperature_unit dp (for devices that are switchable between C/F)
2. unit defined on temperature dp
3. unit defined on current_temperature dp (here).

In this case the number 2 priority has already taken effect. Since this one is the same, it is no harm done, but if it was different it would be ignored. So to avoid confusion by others basing their config on this one in future it is better to remove it.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,167 @@
+name: Infrared Panel Heater
```
make-all: HA naming convention is to only capitalise the first word of device and entity names (unless there is a specific reason like abbreviations and brand names involved).

This also applies to some entity names below.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,167 @@
...
+            value: Low
+          - dps_val: 1
+            value: Average
+          - dps_val: 2
+            value: Above avg
+          - dps_val: 3
+            value: Comfort
+          - dps_val: 4
+            value: High
+  - name: Surface Temperature
```
make-all: Previously I asked to drop "Setpoint" from this, but I think it is not setting the surface temperature directly, but rather setting a limit for the surface temperature. So in hindsight I think a better suggestion would be changing it to "Surface temperature limit" or "Maximum surface temperature" to indicate that clearly.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,167 @@
...
+            value: 18h
+          - dps_val: 10
+            value: 20h
+          - dps_val: 11
+            value: 22h
+          - dps_val: 12
+            value: 24h
+  - name: Program Mode
+    entity: switch
+    translation_key: timer
```
make-all: Just remove this translation_key, the name above anyway would override it even if the key existed. Only select, number and time entities define a translation_key for timer.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,167 @@
...
+        type: string
+        optional: true
+        mapping:
+          - dps_val: ECO
+            value: eco
+          - dps_val: Comfort
+            value: comfort
+          - dps_val: Night
+            value: sleep
+          - dps_val: Off
```
make-all: This needs quotes to avoid  the yaml parser interpreting it as boolean false.

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
```diff
@@ -0,0 +1,167 @@
...
+            value: none
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 7
+        type: boolean
+        name: lock
+  - entity: sensor
+    name: Work state
```
make-all: I think "Work state" is too vague, taken directly from the Tuya protocol. `name: Surface` or "Surface status" ("Surface temperature" is already taken by the actual temperature reading)

make-all (changes requested):

custom_components/tuya_local/devices/termoplaza_panel_heater.yaml
Schm1tz1: Ok, will remove the `unit: C ` in the whole file then.

make-all: Only in this location (current_temperature). The rest of the locations are OK, as they are separate entities.

Schm1tz1:
Ok, looks like I applied all the suggested changes for now. Not 100% happy with the timer integration as HA did not integrate the entity now called "Turbo mode timer" when setting `translation_key: timer` without a name. It failed with an error indicating a missing name AFAICS.

Schm1tz1:
Thanks 🚀

PR #4577, opened 2026-02-25
Add Edison Smart dual-zone CCT ceiling light profile (TREO Mini Spotlight)

hrko:
## Summary
This PR adds a new Tuya Local device profile for Edison Smart dual-zone CCT ceiling lights, verified with:

- Brand: Edison Smart
- Model: TREO Mini Spotlight
- Product ID: `kjuo8stpv9adjkzo`

The profile is intentionally named generically because similar Edison Smart products appear to share the same feature set.

## Added profile
- `custom_components/tuya_local/devices/edisonsmart_dual_zone_cct_ceiling_light.yaml`

This profile was created based on `cct_lightbulb`.

Main mapping:
- DP 20: power (`switch`)
- DP 21: work mode (`work_mode`)
- DP 22: brightness
- DP 23: color temperature
- DP 25: scene (optional)
- DP 26: timer seconds (optional)
- DP 104: lighting zone mode as `light.effect`
  - `up` -> `UP or SIDE`
  - `all` -> `ALL`
  - `down` -> `DOWN`

*Note: The effect names (`UP or SIDE`, `ALL`, `DOWN`) are derived directly from the UI labels used in the Smartlife app.*

<details>
<summary>Smartlife App UI Screenshots</summary>

| UP or SIDE | ALL | DOWN |
|:---:|:---:|:---:|
| image | image | image |

</details>

## Datapoint evidence (from HA logs)
When adding the tested device:

- Product ID log:
  - `Adding TREOミニスポットライト device with product id kjuo8stpv9adjkzo`

- DPS logs:
  - `{"20": true, "21": "white", "22": 751, "23": 0, "25": "000e0d0000000000000000c80000", "26": 0, "104": "up"}`
  - `{"20": true, "21": "white", "22": 751, "23": 0, "25": "000e0d0000000000000000c80000", "26": 0, "104": "down"}`
  - `{"20": true, "21": "white", "22": 751, "23": 0, "25": "000e0d0000000000000000c80000", "26": 0, "104": "all"}`

## Local validation
Confirmed in a local environment:

- Home Assistant: `2026.1.1`
- tuya-local: `2026.2.0`

## Manuals / webpages
- TREO Mini Spotlight (tested): https://edison-led.co.jp/edison-smart/manual/tre02/
- TREO Ceiling Light: https://edison-led.co.jp/edison-smart/manual/tre03/
- TREO Mini Ceiling Light: https://edison-led.co.jp/edison-smart/manual/tre01/
- Dual Ceiling Light (upper light instead of side light): https://edison-led.co.jp/edison-smart/manual/dbc-rf/

## Potential compatibility (not verified)
The following Edison Smart products likely have similar functionality and may work with this profile, but were not physically tested:

- TREO Ceiling Light
- TREO Mini Ceiling Light
- Dual Ceiling Light (upper lighting instead of side lighting)

## Validation
- `pytest tests/test_device_config.py` (passes)
- `yamllint custom_components/tuya_local/devices/edisonsmart_dual_zone_cct_ceiling_light.yaml` (passes)
- `ruff check .` (passes)
- `ruff check --select I .` (passes)
- `ruff format --check .` (passes)


custom_components/tuya_local/devices/edisonsmart_dual_zone_cct_ceiling_light.yaml
```diff
@@ -0,0 +1,55 @@
...
+        name: color_temp
+        type: integer
+        range:
+          min: 0
+          max: 1000
+        mapping:
+          - target_range:
+              min: 2700
+              max: 6500
+      - id: 25
```
make-all: Scene should be a text entity so it can be set by someone who has figured out the format, or captured the raw strings used..

custom_components/tuya_local/devices/edisonsmart_dualzoneceiling_cctlight.yaml
```diff
@@ -0,0 +1,55 @@
...
+              min: 2700
+              max: 6500
+      - id: 25
+        name: scene
+        type: hex
+        optional: true
+      - id: 104
+        type: string
+        name: effect
+        mapping:
```
make-all, line 31: Looking at the range of lights under the TREO range (which have high probability of being the same control module) I think "Direct", "Indirect", "Both" may be better options.

custom_components/tuya_local/devices/edisonsmart_dual_zone_cct_ceiling_light.yaml
```diff
@@ -0,0 +1,55 @@
+name: Dual-zone CCT ceiling light
```
make-all: Device names get prepended in various locations in the UI, so it is better to keep them short, I suggest "Dual zone light". Avoid hyphen, as it becomes indistinguishable from space over when slugified, so can lead to unexpected naming clashes.

make-all (changes requested):

hrko:
Thanks for the review. I’ve updated the device config accordingly.

- Renamed the device to Dual zone light.
- Moved scene (DP 25) out of the light entity and exposed it as a configurable text entity (translation_key: scene, category: config, hidden: true), so raw scene values
  can be set directly.
- Corrected the effect mapping to match the actual behavior:
    - up = Indirect
    - all = Both
    - down = Direct
- Changed work_mode (DP 21) to a select entity, since it takes two known values:
    - white: mode for adjustable brightness and color temperature
    - scene: mode that uses the scene value to run predefined behavior (specific brightness/color temperature and effects such as blinking)

custom_components/tuya_local/devices/edisonsmart_dual_zone_cct_ceiling_light.yaml
```diff
@@ -28,20 +25,35 @@ entities:
...
           - dps_val: up
-            value: UP or SIDE
+            value: Indirect
           - dps_val: all
-            value: ALL
+            value: Both
           - dps_val: down
-            value: DOWN
+            value: Direct
+  - entity: select
```
make-all: Since there are only two valid values, you could make this a "Scene" switch.

hrko: Thanks, I’ve updated DP21 to a `Scene` switch.

Issue #4559, opened 2026-02-23
Request support for Avatto S16Pro: Smart IR+RF Remote Control (Smart IR+RF Remote Control)


kongo09:
### Log message

```text
2026-02-23 17:46:20.806 DEBUG (MainThread) [custom_components.tuya_local.cloud] Found device: 1CH WiFi USB Switch Module CBU
2026-02-23 17:46:20.807 DEBUG (MainThread) [custom_components.tuya_local.cloud] Found device: Thermostat
2026-02-23 17:46:20.807 DEBUG (MainThread) [custom_components.tuya_local.cloud] Found device: Smart IR+RF Remote Control
2026-02-23 17:46:20.807 DEBUG (MainThread) [custom_components.tuya_local.cloud] Found device: Smart IR
2026-02-23 17:46:20.807 DEBUG (MainThread) [custom_components.tuya_local.config_flow] Device count: 4
2026-02-23 17:46:20.808 DEBUG (MainThread) [custom_components.tuya_local.config_flow] Hub count: 1
2026-02-23 17:46:30.292 DEBUG (MainThread) [custom_components.tuya_local.config_flow] Scanning network to get IP address for bf62226efbce5ced5bm6uv.
2026-02-23 17:46:35.347 DEBUG (MainThread) [custom_components.tuya_local.config_flow] Found: {'ip': '192.168.179.177', 'version': '3.3', 'id': '***', 'product_id': 'keym4vvjhx4sd9kk', 'data': {'ip': '192.168.179.177', 'gwId': '***', 'active': 2, 'encrypt': True, 'productKey': 'keym4vvjhx4sd9kk', 'version': '3.3', 'name': '', 'key': '', 'mac': '', 'id': '***', 'ability': 0, 'token': '', 'wf_cfg': '', 'dev_type': 'default', 'origin': 'broadcast'}}
2026-02-23 17:46:41.897 DEBUG (MainThread) [custom_components.tuya_local.device] Refreshing device state for Test
2026-02-23 17:46:41.897 DEBUG (MainThread) [custom_components.tuya_local.device] Setting protocol version for Test to 3.3
2026-02-23 17:46:41.916 DEBUG (SyncWorker_0) [custom_components.tuya_local.device] Test refreshed device state: {"Error": "Invalid JSON Response from Device", "Err": "900", "Payload": "json obj data unvalid", "invalid_json": "json obj data unvalid"}
[...]

PR #4578, opened 2026-02-25
feat (remote): add RF sub-GHz support and fix IR/RF blaster connectivity

kongo09:
## Summary

Adds RF (sub-GHz) command support to the remote entity platform, and fixes a connectivity issue where IR/RF blasters were incorrectly treated as unreachable.

### Device connectivity fix (`device.py`)

IR/RF blasters are fire-and-forget devices that do not respond to status queries. Two failure modes have been addressed:

- **Null state**: when `status()` returns `None`, the device is now marked as reachable with no data, rather than being treated as a connection failure
- **Error 900**: when the device returns a non-JSON response (Tuya error code 900), it is similarly marked as reachable with no data, both in the polling path and the command path

### RF command support (`remote.py`)

- **Learning**: pass `command_type: rf` to `remote.learn_command` to put the blaster into RF learning mode. Learned codes are stored with an `rf:` prefix
- **Sending**: codes with the `rf:` prefix are sent using the RF payload format (`rfstudy_send` with `rf_type: sub_2g`). Inline codes can be passed as `rf:<base64>`
- Fixes a bug where `async_delay_save` was passed a dict instead of a callable

### New device

Adds the **Avatto S16Pro** Smart IR+RF Remote Control to the `s11_rfir_remote` profile.

### Documentation (`README.md`)

- Updated the IR hub paragraph to lead with a positive support statement
- Added a *Using IR/RF blasters* section documenting the learn/send workflow

## Test plan

- [ ] All 40 existing `test_device.py` tests pass
- [ ] Existing `test_remote.py` initialisation tests unaffected
- [ ] Learn an RF command using `command_type: rf` and verify it can be replayed
- [ ] Verify Avatto S16Pro is detected correctly

### DIFF
--- a/ACKNOWLEDGEMENTS.md
+++ b/ACKNOWLEDGEMENTS.md
@@ -1275,3 +1275,4 @@ Further device support has been made with the assistance of users. Please consid
 - [diogolccsilva](https://github.com/diogolccsilva) for contributing support for Beltax BAC-1009 air conditioner.
 - [kong09](https://github.com/kong09) for contributing improvements to thermostat icons.
 - [hrko](https://github.com/hrko) for contributing support for Edison Smart Treo dual zone ceiling light.
+- [akoky](https://github.com/akoky) for assisting with support for Meian SW02 water leak detector.
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -1329,6 +1329,7 @@ entity types as sub devices.
 - LoraTap QCSC400ZB-V2 curtain switch
 - LoraTap SC500ZB-V2 curtain controller
 - Loratap SS9600ZB 6 button remote control
+- Meian SW02 water leak detector
 - Moes BRT-100-TRV thermostat radiator valve
 - Moes dual dimmer module
 - Moes RGBCW lightbulb


kongo09:
This PR would close https://github.com/make-all/tuya-local/issues/4559

kongo09:
For transparency: I used Claude to work on the code. I've manually checked the changes, approved or modified them and tested everything.

kongo09:
Thanks for running the workflows, I'll take a look

kongo09:
The failing Python Test is not due to the PR but also happens on master:

```
[tests (3.13): tests/devices/base_device_tests.py#L142](https://github.com/kongo09/tuya-local/commit/31101a3de6aab3fb4e2217c884abd141bac6a7a4#annotation_46481358689)
TestRGBCWLightbulb.test_entity_category AssertionError: <EntityCategory.CONFIG: 'config'> is not None : text_music is config, expected None
```

The linting problems should be fixed.

kongo09:
Tested now not only on dev machine but also in live environment. I discovered that due to caching there was a problem with learning more than 1 command in a row, but that is fixed now.

custom_components/tuya_local/devices/s11_rfir_remote.yaml
```diff
@@ -8,6 +8,10 @@ products:
     name: RF/IR remote
   - id: 0gawnyfgyneswent
     model: IR Controller Pro
+  - id: x0lyfgjuguuh1vof
```
make-all: Does this model have the extra scene buttons that the S11 has over the basic one, or are you just making this reassignment for naming reasons?

custom_components/tuya_local/device.py
```diff
@@ -504,35 +504,57 @@ def _reset_cached_state(self):
...
+                    )
+                else:
+                    _LOGGER.debug(
+                        "%s protocol error %s: %s",
+                        self.name,
+                        new_state.get("Err"),
+                        new_state.get("Error", "message not provided"),
+                    )
+                if new_state.get("Err") == "900":
+                    # Error 900 means the device responded but sent non-JSON
```
make-all: I don't agree with making an exception for this. Robust error detection is required for detecting the protocol version, treating errors as possible signs of "no data" will result in worse behaviour for many other devices.

custom_components/tuya_local/device.py
```diff
@@ -504,35 +504,57 @@ def _reset_cached_state(self):
...
+                    _LOGGER.debug(
+                        "%s returned error 900; treating as reachable with no data",
+                        self.name,
+                    )
+                    new_state = None
+        else:
+            # Device connected but returned no data (e.g. fire-and-forget devices
+            # like IR/RF blasters that do not respond to status queries).
+            # Mark the cache as updated so the device is considered reachable.
             self._cached_state["updated_at"] = time()
```
make-all: I don't think it is correct to update the timestamp on an empty response. As far as I can see the request should go through already.

custom_components/tuya_local/devices/s11_rfir_remote.yaml
kongo09: Good catch - I was convinced that I saw the scenes in the data. However, spending more time on the Tuya Cloud console and inside the Tuya app and also trying to locally send data to activate scenes all point to the scenes not existing. I'll fix this.

custom_components/tuya_local/device.py
kongo09: You're right, and I've removed the error 900 exception from _refresh_cached_state — that location was wrong because it silently discarded the error before it could inform protocol detection.

However, I've kept equivalent handling in _retry_on_failed_connection, where it's more appropriate. The distinction matters: protocol version mismatches typically produce error 904 ("Unexpected Payload") or 914 ("Check device key/version"). Error 900 specifically ("Invalid JSON Response") is the normal response from IR/RF remote devices to any status query — they never return valid DPS data regardless of protocol version, because they only accept commands.

Without this handling, devices using basic_ir_remote.yaml (and similar) cannot be set up at all: the connection test always fails because these devices will return 900 on every protocol tried. I confirmed this with an Avatto S16 Pro during testing.

custom_components/tuya_local/device.py
kongo09: You're right. I've removed the updated_at update from the null-state branch. Devices that are reachable but return no status data (like IR remotes) do so via error 900, which is handled in _retry_on_failed_connection. A null return from status() means no response at all, and marking that as "reachable" was incorrect.

PR #4581, opened 2026-02-25
Fix ITC-308-WIFI discovery: add product ID and mark DPS optional

dholt:
## Summary
- Adds product ID `rt9nqdxeiqo7dnz9` for newer ITC-308-WIFI units
- Marks all DPS except `104` (current_temperature) as `optional: true`

## Problem
The ITC-308-WIFI only reports DPS `{"104": <value>}` during initial device discovery. Since all other DPS in the config were required (non-optional), the `matches()` function returned `False` due to missing required DPS, and the device was incorrectly matched to `inkbird_ipt2ch_thermostat` instead.

This caused wrong entity mappings — e.g., the temperature setpoint (DPS 106) was mapped to "Low temperature alarm 1" from the IPT-2CH config.

## Fix
Mark all DPS as `optional` except DPS 104, which is the only one reported during discovery. This matches the pattern used by the IPT-2CH config, which successfully discovers because it marks its DPS as optional.

Also adds product ID `rt9nqdxeiqo7dnz9` which ships on newer ITC-308-WIFI units and was not previously listed.

## Device info
- **Device**: Inkbird ITC-308-WIFI (purchased Feb 2026)
- **Product ID**: `rt9nqdxeiqo7dnz9`
- **Discovery DPS**: `{"104": 241}` (only current_temperature reported)
- **Full DPS after setup**: 12, 101, 102, 104, 106, 108, 109, 110, 111, 112, 113, 115, 116, 117, 118

## Testing
- [x] Confirmed device only reports DPS 104 during discovery (from HA logs)
- [x] Verified `inkbird_itc308_thermostat` now appears in device type dropdown
- [x] Current temperature, setpoint, and hvac_action all reporting correctly
- [x] Setpoint control working (set from HA, confirmed on physical device)

### DIFF
--- a/custom_components/tuya_local/devices/inkbird_itc308_thermostat.yaml
+++ b/custom_components/tuya_local/devices/inkbird_itc308_thermostat.yaml
@@ -3,11 +3,16 @@ products:
   - id: B0eP8qYAdpUo4yR9
     manufacturer: Inkbird
     model: ITC-308
+  - id: rt9nqdxeiqo7dnz9
+    manufacturer: Inkbird
+    model: ITC-308-WIFI
 entities:
   - entity: climate
     dps:
       - id: 101
         type: string
+        optional: true
+        force: true
         name: temperature_unit
       - id: 104
         type: integer
@@ -20,6 +25,8 @@ entities:
                 value_redirect: current_temperature_f
       - id: 106
         type: integer
+        optional: true
+        force: true
         name: temperature
         range:
           min: -500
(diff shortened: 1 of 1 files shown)


make-all:
You seem to have done something to this file such as marking it as binary that caused the changes not to show up.

dholt:
Fixed — the previous commit had inadvertently converted the file from CRLF to LF line endings, causing the entire file to show as changed. Force-pushed a clean commit that preserves the original CRLF endings. The diff now shows only the 31 actual additions.

make-all:
The product id has been added separately since this still has a mysterious merge conflict, and some of the changes have risk of causing regression for other matching devices.

A possible cause of the device not showing its data is your protocol version mismatching.

PR #4584, opened 2026-02-26
Add support for Kogan Wi-Fi Blinds Driver

m-sct:


m-sct:
hello @make-all, first time contributing to the repo

could i please confirm what is the process for merging into main? thanks

custom_components/tuya_local/devices/kogan_kasmartblib_blinds_driver.yaml
```diff
@@ -0,0 +1,44 @@
+name: Blind controller
+products:
+  - id: kasmartblib
```
make-all, line 3: This is not the product_id from the Tuya discovery protocol

make-all:
I have modified simple_blinds to fully cover this device while remaining compatible with blinds that do not report current_position.

PR #4585, opened 2026-02-26
Add support for two generic Wi-Fi window openers

m-sct:


m-sct:
hello @make-all, the latest commit should have addressed the issues with the unit test

custom_components/tuya_local/devices/generic_wifi_window_opener_type_1.yaml
```diff
@@ -0,0 +1,30 @@
+name: Generic Wi-Fi window opener type 1
+entities:
+  - entity: cover
+    class: window
+    dps:
+      - id: 1
+        name: work_state
```
make-all, line 7: Just make this optional and one config can cover both.

custom_components/tuya_local/devices/generic_wifi_window_opener_type_1.yaml
```diff
@@ -0,0 +1,30 @@
+name: Generic Wi-Fi window opener type 1
```
make-all, line 1: What makes you say this is generic?

It does not quite follow the usual Tuya allocations for cover functions, which have command on dp id "1", and position on "2", with current_position optionally on "3", and the useless state on "7" (useless because it only indicates "opening" and "closing", and continues to indicate those active states when the action has finished instead of switching to a static state like "opened" or "closed").

PR #4590, opened 2026-02-27
Add support for Caleo C947 floor heating

### LINKED ISSUE #4279: Request support for CALEO C947 floor thermostat
### Log message

```text
Device matches klarstein_wonderwall_irheater with quality of 22%. DPS: {"updated_at": 1767760845.2207594, "1": true, "2": 235, "3": 236, "4": "manual", "9": false, "11": 0, "12": "c", "14": 237, "15": 500, "16": 50, "17": 500, "18": 410, "19": -50, "101": false, "102": 100, "103": false, "104": true, "106": "5_2"}
```

### Product ID

fekemcg4hxekb8gi

### Product Name

CALEO C947 floor thermostat

### DPS information

```text
I added device as "klarstein_wonderwall_irheater" with 3.4 protocol version, but temperature sensors must be divided by 10 (real temp 23.5, but it shows now 235)

**Tested by me DPS:**
"1": power status (true = turned on, false = turned off)
 "2": 235 - target temperature (must be divided by 10)
 "3": 236 - current temperature
 "4": auto\manual\eco mode
 "9": child lock of sensor panel (true = enabled, false = disabled)
 "12": Celsius or Fahrenheit (c\f)
 "101": frost protection mode (true = enabled, false = disabled)
 "102": backlight brightness (0-100)
 "103": current heating status (false = idle, true = heating now)
 "104": night mode (true = enabled, false = disabled)
 "106": work day model (5_2/6_1/7)




```

It is floor thermostat. I want to turn on/off, set temperature, get current temperature and current status (idle or heating now)

sstyle:


custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
```diff
@@ -0,0 +1,81 @@
...
+          min: 0
+          max: 100
+  - entity: switch
+    name: Power
+    category: diagnostic
+    dps:
+      - id: 103
+        type: boolean
+        name: switch
+        readonly: true
```
make-all: A read-only switch does not make sense.
Use a binary_sensor, with `class: power` instead of name.

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
```diff
@@ -0,0 +1,81 @@
...
+        name: temperature
+        precision: 2
+        unit: C
+        range:
+          min: 5
+          max: 30
+        mapping:
+          - scale: 10
+            step: 5
+            range:
```
make-all, line 29: There is no point in setting an incorrect range on the dp level only to override it in the default and only mapping. It is better to set the correct values above and remove this range.

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: heat
+      - id: 2
+        type: float
```
make-all, line 19: Should be an integer, Tuya devices do not generally use floating point.

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
```diff
@@ -0,0 +1,81 @@
...
+          min: 5
+          max: 30
+        mapping:
+          - scale: 10
+            step: 5
+            range:
+              min: 50
+              max: 300
+      - id: 3
+        type: float
```
make-all, line 33: should be integer

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
```diff
@@ -0,0 +1,81 @@
...
+            value: idle
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 9
+        type: boolean
+        name: lock
+  - entity: number
+    name: Brightness
```
make-all, line 63: Use a light entity with brightness dp (and probably `translation_key: display` instead of the explicit icon)

make-all (changes requested):

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
sstyle: this device has 0.5C step, so integer is not possible

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
sstyle: same as before - 0.5C step, so should use mapping

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
sstyle: Also does not.This device has 0.1C step

Image

sstyle (commented):
Done

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
sstyle: Removed

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
sstyle: Sorry, can not understand how to do that. I just copied this one

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
make-all: The range is for the Tuya protocol side only. There are not two separate ranges for the Tuya protocol side and the HA side, there is just one range and it gets scaled by the scale factor for the HA side mapping.
So you should remove this range, and change the range above to `min: 50`, `max: 300`

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
make-all: The device only has a 0.5 step on the HA side after the `scale: 10` in the mapping is applied. On the Tuya side the step is 5, and the representation is an integer.

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
make-all: This is also only on the HA side after the scale is applied. The definition here is for the Tuya protocol side before scaling, where the type is integer.

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
make-all:
```yaml
  - entity: light
    translation_key: display
    category: config
    dps:
      - id: 102
        type: integer
        name: brightness
        range:
           min: 0
           max: 100
```

custom_components/tuya_local/devices/caleo_c947_floorheating.yaml
sstyle: ok, but how to fix this in HA? I saw 217 instead of 21.7

PR #4606, opened 2026-02-28
add icons for thermostat preset modes

kongo09:
Adds per-preset icons to the `thermostat` translation key in `icons.json`, using the `state_attributes` → `preset_mode` → `state` mechanism already in use for fan entities.

All device YAML files using the `thermostat` translation key were audited to identify the full set of preset mode values (11 distinct values across 31 files). Icons were chosen to reflect the intent of each preset:

| Preset | Icon |
|---|---|
| `auto` | `mdi:thermostat-auto` |
| `away` | `mdi:home-export-outline` |
| `comfort` / `comfort_1` / `comfort_2` | `mdi:home-heart` |
| `eco` | `mdi:leaf` |
| `home` | `mdi:home` |
| `manual` | `mdi:hand-back-right` |
| `none` | `mdi:thermostat` |
| `program` | `mdi:calendar-clock` |
| `temp_override` | `mdi:calendar-lock` |

Also adds `mdi:thermostat` as the default entity icon for the `thermostat` translation key, which previously had no default.

Example result:

image


make-all:
The presets defined by the HA climate platform (home, away, comfort, eco) already have icons defined by upstream. Despite some of these being improvements in my opinion, I don't think they should be overridden. none is also an upstream defined preset, it does not have a specific icon defined, but is probably appropriate that it uses the default icon.

kongo09:
Understood.

In my own integration for Philips AirPurifiers I've overridden everything with the Philips icons. But I understand your strategy here.

Let me remove the pre-defined overrides so that only the missing icons remain.

PR #4607, opened 2026-02-28
Improve protocol handling in config flow

kongo09:
When adding a device and the protocol is unknown, the config flow previously defaulted to protocol 3.3. If the device used a different protocol — such as 3.4 or 3.5 — the connection test would fail with a generic error and no guidance.

### Changes

**Default protocol changed to `auto`** (`config_flow.py`)\
The fallback default is now `auto` instead of `3.3`. When cloud or local data is available, the identified protocol version still takes precedence. This only affects a setup or cases where the version is unknown.

**Improved connection error message** (`translations/en.json`)\
The error shown when connection fails now includes: *"If you selected a specific protocol version, try 'auto' instead."*

**Fixed `auto` mode protocol detection** (`config_flow.py`)\
`auto` mode previously reused the same `tinytuya.Device` object across all protocol attempts. This caused protocol 3.4/3.5 handshakes to fail silently: the shared object carries stale session state from the earlier 3.3/3.1/3.2 attempts. The connection test now creates a fresh device object for each protocol and closes the socket between attempts. On success, the detected protocol is saved to the config entry so future HA restarts connect directly without re-probing.

**Suppressed alarming log messages during protocol probing** (`device.py`)\
Connection failures for devices named `"Test"` (used internally during config flow testing) are now logged at `DEBUG` level instead of `ERROR`, avoiding misleading error messages in the HA log during normal auto-detection.

### Tests

The code has been tested in a live production environment and confirmed to work in practice.


make-all:
This was the default some time ago, and it led to support headaches because everybody leaves it at the default, and 'auto' is really only a last resort thing that has a number of undesired side-effects when it comes to error handling and recovery.

I do not want to go back to that, and I do not want to promote use of 'auto'.

kongo09:
Understood.

Would it be acceptable for you to stick to 3.3 as default, but include the fixes of the auto mode (it doesn't not work reliably right now) and improve the error message? I would change the PR then.

make-all:
Suppressing the error message would also not be good, as it makes it difficult to find the reason for failures during initial connection.

But the changes to improve the detection would be useful because it seems the newer protocols do have a tendancy to get the device stuck so it never connects.

make-all:
Actually your idea of saving the detected protocol version seems good, and may make it acceptable to default to auto if the issues with detection of newer protocols is fixed. But I think there should be some feedback to the user on the next screen, so they are not surprised by a different value having been saved that what they set.

kongo09:
No, if you set a specific value there is no fallback to auto. It will fail but the error message will suggest that auto might be an alternative to try. 

So there should not be any user surprise.

custom_components/tuya_local/config_flow.py
```diff
@@ -349,6 +349,16 @@ async def async_step_local(self, user_input=None):
             self.device = await async_test_connection(user_input, self.hass)
             if self.device:
                 self.data = user_input
+                # If auto mode found a working protocol, save it so future
+                # HA restarts connect directly without re-cycling all versions.
+                if (
```
make-all, line 355: This code does not do anything, because device._protocol_configured will always retain what was configured.

Originally I left "auto" as "auto" in the config so devices would continue working after a firmware upgrade. But firmware upgrades are rare, and "auto" does not work as reliably as originally hoped, and causes other undesirable side effects like turning a 20 second reconnection interval into a 10 minute interval, so I am OK with only using "auto" for the detection during initial config, and saving the actual discovered protocol, but as mentioned in the previous comment, I think the user should be informed of this as they do the configuration so they are not surprised by auto not working long term.

custom_components/tuya_local/device.py
```diff
@@ -652,7 +652,10 @@ async def _retry_on_failed_connection(self, func, error_message):
                         self._api_protocol_working = False
                         for entity in self._children:
                             entity.async_schedule_update_ha_state()
-                    if self._api_working_protocol_failures == 1:
+                    if (
```
make-all: As I mentioned in the previous comment, I don't think this change is a good idea, as it will hide the details of connection failures during the initial config when they are most needed.

But maybe a narrower condition can be used, such as only hiding error code 914 when `self._protocol_configured == "auto"`

custom_components/tuya_local/config_flow.py
make-all: I seem to have missed that due to the use of separate devices for each protocol attempt, the devices are being created with individual protocol versions. So the check will work, only the user feedback is missing.

kongo09: I've added a message to the user informing about the auto-detected protocol:

Image

This will obviously only be visible in the English language. I'm not sure if you would be happy to add this to the other languages with auto-translation?

kongo09: Without auto-detection, you see:

Image

custom_components/tuya_local/device.py
kongo09: Ok, I'll narrow it down

kongo09:
The failing tests seem to be related to something else. However, there are warnings about the missing translations. What's your policy? Just leave that open and wait for the community to fix it, or use AI to auto-translate?

kongo09:
Auto-detection tested with four different devices across protocol versions 3.1, 3.3 and 3.4 including an IR blaster that doesn't reply on status. All work flawlessly now while before some got stuck. Maybe it's worth trying this as a default again to find out where it breaks.

make-all:
Yes, defaulting to auto should be OK, since it is now only used during initial setup, and the actual detected protocol will get saved in the config registry. This will avoid most of the issues with using auto.

PR #4609, opened 2026-02-28
Show manufacturer and model in config flow and HA device registry

kongo09:
The device type selector in the config flow previously showed only the raw
YAML filename (e.g. `hysen_hy101rf_thermostat`), and the HA device page always
showed `"Tuya"` as the manufacturer with no model information — even when the
YAML file contains this data.

### Changes

**Richer device type selector** (`config_flow.py`, `helpers/device_config.py`)\
The type selector now shows one option per distinct manufacturer/model
combination found in the YAML's `products` list:

> `Hysen HY101RF (hysen_hy101rf_thermostat)`\
> `Kekotek KTF017802 (hysen_hy101rf_thermostat)`

When a product entry has no manufacturer or model, the YAML device type name
is used as a fallback:

> `Thermostat (some_thermostat)`

The `config_type` in parentheses ensures every option is always unique,
regardless of future community additions to the device library.

The user's selection is stored in the config entry as separate `manufacturer`
and `model` fields, which are then used directly by the device registry — no
runtime lookup needed.

**HA device registry** (`device.py`)\
The device page now shows the correct manufacturer and model selected during
setup. When no specific product was chosen, the YAML device type name
(e.g. `"Thermostat"`) is used as the model fallback.

**Backward compatible**\
Existing config entries without stored manufacturer/model continue to work
unchanged: manufacturer falls back to `"Tuya"` and model is left unset.

### Files changed

- `const.py` — new `CONF_MANUFACTURER` and `CONF_MODEL` constants
- `helpers/device_config.py` — new `product_display_entries()` method on `TuyaDeviceConfig`
- `config_flow.py` — updated `async_step_select_type` and `async_step_choose_entities`
- `device.py` — updated `TuyaLocalDevice.__init__`, `device_info`, and `setup_device`

### Tests

The changes have been tested in a production environment.

### DIFF
--- a/ACKNOWLEDGEMENTS.md
+++ b/ACKNOWLEDGEMENTS.md
@@ -1273,7 +1273,7 @@ Further device support has been made with the assistance of users. Please consid
 - [mxwmnn](https://github.com/mxwmnn) for contributing improvements to ZG-205W mmWave presence sensors.
 - [BarRaider](https://github.com/BarRaider) for assisting with support for Dreamlight RGBIC LED strip.
 - [diogolccsilva](https://github.com/diogolccsilva) for contributing support for Beltax BAC-1009 air conditioner.
-- [kong09](https://github.com/kong09) for contributing improvements to protocol detection, thermostat icons, support for RF remote controls.
+- [kongo09](https://github.com/kongo09) for contributing improvements to protocol detection, thermostat icons, support for RF remote controls.
 - [hrko](https://github.com/hrko) for contributing support for Edison Smart Treo dual zone ceiling light.
 - [akoky](https://github.com/akoky) for assisting with support for Meian SW02 water leak detector.
 - [rwblokzijl](https://github.com/rwblokzijl) for assisting with support for Haier air-water combo heat pump.
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ Although this is documented in the BLE lock documentation from Tuya, Zigbee
 and WiFi locks often use the same naming for datapoints, which may be
 compatible with this scheme.
 
-## Using IR/RF blasters
+## IR/RF blasters
 
 Tuya IR and RF blasters are exposed as remote entities and support learning and
 sending commands via the standard Home Assistant remote services.
@@ -286,6 +286,12 @@ used when learning. You can also send codes directly without learning first:
 
 - **IR inline code**: prefix with `b64:` followed by the base64-encoded IR code
 - **RF inline code**: prefix with `rf:` followed by the base64-encoded RF code
+
+### UI
+
+If you would like to expose the learnt commands as buttons in the user interface
+you might want to take a look at the [Remote buttons](https://github.com/kongo09/remote_buttons)
+integration, which is compatible with Tuya Local.
 
 ## Contributing
 


make-all:
I think it is better to only use the product information when it is confirmed by the product id. Currently this is available only during the cloud config flow, but we could also do the local discovery during the manual flow and grab product id and IP address there as well.

When there is a match, we should show that product's details as the default selected config, and save the manufacturer and model in the registry. We could also use local discovery to update existing registry entries.

This will not work on setups that split devices onto a different subnet than the HA instance is running on, so some fallback is needed, but rather than showing multiple entries for every config with product names that still might not match, my idea for some time has been to show a summary of the device's state alongside the config name.

eg for lights, show switch state, brightness scaled to %, color temperature, for climate show hvac_mode, temperature, current temperature, preset. Basically each entity implementation would have a "summary" function that returns a short line of text to use in this display, and maybe elsewhere in logs.

kongo09:
I can certainly restrict the device info to cases where we have a match with a product id. But what would you expect to see in the other cases? Right now, it's the name of the yaml file, which might be confusing, as it might even relate to a different manufacturer. I was assuming that a user knows the device they're trying to integrate by its name. To be able to see that in the list should give confidence. What's your proposal I do here?

On your second point with local discovery - that sounds like a no-regret move, even if it won't work in all cases. I'm happy to add that. Should I do this here, or in a separate PR?

Your third idea is interesting in order to identify the right device, especially if you have several devices from the same manufacturer and the same model. However, this feels like a bigger exercise, especially if you want to cover a broad range of entities. I think that's honestly outside of the scope of this PR.

Overall, I'd suggest we align on some clear guidance for your first question and I'll work on that. Once we have that working, I can take a look at the local discovery for the manual setup.

make-all:
In the cases where we don't have a product id match, there is no guarantee that listing the products will be any less confusing for the user than the filename, as their product is likely not covered by the products already.

For discovery, I think another PR is better.

kongo09:
Ok, so I'll change this as follows:
- product can be identified by product id -> list the naming as proposed here
- product is only a guess -> list the config filename

Everything else goes to a separate PR

kongo09:
Ok, so if you configure via manual setup and there is no product id, then it shows all the matching yaml configs with all products inside. Chances are, you recognize your manufacturer. If not, you have to pick something that "feels" right. We could add text to the README that encourages users who found a working device that is not matching their manufacturer to report this.

image

In this example screen, you see three thermostats listed, all from the same yaml device file. Mine happens to be the second one, so I would happily select that.

If you don't like the idea of a user selecting the wrong manufacturer and model _knowingly_, we could also add one "generic" entry to each yaml that never has a product id. Or simply present a generic device in the selection box.

Now for the case of the cloud configuration where we have product ids. After the search for a matching model, the code checks if we have a perfect fit of product ids and then only shows those devices in the selection box that match. The DP matches are no longer shown. In theory, there could be a device that has the same product id as another device but is itself listed already without product id. In that case, it wouldn't be shown. Personally, I feel this is rather an edge case and again, we could encourage users to report working devices that where they had to select the wrong manufacturer.

This is how it looks for my thermometer, where two manufacturers have the same product id:

image

To me that feels very user friendly.

kongo09:
I wonder if in the second case we might want to leave out the name of the yaml file? What's your preference?

kongo09:
In case of a generic device where just the DPs match, the device info will also be generic. I could add the description from the yaml, if you like:

image

In case the product id matches, the device info is more meaningful:

image

kongo09:
The failing tests seem unrelated

Issue #4594, opened 2026-02-27
Devices permanently stuck in "Failed setup, will retry: tuya-local device offline" after network/power outage never recover without remove and re-add


TheDarkPreacher:
### Description

After any network interruption or power outage, all Tuya Local devices become permanently stuck in "Failed setup, will retry: tuya-local device offline" and never recover on their own. Neither reloading the integration nor restarting Home Assistant resolves the issue. The only remedy is to remove and re-add every affected device individually — with identical connection information — after which they work fine.
This issue appears related to #1804, which was marked as fixed, but the problem persists.

### Steps to reproduce

Have one or more tuya-local devices set up and working normally
Cause a network interruption or power outage (even briefly)
Observe devices enter "Failed setup, will retry: tuya-local device offline" state
Wait — devices never recover
Try reloading the integration — devices still do not recover
Try restarting Home Assistant — devices still do not recover
Remove and re-add a device with identical IP, device ID, and local key — device works immediately

### Expected behaviour

After a network interruption, devices should continue to retry connection periodically until they succeed, regardless of how many initial failures occurred.

### Additional context

Environment

All devices have static IPs via DHCP reservation, so IP changes are not a factor
20 devices across a mix of lights, switches, and plugs
All using protocol version 3.3

Log Evidence
Debug logging shows two distinct error codes depending on timing:
`Error 905 — Device Unreachable (device hasn't fully come back online yet):
refreshed device state: {"Error": "Network Error: Device Unreachable", "Err": "905", "Payload": null}
protocol error 905: Network Error: Device Unreachable
Retrying after exception AttributeError Network Error: Device Unreachable (0/3)
Setting protocol version to 3.3`

`Error 901 — Unable to Connect (persists indefinitely even after devices are back online):
refreshed device state: {"Error": "Network Error: Unable to Connect", "Err": "901", "Payload": null}
protocol error 901: Network Error: Unable to Connect
Retrying after exception AttributeError Network Error: Unable to Connect (0/3)
Retrying after exception AttributeError Network Error: Unable to Connect (1/3)
Retrying after exception AttributeError Network Error: Unable to Connect (2/3)`

[...]

PR #4611, opened 2026-03-01
fix powercycle issue

miicha:
my lamp does not become responsive in HA after power off until I issue some commands from tuya app. this change forces an updatedps for selected dps (usualla 20 and/or 21 should be enough and can be set in corresponding device yml to true)
this is related to https://github.com/make-all/tuya-local/issues/4594


make-all:
You should probably set the protocol to 3.22 if an updatedps makes some difference to its behaviour. Sending an updatedps command to devices that don't support it can cause them to crash.

custom_components/tuya_local/device.py
```diff
@@ -375,6 +375,22 @@ async def async_receive(self):
...
+                                )
+                                if self._force_dps:
+                                    await self._hass.async_add_executor_job(
+                                        self._api.updatedps,
+                                        self._force_dps,
+                                    )
+                                else:
+                                    await self._hass.async_add_executor_job(
+                                        self._api.updatedps,
+                                    )
```
miicha: I would just remove this and "outsource" the decision, if updatedps should happen to the device config. Does this make sense? For my lamp the updatedps is needed and tinytuya says it is version 3.5. If I set it to 3.22 it no longer works in my python test script.

custom_components/tuya_local/device.py
```diff
@@ -375,6 +375,22 @@ async def async_receive(self):
                                 self.name,
                                 poll["Payload"],
                             )
+                            # Device DPS not initialized (e.g. after power
+                            # cycle) — nudge with updatedps to force init
+                            if "unvalid" in str(poll["Payload"]):
```
make-all: This seems like a strange condition. I think you should be looking at Err rather than the Payload to determine if the error condition is one that might be helped by calling updatedps

custom_components/tuya_local/device.py
make-all: The question does not make sense to me (I don't know what you mean by "outsource" the decision).

miicha: I meant that it will only be performed if there is force: true set for a dps in the device.yaml

miicha:
just tried another approach and fixed initialization as well. this one needs device settings loaded earlier to evaluate the dps with force = true

custom_components/tuya_local/device.py
```diff
@@ -375,6 +375,27 @@ async def async_receive(self):
...
                             )
+                            # Device DPS not initialized (e.g. after power
+                            # cycle) — nudge with a CONTROL command to force
+                            # DPS registration.  Unlike query commands
+                            # (status/updatedps), a set_value (cmd 7) makes
+                            # the firmware register its datapoints.
+                            if "unvalid" in str(poll["Payload"]) and self._force_dps:
+                                nudge_dp = self._force_dps[0]
+                                cached = self._get_cached_state()
+                                nudge_val = cached.get(str(nudge_dp), False)
```
make-all, line 386: I think you should at least check the format of the dp and send the correct type. But sending any value other than the current will alter the device's state, so I don't think it is a good idea to do it this way rather than with updatedps which should not alter the state.

make-all:
I think this change as submitted has high probability of causing more problems than it solves. If you revert it to the updatedps version, it might be able to be reconsidered.

make-all (changes requested):

custom_components/tuya_local/device.py
miicha: I need to test it for a while locally. I switched to this approach because updatedps was not really solving the problem. I hope to find some time for testing this weekend.

make-all:
Closing, as this is now over 2 months old, needs conflict resolution and other changes have been made in the meantime that may fix the original issue.

PR #4612, opened 2026-03-01
Add protocol 3.5 dual lamp

miicha:
add some additional control options for this (and possibly other similar ones) https://www.amazon.de/dp/B0DGQ6WJDB with separate control of RGB and White LED.


custom_components/tuya_local/device.py
```diff
@@ -375,6 +375,22 @@ async def async_receive(self):
                                 self.name,
                                 poll["Payload"],
                             )
+                            # Device DPS not initialized (e.g. after power
+                            # cycle) — nudge with updatedps to force init
+                            if "unvalid" in str(poll["Payload"]):
```
make-all, line 411: poll.get("Err") == 900 seems to be the condition you want here based on logs in issues received with the word "unvalid" in them. This should be more efficient and more reliable in case the underlying dummy text in the payload for this condition is changed by a device manufacturer.

custom_components/tuya_local/devices/rgbcw_ceiling_light_v35.yaml
```diff
@@ -0,0 +1,190 @@
+name: RGBCW ceiling light
```
make-all, line 1: I think the name, and more so the filename, are deceptive, because this is not a normal RGBCW light, it is separate RGB and CCT (CW) lights. The use of dp ids in the 100-199 range for core functions is not generic, so probably using the manufacturer in the filename is more appropriate (but name here should remain generic in case of rebranding, maybe just add a space between RGB and CW).

custom_components/tuya_local/devices/rgbcw_ceiling_light_v35.yaml
```diff
@@ -0,0 +1,190 @@
...
+products:
+  - id: 6qpsweogzjmfvoyg
+    manufacturer: Horevo
+    name: RGBCW ceiling light v3.5
+entities:
+  - entity: switch
+    name: Master power
```
make-all, line 8: Leave out this name, it should be obvious from this being the main switch and separated from the lights that it is for the whole device.

custom_components/tuya_local/devices/rgbcw_ceiling_light_v35.yaml
```diff
@@ -0,0 +1,190 @@
...
+    dps:
+      - id: 20
+        type: boolean
+        name: switch
+  - entity: light
```
make-all, line 13: Consider adding the following to make it obvious that the switch above overrides the on/off switch here:

```yaml
    - id: 20
      type: boolean
      name: available
```

custom_components/tuya_local/devices/rgbcw_ceiling_light_v35.yaml
```diff
@@ -0,0 +1,190 @@
...
+            range:
+              min: 0
+              max: 1000
+          - name: v
+            bytes: 2
+            range:
+              min: 0
+              max: 1000
+  - entity: select
+    name: Scene
```
make-all, line 65: This should use a translation_key instead of name (scene is lowercase in that case). I'm not sure why the CI has not flagged this.

custom_components/tuya_local/devices/rgbcw_ceiling_light_v35.yaml
```diff
@@ -0,0 +1,190 @@
...
+        optional: true
+        unit: s
+        mask: "0000000000FFFF"
+        range:
+          min: 0
+          max: 10000
+        mapping:
+          - scale: 1000
+  - entity: text
+    name: Scene data
```
make-all, line 151: Use `translation_key: scene` here. Because it is a different entity type than the select above, the names do not have to be unique, and it makes sense to have them the same because they are exactly the same setting, just more control here.

custom_components/tuya_local/devices/rgbcw_ceiling_light_v35.yaml
```diff
@@ -0,0 +1,190 @@
...
+    category: config
+    hidden: true
+    dps:
+      - id: 109
+        type: base64
+        name: value
+        optional: true
+  - entity: switch
+    name: Unknown 34
```
make-all, line 160: Unknown dps should generally just be added as attributes on other entities for observation, not made into full entities by themselves.

custom_components/tuya_local/devices/rgbcw_ceiling_light_v35.yaml
```diff
@@ -0,0 +1,190 @@
...
+      - id: 109
+        type: base64
+        name: value
+        optional: true
+        mask: "0000FF0000000000000000000000"
+        range:
+          min: 1
+          max: 100
+  - entity: select
+    name: Power-on behaviour
```
make-all, line 108: dps 33 and 35 should probably have text entities also, to allow full control.

make-all:
Draft seems abandoned

PR #4613, opened 2026-03-02
Arlec PC191HA Series 4 smart plug

TiSocks:
New device yaml for Arlec PC191HA Series 4 smart plug with power monitoring.


make-all:
What is special about this smartplug that it needs its own config file?

It seems to match `smartplugv2_energyv3.yaml` already

TiSocks:
The device did not show up as a plug untill I created the config file.

make-all:
Device has been added to smartplugv2_energyv3.yaml and some more dps made optional to accommodate it (though without a log I cannot confirm that)

PR #4621, opened 2026-03-03
Add support for Arlec FAN ATF4601HA

woutervb:
An older version of the Arlec smartfan, 4 speed, oscilation, temperature, natural flow

### DIFF
--- a/custom_components/tuya_local/devices/arlec_atf4601ha_fan.yaml
+++ b/custom_components/tuya_local/devices/arlec_atf4601ha_fan.yaml
@@ -32,6 +32,7 @@ entities:
         type: boolean
   - entity: select
     translation_key: timer
+    category: config
     dps:
       - id: 101
         type: string
@@ -74,6 +75,7 @@ entities:
   - entity: sensor
     translation_key: time_remaining
     class: duration
+    category: diagnostic
     dps:
       - id: 23
         type: integer


custom_components/tuya_local/devices/arlec_fan_atf4601ha.yaml
```diff
@@ -0,0 +1,84 @@
+name: Fan
+products:
+  - id: bf5e0aadb629a31094y5bc
```
make-all: This does not appear to be a product id, WiFi devices typically have a 16 character product id, it is reported by `tinytuya scan` or the `Query Device Info` cloud API call.

custom_components/tuya_local/devices/arlec_fan_atf4601ha.yaml
```diff
@@ -0,0 +1,84 @@
...
+            value: "9h"
+          - dps_val: "10h"
+            value: "10h"
+          - dps_val: "11h"
+            value: "11h"
+          - dps_val: "12h"
+            value: "12h"
+  - entity: sensor
+    class: temperature
+    category: diagnostic
```
make-all: It is better to leave `category: diagnostic` out for sensors unless they are really of little use. There are many configs not following this for historical reasons, but recent versions of HA anyway split sensors into their own category without any categorization by the integration.

custom_components/tuya_local/devices/arlec_fan_atf4601ha.yaml
woutervb: Updated the id, output from tinytuya scan was
```
Unknown v3.3 Device   Product ID = kccccitzawmgkwkh  [Valid Broadcast]:
```

custom_components/tuya_local/devices/arlec_fan_atf4601ha.yaml
woutervb: Removed all references. Was using the 12 speed fan version as a reference

PR #4626, opened 2026-03-04
Add device support for ednet Smart Plug

kongo09:
### Summary

Adds a new device configuration for the **ednet Smart Plug** (product ID `A0F6MxOOAt2r3OoJ`, category `cz`).

### Device capabilities

- **Switch** (DP 1) — on/off control
- **Countdown timer** (DP 2) — configurable duration (0–86400 seconds) as a number entity with `duration` device class

### Notes

- Uses protocol version 3.1
- DPs confirmed via tinytuya device query
- No existing device file matched this DP layout (similar `simple_switch_timer` files use different DPs for the countdown)


make-all:
This is the same as `mirabella_genio_usb.yaml`, just add the product details to that config

kongo09:
I might have missed this as it is using a `time` entity instead of a `number` entity. The problem is that this renders as a clock input field with am/pm selection in the interface of HA. Other device files use `number` instead. In addition, it is missing the class `duration` which `time` doesn't have. [But `number` has that device class exactly for that purpose](https://developers.home-assistant.io/docs/core/entity/number).

So I'm not sure if the `mirabella_genio_usb.yaml` truly represents what this device actually is.

I see three options
- accept the `time` confusion and simply merge the ednet plug product id into the mirabella file
- change the timer to `number` with device class `duration` and merge the ednet plug into the file
- leave the current definition unchanged to not break anything, but do the right thing for the new ednet plug (which was what my PR tried to propose)

What do you think?

make-all:
If you set your preferences to 24h time, then the time entities work perfectly. A request was already made to HA developers to allow the integration to signal to the UI that the display should use 24h regardless of user preference. 

`class: duration` on a number entity makes no difference to the UI, which still requires either setting the timer in seconds, or scaling the number to minutes or hours and having it show decimal fractions.

The other devices that use number entities are not 24h timers (the time entity cannot be limited in its range, so is only suitable for 24h timers).

kongo09:
Thanks for clarification. I'll merge the product in as you suggested. Please reopen.

PR #4637, opened 2026-03-05
Create dewall_evcharger_7kw.yaml

jamesa1994:
Added Depow Wallbox 7kw - 
Metrics json exposed into HA for power monitoring. 
Clamps on supported currents.

### DIFF
--- a/custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
+++ b/custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
@@ -1,0 +1,159 @@
+name: DEPOW 7kW EV Wallbox
+products:
+  - id: gxrtu5vljdthtd3g
+    manufacturer: DEPOW
+    model: 7kW EV Wallbox
+entities:
+  # --- Main Status & Attributes ---
+  - entity: sensor
+    translation_key: status
+    class: enum
+    category: diagnostic
+    dps:
+      - id: 109
+        type: string
+        name: sensor
+        optional: true
+        mapping:
+          - dps_val: SLEEP
+            value: sleep
+          - dps_val: IDLE
+            value: idle
+          - dps_val: IDLEINS
+            value: plugged_in
+          - dps_val: WORKING
+            value: charging
+          - dps_val: WAIT
+            value: waiting
+          - dps_val: ERRORPAUSE
+            value: fault
+          - dps_val: PAUSE
+            value: paused
+          - dps_val: STOP
+            value: charged
+      - id: 101
+        type: integer
+        name: work_state_code
+        optional: true
+
+  # --- Core Controls ---
+  - entity: switch
+    name: Charge Control
+    icon: "mdi:ev-station"
+    dps:
+      - id: 140
+        type: boolean
+        name: switch
+        optional: true
+
+  - entity: number
+    name: Charge Current
+    category: config
+    class: current
+    dps:
+      - id: 150
+        type: integer
+        name: value
+        unit: A
+        range:
+          min: 6
+          max: 32
+      - id: 152
+        type: integer
+        name: maximum
+        optional: true
+      - id: 107
+        type: string
+        name: steps
+        optional: true
+
+  # --- Device Commands ---
+  - entity: button
+    name: Reset Charger
+    icon: "mdi:refresh"
+    category: config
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
```diff
@@ -0,0 +1,159 @@
+name: DEPOW 7kW EV Wallbox
```
make-all, line 1: Top level name should be without branding, in case other models match the same config.

custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
```diff
@@ -0,0 +1,159 @@
...
+products:
+  - id: gxrtu5vljdthtd3g
+    manufacturer: DEPOW
```
make-all, line 4: This does not match the brand in the filename. Also the filename has the device type and model specification swapped from the usual brand_model_type convention.

custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
```diff
@@ -0,0 +1,159 @@
...
+    name: Charge Control
+    icon: "mdi:ev-station"
+    dps:
+      - id: 140
+        type: boolean
+        name: switch
+        optional: true
+
+  - entity: number
+    name: Charge Current
```
make-all, line 50: Leave out the name, as it shadows the translation for "current" class, and the name is not really adding info that is not obvious to the user from the device's primary function.

custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
```diff
@@ -0,0 +1,159 @@
...
+          - dps_val: STOP
+            value: charged
+      - id: 101
+        type: integer
+        name: work_state_code
+        optional: true
+
+  # --- Core Controls ---
+  - entity: switch
+    name: Charge Control
```
make-all, line 41: Probably this represents the primary function of the device, so does not need a name.

custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
```diff
@@ -0,0 +1,159 @@
...
+        name: maximum
+        optional: true
+      - id: 107
+        type: string
+        name: steps
+        optional: true
+
+  # --- Device Commands ---
+  - entity: button
+    name: Reset Charger
```
make-all, line 72: Entity and device names should follow the HA naming convention of only capitalizing the first word.

custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
```diff
@@ -0,0 +1,159 @@
...
+    icon: "mdi:refresh"
+    category: config
+    dps:
+      - id: 141
+        type: boolean
+        name: button
+        optional: true
+
+  - entity: button
+    name: Reboot Charger
```
make-all, line 82: This name should be removed, as the class describes the function already.

custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
```diff
@@ -0,0 +1,159 @@
...
+    category: config
+    dps:
+      - id: 156
+        type: boolean
+        name: switch
+        optional: true
+
+  # --- Raw JSON Payloads ---
+  - entity: sensor
+    name: Raw Metrics JSON
```
make-all, line 131: Such values are not really suited for sensors, they can be just an attribute on another related entity.

custom_components/tuya_local/devices/dewall_evcharger_7kw.yaml
```diff
@@ -0,0 +1,159 @@
...
+    icon: "mdi:history"
+    dps:
+      - id: 105
+        type: string
+        name: sensor
+        optional: true
+
+  # --- Misc Diagnostics ---
+  - entity: text
+    name: Language Config
```
make-all, line 152: Most likely this is a fixed list of languages, normally a select entity with `translation_key: language` would be used for that.

make-all (changes requested):

PR #4643, opened 2026-03-06
Add support for **YU HOME Yu Turbo laundry rack**

lweikit:
## Summary
Add support for **YU HOME Yu Turbo laundry rack** by introducing a new device config:

- `custom_components/tuya_local/devices/yuhome_laundry_rack.yaml`

This profile covers lift control, lighting, disinfection, wind dry, hot dry, timers, and remaining-time sensors based on the device DPS set.

## Device Reference
Official product page:
- https://yuhome.com.sg/products/automated-laundry-system-yu-turbo

Brand site:
- https://yuhome.com.sg

## DPS Mapping Implemented
- `1` boolean: main power switch
- `2` enum (`up/down/stop`): cover control
- `3` integer (`0-100`): cover position
- `4` boolean: light
- `5` boolean: disinfection (`translation_key: uv_sterilization`)
- `6` boolean: wind dry
- `7` boolean: hot dry
- `8` enum (`up/down/stop`): work state -> cover action
- `9` enum (`1h..12h/cancel`): disinfection timer select
- `10` enum (`1h..12h/cancel`): hot dry timer select
- `11` enum (`1h..12h/cancel`): wind dry timer select
- `12` integer (min): hot dry remaining sensor
- `13` integer (min): wind dry remaining sensor
- `14` integer (min): disinfection remaining sensor

## Notes
- Reused existing translation/icon keys where possible (`timer`, `uv_sterilization`), so no translation or icon file changes were required.
- Added as a new device config instead of extending existing drying-rack configs because required DPS shape differs.

## Validation
- `ruff check .`
- `ruff check --select I .`
- `ruff format --check .`
- `yamllint custom_components/tuya_local/devices`
- `pytest tests/test_device_config.py` (passed in isolated Python 3.12 environment)


custom_components/tuya_local/devices/yuhome_laundry_rack.yaml
```diff
@@ -0,0 +1,196 @@
+name: yuhome laundry rack
```
make-all: Top level name should be generic. Instead you should have a products secstion here with the product details linked to a product_id (from `tinytuya scan` or the Tuya developer portal info)

custom_components/tuya_local/devices/yuhome_laundry_rack.yaml
```diff
@@ -0,0 +1,196 @@
...
+entities:
+  - entity: switch
+    name: Power
```
make-all: Power is not needed as a name, just the switch with default naming should be enough.

lweikit: updated @make-all

custom_components/tuya_local/devices/yuhome_laundry_rack.yaml
lweikit: updated @make-all

PR #4646, opened 2026-03-06
Fix protocol version dropdown type mismatch

### LINKED ISSUE #4645: Unable to select protocol other than "Auto"
### Description

When adding a new device, or editing an existing device, selecting a protocol other than "Auto" results in an empty protocol box and adding/editing the device fails.
This happens for all new devices and all existing devices.

### Steps to reproduce

1. Add a new device
2. Select protocol other than "Auto"

### Expected behaviour

Possibility to select protocols other then "Auto"

### Additional context

No messages in the log for Tuya Local or HA regarding Tuya Local.

I did update HA to 2026.3.0 yesterday. Not sure if this has anything to do with that, I haven't tried to add/change a device for a while now.

### Confirmation

- [x] This report is for tuya-local, **not** for localtuya.
- [ ] If this is for a specific device, the device config filename is mentioned above, or diagnostics are included.

kongo09:
The protocol version dropdown (`vol.In`) used float values from `API_PROTOCOL_VERSIONS`, but defaults were sometimes strings — causing the selected value not to match any option and breaking pre-selection.

### Changes
- **config_flow.py**: Convert all dropdown options and defaults to strings. Convert back to float after form submission (in both `ConfigFlowHandler` and `OptionsFlowHandler`).
- **test_config_flow.py**: Update options flow test to submit protocol as string. Add regression test that verifies the schema accepts string versions and rejects floats.

Closes #4645

### DIFF
--- a/tests/test_config_flow.py
+++ b/tests/test_config_flow.py
@@ -262,6 +262,36 @@ async def test_flow_user_init(hass, mocker):
 
 
 @pytest.mark.asyncio
+async def test_flow_user_init_protocol_options_are_strings(hass, mocker):
+    """Test that protocol version dropdown uses strings, not floats."""
+    result = await hass.config_entries.flow.async_init(
+        DOMAIN, context={"source": "local"}
+    )
+    schema = result["data_schema"]
+    # Validate that string protocol versions are accepted
+    schema(
+        {
+            CONF_DEVICE_ID: "test",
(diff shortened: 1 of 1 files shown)


make-all:
Thanks, I was thinking that HA must have stopped accepting non-string values in their change to use a common ha-dropdown in the frontend (previously it seems config flows were using a separate UI component from select entities which had more relaxed rules).

PR #4653, opened 2026-03-07
add wake_up_light_ii, very similar to moes_zcjk_alarmclock

avg-I:
This Wake Up Light II seems to be either a newer version or a variant of Moes Smart Wake Up Light.
As such, the configuration is almost an exact copy of moes_zcjk_alarmclock.yaml.

The only difference is related to the ambient light feature. While the older device has white color for mode 1, my device loops through the colors in that mode.
Also, a couple of other modes have slightly different color names (not sure about actual colors) according to Smart Life app:
- lime -> green;
- magenta -> purple.

I also decided to define color modes as effects rather than named colors.  First reason is that mode 1 (looping) cannot be fit into the named color scheme.  Second reason is that named colors imply that the light supports color (and brightness) control while having presets for select colors.  The reality is that the colors are fixed, there is no real color (and brightness) control.

The only benefit of named color approach that I can think of was that the presets had the matching colors.
 Other than that, everything else was incorrect and misleading.

#### More about the product.

The original product link does not work now.
Here is a very similar product: https://www.aliexpress.com/item/1005010230164271.html

The device itself and its packaging do not have any brand information.
But in online materials there are some references to Moes.

Product identification data:
```json
      "model": "ACA002-II-WWA",
      "name": "Wake Up Light II",
      "product_id": "ya39qahrcdhvxowz",
      "product_name": "WWA-Wake Up Light II",
```

Its DPs are in the attached file.
[tuya-smartlife-wakeup-light.dps.json](https://github.com/user-attachments/files/25817268/tuya-smartlife-wakeup-light.dps.json)

### DIFF
--- a/custom_components/tuya_local/devices/wake_up_light_ii.yaml
+++ b/custom_components/tuya_local/devices/wake_up_light_ii.yaml
@@ -1,0 +1,241 @@
+name: Alarm clock
+products:
+  - id: ya39qahrcdhvxowz
+    manufacturer: Generic
+    model: Wake Up Light II
+    model_id: ACA002-II-WWA
+entities:
+  - entity: light
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+      - id: 102
+        type: integer
+        name: brightness
+        range:
+          min: 10
+          max: 1000
+  - entity: time
+    icon: "mdi:clock-digital"
+    dps:
+      - id: 103
+        type: string
+        name: hms
+  - entity: select
+    name: Display
+    icon: "mdi:clock-digital"
+    category: config
+    dps:
+      - id: 104
+        type: integer
+        name: option
+        mapping:
+          - dps_val: 0
+            value: "Off"
+          - dps_val: 1
+            value: Day
+          - dps_val: 2
+            value: Night
+          - dps_val: 3
+            value: Auto
+  - entity: switch
+    name: Radio
+    icon: "mdi:radio"
+    dps:
+      - id: 105
+        type: boolean
+        name: switch
+  - entity: number
+    translation_key: volume
+    dps:
+      - id: 106
+        type: integer
+        name: value
+        range:
+          min: 1
+          max: 16
+  - entity: text
+    name: Station list
+    category: config
+    icon: "mdi:playlist-music"
+    hidden: true
+    dps:
+      - id: 107
+        type: base64
+        optional: true
+        name: value
+  - entity: switch
+    name: Alarm 1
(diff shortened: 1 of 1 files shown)


avg-I:
Also, not sure if this is useful and if the information belongs here, but I think that I was able to decode DP 127 (alarm settings or "clock_set").
It has 48 bytes with 12 bytes per alarm.
Definitions of bytes:

0. whether the alarm is enabled (o or 1);
1. alarm time in minutes from midnight (e.g., 390 for 06:30), upper byte;
2. lower byte;
3. bitmask of weekdays when the alarm is active with Monday being the lowest bit;
4. alarm duration in minutes;
5. alarm mode as enumeration:
    0\. sound
    1\. sound + light
    2\. radio
    3\. radio + light
7. selected sound (0..7; 0 - none, 1 - bird, etc);
8. volume (1..16);
9. always `7` for all my alarms, not sure what it means;
10. ramp-up ("sunrise") time for the light, in minutes;
11. light brightness (0..20);
12. whether snooze function is disabled (i.e., 0 - enabled, 1 -disabled).

The configuration items appear in exactly same order in Smart Life app.
There is no configuration item for byte 8.

make-all:
It is identical to Moes Alarm clock. I made the color cycle white, because every other option is a color, and there is not a color that matches "color cycle", but white is all colors combined.

Efeect is not really intended for that, and will not allow grouping with other RGB lights and being able to sync to the closest color.

Color names are chosen for their position on the RGB color wheel, not for the similarity with the name Tuya uses.

avg-I:
@make-all, thank you for explaining the rationale.

What made me think that something was wrong was that having named colors also enabled brightness and color-wheel controls for the ambient light. But its brightness cannot be controlled at all and the colors are fixed.

I also got an impression that named colors are intended to be presets for specific colors for lights with full color support. I can be very wrong about that.

And, sorry to say that, but white for cycling through colors was not intuitive at all.
It makes sense now that you explained it, but it's not obvious.

I guess that HASS doesn't have a perfect match for this kind of lights at the moment.

make-all:
You are right, HA does not support RGB lights without brightness, nor lights that have a list of colors mixed in with other effects. Or maybe a better way to put it, Tuya manufacturers often come up with strange designs that do not fit other platforms such as HA, Matter, Apple, Google and Alexa well.

make-all:
Merged product details with Moes config and made some improvements to the loop handling.

avg-I:
Thank you!
Will test here as well.

PR #4656, opened 2026-03-08
feat(devices): Add ETA Noble Smart humidifier

jiriks74:
Adds a configuration, translations and icons for the ETA Noble Smart humidifier.

Translations were generated by LLM with the device manual as a reference.

The base configuration was generated from API explore output but heavily modified to use better names (eg. the manual mode from device manual was renamed to continuous and auto-humidity renamed to manual), reporting fault codes properly and correcting entity categories.


jiriks74:
I see that this needs cleaning up however I am not familiar enough with the project to do so.

I can't spend much more time on this right now so it will take time to figure out how to clean this up.

If you'd be willing to look into the translation/icon issues it would be very appreciated.

jiriks74:
So far everything seems to be working. I've tested the device the whole afternoon and all the functionality is there (only the overflow error state hasn't been tested).

I'll continue monitoring/testing the condfig further but everything (apart from the translation/icon CI issues) works fine.

custom_components/tuya_local/translations/ro.json
jiriks74, line 1: This file was full of special characters at EOL that were removed (`^M` in NeoVim).

custom_components/tuya_local/translations/bg.json
```diff
@@ -378,7 +378,9 @@
                         "state": {
                             "laundry": "Сухо пране",
                             "purify": "Пречистете",
-                            "continuous": "Непрекъснато"
+                            "continuous": "Непрекъснато",
+                            "manual": "Ръчно",
+                            "auto": "Авто"
```
make-all: auto is already provided by the base humidifier translations

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
```diff
@@ -0,0 +1,194 @@
...
+      - id: 4
+        type: string
+        name: mode
+        mapping:
+          - dps_val: manual
+            value: continuous
+          - dps_val: auto
+            value: auto
+          - dps_val: humidity
+            value: manual
```
make-all: You can use "normal" here to avoid introducing new translations unnecessarily

custom_components/tuya_local/translations/bg.json
```diff
@@ -1211,6 +1218,15 @@
...
             "rangehood": {
                 "name": "Абсорбатор"
+            },
+            "warm_mist": {
+                "name": "Топла пара"
+            },
+            "plasma": {
+                "name": "Плазма"
+            },
+            "night_mode": {
```
make-all: You can use sleep instead to avoid introducing a new translation that is the same meaning

custom_components/tuya_local/translations/bg.json
```diff
@@ -488,6 +493,9 @@
             },
             "rangehood_light": {
                 "name": "Светлина на абсорбатора"
+            },
+            "tank_led": {
```
make-all: You can use the more general `indicator` instead to avoid unnecessarily adding a new translation.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
```diff
@@ -0,0 +1,194 @@
+name: Humidifier
+products:
+  - id: vkh653vaxghgavef
+    manufacturer: ETA
+    model: Noble Smart
+    model_id: fxytco
```
make-all: This looks like the model id from Tuya portal, which is not what is intended here, as that is just a database key of  some sort. model_id is for a user facing model code if any, but most devices do not need it.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
```diff
@@ -0,0 +1,194 @@
...
+    translation_key: sound
+    category: config
+    dps:
+      - id: 8
+        type: boolean
+        name: switch
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 10
```
make-all, line 97: You may need dp 18 in here as the `unit`, with a mapping:

```yaml
        mapping:
          - constraint: unit
            conditions:
              - dps_val: f
                value_redirect: temperature_f
```

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
```diff
@@ -0,0 +1,194 @@
...
+        name: sensor
+        class: measurement
+        unit: C
+      - id: 12
+        type: integer
+        name: temperature_f
+        unit: F
+        optional: true
+  - entity: sensor
+    class: humidity
```
make-all: This entity is a duplicate of the humidity entity's current_humidity, so not needed.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
jiriks74: It was reported when querying the device ability IDs so I have no issues reporting it.

So I guess if the manufacturer's website says `Humidifier ETA Noble Smart 5629 90000 gray` it could be set to `5629 90000`? The name on the box says just Noble Smart so I guess it qualifies as a model ID.


But if it has no use here I'll gladly remove it.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
jiriks74: I understand that this is a duplicate, however this allows users to map the humidity sensor to an area in HA (HA doesn't allow me to select the current_humidity one):


<details>
<summary>Screenshots</summary>

image
image

</details>

While not the best sensor on the planet its good enough for the device and it allows me to use it in Climate and Area sections of HA without buying an extra one.

custom_components/tuya_local/translations/bg.json
jiriks74: Will remove

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
jiriks74: This was used in an attempt to be close to what the manual states while being well descriptive.

I'd consider the `auto` mode to be `normal` since the device turns on with it selected and the manual states the `maintain selected humidity` to be `manual`.

custom_components/tuya_local/translations/bg.json
jiriks74: This was used in an attempt to be close to what the manual states while being well descriptive.

The manual reports it as `Night mode` or `Noční režim` in Czech which I'd consider quite a stretch from `(the) Sleep` or `Spánek` in Czech.

custom_components/tuya_local/translations/bg.json
jiriks74: This was used in an attempt to be close to what the manual states while being well descriptive.

I'm not sure that this would be right since it doesn't really indicate anything, it's just a *pretty LED*. I'll look whether other translations could be used since I agree that this has low chance of being reused and likely has an alternative.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
jiriks74: I'm not quite sure what you mean. I understand that you want to conditionally select the dp for the sensor based on the set value in dp 18?

So just something like:
```yaml
    dps:
      - id: 10
        type: integer
        name: sensor
        class: measurement
        unit: C
        mapping:
          - constraint: unit
            conditions:
              - dps_val: f
                value_redirect: temperature_f
      - id: 12
        type: integer
        name: temperature_f
        unit: F
```
would do it from what I understand.

custom_components/tuya_local/translations/bg.json
jiriks74: The best match I found so far is `light` from `alarm_mode`:

https://github.com/make-all/tuya-local/blob/1a44a2cb0188dd4e9d032f5c08f7e8aed5618c73/custom_components/tuya_local/translations/cz.json#L823-L830

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
make-all: That is correct, mainly if the model name is covering a range of models, and we aren't sure if the config will cover all of them, so those "5629 90000" numbers might be useful to differentiate.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
make-all: Yes, but you need to remove the individual unit markers from there, and add

```yaml
      - id: 18
        type: string
        name: unit
        mapping:
          - dps_val: f
            value: F
          - value: C
```

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
make-all: You can use a template helper for adding that additional sensor. The current policy of this integration is not to add duplicate sensors in the integration itself.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
jiriks74: I seem to be stuck on this.

From what I understand I'm supposed to set it up like this:

```yaml
  - entity: sensor
    class: temperature
    dps:
      - id: 10
        type: integer
        name: sensor
        class: measurement
        mapping:
          - constraint: unit
            conditions:
              - dps_val: f
                value_redirect: temperature_f
      - id: 12
        type: integer
        name: temperature_f

      - id: 18
        type: string
        name: unit
        mapping:
          - dps_val: f
            value: F
          - value: C
```

However nothing changes when I switch the device to Fahrenheit.

make-all: This means it is working correctly.

HA uses your user preferences for display. If the display is changing when you switch units on the device, it would mean that the information provided to HA is incorrect (the value displayed would change, but the units displayed would not, meaning the value is incorrect).

custom_components/tuya_local/translations/bg.json
make-all: This is not correct, that is a select entity, and the name would be "Alarm mode", only one of the options is "Light".

I think "Ambient light" is reusable, and is fairly high up on the list of pending translations that would be reusable across a number of existing devices. Currently most are named just "Ambient" on the assumption that being a light the "light" part is obvious, but probably including light in the name is better.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
make-all: Manual implies that the user has to manually control when to stop the humidification. I'm not sure what auto does here, maybe adjusts the spray level automatically rather than just off and on, but maintaining selected humidity is the normal expectation for a humidifier, so I think "normal" fits.

The aim of this integration is not to match wording in the device's manual, because we are not just supporting one brand of devices here, there are hundreds of manufacturers selling devices with Tuya modules in them, and no consistency between them. To prevent the integration becoming more bloated than necessary with translations and config files that differ only in wording used, we need to break with the manufacturer's choices and enforce our own standards.

This also has value when you have multiple devices of the same type from different manufacturers and want to control them in a group.

custom_components/tuya_local/translations/bg.json
make-all: Maybe the Czech translation could use some work if the current "Sleep" does not mean the same as night mode, as that is the meaning that many other devices give it (reducing lighting brightness and noise to minimum)

custom_components/tuya_local/translations/ro.json
make-all: Most likely there was one extra ^M in there that causes git to treat the file as binary instead of text, I suspect one of my linting tools has a bug that is adding these because they pop up from time to time. I will fix the issue separately, so you can rebase your branch and the merge conflict should go away.

make-all: There don't seem to be any issues like that currently in that file, but it seems to be treated as binary by git, as my Linux VM has it checked out with CRLF line ends while the other translation files have LF line ends. This will cause merge conflicts, as merging of binary files is much more complex than text files.

jiriks74: I agree, that the character should be caused by CRLF instead of LF.

For editors the fix is to use [editorconfig](https://editorconfig.org/).

It has other benefits like allowing you to select tabs/spaces/number of spaces for each file/filetype specifically, etc.

I'm not sure whether cli tooling will repect this bu it would most likely help other contributors since I many changes will be done in an editor.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
jiriks74: > Manual implies that the user has to manually control when to stop the humidification.

> I'm not sure what auto does here, maybe adjusts the spray level automatically rather than just off and on

This device has 3 modes:
- Auto (`AUTO - TEMP` in the manual)
  - The device is fully autonomous
  - Maintains humidity that is automatically selected by the device based on current temperature
- Continuous (`MANUAL` in the manual)
  - Renamed because of existing translations which are quite descriptive of the mode
  - The device doesn't stop humidifying until the user stops it or the water runs out.
- Manual (`AUTO - HUM` in the manual)
  - The device is partially autonomous
  - Maintains a manually selected humidity
  - Seemed like a good translation after selecting the first two

custom_components/tuya_local/translations/bg.json
jiriks74: If that was the goal of the translation than something like `Režím Spánku` (`Sleep Mode`) would be more appropriate since `Spánek` is literally the noun `Sleep`.

`Režim Spánku` => the device is in a sleep mode - while this still is a bit confusing as *sleep mode* might indicate that the device is partially off which allows to start up quicker (like a laptop) it is much metter than just *(the) sleep*

`Spánek` => the device is, most likely, sleeping

custom_components/tuya_local/translations/bg.json
jiriks74: This sounds much better. The translation to Czech could be something like `Podsvícení` which I find quite high up in the list of descriptive translations for this entity.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
make-all: I think `normal` here is a better translation than `manual`:
1. The device manual calls `continuous` "manual" according to your own description above, so it would be confusing to assign the name "manual" to a different mode
2. A humidifier's normal expected operation is to maintain the humidity set by the user.

custom_components/tuya_local/translations/bg.json
jiriks74: Do you want to have the `Sleep` translation changed to `Sleep mode` for all languages so that it's consistent? Or should I change only the Czech one?

custom_components/tuya_local/translations/ro.json
make-all: The issue is not editors. Git by default normalizes line ends in text files, forcing LF line ends on a platform that does not use them will result in more files being accidentally marked as binary.

jiriks74: Ok, so do you want me to drop the `.editorconfig` or is it something you'd like to keep?

jiriks74:
Hi @make-all, could you get me feedback on https://github.com/make-all/tuya-local/pull/4656#discussion_r2991324869 and https://github.com/make-all/tuya-local/pull/4656#discussion_r3042071491 please? I'd like to finish the work on this if you're willing to accept this device and have the time to give me feedback.

custom_components/tuya_local/translations/ro.json
make-all: Drop the .editorconfig

custom_components/tuya_local/translations/bg.json
make-all: You should only change translations for languages you speak.

make-all:
This last force push has made this branch incompatible with the latest head. This can no longer be merged, please resubmit following latest recommendations to split new device contributions from any translation changes proposed.

PR #4648, opened 2026-03-07
Add Curv Smart Alarm System support

TheAdamGalloway:
## Description

Add support for the Curv Smart Alarm System.

This device currently does not auto-match correctly in Tuya Local and falls back to an unrelated device profile during config flow. This PR adds a dedicated YAML definition based on the Tuya Cloud data model and confirmed locally reported DPS.

## Product information

- Product ID: `mrwxwgtnpnemqwa1`
- Model ID: `000003csuo`
- Manufacturer: `Curv`
- Model: `Smart Alarm System`
- [Link](https://support.curv360.com/hc/en-gb/categories/360003358234-Curv-Smart-Alarm-System)

## Log message

```text
2026-03-07 00:03:53.604 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches speaka_sptvcm510_tvmount with quality of 17%. LOCAL DPS: {"updated_at": 1772841824.6704078, "101": "home", "102": "Normal", "103": "English", "104": "normal", "105": 100, "106": 60, "107": 60, "108": 5, "109": 3, "110": 3, "111": "OFF", "112": true, "113": true, "114": false, "115": false, "116": false, "117": true, "118": true, "119": true, "120": true, "121": true, "122": "3232640133788880000", "125": "MODEL:Curv Smart Alarm System\r\nHK UID:4A3853FF005E003701FF504A\r\nIMEI:860665050294106\r\nFIRMWARE VER:1.0.1"}
2026-03-07 00:03:53.604 WARNING (MainThread) [custom_components.tuya_local.config_flow] Include the previous log messages with any new device request to https://github.com/make-all/tuya-local/issues/
```

## Locally observed DPS

`{
  "101": "home",
  "102": "Normal",
  "103": "English",
  "104": "normal",
  "105": 100,
  "106": 60,
  "107": 60,
  "108": 5,
  "109": 3,
  "110": 3,
  "111": "OFF",
  "112": true,
  "113": true,
  "114": false,
  "115": false,
  "116": false,
  "117": true,
  "118": true,
  "119": true,
  "120": true,
  "121": true,
  "122": "3232640133788880000",
  "125": "MODEL:Curv Smart Alarm System\r\nHK UID:4A3853FF005E003701FF504A\r\nIMEI:860665050294106\r\nFIRMWARE VER:1.0.1"
}
`

## Tuya Cloud model

- 101 system_arm_type enum: disarmed, armed, home
[...]

PR #4658, opened 2026-03-08
Add device support for Nivian NVS-A6WG alarm

AlbertoMarziali:
## Description

Add support for the Nivian NVS-A6WG alarm system.

## Product information

Product ID: `p3sdak6d9jznh14x`
Manufacturer: `Nivian`
Model: `NVS-A6WG`
[Link](https://www.nivianhome.com/en/product/nvs-a6wg/)

## Locally observed DPS

`{'101': 'home', '102': 'NoSimCard', '103': 'English', '104': 'normal', '105': 100, '106': 10, '107': 15, '108': 3, '109': 3, '110': 3, '111': 'OFF', '112': True, '113': True, '114': True, '115': True, '116': False, '117': False, '118': False, '119': False, '120': False, '121': False, '122': '*******************', '125': 'MODEL:NVS-A6WG\r\nPID:572382123896274505D7FF33\r\nIMEI:***************\r\nFIRMWARE VER:1.0.9'}`

## Tuya Cloud model

- 101 system_arm_type enum: disarmed, armed, home
- 102 gsm_status enum: NoSimCard, NoNetWork, Normal, GsmOff
- 103 language enum: English, French, Italian, German, Spanish, Danish, Dutch, Portuguese
- 104 dc_status enum: low, normal, high, off
- 105 bat_status value, 0 to 100 percent
- 106 arm_delay value, 0 to 300 seconds
- 107 alarm_delay value, 0 to 300 seconds
- 108 alarm_sound_duration value, 1 to 9 minutes
- 109 ring_times value, 1 to 9
- 110 tel_alarm_cycle value, 1 to 9
- 111 inside_siren_sound enum: ON, OFF
- 112 gsm_en bool
- 113 tel_ctrl_en bool
- 114 arm_sms_en bool
- 115 disarm_sms_en bool
- 116 keyboard_tone_en bool
- 117 arm_delay_tone_en bool
- 118 alarm_delay_tone_en bool
- 119 arm_disarm_tone_en bool
- 120 inside_siren_en bool
- 121 wireless_siren_en bool
- 122 password string
- 125 device_info string
- 124, 126, 127, 128, 129 are raw payload types and are not exposed in this initial config

## How the device functions

- DP 101 controls arm mode, mapped to Home Assistant alarm states:
  - disarmed -> disarmed
  - armed -> armed_away
  - home -> armed_home
- DP 102 reports GSM status

- DP 103 selects system language

- DP 104 reports DC status

- DP 105 reports battery percentage

- DP 106 to 110 are numeric configuration values

- DP 111 to 121 are siren and notification feature toggles

- DP 125 reports device information

## Notes

DP 122 (password) was intentionally not exposed due to sensitivity

Raw DPs 124, 126, 127, 128, and 129 were left out pending payload examples and a suitable entity mapping

### DIFF
--- a/custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
+++ b/custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
@@ -199,7 +199,7 @@ entities:
         mapping:
           - dps_val: "OFF"
             value: false
-          - dps_val:  "ON"
+          - dps_val: "ON"
             value: true
   - entity: switch
     name: Wireless siren


custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
```diff
@@ -0,0 +1,209 @@
...
+  - entity: select
+    translation_key: language
+    category: config
+    dps:
+      - id: 103
+        name: option
+        type: string
+        mapping:
+          - dps_val: English
+            value: English
```
make-all: values should be mapped to translation keys (generally all lowercase versions of the language name, though danish and dutch currently do not exist in the translation files so will need to be added).

custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
```diff
@@ -0,0 +1,209 @@
...
+            value: Portuguese
+  - entity: sensor
+    name: DC Status
+    category: diagnostic
+    dps:
+      - id: 104
+        name: sensor
+        type: string
+  - entity: sensor
+    name: Battery Level
```
make-all: name is not needed, it only hides the class translation

custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
```diff
@@ -0,0 +1,209 @@
...
+    category: diagnostic
+    class: battery
+    dps:
+      - id: 105
+        name: sensor
+        type: integer
+        unit: "%"
+  - entity: number
+    name: Arm Delay
+    category: config
```
make-all, line 75: class: duration

custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
```diff
@@ -0,0 +1,209 @@
...
+      - id: 106
+        name: value
+        type: integer
+        range:
+          min: 0
+          max: 300
+        unit: s
+  - entity: number
+    name: Alarm Delay
+    category: config
```
make-all, line 87: class: duration

custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
```diff
@@ -0,0 +1,209 @@
...
+      - id: 107
+        name: value
+        type: integer
+        range:
+          min: 0
+          max: 300
+        unit: s
+  - entity: number
+    name: Alarm Sound Duration
+    category: config
```
make-all, line 99: class: duration

name just "Alarm duration" is better I think, as long names lead to clutter in the UI.

custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
```diff
@@ -0,0 +1,209 @@
...
+        type: string
+        mapping:
+          - dps_val: disarmed
+            value: disarmed
+          - dps_val: armed
+            value: armed_away
+          - dps_val: home
+            value: armed_home
+  - entity: sensor
+    name: GSM Status
```
make-all: HA naming convention is to only capitalise the first word (unless there is another reason like abbreviations and brand names)

So "GSM status".

Also applies to many other entity names below

custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
```diff
@@ -0,0 +1,209 @@
...
+        type: boolean
+  - entity: switch
+    name: Disarm SMS Enabled
+    category: config
+    dps:
+      - id: 115
+        name: switch
+        type: boolean
+  - entity: switch
+    name: Keyboard Tone Enabled
```
make-all: `translation_key: keytone` fits here (instead of the name)

custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
```diff
@@ -0,0 +1,209 @@
...
+        name: switch
+        type: boolean
+  - entity: switch
+    name: Wireless Siren Enabled
+    category: config
+    dps:
+      - id: 121
+        name: switch
+        type: boolean
+  - entity: sensor
```
make-all: This does not seem useful as a sensor, it may be better as an attribute on another entity.

make-all (commented):
Many of the names are long. I would leave off "enabled" from switches, as that is the obvious function of having a switch. And maybe some of the delays could use clearer names (Exit delay and Entry delay for example rather than Alarm delay and Arm delay)

AlbertoMarziali:
I applied all the suggested changes.

custom_components/tuya_local/devices/nivian_nvs_a6wg_alarm.yaml
```diff
@@ -32,31 +32,26 @@ entities:
...
-            value: German
-          - dps_val: Spanish
-            value: Spanish
-          - dps_val: Danish
-            value: Danish
-          - dps_val: Dutch
-            value: Dutch
-          - dps_val: Portuguese
-            value: Portuguese
+          - dps_val: english
```
make-all: The dps_val should not change based on my advice.
This must exactly match what the device is sending and expects as input (including case).

Also, do not remove languages supported by the device just because they are not in the translation files. They can be added later.

AlbertoMarziali: > The dps_val should not change based on my advice.
> This must exactly match what the device is sending and expects as input (including case).

My bad, I thought it wasn't case sensitive. 

> Also, do not remove languages supported by the device just because they are not in the translation files. They can be added later.

Alright, should I leave Dutch and Danish languages with uppercase value?

AlbertoMarziali: The question is, should we have them "nicely shown" (uppercase) today or keep them shown lowercase until one day the translation is added?

make-all: I will add the translations once this is merged, so keep the values lowercase.

AlbertoMarziali:
Fixed languages.

make-all:
This seems to be identical to #4648 submitted 2 days earlier.
I will merge these into a single config

Issue #4673, opened 2026-03-09
Request support for Emini 16A public EV Charger


lfrlucas:
### Log message

```text
2026-03-09 22:45:57.407 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches nine_ev_charger with quality of 70%. LOCAL DPS: {"updated_at": 1773096348.4854352, "1": 8871, "3": "charger_insert", "4": 6, "9": 2484, "14": "charge_delay", "18": false, "23": "V1.2.0", "25": 0, "27": "online", "101": "GO161EPAM"}
2026-03-09 22:45:57.407 WARNING (MainThread) [custom_components.tuya_local.config_flow] Include the previous log messages with any new device request to https://github.com/make-all/tuya-local/issues/
```

### Product ID

nui1mldb0iedtsoj

### Product Name

Emini 16A Public

### DPS information

```text
{
  "result": {
[...]

PR #4674, opened 2026-03-09
Add device support to Emini 16a public EV charger

lfrlucas:
Issue: https://github.com/make-all/tuya-local/issues/4673

### DIFF
--- a/custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
+++ b/custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
@@ -225,4 +225,3 @@ entities:
             value: false
           - value: false
             hidden: true
-        always_send: true


custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
```diff
@@ -0,0 +1,264 @@
...
+      - id: 1
+        type: integer
+        name: sensor
+        unit: kWh
+        class: total_increasing
+        mapping:
+          - scale: 100
+
+  - entity: number  # charge_cur_set -4- integer A
+    name: Charging current
```
make-all: If this name is removed, the default HA translations can be used here.

custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
```diff
@@ -0,0 +1,264 @@
...
+    class: power
+    dps:
+      - id: 9
+        type: integer
+        name: sensor
+        unit: kW
+        class: measurement
+        mapping:
+          - scale: 1000
+            constraint: work_state
```
make-all: I don't understand the need for this condition. Is the sensor not showing actual power consumption?

custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
```diff
@@ -0,0 +1,264 @@
...
+
+  - entity: switch  # switch -18- boolean
+    icon: "mdi:ev-station"
+    dps:
+      - id: 18
+        type: boolean
+        name: switch
+
+  - entity: sensor  # system_version -23- string
+    name: System version
```
make-all: Static information about the device does not need to be a sensor. It can just be added as an attribute of another entity.

custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
```diff
@@ -0,0 +1,264 @@
...
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+        range:
+          min: -40
+          max: 200
+
+  - entity: sensor  # charge_energy_once -25- integer kW·h
+    name: Last charge
```
make-all, line 203: `class: energy_storage` might work here.

custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
```diff
@@ -0,0 +1,264 @@
...
+      - id: 27
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: offline
+            value: false
+          - dps_val: online
+            value: true
+
+  - entity: sensor  # product_sn -101- string
```
make-all: Does not need to be a sensor

custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
lfrlucas: removed

custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
lfrlucas: Removed this condition. This was to ensure the power would go to 0 after disconnecting the ev charger plug. Because the power is not updating periodically and automatically. But I recently found that it will update for some seconds if the online state sensor is switched from offline to online.  I changed the online state sensor to "Live updates" switch. Having one HA automation toggling this live update switch every minute is enough to get periodically updated power/voltage sensors.

custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
lfrlucas: moved to attribute of status sensor

custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
lfrlucas: done

custom_components/tuya_local/devices/emini_16a_public_evcharger.yaml
lfrlucas: moved to attribute of status sensor

PR #4675, opened 2026-03-10
Add a different Ledkia Fan/Light variant

ankon:
We got the WiFi controller for our fan-light through support, and it seems to expose a bit more functionality than the existing ledkia/windcalm fans (beep controls whether the thing beeps as "ack" for any interaction).

(Feel free to rename the file, I simply couldn't come up with any good name :D)


make-all:
Merged into ledkia_fan_light

PR #4679, opened 2026-03-11
Add support for Generic Mini Central Air Conditioner

roeemz:
This PR adds support for a common generic Tuya-based Mini Central Air Conditioner (Product ID: `ypydgdiic5ztkorj`)

### DIFF
--- a/custom_components/tuya_local/devices/generic_mini_central_ac.yaml
+++ b/custom_components/tuya_local/devices/generic_mini_central_ac.yaml
@@ -1,0 +1,133 @@
+name: Mini central air conditioner
+products:
+  - id: ypydgdiic5ztkorj
+    model: Mini Central AC
+entities:
+  - entity: climate
+    dps:
+      - id: 1
+        name: hvac_mode
+        type: boolean
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: auto
+                value: heat_cool
+              - dps_val: cooling
+                value: cool
+              - dps_val: heating
+                value: heat
+              - dps_val: fan
+                value: fan_only
+              - dps_val: dehum
+                value: dry
+      - id: 2
+        name: temperature
+        type: integer
+        range:
+          min: 16
+          max: 32
+      - id: 3
+        name: current_temperature
+        type: integer
+        mapping:
+          - scale: 10
+      - id: 4
+        name: mode
+        type: string
+        hidden: true
+      - id: 5
+        name: fan_mode
+        type: string
+        mapping:
+          - dps_val: auto
+            value: auto
+          - dps_val: low
+            value: low
+          - dps_val: middle
+            value: medium
+          - dps_val: high
+            value: high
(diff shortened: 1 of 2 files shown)


make-all:
What makes you say this is generic? We have never encountered it before, if it was generic I think it would be more widespread.

tests/devices/test_generic_mini_central_ac.py
make-all, line 1: Tests for individual devices are handled by the generic test_device_config.py, you don't need to add anything to the tests unless you are introducing new functionality.

Currently device tests are broken, due to an update in HA that requires additional setup for async tests which does not work with the IsolatedAsyncTestCase underlying the TuyaDeviceTestCase class, but you should probably just delete this anyway.

custom_components/tuya_local/devices/generic_mini_central_ac.yaml
```diff
@@ -0,0 +1,76 @@
+name: Mini central air conditioner
+products:
+  - id: ypydgdiic5ztkorj
+    manufacturer: Generic
```
make-all: Is this an actual manufacturer name?

If not, leave this out.

custom_components/tuya_local/devices/generic_mini_central_ac.yaml
```diff
@@ -0,0 +1,76 @@
...
+        type: boolean
+        mapping:
+          - dps_val: false
+            value: "off"
+            icon: "mdi:hvac-off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: auto
+                icon: "mdi:hvac"
```
make-all: Don't add icons here, HA now provides default icons that cover these.

custom_components/tuya_local/devices/generic_mini_central_ac.yaml
```diff
@@ -0,0 +1,69 @@
+name: Mini central air conditioner
+products:
+- id: ypydgdiic5ztkorj
```
make-all: normally should be two spaces to indent (suggestion by linter says at least one, but that is for minimum syntax correctness, convention is two)

roeemz: fixed

roeemz:
> What makes you say this is generic? We have never encountered it before, if it was generic I think it would be more widespread.

That's what we have here and it works. It seems to follow a standard DP schema, might help more people that have a similar system.

roeemz:
I have updated the PR to include more diagnostic DPs (101-108)
1. Coil Temperatures (In/Mid/Mid2) with correct scaling.
2. Indoor/Outdoor Fan status.
3. Compressor command and Four-way valve status.

These might provide helpful technical insights into the unit's operation. I've marked them all as optional to ensure broad compatibility, but if they are considered excessive for this repo, I'm happy to remove them.

make-all:
The main update needed here is to update with proper manufacturer/model details for the device.
It is not generic. If it was generic, we never would have reached 1500 supported devices without encountering it before.

PR #4684, opened 2026-03-12
Add configuration for PIR Motion Switch device

BigGeorgeTx:
This PR adds support for the Tuya MFA05 PIR Motion Switch (product ID iii7gagtan21i2of).

The device exposes the following DPS via the local Tuya protocol (3.3):

- 1: Light relay switch (boolean)
- 103: Inching mode (string)
- 104: LED indicator (boolean)
- 106: Motion alarm (boolean)

Motion is represented by DP 106. DP 1 is the actual relay and can be manually toggled; it may mirror motion depending on device settings, but it is not the motion state.

DP 103 is a string-based inching mode value. The device reports opaque string values (e.g., "AQJY"), so it is exposed as a text entity rather than a select.

receives DPS 1, 103, 104, and 106 from the device.

Below is the DPS cache from the Tuya Local device debug output:

"1": true,
"103": "AQJY",
"104": true,
"106": true

The included YAML definition has been tested locally in Home Assistant with Tuya Local and works reliably.

Let me know if anything needs adjustment.

### DIFF
--- a/custom_components/tuya_local/devices/moes_pir_wall_switch.yaml
+++ b/custom_components/tuya_local/devices/moes_pir_wall_switch.yaml
@@ -1,10 +1,9 @@
-name: PIR Wall Switch
+name: PIR wall switch
 
 products:
   - id: iii7gagtan21i2of
     model: MFA05
     manufacturer: MOES
-    name: MOES Smart Motion Sensor Switch Single Pole
 
 entities:
   - entity: light
@@ -14,7 +13,9 @@ entities:
         name: switch
 
   - entity: text
-    name: Inching Mode
+    name: Inching
+    category: config
+    hidden: true
     dps:
       - id: 103
         type: string
@@ -22,16 +23,15 @@ entities:
 
   - entity: light
     translation_key: indicator
+    category: config
     dps:
       - id: 104
         type: boolean
         name: switch
 
   - entity: binary_sensor
-    name: Motion Alarm
-    class: motion          # ← Changed from device_class to class
+    class: motion
     dps:
       - id: 106
         type: boolean
         name: sensor
-        readonly: true


BigGeorgeTx:
My apologies.  I grabbed the wrong file when I created this.  Tomorrow I will try to edit the original PR.

This is the code the is working locally:

Here is the working local version:

name: Motion PIR Wall Switch

products:
  - id: iii7gagtan21i2of

entities:
  - entity: switch
    name: Lights Power
    dps:
      - id: 1
        type: boolean
        name: switch

  - entity: text
    name: Inching Mode
    dps:
      - id: 103
        type: string
        name: value   # ← fixes 'value' warning

  - entity: switch
    name: LED Indicator
    dps:
      - id: 104
        type: boolean
        name: switch  # ← fixes 'switch' warning

  - entity: switch
    name: Motion Alarm
    dps:
      - id: 106
        type: boolean
        name: switch 
        readonly: true

custom_components/tuya_local/devices/mfa05_motionswitch.yaml
```diff
@@ -0,0 +1,42 @@
+# Tuya MFA05 PIR Motion Wall Switch
+# Product ID: iii7gagtan21i2of
+# DP Map:
+#   1   - Relay / Lights Power (boolean)
+#   103 - Inching Mode (string)
+#   104 - LED Indicator (boolean)
+#   106 - Motion Alarm (boolean, readonly)
+
+name: Motion PIR Wall Switch
```
make-all: Keep device names short, as they are prepended in some places in the HA UI. Motion and PIR are redundant, so pick one.

You also need to follow HA naming convention of only capitalising the first word of device and entity names (unless there are abbreviations like PIR).

custom_components/tuya_local/devices/mfa05_motionswitch.yaml
```diff
@@ -0,0 +1,42 @@
...
+products:
+  - id: iii7gagtan21i2of
+    name: MFA05
```
make-all: This is `model`, not name.

custom_components/tuya_local/devices/mfa05_motionswitch.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+entities:
+  - entity: switch
+    name: Lights Power
```
make-all: Make it a light entity, and there is no need for a name

custom_components/tuya_local/devices/mfa05_motionswitch.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+  - entity: text
+    name: Inching Mode
+    dps:
+      - id: 103
+        type: string
+        name: value
+
+  - entity: switch
+    name: LED Indicator
```
make-all: Standard pattern here is to use a light entity, with `translation_key: indicator`

custom_components/tuya_local/devices/mfa05_motionswitch.yaml
```diff
@@ -0,0 +1,42 @@
...
+    dps:
+      - id: 104
+        type: boolean
+        name: switch
+
+  - entity: switch
+    name: Motion Alarm (Raw)
```
make-all: What information is `(Raw)` in the name trying to convey?

This should be a binary_sensor, with `class: motion` (and no name needed in that case)

BigGeorgeTx:
Thanks for the feedback.  I have shortened the device_name, changed the name and added the manufacturer, changed it to a light entity, and made the motion switch a binary sensor, all as requested.  I have tested the revised file locally in HA and it is working.  I also renamed the file to try and better align with the others.

PR #4685, opened 2026-03-13
DEVICE - [Smart Lock] Inverted Logic

square-spade:
Inverted the logic to align with Home Assistant state `true=locked` as my locks are reporting:

> `{47: True}` = `unlocked` 
> `{47: False}` = `locked`

All other DP points and config seem fine/working.

### DIFF
--- a/custom_components/tuya_local/devices/hornbill_y4_lock_inverted.yaml
+++ b/custom_components/tuya_local/devices/hornbill_y4_lock_inverted.yaml
@@ -164,6 +164,11 @@ entities:
         optional: true
         persist: false
         sensitive: true
+      - id: 73
+        type: base64
+        optional: true
+        sensitive: true
+        name: remote_pd_setkey_check
   - entity: event
     name: Alert
     dps:


square-spade:
updated strings and automatic key extraction _[dp 60/73]_
`lock state` and features all woking perfectly for me now.
could be beneficial to add a stage during config flow _[or during initial/failed `lock.unlock` service call]_ explaining how to grab the key from the Tuya IoT platform and allowing input. Let me know if thats something you think would be beneficial.

make-all:
If this is not a Hornbill device, then you should rename the file with the actual brand to avoid confusion. You should also review all the dps to see if they are really valid for your device.

I am concerned here that you may have taken a config that almost works and tweaked it to make the pain points go away, but there may be a lot more incorrect config there lurking that you aren't concerned with.

make-all:
I see you have made Python changes in lock.py to match your specific lock.
From many years of experience, there is no standard for Tuya devices, so there is no guarantee that there is a match with other devices. You are also hard-coding specific dps which vary from device to device along with the formats used.

I cannot accept this PR as-is, any solution needs to be more general than this.

square-spade:
The changes made to the core `lock.py` are not device dependant and are taken straight from the *_[Tuya Dev Docs](https://developer.tuya.com/en/docs/iot/ble)_* for BLE locks. 
I'm not sure how more general I could have made it considering the device DP for `remote_pd_setkey_check` is pulled from the specific device config [else it returns `None`]
I did also ensure fallback regardless, so if the DP failed nothing broke. Just saves the user having to fish the code out of **Tuya IoT platform** which is an effort in itself.
But no stress! I did just finish writing the ability to obtain our own key (using the `Configure keys` flow) which I was about to PR you but I shall not!

square-spade:
> If this is not a Hornbill device, then you should rename the file with the actual brand to avoid confusion.

I saw a PR for a _BLE_Lock_ and it was not approved due to "Hornbill lock is still a better config file" so I don't see the point in adding the name/brand/model if we are all going to use this config.

> "review all DPS to see if valid for my device" 

Confused as to what you mean Since most DPS are `optional: true` and when this is treated as a master device config [and is the standard Tuya BLE implementation] But yes, I obviously checked my device and have all DPS verified.

> "taken a config that almost works and tweaked it... ... lurking thar you aren't concerned with". 

All I changed was inverting the `lock_state` as I found the original config to not be aligned with home assistant [a Tuya true is `unlocked` while HA is `locked`] so again not sure what you mean by this. 

My lock is working perfectly with this config. Even my patch that will get the unlock code hasn't had any issues so far. Maybe you could elaborate on your concerns as I'm not fully understanding

PR #4693, opened 2026-03-14
Add support for HeatBot 15L water heater

aryanhasgithub:
This PR adds support for the HeatBot(15L) by A.O. Smith, product id: edqofyucyuo4zk8a. It includes a water heater entity and a energy usage entity(raw as the DP only provides the energy since the device has been paired to the app this is easily solvable with a utility meter helper for a end user though).

### DIFF
--- a/ACKNOWLEDGEMENTS.md
+++ b/ACKNOWLEDGEMENTS.md
@@ -1288,3 +1288,4 @@ Further device support has been made with the assistance of users. Please consid
 - [lweikit](https://github.com/lweikit) for contributing support for Yu Home Yu Turbo laundry drying rack.
 - [AlbertoMarziali](https://github.com/AlbertoMarziali) for contributing support for Nivian NVS-A6WG alarm panel.
 - [TheAdamGalloway](https://github.com/TheAdamGalloway) for contributing support for Curv alarm panel that was merged to the Nivian above.
+- [LeeRuMax](https://github.com/LeeRuMax) for assisting with support for Avonflow AFD14-H thermostat.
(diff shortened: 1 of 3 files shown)


custom_components/tuya_local/devices/heatbot_15l_water_heater.yaml
```diff
@@ -0,0 +1,36 @@
...
+        unit: C
+        range:
+          min: 25
+          max: 75
+      - id: 10
+        type: integer
+        name: current_temperature
+  - entity: sensor
+    class: energy
+    name: Energy Usage
```
make-all: Leave out the name so HA's built in translations for energy class are used

custom_components/tuya_local/devices/heatbot_15l_water_heater.yaml
```diff
@@ -0,0 +1,36 @@
...
+    translation_key: water_air
+    dps:
+      - id: 1
+        type: boolean
+        name: operation_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: "heat"
```
make-all: heat is not a valid operation mode for water heaters. It should be heat_pump, electric or gas (other possible values eco, performance, high_demand imply multiple modes available)

custom_components/tuya_local/devices/heatbot_15l_water_heater.yaml
```diff
@@ -0,0 +1,36 @@
+name: Electric water heater
+products:
+  - id: edqofyucyuo4zk8a
+    manufacturer: A.O. Smith
+    model: HeatBot 15L
+entities:
+  - entity: water_heater
+    translation_key: water_air
```
make-all: I don't think this translation key is appropriate, as this only seems to heat water.

aryanhasgithub:
All fixed I removed the translation key and the name for the energy sensor and set the operation mode as electric.

PR #4697, opened 2026-03-15
Update Tongou SA1-wifi to support more DPs

lolmaxlevel:
This PR updates the configuration for the Tongou SA1 energy meter. Based on a full datapoint dump, I have exposed several previously missing features, primarily related to device protection and monitoring.

**Changes:**
Protection Alarms (Switches): Added switches to toggle alarms for Over Voltage, Under Voltage, Over Current, Over Power, and Temperature. Mapped Alarm/Closed values to boolean switches for better UX.

**Protection Thresholds:** Added number entities to set specific threshold values for all protection parameters (Voltage, Current, Power, Temperature).

**New Sensors:**
Added Internal Temperature sensor.
Added Power Factor sensor.
Added Last Event sensor with full mapping for human-readable status messages and dynamic icons (e.g "Over Voltage Trip", "Manual OFF"), **although I haven't added translations for other languages.**

Device Info:
Manufacturer: Tongou
Model: TOSA1-0150XJWT2A
ID: gnr22vuyz8uis3gr

Tested on my device, all new controls work as expected.

### DIFF
--- a/custom_components/tuya_local/devices/tongou_sa1_wifi_energy_meter.yaml
+++ b/custom_components/tuya_local/devices/tongou_sa1_wifi_energy_meter.yaml
@@ -157,7 +157,6 @@ entities:
   - entity: sensor
     class: temperature
     category: diagnostic
-    name: Internal temperature
     dps:
       - id: 131
         type: integer
@@ -167,7 +166,6 @@ entities:
         mapping:
           - scale: 10
   - entity: binary_sensor
-    name: Fault
     class: problem
     category: diagnostic
     dps:


custom_components/tuya_local/devices/tongou_sa1_wifi_energy_meter.yaml
```diff
@@ -144,3 +144,277 @@ entities:
...
+    dps:
+      - id: 131
+        type: integer
+        optional: true
+        name: sensor
+        unit: C
+        mapping:
+          - scale: 10
+  - entity: sensor
+    class: enum
```
make-all: There is no name on this entity. Unlike other classes, "enum" is not really descriptive of what the sensor is measuring.

But it looks like the values are fault conditions. In this case, the standard pattern adopted by this integration is to use a `binary_sensor` with `class: problem` to indicate whether there is a fault present, and when the original value was a string, use a description attribute, which unlike an enum sensor does not require a mapping.

If there is already a problem sensor on the device, then a name will be required to distinguish the different faults, or list both as attributes under the same entity, and have a condition on the sensor dp that only returns false when both dps are indicating an OK state.

custom_components/tuya_local/devices/tongou_sa1_wifi_energy_meter.yaml
```diff
@@ -144,3 +144,277 @@ entities:
...
+            icon: "mdi:timer-off"
+          - dps_val: "Electricity_Reset"
+            value: "Energy Reset"
+            icon: "mdi:restore"
+
+  # alarm switches (Switches: Alarm <-> Closed)
+
+  # DP 101: Energy Shortage
+  - entity: switch
+    name: Alarm Low Energy
```
make-all: I would reorder these names to put "alarm" last. Also all new entities should follow HA naming convention of only capitalising the first word (unless there are abbreviations involved)

custom_components/tuya_local/devices/tongou_sa1_wifi_energy_meter.yaml
```diff
@@ -144,3 +144,277 @@ entities:
...
+            value: true
+          - dps_val: "Closed"
+            value: false
+
+  # Thresholds for alarms
+
+  # DP 114: Current Threshold
+  - entity: number
+    class: current
+    name: Threshold Over Current
```
make-all: Again, the "threshold" should be last in the name to sound more natural in English.

You could consider using
```yaml
  - entity: number
    class: current
    translation_key: sensor_threshold
    translation_placeholders:
      sensor: Over current
```
This will translate the "threshold" part of the name and order it appropriately for the language, though since "Over current" is passed in as a placeholder it will not get translated.

custom_components/tuya_local/devices/tongou_sa1_wifi_energy_meter.yaml
lolmaxlevel: done. added problem class with description attribute.

custom_components/tuya_local/devices/tongou_sa1_wifi_energy_meter.yaml
lolmaxlevel: done. added translation keys with placeholders. For temp threshold used translation `maximum_temperature` if thats ok.

custom_components/tuya_local/devices/tongou_sa1_wifi_energy_meter.yaml
lolmaxlevel: done.

lolmaxlevel:
Fixed everything, waiting feedback on using `maximum_temperature` as translation for temp threshold.

PR #4698, opened 2026-03-16
Frizzlife-lp365

rwalker777:
Add support for Frizzlife LP365 (non-P) and support for metric/imperial units.


make-all:
It is not appropriate to completely replace a working config with your own completely rewritten version with nothing in common like this without any prior discussion about why this is required.

rwalker777:
@make-all Not sure how you want me to approach this... here is the HA thread where we refined this:
https://community.home-assistant.io/t/frizzlife-smart-water-monitor/875335

The current config is very minimal and doesn't support the non-P version or the newer -P versions without temperature probe.  Please advise how you would like me to submit.

PR #4702, opened 2026-03-16
 Update GeyserWise device to also support the GeyserWise Max Solar co…

RandMental:
#Update GeyserWise device to also support the GeyserWise Max Solar controller  

## Correct variable name 
- entity: water_heater: 
   Id: 103 - changed name to "setpoint_temperature" as it is a config value and not a measurement.
## ID 10 is a measurement, not a Config value - but leave it here for compatibility.  
   Id: 10 - add  Unit: C        
   

## Actual geyser water temperature Id: 10
   Id: 10 is only available under the water heater entity under the config category as indicated above.  Add it as a Diagnostic Sensor entity, similar to Collector temperature, so it can be used separately from the Water heater entity for simpler external automations.   

  - entity: sensor
    name: Geyser temperature
    class: temperature
    category: diagnostic
    dps:
      - id: 10
        type: integer
        name: sensor
        optional: true
        unit: C
        class: measurement

### DIFF
--- a/custom_components/tuya_local/devices/geyserwise_water_heater.yaml
+++ b/custom_components/tuya_local/devices/geyserwise_water_heater.yaml
@@ -26,10 +26,11 @@ entities:
             value: true
       - id: 10
         type: integer
+        unit: C
         name: current_temperature
       - id: 103
         type: integer
-        name: temperature
+        name: setpoint_temperature
         unit: C
         range:
           min: 30
@@ -190,6 +191,17 @@ entities:
     category: diagnostic
     dps:
       - id: 108
+        type: integer
+        name: sensor
+        optional: true
+        unit: C
+        class: measurement
+  - entity: sensor
+    name: Geyser temperature
+    class: temperature
+    category: diagnostic
+    dps:
+      - id: 10
         type: integer
         name: sensor
         optional: true


make-all:
The change appears to be unnecessary and will prevent setting the temperature on the device, as you changed the name from the standard name for the temperature setting function of a water_heater, to a name that is unrecognized by the water heater.

RandMental:
Hi Jason

The Geyserwise App and display report two standard temperature measurements,  thus adding the one in red below to expose the same variable as we have for the collector make sense and is necessary
.

- entity: sensor
  name: Collector temperature
  class: temperature
  category: diagnostic
  dps:
    - id: 108
      type: integer
      name: sensor
      optional: true
      unit: C
      class: measurement
- entity: sensor
  name: Geyser temperature
  class: temperature
  category: diagnostic
  dps:
    - id: 10
      type: integer
      name: sensor
      optional: true
      unit: C
      class: measurement



 


Below is the Tuya App config,  the standard configuration has no single “temperature” value for the device.  Using the ID: 103 named “temperature” as the setpoint for the HA water_heater controller entity is not correct, as there are 4 temperature setpoints:  

ID: 103  - Block1 (00h00-06h00)     called “temperature in the Tuya Local device ”          (To correct myself this should rather be called “Night temperature” )

ID: 104  - Block2 (06h00 – 12h00)   called “Morning temperature” 

ID: 105 – Block 3 (12h00 – 18h00)   called “ Afternoon temperature”

ID: 106 – Block 4 (18h00 – 00h00)  called “Evening temperature”

 
  

Thus, mapping the Geyserwise HW device controller on the single setpoint HA  water_heater controller is not correct and does not work correctly in practice.

regards,

Deon

make-all:
I don't care about what the manufacturer's app looks like, I only care about integrating with HA.

Your change only removes functionality of setting the temperature from the water_heater, and duplicates the current temperature. Neither of those are necessary changes.

The device has 4 different temperature settings for different time ranges (plus a differential or separate solar temperature boost depending on whether it is a piped or PV model), but no feedback on which of those is the active time range. So a perfect implementation of the water_heater is not possible, instead the overnight temperature was chosen by virtue of being the first in the list, but to avoid confusion that is duplicated explicitly as "Overnight temperature".

RandMental:
Further to the explanation below adding the Red Geyser temperature section does no harm and allows one to build the external automation without using the water-heater entity which anyway does not work correctly.  

Please add that section. 



 

From: ***@***.*** ***@***.***> 
Sent: Monday, March 16, 2026 1:11 PM
To: 'make-all/tuya-local' ***@***.***>; 'make-all/tuya-local' ***@***.***>
Cc: 'Author' ***@***.***>
Subject: RE: [make-all/tuya-local] Update GeyserWise device to also support the GeyserWise Max Solar co… (PR #4702)

 

Hi Jason

The Geyserwise App and display report two standard temperature measurements,  thus adding the one in red below to expose the same variable as we have for the collector make sense and is necessary
.

- entity: sensor
  name: Collector temperature
  class: temperature
  category: diagnostic
  dps:
    - id: 108
      type: integer
      name: sensor
      optional: true
      unit: C
      class: measurement
- entity: sensor
  name: Geyser temperature
  class: temperature
  category: diagnostic
  dps:
    - id: 10
      type: integer
      name: sensor
      optional: true
      unit: C
      class: measurement






Below is the Tuya App config,  the standard configuration has no single “temperature” value for the device.  Using the ID: 103 named “temperature” as the setpoint for the HA water_heater controller entity is not correct, as there are 4 temperature setpoints:  

ID: 103  - Block1 (00h00-06h00)     called “temperature in the Tuya Local device ”          (To correct myself this should rather be called “Night temperature” )

ID: 104  - Block2 (06h00 – 12h00)   called “Morning temperature” 

ID: 105 – Block 3 (12h00 – 18h00)   called “ Afternoon temperature”

ID: 106 – Block 4 (18h00 – 00h00)  called “Evening temperature”

 
  

Thus, mapping the Geyserwise HW device controller on the single setpoint HA  water_heater controller is not correct and does not work correctly in practice.

regards,

Deon

RandMental:
Hi Jason

I really don’t understand the point of not improving a device’s integration options,  if I advise that it does not work correctly?

If you want to keep ID:103 as temperature to make an integration work, that is fine,  but know the solution does not work as the device’s internal controller will override the HA during the other time blocks.

Can I ask that you please add ID 10,  Geyser’s actual temperature so one can access it without using the water_heater entity.  It won’t break the current integration and will provide HA with this screen and easier access to the Actual Temperature

PR #4705, opened 2026-03-17
Device config for wifi refrigerator thermometer

decompil3d:
Adding a device config for the MC82 WiFi refrigerator thermometer. Example device on Amazon: https://www.amazon.com/dp/B0GH7LNTK4

### DIFF
--- a/custom_components/tuya_local/devices/bobyun_gatepro_opener.yaml
+++ b/custom_components/tuya_local/devices/bobyun_gatepro_opener.yaml
@@ -1,8 +1,8 @@
 name: Gate opener
 products:
-  -id: rzqbb9bm0mymxyqh
-   manufacturer: BoboYun Tech
-   model: gatePro
+  - id: rzqbb9bm0mymxyqh
+    manufacturer: BoboYun Tech
+    model: gatePro
 entities:
   - entity: cover
     class: gate


custom_components/tuya_local/devices/wifi_refrigerator_thermometer.yaml
```diff
@@ -0,0 +1,181 @@
+name: WiFi Refrigerator Thermometer
```
make-all: HA naming convention is to only capitalise the first word of device and entity names.

Also, I would leave out WiFi here, as it is theoretically possible that Zigbee or Bluetooth devices could match this config also.

custom_components/tuya_local/devices/wifi_refrigerator_thermometer.yaml
```diff
@@ -0,0 +1,181 @@
...
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: cancel
+            value: false
+          - dps_val: alarm
+            value: true
+
+  - entity: number
+    name: Upper Temp Limit (C)
```
make-all: Leave out the (C), spell out words in full.

This case could use `translation_key: maximum_temperature` instead of an explicit name.

custom_components/tuya_local/devices/wifi_refrigerator_thermometer.yaml
```diff
@@ -0,0 +1,181 @@
...
+        name: value
+        unit: C
+        range:
+          min: -400
+          max: 1000
+        mapping:
+          - scale: 10
+
+  - entity: number
+    name: Lower Temp Limit (C)
```
make-all: As above, could use `translation_key: minimum_temperature`

custom_components/tuya_local/devices/chtoocy_mc82_thermometer.yaml
```diff
@@ -0,0 +1,181 @@
...
+        type: integer
+        name: value
+        unit: min
+        range:
+          min: 1
+          max: 10
+
+  - entity: select
+    name: Alarm Repeat Interval
+    category: config
```
make-all, line 89: Use `translation_key: timer` (but keep the name) and format the values as 1h, 2h etc (same as original in Tuya protocol)

custom_components/tuya_local/devices/wifi_refrigerator_thermometer.yaml
```diff
@@ -0,0 +1,181 @@
...
+  - entity: select
+    translation_key: temperature_unit
+    category: config
+    dps:
+      - id: 102
+        type: string
+        name: option
+        mapping:
+          - dps_val: c
+            value: Celsius
```
make-all: use all lower case to match the translation keys

custom_components/tuya_local/devices/wifi_refrigerator_thermometer.yaml
```diff
@@ -0,0 +1,181 @@
...
+          - dps_val: f
+            value: Fahrenheit
+
+  - entity: switch
+    name: Alarm Sound
```
make-all: Use `translation_key: sound`

Issue #4707, opened 2026-03-17
Request support for IPENGEN EV Charger Station 32A


goblinmaks:
### Log message

```text
LOCAL DPS: {"updated_at": 1773734873.0730245, "3": "0", "4": 32, "10": 0, "17": 0, "18": false, "24": 17, "103": 0, "104": 0, "105": 0, "106": 0, "107": 0, "108": 0, "109": 218, "112": 0, "114": 0, "117": 0, "119": true}
```

### Product ID

cyoerhaalrfq5zds

### Product Name

IPENGEN GO-E31-32A

### DPS information

```text
{
  "result": {
    "model": "{\"modelId\":\"e1k56ymg\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"默认服务\",\"properties\":[
{\"abilityId\":1,\"accessMode\":\"ro\",\"code\":\"forward_energy_total\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-dp_lightning\",\"cloudless\":1,\"trigger\":\"direct\"},\"name\":\"正向电能\",\"typeSpec\":{\"type\":\"value\",\"max\":99999999,\"min\":0,\"scale\":2,\"step\":1,\"unit\":\"kW·h\"}},
{\"abilityId\":3,\"accessMode\":\"ro\",\"code\":\"work_state\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-zhuangtai\"},\"name\":\"Device State\",\"typeSpec\":{\"type\":\"enum\",\"range\":[\"0\",\"1\",\"2\",\"3\",\"4\",\"5\",\"6\",\"7\",\"8\"]}},
{\"abilityId\":4,\"accessMode\":\"rw\",\"code\":\"charge_cur_set\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-shezhi\",\"scope\":\"\"},\"name\":\"设置电流\",\"typeSpec\":{\"type\":\"value\",\"max\":32,\"min\":8,\"scale\":0,\"step\":1,\"unit\":\"A\"}},
[...]

PR #4708, opened 2026-03-17
[EV] Add support for IPENGEN GO-E31-32A

goblinmaks:
Add support for IPENGEN GO-E31-32A


goblinmaks:
#4707

make-all:
Was dp 102 dropped intentionally? It is listed in #4707

PR #4718, opened 2026-03-19
Update entity uniqueness validation in tests

rwalker777:
Refactor entity uniqueness check to allow mutually-exclusive duplicates based on available DPs.


make-all:
There is no mutual exclusivity between entities. If they exist then they are in the HA registry.

rwalker777:
Did you look at this at all?  It does not create 2 entities in HA.

make-all:
If the change to the test is "needed"to silence the CI, then duplicate entities have already been created.

This change is only ignoring a real problem, so cannot be accepted into the project.

PR #4722, opened 2026-03-20
Adding support for commonly found SR223 solar controller

TomPenrose:
Adding basic support for the SR223 Solar controller / water heater. It seems to be sold under many brand names, but all with the same underlying hardware. Manual for the controller can be found here: https://shuangri.com/uploads/soft/240830/SR223-EN(Wifi).pdf

Initially adding support for the most basic installation layout (Solar thermal collector and tank)

Current config supports:
- Temperature of Collector T1
- Temperature of Tank T2
- Pump speed for Pump R1 (Note, this also works in PWM mode for variable pumps)
- WiFi Signal Strength

### DIFF
--- a/custom_components/tuya_local/devices/shuangri_sr223_solarcontroller.yaml
+++ b/custom_components/tuya_local/devices/shuangri_sr223_solarcontroller.yaml
@@ -1,11 +1,12 @@
-name: SR223 Solar Water Heater
+name: Solar water heater
 products:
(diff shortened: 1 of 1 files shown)


TomPenrose:
Apologies @make-all, Learning curve for me here. If i'm understanding correctly, looks like I was using an older version of the config syntax. Updated and tested with the device and seems to be working well. Hopefilly this passes the tests now :)

PR #4723, opened 2026-03-20
added yaml file for voltcraft sem6500 devices

MatthiasSeu:
added yaml file for voltcraft sem6500 devices


MatthiasSeu:
Hopefully I have fixed all now.

MatthiasSeu:
Ready for review, all checks should pass now

make-all:
This seems to be a 100% match for smartplugv2_energy already.

Is there a specific reason why a new config was needed?

MatthiasSeu:
Yes - its redundant from the functionality - but not for the energy dashboard requirements:

<h3 class="text-text-100 mt-2 -mb-1 text-base font-bold">What the Energy Dashboard requires</h3>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">Home Assistant's Energy Dashboard only accepts <strong>sensors with:</strong></p>
<div class="overflow-x-auto w-full px-2 mb-6">
Requirement | Value
-- | --
state_class | total_increasing or total
device_class | energy
Unit | Wh or kWh

</div>

smartplugv2_energy:
class: measurement   # ← PROBLEM
unit: Wh             # ← OK

voltcraft_sem6500.yaml
class: total_increasing  # ← correct
unit: kWh                # ← correct

Maybe a correction of the existing smartplugv2_energy.yaml and ID inclusion of the Voltcraft IDs would lead to the same result.

make-all:
I have done that, but in general it is not expected to work, so if there are issues with it, then create a Reiman Sum Integration helper based on the power sensor.

PR #4733, opened 2026-03-23
Add AGL Mini Smart RX Wi-Fi device config

alisterlf:
## Summary
- add an AGL Mini Smart RX Wi-Fi device config for product id `e8ztbq4effpb0nmw`
- expose the cover-style control plus the related config switches, numbers, notifications, and status sensors
- list the new device in `DEVICES.md`
- align the top-level naming with the repo guidance and fix the cover control / card unlock naming based on the live Tuya DPs

## Validation
- parsed `custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml` with `npx --yes js-yaml`
- could not run `pytest`, `ruff`, or `yamllint` because Python tooling is not installed in this environment

### DIFF
--- a/custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml
+++ b/custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml
@@ -13,9 +13,9 @@ entities:
         optional: true
         mapping:
           - dps_val: true
-            value: open
-          - dps_val: false
             value: close
+          - dps_val: false
+            value: open
       - id: 3
         type: boolean
         name: action
@@ -59,7 +59,7 @@ entities:
         type: boolean
   - entity: switch
     category: config
-    name: Allow IC card unlock
+    name: Allow RF remote unlock
     icon: 'mdi:nfc-variant'
     dps:
       - id: 109
--- a/custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml
+++ b/custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml
@@ -1,0 +1,222 @@
+name: Gate lock
+products:
+  - id: e8ztbq4effpb0nmw
+    manufacturer: AGL
+    name: Mini Smart RX Wi-Fi
+entities:
+  - entity: cover
+    class: gate
+    dps:
+      - id: 118
+        type: boolean
+        name: control
+        optional: true
+        mapping:
+          - dps_val: true
+            value: open
+          - dps_val: false
+            value: close
+      - id: 3
+        type: boolean
+        name: action
+        mapping:
+          - dps_val: true
+            value: closed
+          - dps_val: false
+            value: opened
+  - entity: select
+    category: config
+    name: Beep
+    icon: 'mdi:volume-high'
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml
```diff
@@ -0,0 +1,222 @@
...
+        optional: true
+        mapping:
+          - dps_val: true
+            value: close
+          - dps_val: false
+            value: open
+      - id: 3
+        type: boolean
+        name: action
+        mapping:
```
make-all, line 22: Doesn't the gate take time to open and close?Perhaps this should this be combined with the command to determine when a command is in-progress, or at least return a null (unknown) status to avoid controls becoming disabled in the UI when they should not be?

custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml
```diff
@@ -0,0 +1,222 @@
...
+        type: integer
+        range:
+          min: 1
+          max: 5
+        unit: s
+  - entity: number
+    category: config
+    icon: 'mdi:clock-end'
+    name: Auto close timeout
+    dps:
```
make-all, line 100: `class: duration`

custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml
```diff
@@ -0,0 +1,222 @@
...
+    name: Allow inside button unlock
+    icon: 'mdi:gesture-tap-button'
+    dps:
+      - id: 107
+        name: switch
+        type: boolean
+  - entity: number
+    category: config
+    name: Fast close delay
+    icon: 'mdi:timer'
```
make-all, line 87: `class: duration`, probably then the icon wlll be unnecessary

custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml
```diff
@@ -0,0 +1,222 @@
...
+        type: integer
+        range:
+          min: 1
+          max: 1200
+        unit: s
+  - entity: number
+    category: config
+    icon: 'mdi:alarm'
+    name: Open door alarm delay
+    mode: box
```
make-all, line 112: It is usually better to leave the decision of what representation to use to HA. Is there a good reason for requiring box here?

custom_components/tuya_local/devices/agl_mini_smart_rx_wi_fi.yaml
```diff
@@ -0,0 +1,222 @@
...
+    name: Mobile notifications
+    icon: 'mdi:message-alert'
+    dps:
+      - id: 122
+        name: switch
+        type: boolean
+  - entity: select
+    category: config
+    name: Backup options
+    icon: 'mdi:backup-restore'
```
make-all, line 188: I would mark this as hidden, as it isn't obvious how to make use of it from a local connection.

make-all:
Unresponsive contributor

PR #4734, opened 2026-03-23
Adding new device - KKT Kolbe (kitchen hood)

kordek212:
kitchen hood - KKT Kolbe

### DIFF
--- a/custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
+++ b/custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
@@ -1,4 +1,4 @@
-name: Rangehood
+name: Kitchen Hood
 products:
   - id: bgvbvjwomgbisd8x
     manufacturer: KKT Kolbe


custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
```diff
@@ -0,0 +1,54 @@
...
+          min: 29
+          max: 1000
+      - id: 107
+        type: string
+        name: color_mode
+        mapping:
+          - dps_val: white
+            value: white
+          - dps_val: colour
+            value: hs
```
make-all: There is no color adjustment, so this mode is impossible. There are also no dps for controlling scene and music, so those also seem impossible, leaving only white as an option.
In that case, better to rename this dp to something other than color_mode so it can be observed but does not interfere with operation of the light in HA.

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
```diff
@@ -0,0 +1,54 @@
...
+        type: boolean
+        name: switch
+      - id: 102
+        type: integer
+        name: speed
+        range:
+          min: 1
+          max: 9
+  - entity: switch
+    name: Okap Led
```
make-all: This is duplicating the light below

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
```diff
@@ -0,0 +1,54 @@
...
+    dps:
+      - id: 104
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Okap Led 2
```
make-all: This should probably be a light entity rather than a switch. A more descriptive name would also help.

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
```diff
@@ -0,0 +1,54 @@
+name: KKT Kolbe SOLO HCM
+products:
+  - id: bgvbvjwomgbisd8x # Ensure this matches your exact product key
+    name: KKT Kolbe SOLO HCM Extractor Hood
+entities:
+  - entity: fan
+    name: Okap
+    class: fan
```
make-all: Fan entities do not have device classes.

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
```diff
@@ -0,0 +1,54 @@
+name: KKT Kolbe SOLO HCM
```
make-all: The top level name should be unbranded and a simple description of what the device is ("Rangehood" for example)

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
```diff
@@ -0,0 +1,54 @@
...
+products:
+  - id: bgvbvjwomgbisd8x # Ensure this matches your exact product key
+    name: KKT Kolbe SOLO HCM Extractor Hood
```
make-all: This should be split into `manufacturer` and `model`. No need to repeat what is in the name above.

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
```diff
@@ -0,0 +1,54 @@
...
+entities:
+  - entity: fan
+    name: Okap
```
make-all: Name should be in English, or a translation_key used.

kordek212: Updated. Entity names are now in English (Hood fan, Main light, Auxiliary light).

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
kordek212: Done. Product metadata is now split into manufacturer and model fields (manufacturer: KKT Kolbe, model: SOLO HCM).

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
kordek212: Updated. Top-level device name is now unbranded and generic.

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
kordek212: Removed. The fan entity no longer declares class: fan.

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
kordek212: Changed. The secondary LED control (dp 4) is now a light entity with an English descriptive name (Auxiliary light).

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
kordek212: Fixed. Removed the duplicate switch entity that overlapped with the light control.

custom_components/tuya_local/devices/kkt_kolbe_solo_hcm.yaml
kordek212: Updated. Removed unsupported colour/scene/music mappings and renamed dp 107 from color_mode to mode, keeping only white so it can be observed without interfering with HA light behavior.

PR #4735, opened 2026-03-23
Add device configuration for Smart Light RGB48 (RGBIC string lights)

starfire0606:
Adds a Tuya Local device profile for the Smart Light RGB48, sold under various brand names as RGBIC outdoor/indoor string lights controlled via the SmartLife app.
Device info:

Product name: Smart Light RGB48
Product ID: 2znvfkx64d4fkzxp
Protocol: 3.4
Tested with 48-bulb RGBIC outdoor string lights

DPs implemented:

DP 20: switch (boolean)
DP 21: color mode (white/colour/scene/music) — colour and music hidden; selecting colour directly causes repeated hs color errors in HA logs as there is no direct RGB DP
DP 22: brightness (10–1000)
DP 23: color temperature (0–1000, ~2700–6500K)
DP 26: countdown timer (0–86400s)
DP 56: scene data (base64, animated multicolor scenes saved via Smart Life DIY editor)
DP 58: LED count / segment configuration (1–5000)
DP 59: draw tool (base64, per-bulb static or gradient color control)
DP 101: segment mode (RGB/RGBIC/White/Scene)

Notes:

DP 56 and DP 59 are both write-only in practice — the device never reports them in status polls
DP 56 accepts animated DIY scene payloads from Smart Life. All DP 56 scenes animate; static per-bulb color is NOT possible via this DP
DP 59 implements the full per-bulb paint protocol used by Smart Life's draw tool. Supports static multicolor (effect=0) as well as gradient (effect=1). Payload format per bulb: [0x01, 0x02, effect, hue_hi, hue_lo, sat, 0x64, 0x00, 0x00, 0x81, 0x00, bulb_index] — send init frame [0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x64, 0x00, 0x00] first, then one payload per bulb (index 0–47) with ~50ms delay
DP 24 (HSV color hex) is present but does not persist — device reverts on next poll. Use DP 59 for persistent color control as the SmartLife app does natively

Entities created:

light — on/off, brightness, color temp, scene/music effects via DP 20/21/22/23
select — Segment Mode (RGB/RGBIC/White/Scene) via DP 101
number — LED Count configuration via DP 58
time — Countdown Timer via DP 26
text — Scene Data (base64, animated scenes) via DP 56
text — Draw Tool (base64, static/gradient per-bulb color) via DP 59


custom_components/tuya_local/devices/smart light rgb48.yml
```diff
@@ -0,0 +1,113 @@
...
+        type: integer
+        name: color_temp
+        range:
+          min: 0
+          max: 1000
+        mapping:
+          - target_range:
+              min: 2700
+              max: 6500
+      - id: 24
```
make-all: If this does not work, why include it?

Wouldn't it be better to figure out the dp 59 format so that could be used here?

The Query Things Data Model result may even describe the format in the description field (probably in Chinese, but machine translation does a reasonable enough job of making sense of it)

custom_components/tuya_local/devices/smart light rgb48.yml
```diff
@@ -0,0 +1,113 @@
...
+# Protocol: 3.4
+#
+# Notes:
+# - Direct color control via DP 24 does not persist reliably on this device.
+#   Colors set via DP 59 (base64 scene payload) work correctly and persistently.
+#   This matches how the SmartLife app controls the device natively.
+# - DPs 47 and 53 are present in device responses but their function is unknown.
+#   Included as optional hidden attributes to improve device matching quality.
+
+name: RGBIC Patio String Lights
```
make-all: Follow HA naming convention of only capitalising the first word of device or entity names (abbreviatons excepted). And try to keep names short, as long names in the HA UI can be problematic, particularly on mobile interfaces.

custom_components/tuya_local/devices/smart light rgb48.yml
```diff
@@ -0,0 +1,113 @@
...
+products:
+  - id: 2znvfkx64d4fkzxp
+    name: Smart Light RGB48
```
make-all: Product should have manufacturer and model info rather than a name

custom_components/tuya_local/devices/smart light rgb48.yml
```diff
@@ -0,0 +1,113 @@
...
+              min: 0
+              max: 1000
+          - name: v
+            bytes: 2
+            range:
+              min: 0
+              max: 1000
+      - id: 26
+        type: integer
+        name: timer
```
make-all: This should be a separate entity.

custom_components/tuya_local/devices/smart light rgb48.yml
```diff
@@ -0,0 +1,113 @@
...
+        optional: true
+      - id: 47
+        type: integer
+        name: unknown_47
```
make-all: Details of these should be in the Query Things Data Model cloud API call from the developer portal API Explorer

custom_components/tuya_local/devices/rgb48_lightstring.yaml
```diff
@@ -0,0 +1,113 @@
...
+        optional: true
+        hidden: true
+      - id: 53
+        type: integer
+        name: unknown_53
+        optional: true
+        hidden: true
+  - entity: select
+    name: Segment Mode
+    icon: "mdi:led-strip-variant"
```
make-all, line 54: Should this be `category: config`? Is it setting the type of strip connected as an installation option, or duplicating the color_mode of the light, to be used in operation?

custom_components/tuya_local/devices/smart light rgb48.yml
```diff
@@ -0,0 +1,113 @@
...
+          - dps_val: "SCENE"
+            value: "Scene"
+  - entity: text
+    name: Holiday Scene
+    category: config
+    dps:
+      - id: 59
+        type: base64
+        name: value
+  - entity: sensor
```
make-all: sensor or number? Can the device really count the LEDs on its own, or should this be for setting the strip length?

custom_components/tuya_local/devices/smart light rgb48.yml
starfire0606: Removed DP 24 entirely as suggested. Verified via Query Things Data Model that colour_data is type string, not hex as originally defined, which explains the persistence issue. Color control is handled via DP 59 (draw_tool) and DP 56 (rgbic_linerlight_scene).

custom_components/tuya_local/devices/smart light rgb48.yml
starfire0606: Fixed — changed to “RGBIC string light” following HA naming convention.

custom_components/tuya_local/devices/smart light rgb48.yml
starfire0606: Updated to use model: Smart Light RGB48 without a name field. Device is sold under various generic brand names on Amazon so no reliable manufacturer can be confirmed.

custom_components/tuya_local/devices/smart light rgb48.yml
starfire0606: Fixed — moved DP 26 (countdown) to its own number entity with the correct range (0-86400s) as confirmed by the data model.

custom_components/tuya_local/devices/smart light rgb48.yml
starfire0606: Checked Query Things Data Model via API Explorer — neither DP 47 nor 53 appear in the data model for this product. Removed both from the config entirely.

custom_components/tuya_local/devices/rgb48_lightstring.yaml
starfire0606: Added category: config as suggested — it’s an installation setting for the strip type.

custom_components/tuya_local/devices/smart light rgb48.yml
starfire0606: Confirmed via Query Things Data Model: DP 58 is led_number_set — “represents the number of LEDs in the string, used to divide string length and segment count.” Changed to number entity with range 1-5000 and category: config.​​​​​​​​​​​​​​​​

starfire0606:
Also fixed DP 56 and 59 from type: raw to type: base64 as raw is not a valid type in tuya-local.

PR #4740, opened 2026-03-24
Update moes_star_projector.yaml

pacorola:
Revised scene support (Default ones) and eliminated startup error caused by using two pds on a single command for scenes,

### DIFF
--- a/custom_components/tuya_local/devices/moes_star_projector.yaml
+++ b/custom_components/tuya_local/devices/moes_star_projector.yaml
@@ -26,17 +26,17 @@ entities:
         format:
           - name: h
             bytes: 2
-            range: 
+            range:
               min: 0
               max: 360
           - name: s
             bytes: 2
-            range: 
+            range:
               min: 0
               max: 1000
           - name: v
             bytes: 2
-            range: 
+            range:
               min: 0
               max: 1000
       - id: 25


pacorola:
Finally! Everything OK. To summarize, que previous yaml worked flawlessly, but triggered errors in home assistant log when readiing scenes.

.yamllint.yaml
```diff
@@ -3,3 +3,6 @@ extends: default
 rules:
   document-start: disable
   new-lines: disable
+  line-length:
```
make-all, line 6: This is not the proper way to address lint rule violations

custom_components/tuya_local/devices/moes_star_projector.yaml
```diff
@@ -19,29 +22,53 @@ entities:
           - dps_val: colour
             value: hs
           - dps_val: scene
-            value: Scene
+            value: scene
```
make-all, line 22: Why this change? Scene is not a standard mode / effect so there is no translation key to deal with here.

custom_components/tuya_local/devices/moes_star_projector.yaml
```diff
@@ -19,29 +22,53 @@ entities:
...
       - id: 24
         name: rgbhsv
         type: hex
         format:
           - name: h
             bytes: 2
-            range:
-              min: 0
-              max: 360
+            range: {min: 0, max: 360}
```
make-all: Why this change?

custom_components/tuya_local/devices/moes_star_projector.yaml
```diff
@@ -19,29 +22,53 @@ entities:
...
-              max: 1000
+            range: {min: 0, max: 1000}
+
       - id: 25
         name: scene_data
         type: hex
+
+  # Selector nativo de escenas internas
+  - entity: select
+    name: Scenes
```
make-all, line 46: You should use `translation_key: scene` here

custom_components/tuya_local/devices/moes_star_projector.yaml
```diff
@@ -2,14 +2,17 @@ name: Star projector
 products:
   - id: cw7kinnselbesfp9
     manufacturer: Moes
+
 entities:
+  # Luz principal RGB
```
make-all: Additional comments are not needed, especially ones in foreign languages

custom_components/tuya_local/devices/moes_star_projector.yaml
```diff
@@ -2,14 +2,17 @@ name: Star projector
...
   - entity: light
     name: Background
     icon: "mdi:projector"
     dps:
       - id: 103
         type: boolean
         name: switch
+
```
make-all: Additional whitespace is not needed

.yamllint.yaml
pacorola: Oh! It was suggested by Copilot, I don't know how else to fix such error.

custom_components/tuya_local/devices/moes_star_projector.yaml
pacorola: Initially python test requested the translation key.

custom_components/tuya_local/devices/moes_star_projector.yaml
pacorola: Thanks, removed

custom_components/tuya_local/devices/moes_star_projector.yaml
pacorola: But there its just stylistic change. Just in case, can revert if you consider it to be better

custom_components/tuya_local/devices/moes_star_projector.yaml
pacorola: Must have been a copilot fix for linting errors, but I reverted it now.

custom_components/tuya_local/devices/moes_star_projector.yaml
pacorola: It was there already.

custom_components/tuya_local/devices/moes_star_projector.yaml
pacorola: Deleted all unnecesary spaces. ¡New file incoming!

pacorola:
Thanks for your comments @make-all !!!!   Let's hope its now on spec pending line limits for scenes (those long character strings cannot be avolided)

make-all:
The long lines can be dealt with the same way as in rgbcw_lightbulb.yaml or any other file dealing with long hex or base64 strings.

custom_components/tuya_local/devices/moes_star_projector.yaml
make-all: So remove the name so the translation_key is effective

custom_components/tuya_local/devices/moes_star_projector.yaml
make-all: That was for the entity below, I assume. There is no translation key with coverage of light effect naming.

make-all:
I don't think any of these are necessary changes. There is a text entity giving full access to scenes already, so you can use an automation to send any specific scenes you want (including controlling the automation from an input_select helper).

PR #4742, opened 2026-03-24
Added support for monTEK ceiling TV mount

jmahne222:


custom_components/tuya_local/devices/montek_tm1011e_tvmount.yaml
```diff
@@ -0,0 +1,117 @@
...
+    icon: "mdi:numeric-3-circle-outline"
+    dps:
+      - id: 103
+        type: string
+        name: button
+        mapping:
+          - dps_val: mem3
+            value: true
+  - entity: sensor
+    name: Model
```
make-all, line 99: Static information like this does not need to be a sensor.
It can be an attribute on another entity.

custom_components/tuya_local/devices/montek_tm1011e_tvmount.yaml
```diff
@@ -0,0 +1,117 @@
...
+    category: diagnostic
+    icon: "mdi:tag"
+    dps:
+      - id: 102
+        type: string
+        name: sensor
+  - entity: binary_sensor
+    name: Status
```
make-all, line 107: Leave out the name so HA default translations are used instead.

Also add another attribute with the raw `fault_code` so the information is not lost.

custom_components/tuya_local/devices/montek_tm1011e_tvmount.yaml
```diff
@@ -0,0 +1,117 @@
...
+    icon: "mdi:stop"
+    dps:
+      - id: 101
+        type: string
+        name: button
+        mapping:
+          - dps_val: move_stop
+            value: true
+  - entity: button
+    name: Move up
```
make-all, line 19: Try to follow the same naming as speaka_sptvcm510_tvmount (just "Up" and "Down" for these)

custom_components/tuya_local/devices/montek_tm1011e_tvmount.yaml
```diff
@@ -0,0 +1,117 @@
...
+          - dps_val: mem3
+            value: true
+  - entity: sensor
+    name: Model
+    category: diagnostic
+    icon: "mdi:tag"
+    dps:
+      - id: 102
+        type: string
+        name: sensor
```
make-all, line 105: This should be static info, so does not need to be a sensor.

The value could also be used to combine this and the speaka_sptvcm510_tvmount config into a single file, since they only differ in the movement options available.

make-all (changes requested):

make-all:
Non responsive contributor

PR #4752, opened 2026-03-25
feat: add Reventon / Holtop Smart ERV heat recovery ventilator

numanx11:
Adds device config for the Reventon / Holtop Smart ERV (product id: ddrpa4plk6fsvz9l) with support for fan on/off, supply and exhaust fan speed (3-step and 10-step modes), bypass temperature controls, filter management, CO2/TVOC/humidity/PM2.5 sensors, indoor/outdoor temperature sensors, and diagnostic entities (fault, defrost, filter status).

### DIFF
--- a/custom_components/tuya_local/devices/reventon_holtop_smart_erv.yaml
+++ b/custom_components/tuya_local/devices/reventon_holtop_smart_erv.yaml
@@ -41,6 +41,13 @@ entities:
                     value: 100
               - dps_val: false
                 value_redirect: speed_3step
+                mapping:
+                  - dps_val: "0"
+                    value: 0
+                  - dps_val: "1"
+                    value: 50
+                  - dps_val: "2"
+                    value: 100
       - id: 104
         name: speed_3step
         type: string
@@ -90,6 +97,13 @@ entities:
                     value: 100
               - dps_val: false
                 value_redirect: speed_3step
+                mapping:
+                  - dps_val: "0"
+                    value: 0
+                  - dps_val: "1"
+                    value: 50
+                  - dps_val: "2"
+                    value: 100
       - id: 105
         name: speed_3step
         type: string
@@ -134,7 +148,7 @@ entities:
 
   - entity: number
     name: Bypass temperature delta
-    class: temperature
+    class: temperature_delta
     category: config
     dps:
       - id: 108
@@ -146,7 +160,7 @@ entities:
           max: 15
 
   - entity: select
-    name: Filter life setting
+    name: Filter life
     icon: mdi:air-filter
     category: config
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/reventon_holtop_smart_erv.yaml
```diff
@@ -0,0 +1,295 @@
...
+    class: carbon_dioxide
+    dps:
+      - id: 111
+        name: sensor
+        type: integer
+        unit: ppm
+        class: measurement
+
+  - entity: sensor
+    name: TVOC level
```
make-all: Use `translation_key: air_quality` here, as while this may be based on TVOC only, it is not really a level. The translation key can replace both name and icon

custom_components/tuya_local/devices/reventon_holtop_smart_erv.yaml
```diff
@@ -0,0 +1,295 @@
...
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 119
+        name: fault_code
+        type: bitfield
+
+  - entity: binary_sensor
+    name: Defrost active
```
make-all: Use `translation_key: defrost` instead of this name

custom_components/tuya_local/devices/reventon_holtop_smart_erv.yaml
```diff
@@ -0,0 +1,295 @@
...
+  - entity: binary_sensor
+    name: Filter cleaning required
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 123
+        name: sensor
+        type: boolean
+
+  - entity: sensor
```
make-all: This should be combined with the fan speeds above to select which setting is active within a single entity,

custom_components/tuya_local/devices/reventon_holtop_smart_erv.yaml
numanx11: Changed from name: TVOC level to translation_key: air_quality

custom_components/tuya_local/devices/reventon_holtop_smart_erv.yaml
numanx11: Changed from name: Defrost active to translation_key: defrost

custom_components/tuya_local/devices/reventon_holtop_smart_erv.yaml
numanx11: Fan entities with conditional speed - Replaced the switch + 4 select entities + fan speed mode sensor with 2 fan entities:

Supply fan (translation_key: ventilation) - includes switch (dp 101) + conditional speed mapping using dp 125 to select between 10-step (dp 102) and 3-step (dp 104)
Exhaust fan - speed-only (no switch), same conditional mapping using dp 125 to select between 10-step (dp 103) and 3-step (dp 105)

PR #4753, opened 2026-03-25
feat: add Tuya QCSS600ZB 3 gang ZigBee remote control

numanx11:
Adds device config for the Tuya QCSS600ZB 3 gang ZigBee remote (product id: famkxci2) with three button event entities supporting single click, double click, and long press actions, plus a battery level diagnostic sensor.


make-all:
This can use the existing Loratap six switch config (just hide the entities for the buttons you don't have)

Issue #3738, opened 2025-09-14
Request for additional functionality to be pulled in KRain KRX8


YSIGE123:
### Description

I have added some key functionality to the irrigation controller which I have tested and works (yaml attached).

  - entity: switch
    name: Allow Program Starts
    icon: mdi:timetable
    dps:
      - id: 121
        type: boolean
        name: switch
  - entity: switch
    name: Use Watering Delay
    icon: mdi:timetable
    dps:
      - id: 136
        type: boolean
        name: switch

I've also identified the id 126 which is a selection setting - integer 1-15. This selection sets the number of days watering is delayed when id 136 is turned on. After the prescribed number of days id 136 turns itself off again. id 126 name: Default Watering Delay (Days). Could this please be added to the device?

[tuya_local-01K549BJYVPJJY27647WWK0T7J-Irrigation Controller-f2686c0129b184acec23c3593abc7b29.json](https://github.com/user-attachments/files/22321043/tuya_local-01K549BJYVPJJY27647WWK0T7J-Irrigation.Controller-f2686c0129b184acec23c3593abc7b29.json)

    "name": "Irrigation Controller",
    "type": "krain_krx8_irrigation_controller",
    "device_id": "**REDACTED**",
    "device_cid": "",
    "local_key": "**REDACTED**",
    "host": "**REDACTED**",
    "protocol_version": 3.4,
    "tinytuya_version": "1.17.4",
    "api_version_set": 3.4,
    "api_version_used": 3.4,
    "api_working": true,
    "status": {},
    "cached_state": {
      "updated_at": 1757856693.6597211,
      "1": false,
      "101": false,
      "102": false,
      "103": false,
      "104": false,
      "105": false,
      "106": false,
      "107": false,
      "108": false,
      "117": false,
      "118": false,
      "119": false,
      "120": false,
      "130": true,
      "131": true,
      "132": false,
      "135": true,
      "136": false, 
      "137": false,
      "134": "AAAAAAAAAAAAAFl0AAABABAABhezAgAAAAAAAAAAGQkOFR8hAwAA",
      "126": 9, 
      "121": false 
    },

[krain_krx8_irrigation_controller.yaml](https://github.com/user-attachments/files/22320844/krain_krx8_irrigation_controller.yaml)

### Confirmation

- [x] I am sure this is a bug or improvement, that is well enough described that it can be implemented.
- [x] This report is for tuya-local, **not** for localtuya.
[...]

PR #4759, opened 2026-03-25
KRX6

camcamnz:
Added KRain KRX6 Device

### DIFF
--- a/custom_components/tuya_local/devices/krain_krx6_irrigation_controller.yaml
+++ b/custom_components/tuya_local/devices/krain_krx6_irrigation_controller.yaml
@@ -201,4 +201,3 @@ entities:
         type: base64
         optional: true
         name: value
-        


camcamnz:
I agree that they should be similar, with the KX6 just having 2x less Zones and 1x less Program, and that was my starting point. 
However for example DP117 on the KX6 starts Program A straight away. It is implied in the config for the KX8 that this enables that program for automatic use/scheduling (from the logo and the fact it is in the config section), which is done by DP122 on the KX6.
I suspect that the KX8 is actually the same, and probably should be updated the same and possibly merged, but without having one to test I cant be sure.

I updated the naming to clarify/reflect these differences.

make-all:
In the last comment on #3738, I expressed my uncertainty about how the Start / Enable Program dps worked, so the current state of the KRX8 config is just a guess and open to improvement. Most likely these work the same across both models.

I will improve the KRX8 config based on your submission, and make outputs 7 and 8 optional (actual info about the device had up to 16 outputs, so I guess there are other variants with more than 8 also) to also cover the KRX6.

PR #4763, opened 2026-03-26
Create magnovent_cefiro_with_lights.yaml

R4gN4rk:
New device (fan with presets) added.
Website: https://magnovent.eu/cefiro-modern-ceiling-fan/

### DIFF
--- a/custom_components/tuya_local/devices/magnovent_cefiro_with_lights.yaml
+++ b/custom_components/tuya_local/devices/magnovent_cefiro_with_lights.yaml
@@ -1,8 +1,8 @@
-name: Magnovent Cefiro with light
-# products:
-#   - id: lb9da3os0ihifq7u
-#     manufacturer: Magnovent
-#     model: Cefiro
+name: Ceiling fan with light
+products:
+  - id: lb9da3os0ihifq7u
+    manufacturer: Magnovent
+    model: Cefiro
 entities:
   - entity: fan
     translation_only_key: fan_with_presets


custom_components/tuya_local/devices/magnovent_cefiro_with_lights.yaml
```diff
@@ -0,0 +1,107 @@
...
+          - dps_val: custom
+            value: custom
+      - id: 3
+        type: integer
+        name: speed
+        range:
+          min: 1
+          max: 6
+        mapping:
+          - target_range:
```
make-all: Target range is not needed here. It is identical to the range

custom_components/tuya_local/devices/magnovent_cefiro_with_lights.yaml
```diff
@@ -0,0 +1,107 @@
+name: Magnovent Cefiro with light
+# products:
+#   - id: UNKNOWN
```
make-all: It would be better to find it, either in the tuya portal, or `tinytuya scan` result

custom_components/tuya_local/devices/magnovent_cefiro_with_lights.yaml
```diff
@@ -0,0 +1,107 @@
...
+          min: 1
+          max: 100
+      - id: 17
+        name: color_temp
+        type: integer
+        range:
+          min: 0
+          max: 100
+        mapping:
+          - invert: true
```
make-all: There should only be one default entry in the mapping list. It is unusual for Tuya lights to need the color temp to be inverted so maybe try without that first.

custom_components/tuya_local/devices/magnovent_cefiro_with_lights.yaml
R4gN4rk: Understood. I've tested it without "target_range" and you're absolutely right. Corrected.

custom_components/tuya_local/devices/magnovent_cefiro_with_lights.yaml
R4gN4rk: I did not know how to find it, but I looked for it in tuya portal: bf3c35156d68b66866n2to
I'll update the file with this info.

custom_components/tuya_local/devices/magnovent_cefiro_with_lights.yaml
R4gN4rk: You are right again. Same behavior after removal of "invert: true" line. 
I'll correct it.

custom_components/tuya_local/devices/magnovent_cefiro_with_lights.yaml
make-all: No, that is a device id, product id should be listed by the Query Device Details API call as product id, and shows up in some other places as productKey.  There are lots of other ids used in the cloud API, but none of them are useful to identify devices locally.

R4gN4rk: Okey, clear now. This is the product_id: lb9da3os0ihifq7u
I'll write it in the code.
Thanks for your support!

Issue #4682, opened 2026-03-12
Request support for Air purifier MORENTO MR5866


alfry82:
### Log message

```text
Registratore: custom_components.tuya_local.config_flow
Fonte: custom_components/tuya_local/config_flow.py:435
Integrazione: Tuya Local (documentazione, problemi)
Prima occorrenza: 14:05:49 (2 occorrenze)
Ultimo accesso: 14:28:09

Partial cloud device spec: [{"id": 1, "name": "switch", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 2, "name": "pm25", "type": "Integer", "format": "{\"unit\":\"ug/m3\",\"min\":0,\"max\":999,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 3, "name": "mode", "type": "Enum", "format": "{\"range\":[\"auto\",\"sleep\"]}", "enumMap": {}}, {"id": 5, "name": "filter_life", "type": "Integer", "format": "{\"unit\":\"%\",\"min\":0,\"max\":99,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 7, "name": "child_lock", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 8, "name": "light", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 11, "name": "filter_reset", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 18, "name": "countdown_set", "type": "Enum", "format": "{\"range\":[\"cancel\",\"2h\",\"5h\"]}", "enumMap": {}}, {"id": 19, "name": "countdown_left", "type": "Integer", "format": "{\"unit\":\"min\",\"min\":0,\"max\":480,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 21, "name": "air_quality", "type": "Enum", "format": "{\"range\":[\"good\"]}", "enumMap": {}}]

Registratore: custom_components.tuya_local.config_flow
Fonte: custom_components/tuya_local/config_flow.py:445
Integrazione: Tuya Local (documentazione, problemi)
Prima occorrenza: 14:05:49 (2 occorrenze)
Ultimo accesso: 14:28:09

Device matches treatlife_ds02_fan with quality of 38%. LOCAL DPS: {"updated_at": 1773320730.6875203, "1": true, "2": 9, "3": "auto", "5": 78, "7": false, "8": true, "18": "cancel", "21": "good"}
[...]

PR #4764, opened 2026-03-26
Add support for Morento air purifier MR5866

alfry82:


alfry82:
Add support for Morento MR5866 air purifier

This PR adds a working tuya-local configuration for the Morento MR5866.

Closes #4682

custom_components/tuya_local/devices/morento_mr5866_air_purifier.yaml
```diff
@@ -33,7 +33,7 @@ entities:
 
   - entity: sensor
     class: pm25
-    translation_key: pm25
+    name: pm25
```
make-all: Just remove the name entirely, it is not needed when the class is the same.

custom_components/tuya_local/devices/morento_mr5866_air_purifier.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - id: 2
+        name: sensor
+        type: integer
+        class: measurement
+        unit: ugm3
+
+  - entity: sensor
+    translation_key: filter_life
+    category: diagnostic
+    icon: "mdi:air-filter"
```
make-all: icon is not needed when there is a translation_key, as that should already define an icon as well.

custom_components/tuya_local/devices/morento_mr5866_air_purifier.yaml
```diff
@@ -0,0 +1,113 @@
...
+    translation_key: timer
+    category: config
+    icon: "mdi:timer-outline"
+    dps:
+      - id: 18
+        name: option
+        type: string
+        mapping:
+          - dps_val: cancel
+            value: Cancel
```
make-all: The translation_key above also covers these, but only if the keys are used as the values (keys are all lower case, and for timer values are cancel, 2h, 5h, 8h...

PR #4777, opened 2026-03-29
Update echief_sgh01_planter

MatsNl:
Sensor as config class gave me an error. Alternatively it can be set to diagnostic. I've also added the unit: d for the remaining days.


make-all:
I think it should be diagnostic, because it is related to the timer above it which is config.

MatsNl:
Ja you basically configure how often you want to feed the plants. And then the sensor runs back from that duration to zero. Does that fit better with diagnostic? Happy to change it.

PR #4785, opened 2026-03-30
Add Tongou TOVTH-216WTTDA Temp/Humidity Switch

terryfork:


custom_components/tuya_local/devices/tongou_tovth-216_wttda_switch.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - entity: sensor
+    class: humidity
+    dps:
+      - id: 46
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: switch
+    name: Switch1
```
make-all: Instead of explicit name, use 
```
translation_key: switch_x
translation_placeholders:
  x: "1"
```

PR #4788, opened 2026-03-31
Add Milano ceiling fan

sschwieb:


custom_components/tuya_local/devices/milano_ceiling_fan.yaml
```diff
@@ -0,0 +1,113 @@
...
+  - entity: select
+    translation_key: timer
+    dps:
+      - id: 22
+        name: option
+        type: string
+        mapping:
+          - dps_val: cancel
+            value: cancel
+            available: std
```
make-all: Is this necessary?
I think you have copy-pasted this from another config that supports multiple models. But this model should be using just one specific set of values, so this mapping can be simplified to only support what is actually used.

sschwieb: Good point, let me take a closer look at what's needed and what not!

sschwieb:
@make-all , I cleaned up the PR, and tested it against the 3 Milano fans that I own. So I hope it will work with other models & for other users as well...

custom_components/tuya_local/devices/milano_ceiling_fan.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+  - entity: select
+    translation_key: timer
+    dps:
+      - id: 22
+        name: option
+        type: string
+        mapping:
+          - dps_val: cancel
+            value: off
```
make-all: This should be cancel, as that is the value that has translations defined for it.

custom_components/tuya_local/devices/milano_ceiling_fan.yaml
```diff
@@ -0,0 +1,101 @@
...
+            value: "8h"
+          - dps_val: "9h"
+            value: "9h"
+
+  - entity: sensor
+    name: timer_remaining
+    category: diagnostic
+    dps:
+      - id: 23
+        name: value
```
make-all: sensor, not value

PR #4789, opened 2026-04-01
Add wmdl 2c dual channel meter

medisoft:
Adds support for WMDL-2C device

This PR includes the yaml file for a (WMDL-2C) Smart Energy Meter WIFI Bidirection 2 Channel


medisoft:
fixed the issue with yamllint

medisoft:
fixed the force: true parameter.

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
+---
+name: Dual channel energy meter
+products:
+  - id: zcmq0kvd01bn6s1n
+    manufacturer: WMDL
+    model: (WMDL-2C) Smart Energy Meter WIFI Bidirection 2 Channel
```
make-all: You can use `model_id` instead of putting the ID in parentheses.

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+entities:
+  - entity: sensor
+    name: Power Phase A
```
make-all: Instead of an explicit English name, use 
```
translation_key: power_x
translation_placeholders:
  x: A
```

Same for voltage and current, and phase B.

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+    class: power
+    dps:
+      - id: 19
+        type: integer
+        name: sensor
+        unit: W
+        class: measurement
+        precision: 1
```
make-all: Specifying precision should not be necessary.

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+      - id: 112
+        type: integer
+        name: sensor
+        unit: A
+        class: measurement
+        force: true
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: Total Forward Energy
```
make-all: Use `translation_key: energy_consumed` instead of this name

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+      - id: 119
+        type: integer
+        name: sensor
+        unit: kWh
+        class: total_increasing
+        precision: 2
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Total Reverse Energy
```
make-all: Use `translation_key: energy_produced` instead of this name

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+    dps:
+      - id: 107
+        type: integer
+        name: value
+        unit: W
+        range:
+          min: 0
+          max: 150000
+        mapping:
+          - dps_val:
```
make-all: dps_val requires a value to be specified. If you meant to specify an empty string, then you need to use `""`, or if you means to specify a missing value, you need to use `null`, but in that case the dp also needs to be specifed as `optional: true` to indicate it is not always present, and in most cases it should just be left as is (Unknown rather than Unavailable) so that it can still be set from an action.

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+            invalid: true
+          - scale: 1
```
make-all: scale: 1 is not needed

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+      - id: 121
+        type: integer
+        name: sensor
+        unit: kWh
+        class: total_increasing
+        precision: 2
+        mapping:
+          - scale: 100
+  - entity: number
+    name: Limit Power Phase A
```
make-all: HA naming convention is to only capitalise the first word of an entity name (in this case, A is an abbreviation so exempt from that rule). Since the translation key changes above will result in the word "phase" being dropped from other names (to avoid just that word being in English) this should maybe also drop the "phase". I would also reverse it as "Power limit A".

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+        range:
+          min: 0
+          max: 150000
+        mapping:
+          - dps_val:
+            invalid: true
+          - scale: 1
+            step: 100
+  - entity: sensor
+    name: Forward Energy A
```
make-all: Instead of explicit name and icon, use
```
translation_key: energy_consumed_x
translation_placeholders:
  x: A
```

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+      - id: 17
+        type: integer
+        name: sensor
+        unit: kWh
+        class: total_increasing
+        precision: 2
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Reverse Energy A
```
make-all: As above but `produced` instead of consumed

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,236 @@
...
+      - id: 117
+        type: integer
+        name: sensor
+        unit: kWh
+        class: total_increasing
+        precision: 2
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Cur State A
```
make-all: Spell out words in full.

make-all (changes requested):

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: Fixed

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: Fixed

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: removed the precision

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: updated

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: updated

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: ok, changed

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: applied

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: applied

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: ok

.gitignore
```diff
@@ -17,3 +17,4 @@ __pycache__/
 .history
 .DS_Store
 Thumbs.db
+/.venv/
```
make-all: This is already covered in the .gitignore file by line 5 in the current version of this file

medisoft: removed

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: done

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,248 @@
...
+    dps:
+      - id: 107
+        type: integer
+        name: value
+        unit: W
+        range:
+          min: 0
+          max: 150000
+        mapping:
+          - dps_val: none
```
make-all: The type is integer, so the dps_val cannot be "none". This mapping can just be removed, only the default with step is needed.

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,248 @@
...
+      x: A
+    class: voltage
+    category: diagnostic
+    dps:
+      - id: 20
+        type: integer
+        name: sensor
+        unit: V
+        class: measurement
+        precision: 1
```
make-all: precision is not doing anything here. The default is the number of decimal places generated by the scale.

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,248 @@
...
+      - id: 115
+        type: integer
+        name: sensor
+        unit: kWh
+        class: total_increasing
+        precision: 2
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Forward Energy B
```
make-all: These should also  use the energy_consumed_x / energy_produced_x translation keys, like the A channel

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,248 @@
...
+      - id: 117
+        type: integer
+        name: sensor
+        unit: kWh
+        class: total_increasing
+        precision: 2
+        mapping:
+          - scale: 100
+  - entity: sensor
+    name: Current State A
```
make-all: HA naming convention is to only capitalise the first word of entity names.

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
```diff
@@ -0,0 +1,248 @@
+---
+name: Dual channel energy meter
+products:
+  - id: zcmq0kvd01bn6s1n
+    manufacturer: WMDL
+    model: Smart Energy Meter WIFI Bidirectional 2 Channel
```
make-all: This seems like it is not really a model name, more a description of the device. So maybe it is better removed, and the model_id promoted to model.

make-all (changes requested):

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: Interesting, is there any field for a description? Or that is not important?

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: done!

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: done!

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: removed

custom_components/tuya_local/devices/wmdl_2c_dual_channel_meter.yaml
medisoft: removed.

PR #4791, opened 2026-04-01
Add Tuya smart soil tester

ameer1234567890:


custom_components/tuya_local/devices/tuya_smart_soil_tester.yaml
```diff
@@ -0,0 +1,80 @@
...
+        unit: C
+        class: measurement
+        range:
+          min: -90
+          max: 600
+        mapping:
+          - scale: 1
+            step: 1
+  - entity: sensor
+    class: humidity
```
make-all: I think this should be moisture, not humidity.

custom_components/tuya_local/devices/tuya_smart_soil_tester.yaml
```diff
@@ -0,0 +1,80 @@
...
+    dps:
+      - id: 5
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+        range:
+          min: -90
+          max: 600
+        mapping:
```
make-all: `scale: 1`, `step: 1` is the default, so this mapping is not needed. However, the range above may suggest a `scale: 10` is required here.

make-all:
It seems that this just adds an illuminance sensor to the existing `ble_smart_plant_moisture.yaml` config, so the change could be made there, by making that dp optional and the marking the entity `hidden: unavailable`.

make-all:
merged with ble_smart_plant_moisture.yaml (for SGS01)

PR #4794, opened 2026-04-01
docs(smart_light_rgb48): document DP56/DP59 protocol behavior

starfire0606:
Adding protocol documentation discovered through reverse engineering. 

After extensive testing with this device in Home Assistant via local Tuya, I've documented the behavior of DP 56 and DP 59 in the file header comments:

-DP 56 accepts animated DIY scene payloads from Smart Life but does not support static per-bulb color — all scenes animate regardless of mode bytes -DP 59 implements the full per-bulb paint protocol used by Smart Life's draw tool, supporting static multicolor via effect=0 in the payload header -The per-bulb payload format is fully documented in the header comments including the init frame requirement and 50ms delay between bulb writes -Confirmed hidden: true on colour mode is correct behavior — selecting it directly causes repeated hs color errors in HA logs

### DIFF
--- a/custom_components/tuya_local/devices/rgb48_lightstring.yaml
+++ b/custom_components/tuya_local/devices/rgb48_lightstring.yaml
@@ -3,22 +3,24 @@
 # Protocol: 3.4
 #
 # Notes:
-# - DP 56 (rgbic_linerlight_scene): base64 scene payload for animated multicolor scenes
-#   saved via Smart Life DIY scene editor. Write-only — device never reports this DP in
-#   status polls. All DP 56 scenes animate; static per-bulb color is NOT possible via
-#   this DP. Use text.set_value to apply saved scenes.
+# - DP 56 (rgbic_linerlight_scene): base64 scene payload for animated
+#   multicolor scenes saved via Smart Life DIY scene editor. Write-only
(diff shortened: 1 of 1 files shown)


starfire0606:
Note for anyone implementing static per-bulb color control:
The DP 59 per-bulb paint protocol requires sending 48 sequential payloads (one per bulb) with ~50ms delay between each. This cannot be done natively in a Home Assistant automation — you'll need AppDaemon, Node-RED, or a Python script to loop through all 48 bulbs.

Example AppDaemon implementation:

Send init frame: base64([0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x64, 0x00, 0x00]) to DP 59
For each bulb index 0–47, send: base64([0x01, 0x02, 0x00, hue_hi, hue_lo, sat, 0x64, 0x00, 0x00, 0x81, 0x00, bulb_index]) to DP 59 with 50ms delay

For animated multicolor (effect=1), the same sequence works but with 0x01 as the effect byte — and in that case a single DP 56 scene payload from the Smart Life DIY scene editor is much simpler.

If you just want static multicolor without home automation integration, you can also set it directly in the Smart Life app using the draw tool — it will persist on the device until changed. Not exactly the point of local control, but it works! 😄

PR #4795, opened 2026-04-01
Re-added missing sensor_select section

belleDESiRE:
Seems that this section was accidentality removed a few PRs ago


make-all:
Not accidentally removed, rather never existed (an attribute "sensor_selection" was in the original PR, but got removed because it was unused and an entity existed for it after the review iterations)

belleDESiRE:
> Not accidentally removed, rather never existed (an attribute "sensor_selection" was in the original PR, but got removed because it was unused and an entity existed for it after the review iterations)

Thing is it is/was being referred to on line 36, so my floor temperature was never being reported

PR #4798, opened 2026-04-01
Add support for Feyree 4.3 Wallbox

kriskomp:
Add support for Feyree 4.3 Wallbox
Closes #4717

PR #4800, opened 2026-04-02
Add support for Feyree 4.3 Wallbox

kriskomp:
Fix translation (https://github.com/make-all/tuya-local/pull/4798)

### DIFF
--- a/custom_components/tuya_local/devices/feyree_43_evcharger.yaml
+++ b/custom_components/tuya_local/devices/feyree_43_evcharger.yaml
@@ -1,4 +1,4 @@
-name: EV Wallbox
+name: EV wallbox
 products:
   - id: zk1jkjhxqkfaaalp
     manufacturer: Feyree
@@ -33,7 +33,7 @@ entities:
 
   # --- BŁĘDY ---
   - entity: binary_sensor
-    name: Fault
+    class: problem
     category: diagnostic
     dps:
       - id: 10
@@ -194,6 +194,7 @@ entities:
   - entity: sensor
     name: Phase mode
     icon: "mdi:transmission-tower"
+    category: diagnostic
     dps:
       - id: 108
         type: string
@@ -229,7 +230,6 @@ entities:
       - id: 111
         type: integer
         name: sensor
-        unit: mth
 
   # --- ENERGIA CAŁKOWITA ---
   - entity: sensor
@@ -244,7 +244,7 @@ entities:
 
   # --- USTAWIONY PRĄD ---
   - entity: number
-    name: Charge current limit
+    name: Charge current
     icon: "mdi:current-ac"
     class: current
     dps:
@@ -252,28 +252,23 @@ entities:
         type: integer
         name: value
         unit: A
-        range:
-          min: 6
-          max: 16
         mapping:
           - constraint: maximum
             conditions:
               - dps_val: Max32A
                 value_redirect: value_32a
-                range:
-                  min: 8
-                  max: 32
               - dps_val: Max40A
                 value_redirect: value_40a
-                range:
-                  min: 8
-                  max: 40
               - dps_val: Max50A
                 value_redirect: value_50a
-                range:
-                  min: 8
-                  max: 50
-
+      - id: 113
+        type: string
+        optional: true
+        name: minimum
+        mapping:
+          - dps_val: Max16A
+            value: 6
+          - value: 8
       - id: 113
         type: string
         optional: true
@@ -287,11 +282,13 @@ entities:
             value: 40
           - dps_val: Max50A
             value: 50
-          - dps_val: Max60A
-            value: 60
-          - dps_val: Max80A
-            value: 80
-
+          # The following do not have dps to redirect to so ignore
+          # - dps_val: Max60A
+          #   value: 60
+          # - dps_val: Max80A
+          #   value: 80
+          # default to 16, as there will be no redirect
+          - value: 16
       - id: 115
         type: integer
         optional: true
@@ -318,7 +315,7 @@ entities:
 
   # --- OPÓŹNIENIA ---
   - entity: number
-    name: Charge delay time
+    name: Charge delay
     icon: "mdi:timer-sand"
     dps:
       - id: 118
@@ -330,7 +327,7 @@ entities:
           max: 15
 
   - entity: number
-    name: Charge duration limit
+    name: Charge duration
     icon: "mdi:timer-outline"
     dps:
       - id: 119


custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
+name: Feyree 4.3 Wallbox
```
make-all: Branding should go under `products`, as `manufacturer` and `model` attributes. Leave this name generic in case other products also match the config.

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
...
+products:
+  - id: zk1jkjhxqkfaaalp
+
+entities:
+
+  # --- STATUS ---
+  - entity: select
+    name: Device State
```
make-all: "State" normally refers to a sensor, not a configurable select. Most of the states below look like something the device would report, not the user would set.

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
...
+            value: "Finished"
+          - dps_val: "wait_charing"
+            value: "Waiting Charging"
+          - dps_val: "error"
+            value: "Error"
+          - dps_val: "stopusing"
+            value: "Stopped"
+
+  # --- BŁĘDY ---
+  - entity: sensor
```
make-all: This should follow the standard `problem` class binary_sensor pattern, with just a true/false value reported for detection of errors, and the full value available as a "fault_code" attribute.

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
...
+    category: diagnostic
+    dps:
+      - id: 10
+        type: integer
+        name: sensor
+
+  # --- ENERGIA SESJI ---
+  - entity: sensor
+    name: Session Energy
+    class: energy
```
make-all: May be better as `energy_storage`  class if the purpose is only to measure the current session, not for long term energy usage tracking.

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
...
+      - id: 25
+        type: integer
+        name: sensor
+        unit: kWh
+        mapping:
+          - scale: 100
+
+  # --- NAPIĘCIA ---
+  - entity: sensor
+    name: Voltage A
```
make-all: Instead of explicit name, use 
```
translation_key: voltage_x
translation_placeholders:
  x: A
```

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
...
+        type: string
+        name: option
+        mapping:
+          - dps_val: "Single_phase"
+            value: "Single phase"
+          - dps_val: "Three_phase"
+            value: "Three phase"
+          - dps_val: "No_phase"
+            value: "No phase"
+          - dps_val: "Phase_err"
```
make-all: This seems like a sensor reporting value, not a selectable option.

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
...
+    dps:
+      - id: 112
+        type: integer
+        name: sensor
+        unit: kWh
+        mapping:
+          - scale: 10
+
+  # --- MAKSYMALNY PRĄD ---
+  - entity: select
```
make-all: Is this really settable by the user? It seems like something that should depend on the unit you have bought, and just be reported by this dp

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
...
+      - id: 118
+        type: integer
+        name: value
+        unit: h
+        range:
+          min: 0
+          max: 15
+
+  - entity: number
+    name: Define Time
```
make-all: name is not clear

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
...
+          - dps_val: "OpenCharging"
+            value: "Start Charging"
+          - dps_val: "CloseCharging"
+            value: "Stop Charging"
+          - dps_val: "WaitOperation"
+            value: "Waiting"
+
+  # --- WERSJE ---
+  - entity: sensor
+    name: Hardware Version
```
make-all: This is static information, and does not need a sensor.
It is probably better attached to the status sensor as an attribute.

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,304 @@
...
+    name: Software Version
+    icon: "mdi:chip"
+    dps:
+      - id: 126
+        type: string
+        name: sensor
+
+  # --- LOG FLAG ---
+  - entity: binary_sensor
+    name: Log Flag
```
make-all: Not sure what the purpose of this is.

kriskomp: Hello, @make-all 
All fixes done on file https://github.com/kriskomp/tuya-local/blob/fix-feyree-translation/custom_components/tuya_local/devices/feyree_4_3.yaml

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,337 @@
...
+            value: charging
+          - dps_val: wait_rfid
+            value: waiting_for_authorization
+          - dps_val: finish
+            value: charged
+          - dps_val: wait_charing
+            value: waiting
+          - dps_val: error
+            value: fault
+    attributes:
```
make-all: This line is not needed, just include the below entries in the dps list, if the names are not matching standard entity functions, they will be automatically added as extra attributes.

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,337 @@
...
+products:
+  - id: zk1jkjhxqkfaaalp
+    manufacturer: Feyree
+    model: "4.3 22kW 3phase wallbox Socket Version"
+
+entities:
+
+  # --- STATUS ---
+  - entity: sensor
+    name: Device state
```
make-all: Use translation_key: status instead of this name

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,337 @@
...
+        type: string
+        optional: true
+      - dps: 126
+        name: software_version
+        type: string
+        optional: true
+
+  # --- BŁĘDY ---
+  - entity: binary_sensor
+    name: Fault
```
make-all: Remove the name and icon, let the HA class translations and default class icons be used here.

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,337 @@
...
+    category: diagnostic
+    dps:
+      - id: 111
+        type: integer
+        name: sensor
+        unit: mth
+
+  # --- ENERGIA CAŁKOWITA ---
+  - entity: sensor
+    name: Total session energy
```
make-all: Probably best to remove this name and let the default HA translation by used instead.

custom_components/tuya_local/devices/feyree_4_3.yaml
```diff
@@ -0,0 +1,337 @@
...
+        name: sensor
+        unit: kWh
+        mapping:
+          - scale: 10
+
+  # --- MAKSYMALNY PRĄD ---
+  - entity: sensor
+    name: Max current setting
+    icon: "mdi:current-ac"
+    class: enum
```
make-all: class enum requires all the possible values to be listed in a mapping.
This should probably be a `maximum` attribute on the "Charge current limit" entity below, with string values mapped to integers.

PR #4803, opened 2026-04-02
Add support for ATORCH S1BWP-T/H Temperature Controller

rezendeneto:
Adding support for [ATORCH S1BWT Temperature Controller](https://aliexpress.com/item/1005008523536156.html)

### DIFF
--- a/custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
+++ b/custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
@@ -1,7 +1,7 @@
 name: Temperature controller
 products:
   - id: 3kw82iy6w4udblee
-    manufacturer: ATORCH
+    manufacturer: ATorch
     model: S1BWP-T/H
 
 entities:
@@ -209,6 +209,11 @@ entities:
         class: total_increasing
         mapping:
           - scale: 1000
+      - id: 127
+        type: integer
+        name: historical
+        mapping:
+          - scale: 100
 
   - entity: number
     name: Electricity price
@@ -224,7 +229,7 @@ entities:
           - scale: 100
 
   - entity: number
-    name: Over-voltage protection
+    name: High voltage limit
     class: voltage
     category: config
     dps:
@@ -239,7 +244,7 @@ entities:
           - scale: 10
 
   - entity: number
-    name: Over-current protection
+    name: Current limit
     class: current
     category: config
     dps:
@@ -254,7 +259,7 @@ entities:
           - scale: 100
 
   - entity: number
-    name: Over-power protection
+    name: Power limit
     class: power
     category: config
     dps:
@@ -313,42 +318,32 @@ entities:
           - scale: 1000
 
   - entity: binary_sensor
-    name: Temperature sensor
+    name: Temperature
     class: problem
     dps:
       - id: 126
         type: boolean
         name: sensor
 
-  - entity: sensor
-    name: Historical energy
-    dps:
-      - id: 127
-        type: integer
-        name: sensor
-        unit: kWh
-        class: total_increasing
-        mapping:
-          - scale: 100
-
   - entity: switch
-    name: Over-limit disconnect
+    name: Over limit breaker
     category: config
     dps:
       - id: 128
         type: boolean
         name: switch
 
-  - entity: switch
-    name: Real-time refresh
+  - entity: button
+    name: Refresh
+    icon: "mdi:refresh"
     category: config
     dps:
       - id: 129
         type: boolean
-        name: switch
+        name: button
 
   - entity: number
-    name: Over-limit disconnect delay
+    name: Over limit breaker delay
     class: duration
     category: config
     mode: box
@@ -364,7 +359,7 @@ entities:
           - scale: 10
 
   - entity: number
-    name: Low-voltage protection
+    name: Low voltage limit
     class: voltage
     category: config
     dps:
@@ -379,7 +374,7 @@ entities:
           - scale: 10
 
   - entity: sensor
-    name: Recovery countdown
+    name: Auto reset countdown
     class: duration
     category: config
     dps:
@@ -422,7 +417,7 @@ entities:
             value: Relay error
 
   - entity: number
-    name: Over-limit recovery time
+    name: Auto reset time
     class: duration
     category: config
     mode: box


custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
+name: Temperature controller
+products:
+  - id: 3kw82iy6w4udblee
+    name: ATORCH S1BWP-T/H
```
make-all: name here has no effect. ANyway this is just a combination of manufacturer and model info that is already available.

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+          - dps_val: false
+            value: idle
+          - dps_val: true
+            constraint: hvac_mode
+            conditions:
+              - dps_val: hot
+                value: heating
+              - dps_val: colding
+                value: cooling
+              - dps_val: socket
```
make-all: The off case should not be needed, as there is code to handle showing the hvac_action as off when the hvac_mode is off.

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+                value_redirect: cooling_temp_low_f
+      - id: 123
+        type: integer
+        name: cooling_temp_high_f
+        hidden: true
+        mapping:
+          - scale: 10
+
+  - entity: switch
+    name: Switch
```
make-all: name not needed.

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+        name: switch
+      - id: 101
+        type: string
+        name: available
+        mapping:
+          - dps_val: socket
+            value: true
+          - value: false
+
+  - entity: sensor
```
make-all, line 169: Duplicate of climate attributes is not needed

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+        type: integer
+        name: sensor
+        unit: °C
+        class: measurement
+        mapping:
+          - scale: 10
+
+  - entity: sensor
+    name: Current
+    class: measurement
```
make-all: You should use class: current here, and no name. "measurement" is not a valid device class for a sensor, but belongs on the dp level as state class (where you already have it).

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+        type: integer
+        name: sensor
+        unit: A
+        class: measurement
+        mapping:
+          - scale: 1000
+
+  - entity: sensor
+    name: Power
+    class: measurement
```
make-all: Should be `class: power` instead of this name and class

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+      - id: 109
+        type: integer
+        name: sensor
+        unit: W
+        class: measurement
+        mapping:
+          - scale: 100
+
+  - entity: sensor
+    name: Voltage
```
make-all: Should be `class: voltage`

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+      - id: 110
+        type: integer
+        name: sensor
+        unit: V
+        class: measurement
+        mapping:
+          - scale: 100
+
+  - entity: sensor
+    name: Energy
```
make-all: Should be `class: energy`

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+        range:
+          min: 0
+          max: 99999
+        mapping:
+          - scale: 100
+
+  - entity: number
+    name: Over-voltage protection
+    category: config
+    dps:
```
make-all, line 235: class: voltage might be useful here

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+        unit: V
+        range:
+          min: 0
+          max: 2750
+        mapping:
+          - scale: 10
+
+  - entity: number
+    name: Over-current protection
+    category: config
```
make-all, line 249: class: current

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+        unit: A
+        range:
+          min: 0
+          max: 2000
+        mapping:
+          - scale: 100
+
+  - entity: number
+    name: Over-power protection
+    category: config
```
make-all, line 264: class: power

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+  - entity: select
+    translation_key: temperature_unit
+    category: config
+    dps:
+      - id: 118
+        type: string
+        name: option
+        mapping:
+          - dps_val: c
+            value: Celsius
```
make-all: values should be all lower case to match translation keys

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+    name: Real-time refresh
+    category: config
+    dps:
+      - id: 129
+        type: boolean
+        name: switch
+
+  - entity: number
+    name: Over-limit disconnect delay
+    category: config
```
make-all, line 348: class: duration

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+        type: integer
+        name: value
+        unit: s
+        range:
+          min: 0
+          max: 20
+
+  - entity: number
+    name: Low-voltage protection
+    category: config
```
make-all, line 364: class: voltage

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+        name: value
+        unit: V
+        range:
+          min: 0
+          max: 2750
+        mapping:
+          - scale: 10
+
+  - entity: sensor
+    name: Recovery countdown
```
make-all: class: duration

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+    dps:
+      - id: 132
+        type: integer
+        name: sensor
+        unit: s
+        class: measurement
```
make-all: I don't recommend class: measurement here, as long term statistics are not really useful for countdown type sensors.

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+
+  - entity: select
+    name: Warning
+    category: diagnostic
+    dps:
+      - id: 133
+        type: string
+        name: option
+        readonly: true
```
make-all: Should be a sensor if it is read-only. Probably a binary_sensor with class: problem (no name) is appropriate for this, with the string as "description", and mapped to "off" = false, default = true for the "sensor" dp.

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,435 @@
...
+          - dps_val: lvp
+            value: Low voltage
+          - dps_val: err_temp
+            value: Temperature error
+          - dps_val: err_relay
+            value: Relay error
+
+  - entity: number
+    name: Over-limit recovery time
+    category: config
```
make-all, line 422: class: duration

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
rezendeneto: this is to turn off the thermostat (i.e. put it in pass-through/no-control mode) - in this mode, it's possible to manually turn the switch on/off.

as a user (I've been testing the device), I'm good with the "off" option, but a possibility here would be to name it "manual mode" or something along these lines through translations - wdyt?

make-all: hvac_action is reporting the current state from the device. It is not possible to control the device's mode through this.

What you are saying above applies to the hvac_mode. "off" should remain there.

rezendeneto: oh, gotcha! sorry about the confusion here 😅 just fixed it

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
```diff
@@ -0,0 +1,434 @@
...
+        name: switch
+
+  - entity: number
+    name: Over-limit disconnect delay
+    class: duration
+    category: config
+    mode: box
+    dps:
+      - id: 130
+        type: float
```
make-all: Theoretically float could be valid and is supported by the integration, but I have never seen a Tuya device that uses floating point directly, normally they will output/input integers with a scale factor.

Because float is not in use by any known Tuya device, the CI flags it as an error, but if your device is really using floating point here, it can be let through. But please double check the value coming from the device, and/or developer portal documentation, because I think it should be integer here, and maybe a scale mapping is needed (and perhaps adjustment to the range, which should be the device side range when it is scaled)

make-all (dismissed):

custom_components/tuya_local/devices/atorch_s1bwpth_thermostat.yaml
rezendeneto: thanks for letting me know!

> normally they will output/input integers with a scale factor

makes sense! I just realized that this doesn't work properly with `float`. after changing it to integer w/ a scale factor I was finally able to set values such as `0.3s` for example.

PR #4808, opened 2026-04-03
Support for Orbegozo DHW2080 dehumidifier

axbl:
I have added and tested since 5 months Support for Orbegozo DHW2080 dehumidifier.

### DIFF
--- a/custom_components/tuya_local/devices/orbegozo_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/orbegozo_dehumidifier.yaml
@@ -1,4 +1,4 @@
-name: Deshumidifier
+name: Dehumidifier
 products:
   - id: 55e916oobezauqfb
     manufacturer: Orbegozo


custom_components/tuya_local/devices/orbegozo_dehumidifier.yaml
```diff
@@ -35,7 +35,7 @@ entities:
           - dps_val: high
             value: High
   - entity: select
-    name: Mode
+    translation_key: Mode
```
make-all: lower case for translation keys

axbl: thank you, modified my code

axbl: enabled issues and added topics to repository for hacs validation

make-all:
This is very close to alecoair_d12_dehumidifier (100% match on dps level, but humidity range and modes need some adjustment), so I will merge with that.

axbl:
ID 5 doesnt work for Orbegozo.

make-all:
To clarify you mean after upgrading to 2026.4.2?

axbl:
Yes. Fan Speed and Mode are not found.

make-all:
alecoair_d12_dehumidifier has these properly integrated into the humidifier and fan entities.
Probably you need to click on the entity name in the UI to open the full controls if you are not using a specific card that shows a full UI for the entity, or the Tile card with all features enabled.

PR #4809, opened 2026-04-03
Add support for TyloHelo SaunaLogic2 (Finnleo sauna controller)

bitswype:
## Summary

Adds a device definition for the **TyloHelo SaunaLogic2**, a Tuya-based sauna controller (product ID `acl5qrawgmjajabn`, protocol 3.3). This controller is sold under the **Finnleo**, **Helo**, and **Sauna360** brands.

## Entities

| Entity | DP | Type | Description |
|--------|-----|------|-------------|
| climate | 1, 2, 3 | boolean, int, int | Heater on/off, target temp (25-194°F), current temp |
| select | 101 | int (0-8) | Light color: Off, White, Red, Green, Blue, Yellow, Aqua, Purple, Rainbow |
| switch | 103 | boolean | Bluetooth audio on/off |
| sensor | 10 | int | Timer countdown (minutes, read-only) |
| sensor | 4 | string | Sauna mode (diagnostic, e.g. "ONLY_TRAD") |
| sensor | 9 | int | Unknown purpose (diagnostic) |
| sensor | 105 | int | Sauna state: 1=idle, 2=running (diagnostic) |

## Testing

All data points verified via local protocol (v3.3) testing with physical confirmation at the sauna:

- Heater on/off: relay click confirmed, warmth felt
- Temperature set: display updated (device has firmware-enforced minimum ~119°F)
- All 9 light modes: visually confirmed
- Bluetooth on/off: confirmed on sauna display
- Timer write: tested and confirmed **read-only** (device silently ignores writes)
- Two-way: physical panel changes reflected in HA in real-time

## Documentation

Full reverse engineering journey and device specification:
https://github.com/bitswype/finnleo_ha

This device was discovered to be a Tuya OEM (appId 15909) through APK decompilation of the SaunaLogic Android app. Local key obtained by re-pairing to Smart Life. Hidden DPs (101, 103, 105, 4, 9) discovered via local probing — they are not registered in the Tuya cloud product definition.

### DIFF
--- a/custom_components/tuya_local/devices/tylohelo_sl2_sauna.yaml
+++ b/custom_components/tuya_local/devices/tylohelo_sl2_sauna.yaml
@@ -55,6 +55,7 @@ entities:
             value: "white"
             default: true
             hidden: true
+      - id: 101
         name: named_color
         type: integer
         mapping:


make-all:
Accepting this contribution would complicate the licensing of this package. I suggest you either relicense it under the same license as the main project, or I will not be able to accept it.

custom_components/tuya_local/devices/tylohelo_sl2_sauna.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - entity: sensor
+    name: Sauna mode
+    category: diagnostic
+    dps:
+      - id: 4
+        name: sensor
+        type: string
+        optional: true
+  - entity: sensor
+    name: DP 9
```
make-all: There is no point exposing unknown dps as sensors. Just make them attributes of another entity (suggest `unknown_9` as the attribute name).

Query Things Data Model API call from platform.tuya.com should list these if you have access.

custom_components/tuya_local/devices/tylohelo_sl2_sauna.yaml
```diff
@@ -0,0 +1,95 @@
...
+    category: diagnostic
+    dps:
+      - id: 9
+        name: sensor
+        type: integer
+        optional: true
+  - entity: sensor
+    name: Sauna state
```
make-all: Should be moved to `hvac_action` attribute of climate entity (values mapped to `idle` and `heating`)

bitswype:
Thank you for the feedback! I've pushed an update addressing all three points:

1. **License**: Changed from Apache-2.0 to MIT to match the project license.
2. **DP 9**: Moved from a standalone diagnostic sensor to `unknown_9` attribute on the climate entity. Unfortunately this DP is not registered in the Tuya cloud product definition (it's a hidden DP discovered via local probing), so the Query Things Data Model API doesn't list it. We observed it change from 1→0 when a delay start was set from the physical panel, but couldn't fully characterize its behavior.
3. **DP 105 (sauna state)**: Moved from standalone sensor to `hvac_action` on the climate entity, mapped as 1=idle, 2=heating.

PR #4812, opened 2026-04-04
Remove redundant 'available' entity definitions

SimKaiLong:
Ensure fan and light entities remain available even when the main device is powered off, allowing sub-entity commands to act as an implicit power-on trigger, which is a behavior of the ceiling fan.

Updated Logic & Behavior
Persistent Availability: Sub-entities no longer move to an unavailable state when the master power is off.

State Decoupling: This prevents the device from blindly restoring its "last state" (where both fan and light might have been active) when only one component is required.

Use Case:
This is specifically designed for non-disruptive automations, such as a morning transition from A/C to ceiling fans. By interacting directly with the fan speed entity, the system can initiate a "clean start" for the fan without accidentally triggering the light, even if the light was on during the previous session.


make-all:
If I understand correctly, the device's behaviour is not to use dp 1 as a main power switch, but as a 2-in-1 switch, and automatically turns the device on if either the fan or the light is turned on. Maybe the naming of that switch also needs to change then, to reflect it's purpose (I can't really think of a good name, 2-in-1 is out because it starts with a number, maybe "Dual control"?

make-all:
On second thoughts, changing the name will reassign the entity id, so it would be less disruptive to leave it as it is, even if it is a bit misleading. The entity already has a config category, so it is deprioritized compared to fan and light entities that can work standalone.

SimKaiLong:
Yes that is correct. Thanks for the merge!

PR #4814, opened 2026-04-04
Added support for Hunter Pacific 6 speed WiFi fan controller (WIFIM099 6SP_WIFI_REMOTER)

thetallbloke:
From Home Assistant Logs:
15:24:56.201 - Adding 6SP_WIFI_REMOTER device with product id 8tai0lslpwgaexzv
15:24:56.979 - Device matches hunterpacific_6speed_fanlight with quality of 101%

Not sure why 101%, but tested and working with my 5 units on both protocol v3.3 and v3.5


make-all:
101% indicates that the product id matches and there are no conflicts with the dps.

It is 101 to put it above configs that match the dps 100%, but do not have a product id match.

PR #4815, opened 2026-04-04
add (config): add costway FD10584DE-WH config

tkaefer:
Following the advice from https://github.com/make-all/tuya-local/pull/4006#issuecomment-3565353273 (plus some time to spare), I've created a dedicated config for the Costway FD10584DE-WH portable air conditioner with heating.

As with https://github.com/make-all/tuya-local/pull/4006 this one is based on the ballu aura AC - but with the necessary tweaks.

I hope this one is more suitable.


tkaefer:
Sorry for the wrong config.

I now let pytest run for `pytest tests/test_device_config.py` and `pytest tests/` locally - no issues.

PR #4716, opened 2026-03-19
feat (devices): add Aquastrong heat pump

zackify:
## Description

Adds support for the Aquastrong heat pump pool/spa heater. Used claude to debug, fully tested in HA and with cli as well. Thankfully there was a similar device partially working so it was easy to debug.

## Device Details

- **Manufacturer:** Aquastrong
- **Product ID:** `g9m7honkxjweukvt`
- **Category:** Heat Pump
- **Amazon:** https://www.amazon.com/dp/B0DRV95LP5

## Datapoint Mapping

| DP ID | Type | Mode | Entity | Description |
|-------|------|------|--------|-------------|
| 1 | boolean | rw | hvac_mode / binary_sensor | Power on/off |
| 2 | string | rw | work_mode / select | Mode: make_hot, make_cold, auto |
| 4 | integer | rw | temperature | Temperature setpoint (5–105°F) |
| 5 | string | rw | preset_mode | Preset: ECO, Normal, Boost |
| 13 | string | rw | select | Heat curve (OFF, H1–H8, L1–L8) |
| 14 | integer | ro | sensor | EEV opening (steps) |
| 16 | integer | ro | sensor | Current draw (÷10 = amps) |
| 20 | integer | ro | sensor | Compressor frequency (Hz) |
| 21 | integer | ro | sensor / current_temperature | Inlet temperature (°F) |
| 22 | integer | ro | sensor | Outlet temperature (°F) |
| 23 | integer | ro | sensor | Coil temperature (°F) |
| 24 | integer | ro | sensor | Discharge temperature (°F) |
| 26 | integer | ro | sensor | Ambient temperature (°F) |
| 27 | boolean | ro | hvac_action | Running/idle state |
| 31 | boolean | ro | binary_sensor | Backflow problem |
| 33 | boolean | ro | binary_sensor | Defrost active |
| 35 | integer | ro | sensor | High pressure (psi) |
| 37 | integer | ro | sensor | Indoor coil temperature (°F) |
| 38 | integer | ro | sensor | Water heater current temp (°F) |
| 40 | integer | ro | sensor | Fan motor frequency (Hz) |
| 41 | integer | ro | sensor | Suction temperature (°F) |

## Implementation Notes

- This device uses **non-standard mode strings**: `make_hot`, `make_cold`, `auto` — different from the similar SolarEast/Adlår Castra Aurora II which uses `heating`, `cold`, `floor_heating`, etc.
[...]

PR #4820, opened 2026-04-04
Add AQUASTRONG HEX pool heat pump profile

ps851112:
This PR adds a new Tuya Local device profile for AQUASTRONG HEX-series pool heat pumps.

Tested on:

AQUASTRONG HEX100
Local protocol: 3.5
Stable local DPS polling and control verified
Compatibility target:

HEX012
HEX016
HEX018
HEX022
HEX035
HEX055
HEX075
HEX100
What is included:

Climate entity (with pool/hot-tub style icon)
Heat pump mode select
Defrost binary sensor
Diagnostic sensors:
Compressor frequency
Fan motor frequency
Inlet/outlet/coiler/discharge/ambient/incoiler/suction temperatures
EEV opening
Electrical telemetry mappings (current/voltage/power/energy)
Notes:

Temperature mapping is in Fahrenheit for this profile variant.
Setpoint range is 50-104 F.
On the tested unit, electrical DPS keys are present but reported zero values during local polling, even under load.

### DIFF
--- a/custom_components/tuya_local/devices/aquastrong_hex_heatpump.yaml
+++ b/custom_components/tuya_local/devices/aquastrong_hex_heatpump.yaml
@@ -20,11 +20,7 @@ entities:
             conditions:
               - dps_val: make_cold
                 value: cool
-              - dps_val: cold
-                value: cool
               - dps_val: make_hot
-                value: heat
-              - dps_val: heating
                 value: heat
               - dps_val: auto
                 value: auto
@@ -65,10 +61,6 @@ entities:
           - dps_val: make_cold
             value: cool
           - dps_val: make_hot
-            value: heat
-          - dps_val: cold
-            value: cool
-          - dps_val: heating
             value: heat
           - dps_val: auto
             value: auto
--- a/custom_components/tuya_local/devices/aquastrong_hex_heatpump.yaml
+++ b/custom_components/tuya_local/devices/aquastrong_hex_heatpump.yaml
@@ -39,6 +39,17 @@ entities:
         name: current_temperature
         type: integer
         unit: F
+      - id: 20
+        name: hvac_action
+        type: integer
+        mapping:
+          - dps_val: 0
+            value: idle
+          - value: heating
+            constraint: work_mode
+            conditions:
+              - dps_val: make_cold
+                value: cooling
       - id: 5
         name: preset_mode
         type: string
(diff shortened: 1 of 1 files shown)


make-all:
Duplicate of #4716

custom_components/tuya_local/devices/aquastrong_hex_heatpump.yaml
```diff
@@ -0,0 +1,230 @@
+name: Pool heat pump
+products:
+  - id: nxuflhfprft3btpt
+    manufacturer: AQUASTRONG
+    model: HEX series
+    model_id: HEX012/HEX016/HEX018/HEX022/HEX035/HEX055/HEX075/HEX100
+# Mapping validated on AQUASTRONG HEX100 over local protocol 3.5.
+entities:
+  - entity: climate
+    name: Pool heat pump
```
make-all: Use `translation_key: pool_heatpump` instead of this name and icon

custom_components/tuya_local/devices/aquastrong_hex_heatpump.yaml
```diff
@@ -0,0 +1,230 @@
...
+        mapping:
+          - dps_val: ECO
+            value: eco
+          - dps_val: Normal
+            value: normal
+          - dps_val: Boost
+            value: boost
+
+  - entity: select
+    translation_key: heat_pump_mode
```
make-all: This is strange, the device name is "Pool heatpump", but here it seems to be hot water and floor heating/cooling. combo device (which should have 2 main entities, one for the hot water (water_heater), and one for the heat/cool functions (climate)

make-all (changes requested):

custom_components/tuya_local/devices/aquastrong_hex_heatpump.yaml
ps851112: fixed. heat, cool, auto added. removed others.

ps851112:
Updated this PR with the requested changes and validated behavior on hardware.

What changed

Updated climate metadata to use translation_key: pool_heatpump (removed custom name/icon).
Reduced mode scope to heat, cool, and auto only.
Removed floor heating, hot water, and combo mode mappings from the mode select.
Reworked climate hvac_mode mapping to use DP1 power with DP2 work_mode constraint, so mode state reflects actual device mode.
Added hidden work_mode DP in climate for correct mode translation logic.

Verified mode switching now behaves correctly:
off -> heat
heat -> cool
cool -> auto

Verified all Diagnostic sensors.

custom_components/tuya_local/devices/aquastrong_hex_heatpump.yaml
ps851112: done

custom_components/tuya_local/devices/aquastrong_hex_heatpump.yaml
```diff
@@ -0,0 +1,260 @@
...
+        type: string
+        mapping:
+          - dps_val: ECO
+            value: eco
+          - dps_val: Normal
+            value: normal
+          - dps_val: Boost
+            value: boost
+
+  - entity: select
```
make-all: There should be no need to duplicate this here. It is identical to the hvac_mode settings of the climate entity

zackify:
For the ambient temperature, are you getting updated values? I had to ping these devices every minute with a set point command to get updated data on amps and temp sensors.

PR #4821, opened 2026-04-05
Add Holman WX4 4-zone irrigation controller

Francois3d:
## Device
- **Manufacturer:** Holman
- **Model:** WX4 4-Zone Wi-Fi Irrigation Controller
- **Tuya Product ID:** `5jotr6jtclzapmik`
- **Reference:** https://www.holmanindustries.com.au/products/wx4-wi-fi-4-zone-irrigation-controller/

## What's implemented
- Zone valve control (on/off) for all 4 zones — DPs 108, 155, 181, 182
- Manual run timer (0–60 min) for Zones 1 & 2 — DPs 107, 154 (Zones 3 & 4 DPs not yet confirmed)
- Last water flow sensors for all 4 zones — DPs 103, 152, 183, 184 (raw value ÷ 10 = litres)
- Last runtime sensors for Zones 1 & 2 — DPs 193, 194 (Zones 3 & 4 DPs not yet confirmed)
- Battery percentage — DP 7
- Alarm status — DP 120
- Physical soil/rain sensor presence diagnostics — DPs 115, 117, 125, 161, 162
- Units select (DP 119) and Time Format select (DP 114) — borrowed from WX1 pattern, marked `optional: true`, not personally confirmed on WX4
- Raw data DPs 172, 173, 175, 176 exposed as text/base64 sensors for advanced HA template use (live countdowns, rain delay selector, timer sync)

## Known limitations / not implemented
- Manual timer and runtime DPs for Zones 3 & 4 not yet identified — contributions welcome
- DP 176 scheduling blob not decoded — device scheduling is complex; recommend using HA automations instead
- Tested in metric mode only (Australia)

## Testing
Tested on HA 2026.4.x with tuya_local 2026.3.0. All 4 zones open/close correctly. Water flow, runtime, and battery sensors confirmed working.

### DIFF
--- a/custom_components/tuya_local/devices/holman_wx4_irrigation_controller.yaml
+++ b/custom_components/tuya_local/devices/holman_wx4_irrigation_controller.yaml
@@ -71,11 +71,12 @@ entities:
     name: Last water flow 1
     class: water
     category: diagnostic
-    icon: "mdi:water"
+    icon: mdi:water
     dps:
       - id: 103
         name: sensor
         type: integer
+        readonly: true
         unit: L
         range:
           min: 0
@@ -87,11 +88,12 @@ entities:
     name: Last water flow 2
     class: water
     category: diagnostic
-    icon: "mdi:water"
+    icon: mdi:water
     dps:
       - id: 152
         name: sensor
         type: integer
+        readonly: true
         unit: L
         range:
           min: 0
@@ -103,11 +105,12 @@ entities:
     name: Last water flow 3
     class: water
     category: diagnostic
-    icon: "mdi:water"
+    icon: mdi:water
     dps:
       - id: 183
         name: sensor
         type: integer
+        readonly: true
         unit: L
         range:
           min: 0
@@ -119,11 +122,12 @@ entities:
     name: Last water flow 4
     class: water
     category: diagnostic
-    icon: "mdi:water"
+    icon: mdi:water
     dps:
       - id: 184
         name: sensor
         type: integer
+        readonly: true
         unit: L
         range:
           min: 0
@@ -137,11 +141,12 @@ entities:
     name: Last runtime 1
     class: duration
     category: diagnostic
-    icon: "mdi:timer"
+    icon: mdi:timer
     dps:
       - id: 193
         type: integer
         name: sensor
+        readonly: true
         optional: true
         unit: s
 
@@ -149,11 +154,12 @@ entities:
     name: Last runtime 2
     class: duration
     category: diagnostic
-    icon: "mdi:timer"
+    icon: mdi:timer
     dps:
       - id: 194
         type: integer
         name: sensor
+        readonly: true
         optional: true
         unit: s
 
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/holman_wx4_irrigation_controller.yaml
```diff
@@ -0,0 +1,308 @@
...
+    category: diagnostic
+    icon: "mdi:alert-circle"
+    dps:
+      - id: 120
+        name: sensor
+        type: integer
+
+  # Physical add-on sensor presence (hidden when not fitted)
+  - entity: binary_sensor
+    name: T&H soil sensor present 1
```
make-all: I'm not sure what the point of these "present" sensors is, as the corresponding measurement sensors are not present in the config. Normally they should be attached as `available` attributes to the corresponding sensor.

custom_components/tuya_local/devices/holman_wx4_irrigation_controller.yaml
```diff
@@ -0,0 +1,308 @@
...
+  # DP 172: zone timer sync blob - decoded in HA to keep manual
+  #         timer sliders in sync when changed via the Tuya app.
+  # DP 173: live countdown blob - decoded in HA templates to show
+  #         remaining watering time per zone.
+  # DP 175: rain delay blob - encoded/decoded in HA to expose an
+  #         Off / 24h / 48h / 72h rain delay selector.
+  # DP 176: master scheduling blob - not decoded; device scheduling
+  #         is replaced by HA automations for greater flexibility.
+  - entity: text
+    name: Zone Timers Raw Data
```
make-all: The entity names do not need things like "raw data" and "encoded" appended. The name should just describe what is measured, not how it is represented.

They should also follow HA naming convention, which is to only capitalize the first word of an entity or device name (abbreviations excepted)

custom_components/tuya_local/devices/holman_wx4_irrigation_controller.yaml
```diff
@@ -0,0 +1,308 @@
...
+    category: config
+    hidden: true
+    dps:
+      - id: 175
+        type: string
+        name: value
+        optional: true
+
+  - entity: sensor
+    name: Timestamp or schedule encoded
```
make-all: I don't think "Timestamp or schedule" is a good name, pick one of them.

Francois3d:
Thanks for the review feedback. I've pushed an update addressing all three points:

1. **Removed the standalone presence/power binary sensors** (DPs 115, 117, 125, 161, 162). These should be added as `available` attributes on corresponding soil temperature, humidity, and rain measurement entities — I've added a comment in the file noting this for when those DPs are confirmed.

2. **Renamed the raw data entities** to describe what is measured rather than how it's represented:
   - `Zone Timers Raw Data` → `Zone timers`
   - `Timer setting 1 encoded` → `Zone countdown`
   - `Rain Delay Raw Data` → `Rain delay`
   - `Timestamp or schedule encoded` → `Schedule`

3. **Fixed capitalisation** throughout to follow HA naming convention (first word only).

custom_components/tuya_local/devices/holman_wx4_irrigation_controller.yaml
```diff
@@ -71,11 +71,12 @@ entities:
     name: Last water flow 1
     class: water
     category: diagnostic
-    icon: "mdi:water"
+    icon: mdi:water
     dps:
       - id: 103
         name: sensor
         type: integer
+        readonly: true
```
make-all, line 79: Why are you adding this back?

PR #4823, opened 2026-04-05
Add Mowe MW368F Atlas 36 smart ceiling fan

eijnehc:
## New device: Mowe Atlas 36" Smart Ceiling Fan (MW368F)

**Product page:** https://mowesmarthome.com/product/smart-ceiling-fan/
**Tuya product id:** `12giv5ur6oye6kbo`
**Manufacturer:** Mowe
**Model:** Atlas 36 Smart Ceiling Fan
**Model ID:** MW368F

### Entities

| Entity | Description |
|--------|-------------|
| `fan` | Main fan — on/off, preset modes (nature/sleep/normal), speed (1–6), direction |
| `light` | Integrated ceiling light — on/off, brightness (0–100, step 2) |
| `select` (timer) | Countdown timer — cancel, 1h, 2h, 4h, 8h (supports two DP value formats: std and hour) |

### DPs

| DP | Type | Name | Notes |
|----|------|------|-------|
| 1 | boolean | switch | Fan on/off |
| 2 | string | preset_mode | nature / sleep / normal |
| 3 | integer | speed | 1–6, optional |
| 8 | string | direction | optional |
| 15 | boolean | switch | Light on/off |
| 16 | integer | brightness | 0–100, step 2 |
| 22 | string | timer | Two value sets: std (cancel/1h/2h/4h/8h) and hour (off/1hour/2hour/4hour/8hour) |

### Tested on
- Home Assistant OS 17.1 / Core 2026.4.1
- tuya-local (HACS custom integration)

### DIFF
--- a/custom_components/tuya_local/devices/mowe_mw368f_ceiling_fan.yaml
+++ b/custom_components/tuya_local/devices/mowe_mw368f_ceiling_fan.yaml
@@ -1,0 +1,105 @@
+# https://mowesmarthome.com/product/smart-ceiling-fan/
+name: Ceiling fan
+products:
+  - id: 12giv5ur6oye6kbo
+    manufacturer: Mowe
+    model: Atlas 36 Smart Ceiling Fan
+    model_id: MW368F
+entities:
+  - entity: fan
+    translation_only_key: fan_with_presets
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 2
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: nature
+            value: nature
+          - dps_val: sleep
+            value: sleep
+          - dps_val: normal
+            value: normal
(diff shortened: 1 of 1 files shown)


make-all:
This is not a new config, it is a copy of calibo_smart_ceiling_fan with only the product details changed,.

eijnehc:
> This is not a new config, it is a copy of calibo_smart_ceiling_fan with only the product details changed,.

So they are 2 different products, but they work exactly the same, I just have to connect it to calibo_smart_ceiling_fan? I thought adding a new device would be appropriate as Mowe is quite popular in Singapore and I hope ppl don't get confused when integrating. Any advice on this?

make-all:
The advice is that there are already 1500 config files and we don't need to make that 10x worse by duplicating them for every branding variation.

PR #4824, opened 2026-04-05
Added 8ch relay board

Bt338:
## New device: 8ch relay board

**Tuya product id:** aaa2vqyazpodpqhb
**Manufacturer:** 8路开关
**Model:** 8112

### Entities

| Entity | DP | Type | Description |
|--------|-----|------|--------------|
| `switch 1-8` | 1-8 | boolean | relay output — on/off |
| `timer 1-8` | 9-16 | integer | operation timer (in seconds) |
| `relays initial state` | 38 | string | relays state after power-up: Off-all off, On-all on, Memory - last state |
| `light` | 40 | string | mode of state leds: relay-state of the relay, pos-?, none-always off |

### Tested
- Home Assistant 2025.10.1 - 2026.4.1 versions
- tuya-local from HACS

### DIFF
--- a/custom_components/tuya_local/remote.py
+++ b/custom_components/tuya_local/remote.py
@@ -399,7 +399,7 @@ class TuyaLocalRemote(TuyaLocalEntity, RemoteEntity):
             persistent_notification.async_dismiss(
                 self._device._hass, notification_id="learn_command"
             )
-            _LOGGER("%s ending learning mode", self._config.config_id)
+            _LOGGER.debug("%s ending learning mode", self._config.config_id)
             if self._control_dp:
                 await self._control_dp.async_set_value(
                     self._device,


Bt338 (commented):
yamltest expects string as model name

make-all:
> yamltest expects string as model name

You just need to enclose it in quotes to make it a string, no need to prepend an underscore.

custom_components/tuya_local/devices/eight_channel_relay.yaml
```diff
@@ -0,0 +1,207 @@
...
+        name: option
+        mapping:
+          - dps_val: "relay"
+            value: "relay"
+          - dps_val: "pos"
+            value: "pos"
+          - dps_val: "none"
+            value: "none"
+  - entity: time
+    category: config
```
make-all: Maybe you meant to include these as attributes on the timer_1 above?

But actually they would be better as text entities, but with `hidden: true`, since they aren't useful to most users who don't know the data format. Probably the encoding is base64 (it is on most other smart switches that use dps 42 and 43 for these), so changing the type to that will add some additional input checking.

Bt338: It was working fine in HA, but I accept your suggestions

make-all:
Merged with seven_channnel_switch into `multi_channel_switch`

PR #4825, opened 2026-04-05
Add Dekala Lumos sunrise alarm clock (e1xp7w)

martinsotir:
Adds a device config for the Dekala [Lumos™ sunrise alarm clock](https://ebay.us/m/ICOsLj) (model e1xp7w).

Entities covered: color light, ambient light, night light, dual alarms, FM radio, display brightness/seconds dot, wireless charging sensor, light status sensor, clock time, and various switches/buttons.

Note: [other models from Dekala](https://www.amazon.com/stores/Dekala/page/95E2F278-6750-4182-8B26-CB72A5CADA89) might be compatible (but I do test them).

Known limitations:
 - Alarm raw parameters may need one-time initialization (via the official app, or via `Alarm 1 reset` / `Alarm 2 reset`). I tried using force:true but I could not poll the alarm status.
 - light tones can differ slightly from the Tuya app?
 - Some entities not exposed yet:  FM station scan, decoded weather, alarm runtime status, LCD icon visibility

### DIFF
--- a/custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
+++ b/custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
@@ -839,7 +839,7 @@ entities:
         type: string
         mapping:
           - dps_val: "0"
-            value: off
+            value: "off"
           - dps_val: "1"
             value: dim
           - dps_val: "2"
--- a/custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
+++ b/custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
@@ -130,7 +130,7 @@ entities:
         unit: "%"
 
   - entity: switch
-    translation_key: sleep
+    name: Sleep aid
     dps:
       - id: 106
         name: switch
--- a/custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
+++ b/custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
@@ -254,25 +254,6 @@ entities:
         optional: true
         mask: "00FF000000000000000000000000"
 
-  - entity: select
-    name: Alarm 1 repeat
-    category: config
-    dps:
-      - id: 109
-        name: option
-        type: base64
-        optional: true
-        mask: "0000FF0000000000000000000000"
-        mapping:
-          - dps_val: 0
-            value: once
-          - dps_val: 124
-            value: weekdays
-          - dps_val: 3
-            value: weekends
-          - dps_val: 127
-            value: every_day
-
   # Repeat-day bits map Monday..Sunday across the 7 low bits of
   # Data[2].
   - entity: switch
@@ -463,25 +444,6 @@ entities:
         type: base64
         optional: true
         mask: "00FF000000000000000000000000"
-
-  - entity: select
-    name: Alarm 2 repeat
-    category: config
-    dps:
-      - id: 111
-        name: option
-        type: base64
-        optional: true
-        mask: "0000FF0000000000000000000000"
-        mapping:
-          - dps_val: 0
-            value: once
-          - dps_val: 124
-            value: weekdays
-          - dps_val: 3
-            value: weekends
-          - dps_val: 127
-            value: every_day
 
   - entity: switch
     name: Alarm 2 Sunday
--- a/custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
+++ b/custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
@@ -1,4 +1,5 @@
-# `tuya-local` configuration for the Dekala Lumos sunrise alarm clock (`e1xp7w`).
+# `tuya-local` configuration for the Dekala Lumos sunrise alarm
+# clock (`e1xp7w`).
 #
 # Working features:
 # - color, ambient and night lights with brightness/color control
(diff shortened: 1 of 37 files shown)


custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
```diff
@@ -0,0 +1,1090 @@
...
+          - dps_val: 1
+            value: "Mute + light"
+          - dps_val: 2
+            value: "Light only"
+
+  - entity: button
+    name: Alarm snooze
+    translation_only_key: alarm_snooze
+    dps:
+      - id: "7"
```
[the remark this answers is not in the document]
martinsotir: I added translation for en & fr . Tested, no warning, on Tuya local version 2026.3.3.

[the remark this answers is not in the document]
martinsotir: Hi @make-all , I’m not sure how to address the “entity can use translation key x” issue.

Should I update all 10 users’ devices to use the translation I added, or did I misuse the translation feature and should remove the translation keys I introduced?

image

make-all:
I would suggest removing the translation keys from this change, and introduce them as a separate PR that updates all uses across device files after the merge.

custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
```diff
@@ -0,0 +1,1022 @@
...
+  - entity: select
+    name: Display brightness
+    category: config
+    dps:
+      - id: 116
+        name: option
+        type: string
+        mapping:
+          - dps_val: "0"
+            value: off
```
make-all, line 842: This needs quotes

custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
```diff
@@ -0,0 +1,1022 @@
...
+        type: integer
+        optional: true
+        range:
+          min: 10
+          max: 1000
+        mapping:
+          - step: 10
+
+  - entity: light
+    name: Night light
```
make-all: Use `translation_key: nightlight`

custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
```diff
@@ -0,0 +1,1022 @@
...
+        optional: true
+        mask: "FF00000000000000000000000000"
+      - id: 109
+        name: minute
+        type: base64
+        optional: true
+        mask: "00FF000000000000000000000000"
+
+  - entity: select
+    name: Alarm 1 repeat
```
make-all: This seems to overlap with the individual day switches below. I think it is better to just use the switches, as covering everything with a select is 7! complexity, and not covering everything means there will be errors in the log when non-covered options are enabled.

custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
```diff
@@ -0,0 +1,1022 @@
...
+    dps:
+      - id: 109
+        name: switch
+        type: base64
+        optional: true
+        mask: "0000400000000000000000000000"
+
+  - entity: select
+    name: Alarm 1 wake sound
+    category: config
```
make-all: It would be tidier to combine this with volume and duration into a siren entity.

custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
```diff
@@ -0,0 +1,1022 @@
...
+          min: 76
+          max: 108
+        unit: MHz
+      - id: 130
+        name: available
+        type: boolean
+        optional: true
+
+  - entity: number
+    name: Radio decimal
```
make-all: This is ugly, I might need to think about handling this within a single number.

custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
martinsotir: Hi make-all. There is indeed an overlap. However, this was intended to mimic the actual official configuration app, which allows both "presets" for weekends, weekdays, and all days, as well as fine-grained day configurations. During my tests, this worked well: both entities updated in real time in both directions (although the Alarm repeat will indeed switch to "unknown" for custom states). From a UX standpoint, I would like to keep this entity, but I understand if it has to be removed.

Ideally, I would have a way to authorize unknown states with a label such as "custom," for instance. Maybe this could be implemented with a third "virtual" entity not directly mapped to a DPS, used as a condition to make day switches available... But this may be overkill (and beyond my current ability when configuring local-tuya devices!).

make-all: It is not the aim of this integration to replicate the app UI. Additional entities such as this can be implemented at a higher level on top of the minimum entities this integration exposes.

custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
make-all: I have implemented a change to number entity, so this can now be merged with the above by specifying `decimal` as the dp name, and a scale of 10.

martinsotir: I agree with you. I attempted to use the current mapping feature, even mapping this to a text entity, but I was not able to find a more functional solution than this that would allow both reading and user input.

martinsotir: I implemented this. However I have not tested it in homeassistant.

custom_components/tuya_local/devices/dekala_lumos_alarm_clock.yaml
martinsotir: I implemented your suggestion. Please tell me if this was what you had in mind.

PR #4828, opened 2026-04-06
Sanjun Steamroom

aicbaicb:
Add Sanjun Steamer

### DIFF
--- a/custom_components/tuya_local/devices/sanjun_steamer.yaml
+++ b/custom_components/tuya_local/devices/sanjun_steamer.yaml
@@ -61,9 +61,9 @@ entities:
         mapping:
           - dps_val: "00"
             value: Rainbow
-          - dps_val: "01"
-            value: red
-            default: true
+          - dps_val: "07"
+            value: "off"
+          - value: "off"
             hidden: true
       - id: 107
         type: string
--- a/custom_components/tuya_local/devices/sanjun_steamer.yaml
+++ b/custom_components/tuya_local/devices/sanjun_steamer.yaml
@@ -1,0 +1,136 @@
+name: Steam Room
+products:
+  - id: pjh7lsqhyn86almn
+    manufacturer: Sanjun
+    model: Steam Room
+entities:
+  - entity: climate
+    name: Steam Room
+    dps:
+      - id: 104
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: true
+            value: heat
+          - dps_val: false
+            value: "off"
+      - id: 102
+        type: integer
+        name: temperature
+        range:
+          min: 25
+          max: 68
+      - id: 112
+        type: integer
+        name: current_temperature
+        readonly: true
+      - id: 113
+        type: boolean
+        name: temperature_unit
+        mapping:
+          - dps_val: false
+            value: C
+          - dps_val: true
+            value: F
+  - entity: switch
+    name: Power
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+  - entity: light
+    name: Light
+    dps:
+      - id: 105
+        type: boolean
+        name: switch
+  - entity: light
+    name: Colour Lamp
+    dps:
+      - id: 106
+        type: boolean
+        name: switch
+      - id: 107
+        type: string
+        name: effect
+        mapping:
+          - dps_val: "00"
+            value: Rainbow
+          - dps_val: "01"
+            value: Red
+          - dps_val: "02"
+            value: Yellow
+          - dps_val: "03"
+            value: Green
+          - dps_val: "04"
+            value: Cyan
+          - dps_val: "05"
+            value: Blue
+          - dps_val: "06"
+            value: Purple
+          - dps_val: "07"
+            value: White
+  - entity: switch
+    name: Aromatherapy
+    dps:
+      - id: 108
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Wash / Rinse
+    dps:
+      - id: 109
+        type: boolean
+        name: switch
+  - entity: binary_sensor
+    name: Fault
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 111
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+  - entity: button
+    name: RTC Sync
+    icon: "mdi:timer-sync-outline"
+    category: config
+    dps:
+      - id: 114
+        type: boolean
+        name: button
+  - entity: sensor
+    name: Countdown Timer
+    category: diagnostic
+    dps:
+      - id: 103
+        type: base64
+        name: sensor
+        optional: true
+  - entity: sensor
+    name: Schedule / Timing
+    category: diagnostic
+    dps:
+      - id: 110
+        type: base64
+        name: sensor
+        optional: true
+  - entity: sensor
+    name: Steam Time Count
+    category: diagnostic
+    dps:
+      - id: 115
+        type: base64
+        name: sensor
+        optional: true
+  - entity: sensor
+    name: Time / Temperature Range
+    category: diagnostic
+    dps:
+      - id: 116
+        type: base64
+        name: sensor
+        optional: true
--- a/custom_components/tuya_local/devices/sanjun_steamer.yaml
+++ b/custom_components/tuya_local/devices/sanjun_steamer.yaml
@@ -21,10 +21,16 @@ entities:
         range:
           min: 25
           max: 68
+        mapping:
+          - constraint: temperature_unit
+            conditions:
+              - dps_val: true
+                range:
+                  min: 77
+                  max: 154
       - id: 112
         type: integer
         name: current_temperature
-        readonly: true
       - id: 113
         type: boolean
         name: temperature_unit
@@ -34,19 +40,17 @@ entities:
           - dps_val: true
             value: F
   - entity: switch
-    name: Power
     dps:
       - id: 101
         type: boolean
         name: switch
   - entity: light
-    name: Light
     dps:
       - id: 105
         type: boolean
         name: switch
   - entity: light
-    name: Colour Lamp
+    name: Colour lamp
     dps:
       - id: 106
         type: boolean
@@ -58,19 +62,27 @@ entities:
           - dps_val: "00"
             value: Rainbow
           - dps_val: "01"
-            value: Red
+            value: red
+            default: true
+            hidden: true
+      - id: 107
+        type: string
+        name: named_color
+        mapping:
+          - dps_val: "01"
+            value: red
           - dps_val: "02"
-            value: Yellow
+            value: yellow
           - dps_val: "03"
-            value: Green
+            value: lime
           - dps_val: "04"
-            value: Cyan
+            value: aqua
           - dps_val: "05"
-            value: Blue
+            value: blue
           - dps_val: "06"
-            value: Purple
+            value: magenta
           - dps_val: "07"
-            value: White
+            value: white
   - entity: switch
     name: Aromatherapy
     dps:
@@ -78,13 +90,12 @@ entities:
         type: boolean
         name: switch
   - entity: switch
-    name: Wash / Rinse
+    name: Wash rinse
     dps:
       - id: 109
         type: boolean
         name: switch
   - entity: binary_sensor
-    name: Fault
     class: problem
     category: diagnostic
     dps:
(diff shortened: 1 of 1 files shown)


aicbaicb (commented):
Add Sanjun Steam room

Includes Power On, Temperature, Aromatherapy, RGB Lights, White Light, Wash/Rinse, RTC Sync, 

Limitations: Time remaining showing as config in Base64.

Tested and working in Home Assistant
image

image

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+        name: hvac_mode
+        mapping:
+          - dps_val: true
+            value: heat
+          - dps_val: false
+            value: "off"
+      - id: 102
+        type: integer
+        name: temperature
+        range:
```
make-all, line 21: Since temperature_unit is variable, I would expect this range to be conditional on temperature unit.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+          min: 25
+          max: 68
+      - id: 112
+        type: integer
+        name: current_temperature
+        readonly: true
```
make-all: readonly is not needed here, current_temperature is read only by definition

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+          - dps_val: true
+            value: F
+  - entity: switch
+    name: Power
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+  - entity: light
+    name: Light
```
make-all: Name that is the same as the entity type is generally not needed.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+      - id: 113
+        type: boolean
+        name: temperature_unit
+        mapping:
+          - dps_val: false
+            value: C
+          - dps_val: true
+            value: F
+  - entity: switch
+    name: Power
```
make-all: Name is not really needed. The default switch should generally be understood to be main power by most users.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+        name: switch
+  - entity: light
+    name: Colour Lamp
+    dps:
+      - id: 106
+        type: boolean
+        name: switch
+      - id: 107
+        type: string
+        name: effect
```
make-all, line 60: Most of these could be `named_color`. See #4809 for an example.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+          - dps_val: "07"
+            value: White
+  - entity: switch
+    name: Aromatherapy
+    dps:
+      - id: 108
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Wash / Rinse
```
make-all: Avoid punctuation characters in entity names

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+    dps:
+      - id: 109
+        type: boolean
+        name: switch
+  - entity: binary_sensor
+    name: Fault
```
make-all: name not needed.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 111
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
```
make-all, line 107: You also need a default mapping to ensure the type is correct for a binary sensor
```
      - value: true
```

Following the general pattern for problem sensors, you should also add the original raw value as a fault_code attribute so information is not lost
```
    - id: 111
      type: bitfield
      name: fault_code
```

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+  - entity: button
+    name: RTC Sync
```
make-all: "Clock sync" is a better name, as RTC may be too technical for some users to understand. But is it a button to control the sync timing, or is it supposed to be a configuration switch?

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+    icon: "mdi:timer-sync-outline"
+    category: config
+    dps:
+      - id: 114
+        type: boolean
+        name: button
+  - entity: sensor
+    name: Countdown Timer
```
make-all: `translation_key: time_remaining` would be better here.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+    category: diagnostic
+    dps:
+      - id: 103
+        type: base64
```
make-all, line 127: If the base64 cannot be decoded, then this entity should have `hidden: true` at the same level as `category: diagnostic`, as it will not be useful to most users.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+        name: sensor
+        optional: true
+  - entity: sensor
+    name: Schedule / Timing
```
make-all: Pick one name

As with above, `hidden: true` if it is just raw data that most users cannot decipher.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+        type: boolean
+        name: switch
+  - entity: light
+    name: Light
+    dps:
+      - id: 105
+        type: boolean
+        name: switch
+  - entity: light
+    name: Colour Lamp
```
make-all: HA naming convention is to only capitalize the first word of entity and device names.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+    name: Schedule / Timing
+    category: diagnostic
+    dps:
+      - id: 110
+        type: base64
+        name: sensor
+        optional: true
+  - entity: sensor
+    name: Steam Time Count
+    category: diagnostic
```
make-all, line 141: hidden: true

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -0,0 +1,136 @@
...
+    dps:
+      - id: 115
+        type: base64
+        name: sensor
+        optional: true
+  - entity: sensor
+    name: Time / Temperature Range
+    category: diagnostic
```
make-all, line 150: Pick a single name without any punctuation.

hidden: true

make-all (changes requested):

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Good point — I've added a conditional range mapping using a constraint on temperature_unit, applying 77–154°F when DP 113 is true. I'm not certain this syntax is fully supported for range inside conditions — happy to adjust if needed.

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Done. The colour lamp now uses:

effect — exposes Rainbow only (plus a hidden default of red for when a solid colour is active)
named_color — exposes the 7 solid colours using standard HA CSS colour names (red, yellow, lime, aqua, blue, magenta, white)

Purple isn't a standard named_color so I've mapped "06" to magenta as the closest equivalent

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Agreed, added - value: true as default mapping and a second DP 111 entry with name: fault_code and hidden: true.

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Agreed, added hidden: true to all raw base64 sensors.

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Agreed, added hidden: true to all raw base64 sensors.

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Agreed, renamed. Hidden

aicbaicb (commented):
All suggested changes have been made

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Resolved

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Renamed to "Schedule" and hidden

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Renamed to "Wash rinse"

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: name removed

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Renamed to "Clock sync". It's a button: DP 114 triggers a one-time sync of the device clock to the RTC, typically only needed after power loss. A persistent switch wouldn't make sense here.

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: renamed

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Name removed

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: Name removed

custom_components/tuya_local/devices/sanjun_steamer.yaml
aicbaicb: readonly removed

IsaacBreuer:
Thanks @aicbaicb for this. I have opened a pr #5012   as I have decoded the time remaining and allow it to be set as well. 

As my steamroom is not labeled sanjun but I believe all tuya work the same. I am just not sure if all have the same limitations for time as mine has a maximum of 1 hour and the time remaining is actually encoded as 2 bytes first byte is hour and second byte is minutes the hour is only used for 1 hour as this is max (minutes will be 00 when hour is 1 and minutes csn be 0 to 59 when hour is 0)

So I wonder if yours also has the 1 hour limit if so my pr should be good for you as you can see my first approach was to use masks for for hour and min so we don't have to add all possible values in the device yamal. It worked well be had an issue when restarting HA. It was unknown and was not able to set from HA until the device reported a change (like the timer actually changed or was set from app) but with combining the hours and minutes we now can set it even ots unknown,  but that required me to have all 60 values in the mapping.

make-all:
It is certainly not true that "all Tuya work the same". But if your dps all match it is quite likely that your devices are actually the same device just badged differently for different markets.

IsaacBreuer:
OK that's actually why I wanted to hear from someone who has another brand  than the one I have how the timers works for them.

PR #4831, opened 2026-04-06
add support for tomz smart breaker

underscorejasiu:
Hello, added support for [tomz 3 phase breaker](https://aliexpress.com/item/1005010016630671.html). Had been using it in my instance since 2 months and it works fine. Was using same schema for single phase too. Unlike others configs this one supports 3 phases which allows to later use in energy dashboard.
There are also another fields that allows specific configuration of device like cut points etc but I decided i don't want to expose it for ha. If needed I can add them. 
Test are auto generated by ai.


underscorejasiu:
@make-all updated conflicting translations, now all tests passes.  Thanks!

tests/devices/test_tomzn_swc32v4_breaker.py
make-all, line 1: Remove the test changes, we don't add tests for individual devices any more, they are tested by tests/test_device_config.py, and unless there is a new feature added there is no need to add more tests.

custom_components/tuya_local/devices/tomzn_swc32v4_breaker.yaml
```diff
@@ -0,0 +1,189 @@
+name: Breaker SWC32v4
```
make-all: Top level name should be generic in case other products also match

custom_components/tuya_local/devices/tomzn_swc32v4_breaker.yaml
```diff
@@ -0,0 +1,189 @@
...
+products:
+  - id: sh45wydgedautxda
+    manufacturer: TOMZN
+    model: SWC32v4
+
+entities:
+  - entity: switch
+    icon: "mdi:fuse"
```
make-all: Use `translation_key: breaker` instead of customizing the icon

custom_components/tuya_local/devices/tomzn_swc32v4_breaker.yaml
```diff
@@ -0,0 +1,189 @@
...
+    dps:
+      - id: 16
+        type: boolean
+        name: switch
+
+  - entity: sensor
+    name: Total Energy
```
make-all: Remove the name from this sensor so HA's built in translations are not overridden

custom_components/tuya_local/devices/tomzn_swc32v4_breaker.yaml
```diff
@@ -0,0 +1,189 @@
...
+        type: integer
+        unit: kWh
+        class: total_increasing
+        name: sensor
+        mapping:
+          - scale: 100
+
+  #Phase A
+  - entity: sensor
+    name: Phase A Voltage
```
make-all: use
```
    translation_key: voltage_x
    translation_placeholders:
      x: A
```

Likewise for B and C, change `x: A` to `x: B` or `x: C`
Likewise for current and power, change `voltage_x` to `current_x` or `power_x`

custom_components/tuya_local/devices/tomzn_swc32v4_breaker.yaml
```diff
@@ -0,0 +1,189 @@
...
+        type: base64
+        unit: kW
+        name: sensor
+        mask: "0000000000FFFFFF"
+        mapping:
+          - scale: 1000
+
+  #Rest of the sensors
+  - entity: sensor
+    name: Leakage Current
```
make-all: Follow HA naming convention for entity (and device) names: only the first word should be capitalized.

make-all (dismissed):

custom_components/tuya_local/devices/tomzn_swc32v4_breaker.yaml
underscorejasiu: added it as translation key + added translations for it

underscorejasiu:
Replaced names  which used `Leakage Current` with translation key `leakage_current`

underscorejasiu:
Didn't notice some were type number not sensor 🙇🏻  added translations for that type too

custom_components/tuya_local/devices/taxnele_energy_meter.yaml
```diff
@@ -207,7 +207,7 @@ entities:
         name: switch
   - entity: number
     translation_key: timer
-    deprecated: time.timer  # 2026-01-20
+    deprecated: time.timer # 2026-01-20
```
make-all: Why did you change this line?

tests/const.py
```diff
@@ -1543,6 +1543,16 @@
     # "106": False,
 }
 
+TOMZN_SWC32V4_BREAKER_PAYLOAD = {
```
make-all: This can also be removed, as it is not used anywhere

custom_components/tuya_local/devices/taxnele_energy_meter.yaml
underscorejasiu: auto formatting, restored past version

tests/const.py
underscorejasiu: removed

underscorejasiu:
Fixed

PR #4838, opened 2026-04-06
Beok TR9HB-AC2 support

pepitogrillo02:
Add suport for Beok TR9HB-AC2 Fan coil thermostat
image

### DIFF
--- a/custom_components/tuya_local/devices/beok_tr9hb-ac2_thermostat.yaml
+++ b/custom_components/tuya_local/devices/beok_tr9hb-ac2_thermostat.yaml
@@ -11,16 +11,16 @@ entities:
         type: boolean
         name: hvac_mode
         mapping:
-          - dps_val: False
+          - dps_val: false
             value: "off"
-          - dps_val: True
+          - dps_val: true
             constraint: mode
             conditions:
               - dps_val: cool
                 value: cool
               - dps_val: heat
                 value: heat
-              - dps_val: [ventilation, ventil]
+              - dps_val: ventil
                 value: fan_only
       - id: 2
         type: string
@@ -29,13 +29,24 @@ entities:
       - id: 16
         type: integer
         name: temperature
-        unit: C
         range:
-          min: 100
-          max: 350
+          min: 50
+          max: 400
         mapping:
           - scale: 10
             step: 5
+            constraint: temperature_unit
+            conditions:
+              - dps_val: c
+                range:
+                  min: 50
+                  max: 400
+                step: 5
+              - dps_val: f
+                range:
+                  min: 410
+                  max: 1040
+                step: 10
       - id: 23
         type: string
         name: temperature_unit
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/beok_tr9hb-ac2_thermostat.yaml
```diff
@@ -0,0 +1,134 @@
...
+          - dps_val: False
+            value: "off"
+          - dps_val: True
+            constraint: mode
+            conditions:
+              - dps_val: cool
+                value: cool
+              - dps_val: heat
+                value: heat
+              - dps_val: [ventilation, ventil]
```
make-all: Why are two values needed here?

custom_components/tuya_local/devices/beok_tr9hb-ac2_thermostat.yaml
```diff
@@ -0,0 +1,134 @@
...
+        type: string
+        name: mode
+        hidden: true
+      - id: 16
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 100
+          max: 350
```
make-all: Is this range correct for both C anf F?

custom_components/tuya_local/devices/beok_tr9hb-ac2_thermostat.yaml
```diff
@@ -0,0 +1,134 @@
...
+      - id: 36
+        type: string
+        name: hvac_action
+        hidden: true
+        mapping:
+          - dps_val: open
+            value: idle
+            constraint: mode
+            conditions:
+              - dps_val: [cool]
```
make-all: Why is an array used here?

custom_components/tuya_local/devices/beok_tr9hb-ac2_thermostat.yaml
```diff
@@ -0,0 +1,134 @@
...
+        type: string
+        name: unit
+        mapping:
+          - dps_val: c
+            value: C
+          - dps_val: f
+            value: F
+      - id: 108
+        type: integer
+        name: step
```
make-all, line 143: What are you expecting this to affect?

pepitogrillo02: I copied this from the existing beok_tr8b_thermostat definition. I don't know if it has effect or not. If you think it has no effect in practice, I'm happy to remove it.

custom_components/tuya_local/devices/beok_tr9hb-ac2_thermostat.yaml
pepitogrillo02: I previously included "ventilation" for compatibility with similar devices, but I’ve now kept only "ventil", as it is the value reported by my device.

custom_components/tuya_local/devices/beok_tr9hb-ac2_thermostat.yaml
pepitogrillo02: I’ve set different value ranges for C and F in the following commit.

custom_components/tuya_local/devices/beok_tr9hb-ac2_thermostat.yaml
pepitogrillo02: Not needed. Removed.

PR #4841, opened 2026-04-06
Add AquaViva Superior Split waterheater

terryfork:


custom_components/tuya_local/devices/aquaviva_superiorsplit_waterheater.yaml
```diff
@@ -0,0 +1,61 @@
+name: Water heater
+products:
+  - id: cjfj60binlzftdqw
+    manufacturer: Aquaviva
+    model: Superior Split
+entities:
+  - entity: water_heater
```
make-all: Since this includes cooling modes, this would be better as a climate entity (with `translation_key: pool_heatpump`)

remora_heatpump is very close to this device, so follow that. I think only the `quiet_cool` mode is missing from this model, and the temperature range is different. Maybe the water flow detection is not the same (can be removed and just the general problem sensor used)

make-all (changes requested):

custom_components/tuya_local/devices/aquaviva_superiorsplit_waterheater.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - id: cjfj60binlzftdqw
+    manufacturer: Aquaviva
+    model: Superior Split
+entities:
+  - entity: climate
+    translation_key: pool_heatpump
+    dps:
+      - id: 1
+        type: boolean
+        name: operation_mode
```
make-all, line 12: You will need to change this to `hvac_mode` with values mapped to valid hvac_modes.

Since there is not full coverage of the below in hvac_mode, you will also need a preset_mode to cover all.

See remora_heatpump for the simple version; always reporting the hvac_mode as heat (incorrectly) - though you can use auto instead to be less incorrect - this is mainly used because some devices do not like you setting multiple dps at once.

Or brustec_br80_poolheater.yaml for a more correct version (but using different strings).

Due to the match with strings used by remora, I would not be surprised if your device also has issues with multiple dps being set at once (but it should already affect the current config).

custom_components/tuya_local/devices/aquaviva_superiorsplit_waterheater.yaml
```diff
@@ -0,0 +1,61 @@
...
+        range:
+          min: 15
+          max: 40
+        unit: C
+      - id: 4
+        type: string
+        name: mode
+        hidden: true
+  - entity: sensor
+    class: temperature
```
make-all, line 41: This should be the `current_temperature` of the climate entity (or even the previous water_heater entity), not a separate sensor.

PR #4846, opened 2026-04-07
Add support for Spa PB562

Tombost:
Added configuration for JOYONWAY Spa PB562 based on DP analysis : 

Features implemented:

- Climate Control: Target temperature (15°C to 40°C) and current temperature monitoring.
- Equipment Switches: Dedicated controls for Jet Pump, Blower, and Lights.
- Scheduling (Numbers): Ability to set Start and End times for both Heating and Filtration cycles.
_Note: These are implemented as number entities representing minutes from midnight (0-1440 range) with a 30-minute step for better UI usability._
- Energy Monitoring: Real-time Power (W) and Voltage (V) sensors to track heater activity and electrical status.

**Data Points (DPs) used:**

- 3: Target Temperature
- 24: Current Temperature
- 101: Voltage (V)
- 102: Power (W)
- 104: Jet Pump
- 123: Lights
- 125: Blower
- 127/128: Heating Start/End (minutes from midnight)
- 137/138: Filtration Start/End (minutes from midnight)

I have verified that all entities respond correctly and do not conflict with existing device schemas.


custom_components/tuya_local/devices/spa_pb562.yaml
```diff
@@ -0,0 +1,99 @@
+name: Spa PB562 Complete
+products:
+  - id: jbxvi3pbxjjlmec2
```
make-all, line 3: Add manufacturer and model details here

custom_components/tuya_local/devices/spa_pb562.yaml
```diff
@@ -0,0 +1,99 @@
+name: Spa PB562 Complete
```
make-all: Remove the model name from here, as other models may also match. Just use a generic name for the type of device.

custom_components/tuya_local/devices/spa_pb562.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  - entity: switch
+    name: Blower
+    dps:
+      - id: 125
+        type: boolean
+        name: switch
+
+  - entity: light
+    name: Lights
```
make-all: This name is not necessary

custom_components/tuya_local/devices/spa_pb562.yaml
```diff
@@ -0,0 +1,99 @@
...
+    dps:
+      - id: 123
+        type: boolean
+        name: switch
+
+  # --- HEATING SCHEDULE (MINUTES) ---
+  - entity: number
```
make-all, line 52: Use a time entity, since the range is 24h

To display correctly, your user settings will need to be "24h" time display, but then it is much better than number.

custom_components/tuya_local/devices/spa_pb562.yaml
```diff
@@ -0,0 +1,99 @@
+name: Spa PB562 Complete
+products:
+  - id: jbxvi3pbxjjlmec2
+
+entities:
+  # --- THERMOSTAT ---
+  - entity: climate
```
make-all, line 8: Is there no on/off switch for the heating?

make-all (commented):
Since this is only heating, and the medium is water, it may be more appropriate as a water_heater rather than climate entity.

custom_components/tuya_local/devices/spa_pb562.yaml
```diff
@@ -0,0 +1,121 @@
...
+  - entity: switch
+    name: Blower
+    dps:
+      - id: 125
+        type: boolean
+        name: switch
+
+  # --- LIGHT ---
+  - entity: light
+    name: LED
```
make-all: I don't think a name is needed here.

custom_components/tuya_local/devices/spa_pb562.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+  - entity: sensor
+    class: voltage
+    dps:
+      - id: 101
+        type: integer
+        name: sensor
+        unit: V
+
+  - entity: sensor
```
make-all: Please see other configs for the standard handling of fault sensors as binary_sensors with class: problem and fault_code as an extra attribute.

make-all (changes requested):

PR #4852, opened 2026-04-08
Add garage door opener with light

terryrankine:
## Summary

Adds a device config for Tuya garage door openers that include a controllable light (DPS 105).

**Device:** Auto Openers AOSDK Smart Garage Door Opener Kit - Panel
**Manufacturer:** Auto Openers (autoopeners.com.au)
**Product page:** https://autoopeners.com.au/product/auto-openers-aosdk-smart-garage-door-opener-kit-panel/
**App:** Smart GDO (Tuya-based)
**Tuya product name:** Smart GDO only(6T-8T), category `ckmkzq`

## DPS mapping

Confirmed via live device testing (protocol 3.5):

| DPS | Type | Role | Values |
|-----|------|------|--------|
| 101 | string | control | `open` / `close` / `stop` |
| 102 | integer | current_position / position | 0–100 (%) |
| 105 | boolean | light switch | true / false |
| 107 | string | action / state | `opening` / `opened` / `closing` / `closed` / `partial_opening` |

## Entities

- `cover` (class: garage) — open/close with position feedback and target position support
- `light` — garage light built into the opener unit


custom_components/tuya_local/devices/garage_door_with_light.yaml
```diff
@@ -0,0 +1,44 @@
...
+          - dps_val: opened
+            value: open
+          - dps_val: closed
+            value: closed
+          - dps_val: opening
+            value: opening
+          - dps_val: closing
+            value: closing
+          - dps_val: partial_opening
+            value: open
```
make-all, line 38: Incorrectly reporting the cover as open when it is only partially open can result in the open button in the UI becoming disabled (the exact circumstances may be more complicated and involve current_position, but are not documented anywhere, so I prefer to be safe and not falsely report this).
Instead, this should be mapped to null, which will make the action Unknown, and the current_position should then be used on its own by HA.

make-all (changes requested):

terryrankine:
Happy to change it to whatever you want

PR #4866, opened 2026-04-10
Refactor Feyree 4.3 EV Charger configuration for better stability and control logic.

kriskomp:
Refactor Feyree 4.3 EV Charger configuration:Optimized Fault Reporting: Simplified binary_sensor for error detection by removing redundant optional flags and improving the bitfield mapping.Improved Current Control: Implemented a more robust number entity for charge current using value_redirect to handle different hardware power limits (16A/32A/40A/50A) dynamically.Enhanced UI/UX: Added dedicated Start/Stop buttons for more intuitive charging control.Cleaned up code: Fixed linting issues (line lengths and spacing) and categorized entities (diagnostic/config) for better Home Assistant integration.

### DIFF
--- a/custom_components/tuya_local/devices/feyree_43_evcharger.yaml
+++ b/custom_components/tuya_local/devices/feyree_43_evcharger.yaml
@@ -38,9 +38,12 @@ entities:
     dps:
       - id: 10
         type: bitfield
+        optional: true
         name: sensor
         mapping:
           - dps_val: 0
+            value: false
+          - dps_val: null
             value: false
           - value: true
 
@@ -295,7 +298,7 @@ entities:
             value: 40
           - dps_val: Max50A
             value: 50
-          # Jeśli DP 113 nie pasuje do niczego powyżej, suwak domyślnie ustawi 32A.
+          # Jeśli DP 113 nie pasuje do niczego powyżej
           - value: 16
 
       # 5) Definicje docelowych rejestrów (obsługiwane przez przekierowanie)
@@ -387,7 +390,7 @@ entities:
         mapping:
           - dps_val: "OpenCharging"
             value: true
-            constraint: "OpenCharging" # Wymuszamy tę wartość
+            constraint: "OpenCharging"
 
   - entity: button
     name: Stop charging
--- a/custom_components/tuya_local/devices/feyree_43_evcharger.yaml
+++ b/custom_components/tuya_local/devices/feyree_43_evcharger.yaml
@@ -31,14 +31,13 @@ entities:
           - dps_val: error
             value: fault
 
-  # --- BŁĘDY ---
+# --- BŁĘDY ---
   - entity: binary_sensor
     class: problem
     category: diagnostic
     dps:
       - id: 10
         type: bitfield
-        optional: true
         name: sensor
         mapping:
           - dps_val: 0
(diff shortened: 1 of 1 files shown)


make-all:
Perhaps you can explain why each of these changes are necessary, and not just your personal preference

kriskomp:
Of course
As an owner of this device, I have conducted extensive real-world testing, especially within Home Assistant automations and manual dashboard control. These changes are necessary for stability and to mirror the device's native logic:

1. Charge Current (DP 114/113): The previous configuration was flawed as it lacked the readonly: true flag for DP 113. This caused Home Assistant to attempt illegal write commands to a read-only configuration register whenever the charging current was adjusted. These invalid commands led to device communication errors (Protocol Error 914) and MCU lockups. The new version correctly uses DP 113 as a read-only constraint for a dynamic value_redirect, allowing the UI to safely scale the slider (e.g., 8A-32A) while ensuring that write commands are only sent to the appropriate current-control registers (DP 114/115/116/117)."

2. Charging Control (Buttons vs Select): I have added button entities for 'Start' and 'Stop' commands. Using a single select entity is impractical because the device state often remains stuck on the last command after the car is unplugged. This makes it impossible to re-trigger the action without first toggling the state. Buttons accurately reflect how the native Smart Life app functions and ensure reliable operation in automations.

make-all:
1. Seems to describe a bug which needs to be fixed in the code (if it actually exists). HA does not have any API to write to minimum and maximum values on the number entity, they are read-only by definition. So the `readonly: true` is not required. If dp 113 is being written, I need you to capture debug logs so I can track down where the actual bug is.
2. OK, retriggering is a fair reason, what about the Wait option there? I think I asked a question about this on the originall PR, as it does not seem like it should be a user selectable option, which led me to think this should only be a sensor. Or does it start the delayed charging function?

kriskomp:
Thank you for the detailed feedback. Here are my thoughts on your points:

1. Read-only flag (DP 113): I have re-tested this thoroughly. You are correct that the integration should handle this natively. I have reverted this change in my latest commit to keep the code clean and will monitor the logs for any further 914 errors.

2. Charging Control (Buttons vs. Select): As a daily user of this device, I strongly believe that buttons provide a much better experience than a single select entity for several reasons:

Intuition & UX: Buttons for 'Start' and 'Stop' are more intuitive for manual dashboard control. They mirror the logic of the native Smart Life app, where you trigger an action rather than picking a state from a list.

Automation Reliability: For Home Assistant automations (e.g., solar tracking), sending a 'press' command to a button is more reliable than setting a state in a selector, especially when the device state might be out of sync after a physical unplug.

The 'Wait' Issue: You were right about the 'Wait' option—it is a status, not a command. By using buttons, I have limited the control interface to only the two functional actions (OpenCharging / CloseCharging).

Hybrid Approach: I have kept the select entity in the Configuration category for those who prefer it, but the button entities remain as the primary, stateless way to control the charging process.

I believe this approach offers the most robust and user-friendly way to manage the Feyree wallbox.

kriskomp:
After revert my change in set current i have situation
Entitity ja unavailable for a moment
After my fix we don't have such situation.
I will sent you log to confirm it. Revert of change 10 minutes ago was
miatake.



pt., 10 kwi 2026, 15:19 użytkownik Jason Rumney ***@***.***>
napisał:

kriskomp:
Please disregard my last message about the 'unavailable' state. It was a false alarm; the entity became unavailable simply because Home Assistant was restarting during my tests, not due to any issues with the DP 113 configuration.

I have confirmed that the device works perfectly fine without the readonly: true flag on DP 113. My latest commit, which reverts that change, is correct and ready for review :)

kriskomp:
I am using the version after commit [da87965](https://github.com/make-all/tuya-local/commit/da87965b1ab012ea6574c2341861eafeec364178). After restarting, I encounter a configuration error. It appears the sensor_charging_operation in the YAML file is missing the required name: sensor mapping for its DPS.

Rejestrator: custom_components.tuya_local.helpers.config
Źródło: custom_components/tuya_local/helpers/config.py:34
integracja: Tuya Local (dokumentacja, Problemy)
Pierwsze zdarzenie: 15:31:49 (1 wystąpienie)
Ostatnio zalogowany: 15:31:49

Error adding sensor_charging_operation for feyree_43_evcharger.yaml: sensor_charging_operation is missing a sensor dps
2026-04-12 15:31:49.013 DEBUG (SyncWorker_6) [custom_components.tuya_local.device] Wallbox refreshed device state: {"dps": {"10": 0, "15": 0, "25": 445, "101": "no_connet", "102": 2436, "103": 2424, "104": 2272, "105": 0, "106": 0, "107": 0, "108": "Three_phase", "109": 0, "110": 257, "111": 1, "112": 0, "113": "Max32A", "115": 8, "116": 24, "117": 24, "118": 0, "119": 0, "120": "00:00:00", "123": false, "124": "OpenCharging", "125": "FELDWV02", "126": "V1.6", "132": false}}
2026-04-12 15:31:49.066 ERROR (MainThread) [custom_components.tuya_local.helpers.config] Error adding sensor_charging_operation for feyree_43_evcharger.yaml: sensor_charging_operation is missing a sensor dps

PR #4875, opened 2026-04-12
Add device definition for KASMWEKFITEA electric blanket

haywoodmarx:
## Device
- **Model:** KASMWEKFITEA
- **Product Name:** Wifi Electric Fitted Blanket
- **Tuya Product ID:** veafgxwmhwd5iakw
- **Known as:** Kogan SmartHome Electric Fitted Blanket (likely available as white-label variants under other brands)

## Features Exposed

| Entity | Type | Values/Range |
|--------|------|--------------|
| Power | Switch | on/off |
| Body heat level | Select | Level 0–6 |
| Feet heat level | Select | Level 0–6 |
| Body timer | Select | 1–12 hours |
| Feet timer | Select | 1–12 hours |
| Body remaining time | Sensor | 0–86400 seconds |
| Feet remaining time | Sensor | 0–86400 seconds |

## Implementation Note: Heat Level Mapping

The device uses an unusual protocol mapping for heat levels: DPS values `level_1` through `level_7` map to user-facing levels 0–6 (off through maximum). This creates a 1-level offset. The device firmware does not support `level_0`, so the valid range begins at `level_1`.

## Testing
- Validated with Home Assistant 2026.4.2
- Validated with tuya-local v2026.4.1
- All 7 entities functional and verified with actual device


custom_components/tuya_local/devices/kogan_kasmwekfitea_electricblanket.yaml
```diff
@@ -0,0 +1,145 @@
+name: Electric Fitted Blanket
+products:
+  - id: veafgxwmhwd5iakw
+    model: KASMWEKFITEA
```
make-all, line 5: manufacturer: Kogan should be included here. Also in the filename.

Don't put too much detail in the filename, as it is getting too long, just `kogan_kasmwefitea_electricblanket.yaml` without the "fitted" is OK.

custom_components/tuya_local/devices/kasmwekfitea_electric_fitted_blanket.yaml
```diff
@@ -0,0 +1,145 @@
+name: Electric Fitted Blanket
```
make-all: Follow HA naming convention for device and entity names.
Only the first word should be capitalised.

custom_components/tuya_local/devices/kasmwekfitea_electric_fitted_blanket.yaml
```diff
@@ -0,0 +1,145 @@
...
+    name: Body heat level
+    icon: "mdi:thermometer"
+    category: config
+    dps:
+      - id: 2
+        type: string
+        name: option
+        mapping:
+          - dps_val: "level_1"
+            value: "level 0"
```
make-all: Values should be human readable, so "Level" should be capitalised

custom_components/tuya_local/devices/kasmwekfitea_electric_fitted_blanket.yaml
```diff
@@ -0,0 +1,145 @@
...
+          - dps_val: "level_5"
+            value: "level 4"
+          - dps_val: "level_6"
+            value: "level 5"
+          - dps_val: "level_7"
+            value: "level 6"
+
+  - entity: select
+    name: Body timer
+    icon: "mdi:timer"
```
make-all: Use `translation_key: timer` instead of the icon, and for values use "1h", "2h" etc to match the translation keys.

make-all (changes requested):

PR #4876, opened 2026-04-12
Added support for WT24 mmWave Human Presence Sensor (人体存在传感器WT24)

rtxbae:
New device: 人体存在传感器WT24

Relevant product page: https://zigbee.blakadder.com/Tuya_ZG-205Z-B.html
Product review: https://smarthomescene.com/reviews/tuya-zigbee-presence-sensor-zg-205z-review/
Tuya product id: 2jmaefvrsyghdswo
Manufacturer: Tuya

This product is relevant to its sister product https://www.zigbee2mqtt.io/devices/ZG-205Z_A.html

Tested with real product and everything is working as expected:

image


custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
```diff
@@ -0,0 +1,110 @@
+name: mmWave presence sensor
+products:
+  - id: 2jmaefvrsyghdswo
+    name: 人体存在传感器WT24
```
make-all, line 4: Name should not be used here, it is just duplicating model.

custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
```diff
@@ -0,0 +1,110 @@
...
+    model: 人体存在传感器WT24
```
make-all, line 5: This should just be WT24
If possible, please include the manufacturer (brand under which it is sold),

custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - id: 1
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: none
+            value: false
+          - dps_val: presence
+            value: true
+          - dps_val: peaceful
+            value: false
```
make-all, line 19: This should be true for occupancy (false for motion) I think.

custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
```diff
@@ -0,0 +1,110 @@
...
+        type: integer
+        name: value
+        range:
+          min: 1
+          max: 10
+        mapping:
+          - scale: 0
+            step: 1
+  - entity: number
+    name: Near detection distance
```
make-all, line 35: I prefer "Minimum detection distance" (and "Maximum detection distance" for below) for consistency with other similar devices.

custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
```diff
@@ -0,0 +1,110 @@
...
+        name: value
+        optional: true
+        range:
+          min: 0
+          max: 1000
+        mapping:
+          - step: 10
+        unit: cm
+  - entity: sensor
+    name: Target closest distance
```
make-all, line 65: Use `class: distance` with no name or icon

custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
```diff
@@ -0,0 +1,110 @@
...
+          - dps_val: none
+            value: false
+          - dps_val: presence
+            value: true
+          - dps_val: peaceful
+            value: false
+  - entity: number
+    name: Motion sensitivity
+    category: config
+    icon: mdi:motion-sensor
```
make-all, line 23: Icon names must be quoted due to the embedded `:` which confuses the yaml parser

custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
```diff
@@ -0,0 +1,110 @@
...
+    dps:
+      - id: 2
+        type: integer
+        name: value
+        range:
+          min: 1
+          max: 10
+        mapping:
+          - scale: 0
```
make-all, line 32: `scale: 0` is not valid and will cause a divide by zero error.
`step: 1` is not needed, as that is the default, likely the scale should be 1 which is also the default so the whole mapping can be removed.

custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
make-all: `class: distance` can also be used on the number entities above instead of the icon, but leave the names in those cases, because there are multiple.

custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
```diff
@@ -0,0 +1,110 @@
...
+    name: Target closest distance
+    icon: mdi:signal-distance-variant
+    dps:
+      - id: 9
+        type: integer
+        name: sensor
+        class: measurement
+        unit: cm
+  - entity: number
+    name: No person duration
```
make-all, line 74: Standard name would be "Exit delay"

use `class: duration` instead of the icon below

custom_components/tuya_local/devices/wt24_mmWave_presence_sensor.yaml
```diff
@@ -0,0 +1,110 @@
...
+    dps:
+      - id: 103
+        type: integer
+        name: value
+        range:
+          min: 1
+          max: 10
+  - entity: switch
+    category: config
+    name: Switch
```
make-all, line 106: Name not needed, it is obvious to users that the default switch will turn on/off the device.

make-all (changes requested):

make-all:
Non responsive contributor

PR #4880, opened 2026-04-12
add BAC-009 Smart Knob Thermostat climate device support

ludook:
Hi @make-all,
thank you very much for your great add-on.

in this PR you will find a new device : https://www.becasmart.com/product/75.html

image

tested with tuya-local 2026.4.0, on homeassistant 2026.4.1


custom_components/tuya_local/devices/beca_bht009_thermostat_c.yaml
```diff
@@ -0,0 +1,107 @@
...
+        type: integer
+        name: current_temperature
+        mapping:
+          - scale: 10
+      - id: 36
+        type: string
+        name: hvac_action
+        mapping:
+          - dps_val: close
+            value: null
```
make-all: Should be `idle`

make-all (commented):
The _c on the filename is not needed, it was only added on some very early models back in 2019 before we knew how to handle multiple units.
The presence of min and max temperatures suggests that we should be able to handle C and F in the same config if anyone ever comes with a request to support a US version.

custom_components/tuya_local/devices/beca_bht009_thermostat_c.yaml
ludook: I try first with "idle", but when idle is set here, a schedule/time icon is added on right top of climate picture
the same icon as when a schedule is really set on thermostat

example : 
 the first thermostat on this pic is built with tuya-local and with hvac_action config set to idle on dps_val closed
the 3 others are built with zigbee2mqtt, and then so with mqtt climate integration (with hvac action mapped I mean since I can see heating or cooling status on their charts)
image

so to keep homogeneous design, I switch tuya-local hvac_action to null when valv is off

make-all: The aim of this integration is to integrate with HA, not to match your previous MQTT setup. `idle` does not have the meaning you are giving it.

ludook: ok I understand you remark
but it is not my previous mqtt setup, just the default zigbee2mqtt setup for thermostats 
so you think that zigbee2mqtt thermostat integration is wrong with HA spirit on valv status ?
should I open a bug on zigbee2mqttt repo ? (could you please give me the HA doc pointers which explains the right valv behaviour that I should join in the bug ticket ?)

make-all: The HA documentation is https://developers.home-assistant.io/docs/core/entity/climate/#hvac-action

I cannot comment on Zigbee2MQTT, as I am not familiar with what it is doing.

make-all:
Actually I am curious why you would move from a Z2M setup to a Tuya based setup for this device. Native Zigbee is better than dealing with Tuya Zigbee hubs.

custom_components/tuya_local/devices/beca_bht009_thermostat_c.yaml
ludook: I update hvac_action value to idle when valv is off

ludook:
> The _c on the filename is not needed, it was only added on some very early models back in 2019 before we knew how to handle multiple units. The presence of min and max temperatures suggests that we should be able to handle C and F in the same config if anyone ever comes with a request to support a US version.

fixed

ludook:
> Actually I am curious why you would move from a Z2M setup to a Tuya based setup for this device. Native Zigbee is better than dealing with Tuya Zigbee hubs.

because a lot of zigbee instability only on these kind of thermostats (I have 100 zigbee devices moreover and only issues with them, even isolated on their zigbee network and coordinator)
I suspect a memory leak on zigbee stack in those devices, in specific conditions like 'zigbee2mqtt with partial or wrong support' or something not cool like that : works well for a couple weeks and then zigbee freeze. so I need to unplug them from wall to unfreeze zigbee events ....
and after a lot of tests, a lot of researches and discussions with my new AI friends, they suggest that a tuya box with 'official zigbee support' for those devices, should avoid thermostat to freeze their zigbee stack
so I try and pray :-)

PR #4887, opened 2026-04-13
Add JTMSPRO door knob lock device config

kordek212:
Adds new device file` custom_components/tuya_local/devices/jtmspro_door_knob_lock.yaml`

Maps lock DPs for fingerprint/password/card/BLE unlock records, lock motor state, remote unlock checks, offline password handling, and config controls (auto lock, reverse lock, arming, beep volume)

Includes product mapping for bf7920otpjfk9cqr (JTMSPRO door knob lock)


custom_components/tuya_local/devices/hornbill_y4_smart_lock.yaml
```diff
@@ -1,7 +1,7 @@
 name: Keypad lock
 products:
   - id: g7rboh0x
-    manufacturer: Hornbill
+    manufacturer: jtmspro
```
make-all, line 4: Why this change?

make-all (commented):
The product id indicates that this is a WiFi device, but it has a battery sensor.

I am not sure it is worth my time reviewing almost 500 lines of config for a device that is not going to work properly.

PR #4893, opened 2026-04-14
Add custom configuration for Frizzlife LP365

TastePurple:
Existing configuration for LP365**P** exists but this extends support to the LP365


make-all:
Unfortunately I cannot accept having contributions dropped on me to clean up without the necessary device information having been provided.

PR #4899, opened 2026-04-15
Adding new device a Eglo fan with light.

pro-logic:
There seems to be an existing YAML for an Eglo fan without a light, I've created this one for Eglo with light.


pro-logic:
Sorry, after I submitted this PR my fan started playing up so I pulled the PR. 

Upon testing this had nothing to do with this PR so re-opening.

make-all:
Merged with existing eglo_ceiling_fan config, as the product ids are the same

PR #4915, opened 2026-04-17
KRX8 Fixed status for programs and moved entries to/from config

camcamnz:
- Moved Start buttons to main section
- Moved remaining Program switches to config section
- Added mappings for Program switches so that they display properly

### DIFF
--- a/custom_components/tuya_local/devices/krain_krx8_irrigation_controller.yaml
+++ b/custom_components/tuya_local/devices/krain_krx8_irrigation_controller.yaml
@@ -100,7 +100,6 @@ entities:
   - entity: button
     name: Start A
     icon: "mdi:timer-play"
-    category: config
     dps:
       - id: 117
         type: boolean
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/krain_krx8_irrigation_controller.yaml
```diff
@@ -141,52 +137,93 @@ entities:
           - value: true
   - entity: switch
     name: Enable programs
+    category: config
     icon: "mdi:calendar-check"
     dps:
       - id: 121
         type: boolean
         name: switch
+        mapping:
```
make-all: This mapping is wrong, and not needed.

camcamnz:
@make-all why didn't you include the mapping fixes for programs? 
To show why I added this, here is how it looked when updated to the latest release, without the mappings added:
image

And here is the same with my fixes applied:
image

I see lint complained that the values are not true/false, but according to the documentation I can find switches should have a value of on or off, not true or false.

PR #4919, opened 2026-04-17
Add support for Shrivee Triple water timer

macher91:
Support for Shrivee 3-Way Automatic WiFi Connected Watering Timer.

RF Water Timer
Model: HCT-638

Sub-GHz Gateway:
HCG-003


macher91:
@make-all I fixed the issue

PR #4921, opened 2026-04-17
Add support for l10wfe gas water heater

nicholasng99:
image
tested to match controls from Tuya

### DIFF
--- a/custom_components/tuya_local/devices/city_energy_l10wfe_gas_waterheater.yaml
+++ b/custom_components/tuya_local/devices/city_energy_l10wfe_gas_waterheater.yaml
@@ -1,0 +1,155 @@
+name: Gas water heater
+products:
+  - id: di9ybnrv3s2d1axk
+    manufacturer: City Energy
+    model: L10WFE
+entities:
+  - entity: water_heater
+    translation_only_key: water_air
+    dps:
+      - id: 1
+        type: boolean
+        name: operation_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: gas
+      - id: 9
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 35
+          max: 55
+      - id: 10
+        type: integer
+        name: current_temperature
+        unit: C
+  - entity: sensor
+    class: enum
+    translation_key: status
+    category: diagnostic
+    dps:
+      - id: 13
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: standby
+            value: standby
+          - dps_val: burning
+            value: heating
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 20
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 20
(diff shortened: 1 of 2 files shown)


make-all:
Please revert the last commit. This integration is targeting Home Assistant, not HomeKit, and lack of support is an issue for the homekit integration with HA, not this integration.
Probably the issue can be worked around by defining a climate template helper to map the water_heater functionality onto a climate entity in the meantime.

custom_components/tuya_local/devices/city_energy_l10wfe_gas_waterheater.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - id: 13
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: standby
+            value: standby
+          - dps_val: burning
+            value: heating
+  - entity: sensor
+    name: Water flow
```
make-all: Remove the name and let the default class translation be used instead

custom_components/tuya_local/devices/city_energy_l10wfe_gas_waterheater.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - id: 103
+        type: integer
+        name: sensor
+        unit: m³
+        class: measurement
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: Cumulative gas
+    class: volume
```
make-all: Change this to `class: gas`, to make it compatible with the energy dashboard.

custom_components/tuya_local/devices/city_energy_l10wfe_gas_waterheater.yaml
```diff
@@ -0,0 +1,117 @@
...
+    category: diagnostic
+    dps:
+      - id: 105
+        type: integer
+        name: sensor
+        unit: L
+        class: measurement
+  - entity: sensor
+    name: Cumulative water
+    class: volume
```
make-all: Change to class: water to make this compatible with the energy dashboard

custom_components/tuya_local/devices/cityenergy_l10wfe_gaswaterheater.yaml
```diff
@@ -0,0 +1,117 @@
...
+    category: diagnostic
+    dps:
+      - id: 106
+        type: integer
+        name: sensor
+        unit: m³
+        class: total_increasing
+  - entity: button
+    name: Reset water meter
+    category: config
```
make-all, line 100: Probably make these two buttons `hidden: true`, because they are too easy to hit by mistake and generally should not need resetting.

make-all (changes requested):

custom_components/tuya_local/devices/city_energy_l10wfe_gas_waterheater.yaml
nicholasng99: Hi, can i clarify if you mean just this name or all the name entries

make-all: I was singling out that one, since the name does not add info that is not already covered by the class

PR #4922, opened 2026-04-17
Add support trident air dome 70

nicholasng99:
image
tested to match controls from Tuya

### DIFF
--- a/custom_components/tuya_local/devices/trident_airdome70_airpurifier.yaml
+++ b/custom_components/tuya_local/devices/trident_airdome70_airpurifier.yaml
@@ -78,21 +78,6 @@ entities:
           - dps_val: severe
             value: severe
 
-  - entity: binary_sensor
-    class: problem
-    category: diagnostic
-    dps:
-      - id: 22
-        type: bitfield
-        name: sensor
-        mapping:
-          - dps_val: 0
-            value: false
-          - value: true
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/trident_airdome70_airpurifier.yaml
```diff
@@ -0,0 +1,104 @@
...
+    class: pm25
+    dps:
+      - id: 2
+        name: sensor
+        type: integer
+        unit: ugm3
+        class: measurement
+
+  - entity: select
+    name: Fan speed
```
make-all: It is better to merge this into the fan entity as `speed`, and map the values to 33, 66, 100

custom_components/tuya_local/devices/trident_airdome70_airpurifier.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - id: 21
+        name: sensor
+        type: string
+        mapping:
+          - dps_val: great
+            value: excellent
+          - dps_val: good
+            value: good
+          - dps_val: medium
+            value: moderate
```
make-all: maybe `poor` is better than jumping straight from moderate to severe.

make-all (changes requested):

PR #4929, opened 2026-04-19
Add Ideal Clima fancoil support

deltaclock:
For reference, this configuration targets the device TQCT07. This is used by [Ideal Clima ](https://www.idealclima.eu/it/area-tecnica/wi-fi-e-wi-fi-ready) on various units. I can confirm it works perfectly for my Nemo units.

### DIFF
--- a/custom_components/tuya_local/devices/ideal_clima_fancoil.yaml
+++ b/custom_components/tuya_local/devices/ideal_clima_fancoil.yaml
@@ -1,0 +1,76 @@
+name: Fancoil
+products:
+  - id: 5dgguakbmhwzwiko
+    manufacturer: Ideal Clima
+    model: Nemo
+entities:
+  - entity: climate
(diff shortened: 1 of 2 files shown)


deltaclock:
Hey @make-all, about these commits some comments:

- Are you sure the `aircon_extra` key is suitable for this? I see the quiet value but it doesn't match the rest exactly, like low, or high. What kind of icons will be shown if this key is used?
- Fancoils are different from heat pumps or AC units. For example my current home setup features plenty fancoils and one heat pump unit which heats or cools the water. Fancoils are being used to blow hot or cold air inside. While similar to AC units in terms of function, there are definitely some differences. https://en.wikipedia.org/wiki/Fan_coil_unit & https://www.idealclima.eu/en/ (under products). The term is more commonly used in Europe for sure.

make-all:
low and high are provided by the default climate translations

PR #4936, opened 2026-04-20
feat: expose update_dps as a Home Assistant service

facuparedes:
## Summary

- Adds a new `tuya_local.update_dps` service that sends the Tuya `UPDATEDPS` (`0x12`) protocol command to request fresh datapoint values from a device on demand
- Accepts an optional `dps` parameter to specify which DP indexes to refresh (defaults to the device's `force_dps` list, or `[18, 19, 20]` if none configured)
- Properly acquires `_api_lock` to avoid socket conflicts with the persistent connection loop
- Uses `nowait=True` so the response is processed by `async_receive`'s loop, ensuring entities update immediately

## Motivation

Currently, `updatedps()` is only triggered internally via the 30-second cache timeout cycle (alternating with `status()`), and only for devices with DPs marked `force: true` in their config YAML. There is no way for users to trigger an on-demand DPS refresh from automations or scripts.

This is particularly useful for energy monitoring scenarios where users want to:
- Poll power/current/voltage DPs only when a smart plug is actively in use
- Avoid unnecessary traffic when the device is idle
- Control polling frequency dynamically based on device state

### Example automation

```yaml
automation:
  - alias: "Poll energy only when plug is on"
    trigger:
      - platform: time_pattern
        seconds: "/10"
    condition:
      - condition: state
        entity_id: switch.my_plug
        state: "on"
    action:
      - action: tuya_local.update_dps
        data:
          device_id: "bf3a9c1e4ab012df3c"
          dps:
            - 18
            - 19
            - 20
```

## Implementation details

- `nowait=True` is critical: without it, `updatedps()` consumes the device response directly but discards it. Entities only update when `async_receive`'s loop yields data on its 30s cycle. With `nowait=True`, the command is sent and the response stays in the socket buffer for the loop to pick up on its next iteration (~0.1s), allowing immediate entity updates.
- `_api_lock` must be acquired before calling `updatedps` because the persistent connection loop holds this lock while communicating over the shared socket.

## Changes

- `__init__.py`: Added imports, service schema, `async_setup()` with `_handle_update_dps` handler (with proper locking and nowait)
- `services.yaml`: New file with `update_dps` service definition

## Test plan

- [x] Manually tested on a Tuya smart plug — calling the service updates energy sensor entities immediately
- [x] Verified that without `nowait=True`, updates only happen every 30s
- [x] Verified that without `_api_lock`, calls fail intermittently
- [ ] Verify behavior with sub-devices (CID)
- [ ] Verify error handling when device is offline

### DIFF
--- a/custom_components/tuya_local/__init__.py
+++ b/custom_components/tuya_local/__init__.py
@@ -66,7 +66,9 @@ async def async_setup(hass: HomeAssistant, config: dict):
 
         try:
             async with device._api_lock:
-                await hass.async_add_executor_job(device._api.updatedps, dps)
+                await hass.async_add_executor_job(
+                    device._api.updatedps, dps, True
+                )
         except Exception as e:
             _LOGGER.warning("update_dps failed for %s: %s", dev_id, e)
 
--- a/custom_components/tuya_local/__init__.py
+++ b/custom_components/tuya_local/__init__.py
@@ -64,7 +64,11 @@ async def async_setup(hass: HomeAssistant, config: dict):
         if dps is None:
             dps = device._force_dps or [18, 19, 20]
 
-        await hass.async_add_executor_job(device._api.updatedps, dps)
+        try:
+            async with device._api_lock:
+                await hass.async_add_executor_job(device._api.updatedps, dps)
+        except Exception as e:
+            _LOGGER.warning("update_dps failed for %s: %s", dev_id, e)
 
     hass.services.async_register(
         DOMAIN,
--- a/custom_components/tuya_local/__init__.py
+++ b/custom_components/tuya_local/__init__.py
@@ -9,10 +9,13 @@ https://github.com/codetheweb/tuyapi/issues/31.
 
 import logging
 
+import voluptuous as vol
+
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_HOST
-from homeassistant.core import HomeAssistant, callback
-from homeassistant.exceptions import ConfigEntryNotReady
+from homeassistant.core import HomeAssistant, ServiceCall, callback
+from homeassistant.exceptions import ConfigEntryNotReady, HomeAssistantError
+from homeassistant.helpers import config_validation as cv
 from homeassistant.helpers.entity_registry import (
     async_get as async_get_entity_registry,
 )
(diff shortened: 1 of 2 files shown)


make-all:
I intentionally designed this integration to be an easy to use alternative to the low level DIY approach of localtuya. Low level protocol functionality like this is not part of that design, and complicates the integration unnecessarily.

facuparedes:
Fair point, and I appreciate the clear design intent.

That said, here's the use case that brought me here: I have a smart plug monitoring a foam maker. I want to detect when it finishes (current drops below a threshold) and auto-shut off. The 30s passive polling makes this unreliable — sometimes it catches the drop, sometimes it doesn't. And having the polling run 24/7 when the plug is off most of the time feels wasteful.

I looked at localtuya for this, but `rospogrigio/localtuya` is basically dead (1,400 open issues, patches only when HA breaks it), and `xZetsubou/hass-localtuya` reviews PRs every few months at best. tuya-local is where things are heading for local Tuya control, which is why I'm here.

I get that exposing raw DP indexes doesn't fit your philosophy. A few alternatives that might:

- A **`tuya_local.refresh_entity`** service — user passes an entity_id, the integration figures out what to poll internally. No protocol details leak to the user.

- Making the **`force` flag configurable from the UI** — a toggle in device options, so users don't have to edit YAML device configs to get energy data refreshed actively.

- A **configurable poll interval** per device — let users pick 5s/10s/30s/60s from device settings instead of the hardcoded 30s.

Any of these would solve the problem without breaking the "easy to use" contract. Happy to rework the PR if one of these directions works for you.

make-all:
Get a better smartplug that is up to the task. Only a small number of models act like this.

facuparedes:
I understand the decision, thanks for your time. For context though — the plug works fine. It's the Tuya protocol that only pushes energy data on significant changes, which is standard behavior across most budget models. That's exactly why updatedps (0x12) exists as a command. Here's [one example](https://community.home-assistant.io/t/power-consumption-of-tuya-smart-plug-not-automatically-updating/374681), but a quick Google search shows many more users hitting the same limitation across different plug brands.

Anyway, thanks for reviewing the PR. In the meantime, I'll keep my fork synced with upstream with this feature implemented.

Issue #4799, opened 2026-04-01
Request support for Ecostrad iQ Ceramic Electric Radiators


bluefroguk1:
### Log message

```text
Partial cloud device spec: [{"id": 1, "name": "switch", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 2, "name": "mode", "type": "Enum", "format": "{\"range\":[\"hot\"]}", "enumMap": {}}, {"id": 3, "name": "temp_set", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":70,\"max\":300,\"scale\":1,\"step\":5}", "enumMap": {}}, {"id": 24, "name": "temp_current", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":0,\"max\":1000,\"scale\":1,\"step\":10}", "enumMap": {}}, {"id": 27, "name": "temp_correction", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":-5,\"max\":5,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 40, "name": "child_lock", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 42, "name": "countdown_left", "type": "Integer", "format": "{\"unit\":\"min\",\"min\":0,\"max\":1440,\"scale\":0,\"step\":1}", "enumMap": {}}]

Device matches zigbee_dv7vcs_human_presence_sensor with quality of 31%. LOCAL DPS: {"updated_at": 1775075524.9255986, "1": true, "2": "hot", "3": 205, "24": 137, "27": 0, "32": "heating", "40": false, "42": 60, "101": false, "102": 10, "103": 210, "104": "0", "105": "15", "106": "1", "107": "0", "108": "0", "109": "0", "110": "1", "111": 200, "112": 140, "113": 70, "117": false, "118": 2283, "119": 14796, "120": 6311, "121": 20}
Device matches zigbee_dv7vcs_human_presence_sensor with quality of 31%. LOCAL DPS: {"updated_at": 1775076146.0016696, "1": true, "2": "hot", "3": 205, "24": 143, "27": 0, "32": "heating", "40": false, "42": 52, "101": true, "102": 10, "103": 260, "104": "0", "105": "15", "106": "1", "107": "0", "108": "0", "109": "0", "110": "1", "111": 200, "112": 140, "113": 70, "117": false, "118": 2261, "119": 14531, "120": 6213, "121": 20}
```

[...]

PR #4950, opened 2026-04-23
Add device configuration for Ecostrad iQ Ceramic WiFi Controlled Electric Radiators

bluefroguk1:
Add support for Ecostrad iQ Ceramic WiFi electric radiator

Climate
Maps DP 1 to hvac_mode with explicit on/off to heat / off.
Uses DP 3 as target temperature (temperature) with the correct Tuya scale 1/10 range (70–300 raw → 7.0–30.0 °C) and 0.5 °C steps.
Uses DP 24 for current temperature with the same scaling.
Maps DP 2 to preset mode: Tuya hot → heat, and Program / program → program (no separate Eco mapping).
Optional DP 32 for HVAC action when the device reports heating / warm.
Other entities (moved from raw climate attributes to dedicated entities where appropriate)
Lock on DP 40 (child lock).
Number on DP 27 (temperature calibration, −5 to +5 °C).
Sensor on DP 42 (countdown / time remaining, minutes).
Secondary DPs are marked optional where suitable so the device can still be detected when polling is partial.


bluefroguk1:
Relates to https://github.com/make-all/tuya-local/issues/4799

bluefroguk1:
Will resubmit with corrected errors

Issue #4802, opened 2026-04-02
Request support for Ecostrad iQ WiFi Element (Towel Rail)


bluefroguk1:
### Log message

```text
Device matches rumba_bathroom_heater with quality of 55%. LOCAL DPS: {"updated_at": 1775150592.9996524, "1": true, "2": "auto", "16": 70, "24": 177, "27": 0, "32": "heating", "40": false, "102": 700, "103": 0, "104": "off", "106": "fro"}

Partial cloud device spec: [{"id": 1, "name": "switch", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 2, "name": "mode", "type": "Enum", "format": "{\"range\":[\"hot\",\"eco\",\"auto\"]}", "enumMap": {}}, {"id": 16, "name": "temp_set", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":70,\"max\":300,\"scale\":1,\"step\":5}", "enumMap": {}}, {"id": 24, "name": "temp_current", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":0,\"max\":1000,\"scale\":1,\"step\":10}", "enumMap": {}}, {"id": 27, "name": "temp_correction", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":-5,\"max\":5,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 40, "name": "child_lock", "type": "Boolean", "format": "{}", "enumMap": {}}]
```

### Product ID

keq3x2tw6plbbipm

### Product Name

Ecostrad iQ WiFi Element or Ecostrad Fina-E iQ WiFi Electric Towel Rail

### DPS information

```text
{
    "result": [
        {
            "code": "switch",
            "value": true
        },
        {
            "code": "mode",
            "value": "auto"
        },
        {
            "code": "temp_set",
            "value": 70
        },
        {
            "code": "temp_current",
            "value": 177
        },
        {
            "code": "temp_correction",
            "value": 0
        },
        {
            "code": "child_lock",
            "value": false
        }
    ],
    "success": true,
    "t": 1775147584747,
    "tid": "9f93fab92eb111f1a34cee5ee6a19de0"
}

{
    "result": {
        "active_time": 1775136539,
        "biz_type": 0,
        "category": "wk",
        "create_time": 1758198160,
[...]

PR #4951, opened 2026-04-23
Add Support for Ecostrad IQ Ceramic Radiator & Wifi Element

bluefroguk1:
Add support for Ecostrad iQ Ceramic WiFi electric radiator

Climate
Maps DP 1 to hvac_mode with explicit on/off to heat / off.
Uses DP 3 as target temperature (temperature) with the correct Tuya scale 1/10 range (70–300 raw → 7.0–30.0 °C) and 0.5 °C steps.
Uses DP 24 for current temperature with the same scaling.
Maps DP 2 to preset mode: Tuya hot → heat, and Program / program → program (no separate Eco mapping).
Optional DP 32 for HVAC action when the device reports heating / warm.
Other entities (moved from raw climate attributes to dedicated entities where appropriate)
Lock on DP 40 (child lock).
Number on DP 27 (temperature calibration, −5 to +5 °C).
Sensor on DP 42 (countdown / time remaining, minutes).
Secondary DPs are marked optional where suitable so the device can still be detected when polling is partial.

Relates to https://github.com/make-all/tuya-local/issues/4799


bluefroguk1:
also relates to https://github.com/make-all/tuya-local/issues/4802

custom_components/tuya_local/devices/ecostrad_iq_ceramic_radiator.yaml
```diff
@@ -0,0 +1,83 @@
+name: Ecostrad iQ Ceramic Radiator
```
make-all: Top level name should be without branding

custom_components/tuya_local/devices/ecostrad_iq_ceramic_radiator.yaml
```diff
@@ -0,0 +1,83 @@
...
+products:
+  - id: g8szwifd3zgozoxt
+    name: Ecostrad iQ Ceramic WiFi Controlled Electric Radiator
```
make-all: This should be split into manufacturer and model.
Generic device type info that is duplicating the name above can be left out.

custom_components/tuya_local/devices/ecostrad_iqceramicv2_radiator.yaml
```diff
@@ -0,0 +1,83 @@
+name: Ecostrad iQ Ceramic Radiator
+products:
+  - id: g8szwifd3zgozoxt
+    name: Ecostrad iQ Ceramic WiFi Controlled Electric Radiator
+entities:
+  - entity: climate
+    translation_only_key: heater
```
make-all: Use translation_key for new configs as a rule.

translation_only_key is mainly for maintaining backwards compatibliity with entity ids that are already released.

custom_components/tuya_local/devices/ecostrad_iqceramicv2_radiator.yaml
```diff
@@ -0,0 +1,83 @@
...
+          - scale: 10
+            step: 5
+      - id: 2
+        name: preset_mode
+        type: string
+        optional: true
+        mapping:
+          - dps_val: hot
+            value: heat
+          - dps_val: Program
```
make-all, line 38: Why are there two variants of program here? If only supporting one device, only one should be needed, otherwise there needs to be some way to select which one applies.

custom_components/tuya_local/devices/ecostrad_iq_wifi_element.yaml
```diff
@@ -0,0 +1,92 @@
+name: Ecostrad iQ WiFi Element
```
make-all: As above, top level name should be without branding.

custom_components/tuya_local/devices/ecostrad_iq_wifi_element.yaml
```diff
@@ -0,0 +1,92 @@
...
+          - scale: 10
+      - id: 32
+        type: string
+        name: hvac_action
+        mapping:
+          - dps_val: heating
+            value: heating
+          - dps_val: noheat
+            value: idle
+          - dps_val: cooling
```
make-all: Is this for "cooling down"? Maybe "off" is better in that case, as idle implies that it will switch back to heating if the temperature drops.

Or, if this is for a cooling mode where cold water is run through the radiator, then it should indicate `cooling`

make-all (changes requested):

custom_components/tuya_local/devices/ecostrad_iq_ceramic_radiator.yaml
bluefroguk1: Corrected

bluefroguk1:
Thanks for the comments - all fixed (hopefully) 
If any other changes are needed let me know

PR #4957, opened 2026-04-25
ble_water_valve use number instead of time for irrigation time/duration

trasbd:
Irrigation time was set as a `time` (timestamp) instead of a duration. When not using a 24-hour clock, it becomes clear the [time entity](https://www.home-assistant.io/integrations/time/) represents a time of day, not a time span. This PR changes the entity to a `number` with scaling/steps to convert seconds to minutes, matching the Tuya app interface. The entity name has also been updated to clarify intent. Several other devices also use `number` rather than `time` to represent duration.

If `time` is the intended behavior for DP11, some products using this device would likely need a separate configuration.

Resolves issue from #3718

### DIFF
--- a/custom_components/tuya_local/devices/ble_water_valve.yaml
+++ b/custom_components/tuya_local/devices/ble_water_valve.yaml
@@ -101,17 +101,23 @@ entities:
             value: "48h"
           - dps_val: "72h"
             value: "72h"
-  - entity: time
-    name: Irrigation time
+  - entity: number
+    class: duration
+    name: Irrigation duration
     category: config
     dps:
       - id: 11
         type: integer
-        name: second
+        name: value
+        unit: min
         optional: true
+        precision: 0
         range:
           min: 0
           max: 86400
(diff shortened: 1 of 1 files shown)


trasbd:
the file, from commit e100d18, that failed the check is not related to this PR

make-all:
You can set your user settings to 24h if you need to use these entities.

number is not a good substitute for entering times to minute or second resolution.

trasbd:
Reguardless if its formatted correctly, time isnt the correct context for this entity. It should be a number/duration for statistic purposes too. If a user wants to view the average amount of time the valve is open for a given period Home Assistant can and will automatically generate that.

It also causes confusion of intent. Is this the time the valve will turn on? Or is it the time it will turn off? Neither convey that its actually the duration.

Someone who normally uses 12-hour clocks shouldn't have to change their user settings to 24-hours for 1 entity nor should they have to enter 12:30AM to set a 30 minute timer.

PR #4962, opened 2026-04-26
Remove duplicate and add hidden entity

trasbd:
sop10_sprinkler is the same device as ble_water_valve. ID and model were already included in ble_water_valve

_**Leaves the exsisting `Irrigation time` as timestamp for ease of entering minutes and seconds.**_ Addes a **_hidden_** entity, `Irrigation duration`, with the correct representation of the value. Can be enabled by the user for statistics and other purposes.

Follow up to #4957


make-all:
You can't just remove config files that other users have already configured.
I don't know how the SOP10 ended up both listed in the generic config and getting its own config, but now that it has happened it is not an excuse to piggyback your previously rejected change onto.

trasbd:
This was a different solution to the same issue, glad you took a look.

PR #4966, opened 2026-04-26
feat(devices): add Breville Allrounder

elcarabo:
Add device definition for Breville Airrounder air purifier and heater

- Model: LPH408
- Tuya Product ID: 56ssixusjwkhp6yg
- Tested with Home Assistant 2026.4.4 and tuya-local v2026.4.2
- Based on Breville Airrounder Plus config
- Updated modes as fan and heat settings needed further constraints.


custom_components/tuya_local/devices/breville_airrounder_heater.yaml
```diff
@@ -0,0 +1,142 @@
+name: Breville AirRounder
```
make-all: Top level name should be without branding. Other devices may also match.

custom_components/tuya_local/devices/breville_airrounder_heater.yaml
```diff
@@ -0,0 +1,142 @@
...
+products:
+  - id: 56ssixusjwkhp6yg
+    manufacturer: Breville
+    model: LPH408
```
make-all: AirRounder should be part of the model

custom_components/tuya_local/devices/breville_airrounder_heater.yaml
```diff
@@ -0,0 +1,142 @@
...
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: heat_low
+                value: fan_only
+              - dps_val: heat_high
+                value: heat
```
make-all, line 34: There are two heat settings here. One needs to be marked as `hidden: true` to inform the integration which one not to use when the user selects `heat`.

custom_components/tuya_local/devices/breville_airrounder_heater.yaml
```diff
@@ -0,0 +1,142 @@
...
+        unit: C
+      - id: 3
+        type: integer
+        name: current_temperature
+      - id: 4
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: heat_high
+            value: "Low Heat"
```
make-all: It is better to use standard presets defined in the HA translation files.
`eco`, `comfort` or `boost`, `none`, for example.

Or you can use the presets defined for the heater translation key in this integration: `low`, `high`, `fan_only`

custom_components/tuya_local/devices/breville_airrounder_heater.yaml
```diff
@@ -0,0 +1,142 @@
...
+          - dps_val: heat_0
+            value: "High Heat"
+          - dps_val: heat_low
+            value: "Fan Only"
+            hidden: true
```
make-all: Why is this hidden? It should be OK to leave it exposed, otherwise it will seem strange that the device is sometimes in this preset, but it cannot be selected directly (instead the hvac_mode needs to be changed).

custom_components/tuya_local/devices/breville_airrounder_heater.yaml
```diff
@@ -0,0 +1,142 @@
...
+        name: fan_mode
+        mapping:
+          - dps_val: NatureWind_Low
+            value: low
+          - dps_val: NatureWind_High
+            value: medium
+          - dps_val: CoolWind_0
+            value: high
+          - dps_val: SleepWind
+            value: auto
```
make-all, line 84: auto, or sleep?

make-all (changes requested):

elcarabo:
Further updates made. My preference was to keep one climate UI rather than having them split like in the AirRounder Plus setup. It uses fan speed when in fan only mode, then swaps to presets for low and high heat. Added further checks so it swaps to the other mode automatically if selected.

elcarabo:
Fan mode:
Screenshot 2026-07-14 000204
Heat:
Screenshot 2026-07-14 000218

PR #4967, opened 2026-04-26
Add Krisbow Sync Dehumidifier 36W

lboy:
Adds support for Krisbow Sync 0.4L Dehumidifier 36W (Tuya product_id: ymfznwejbhviuhjk)
  https://www.ruparupa.com/p/krisbow-sync-0-4-ltr-dehumidifier-36-watt-putih.html
  Sold in Indonesia at Ruparupa. DPS verified from local diagnostics.
  Entities: humidifier (dehumidifier), temperature sensor, humidity sensor,
  sleep switch, water tank reminder, indicator light with colour modes.

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -419,7 +419,8 @@
 - Skyfan DC fan with light
 - Smart Mist3 TX-1602MF (ZJ-1522A-WiFi)
 - Smartmi Air Circulating fan
-- Stirling FS1-40DC pedestal fan
+- Stirling FS1-40DC pedestal fan
+
 - Sulion Crixus L ceiling fan with light
 - Temple and Webster Alina ceiling fan
 - TMWF02 fan controller
@@ -522,6 +523,7 @@
 - Vivosun DE0003 dehumidifier
 - Wellio D008A 20L dehumidifier
 - Woods MRD25GW and WDD90 dehumidifiers
+- Krisbow Sync Dehumidifier 36W
 
 ### Humidifiers
 
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -504,6 +504,7 @@
 - Juro-Pro 2006 dehumidifier
 - Klarstein DryFy Pro Connect dehumidifier
 - Kogan SmarterHome 7L desiccant dehumidifier
+- Krisbow Sync Dehumidifier 36W
 - Luko dehumidifier
 - Meaco DD8L Pro dehumidifier
 - MeacoDry Arete Two 10L dehumidifier
@@ -523,7 +524,6 @@
 - Vivosun DE0003 dehumidifier
 - Wellio D008A 20L dehumidifier
 - Woods MRD25GW and WDD90 dehumidifiers
-- Krisbow Sync Dehumidifier 36W
 
 ### Humidifiers
 
--- a/custom_components/tuya_local/devices/krisbow_sync_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/krisbow_sync_dehumidifier.yaml
@@ -48,16 +48,16 @@ entities:
         name: switch
       - id: 31
         type: string
-        name: effect
+        name: named_color
         mapping:
           - dps_val: "light_red"
             value: red
           - dps_val: "light_green"
-            value: green
+            value: lime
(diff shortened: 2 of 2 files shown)


custom_components/tuya_local/devices/krisbow_sync_dehumidifier.yaml
```diff
@@ -0,0 +1,109 @@
+name: Krisbow Sync Dehumidifier 36W
```
make-all: Top level name should be without branding.
Other models may also match this config.

custom_components/tuya_local/devices/krisbow_sync_dehumidifier.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
+        unit: "°C"
+
+  - entity: sensor
```
make-all: Duplication of the humidifier current_humidity is not required

custom_components/tuya_local/devices/krisbow_sync_dehumidifier.yaml
```diff
@@ -0,0 +1,109 @@
...
+        unit: "%"
+
+  - entity: switch
+    translation_key: sleep
+    dps:
+      - id: 14
+        type: boolean
+        name: switch
+
+  - entity: select
```
make-all: This can just be a switch, with the mapping being to true and false.

custom_components/tuya_local/devices/krisbow_sync_dehumidifier.yaml
```diff
@@ -0,0 +1,109 @@
...
+        type: string
+        name: option
+        mapping:
+          - dps_val: "on"
+            value: "on"
+          - dps_val: "off"
+            value: "off"
+
+  - entity: light
+    name: Indicator Light
```
make-all: Use `translation_key: indicator`

make-all: Probably also add `category: config`, since this is a config option.

custom_components/tuya_local/devices/krisbow_sync_dehumidifier.yaml
```diff
@@ -0,0 +1,109 @@
...
+    dps:
+      - id: 29
+        type: boolean
+        name: switch
+      - id: 31
+        type: string
+        name: effect
+        mapping:
```
make-all, line 52: Most of these are colors, so should be under `named_color` rather than effect.

custom_components/tuya_local/devices/krisbow_sync_dehumidifier.yaml
```diff
@@ -0,0 +1,109 @@
...
+            value: Purple
+          - dps_val: "light_warm_yellow"
+            value: Warm Yellow
+          - dps_val: "light_orange"
+            value: Orange
+          - dps_val: "colorful"
+            value: Colorful
+  - entity: select
+    translation_key: timer
+    icon: "mdi:timer-outline"
```
make-all: It should not be necessary to override the icon here.

custom_components/tuya_local/devices/krisbow_sync_dehumidifier.yaml
```diff
@@ -0,0 +1,109 @@
...
+    category: config
+    dps:
+      - id: 17
+        type: string
+        name: option
+        mapping:
+          - dps_val: "cancel"
+            value: "Off"
```
make-all: Use the original dps_val also in the value, as that is what the translations are mapped to (cancel, 1h, 2h etc).

DEVICES.md
```diff
@@ -522,6 +523,7 @@
 - Vivosun DE0003 dehumidifier
 - Wellio D008A 20L dehumidifier
 - Woods MRD25GW and WDD90 dehumidifiers
+- Krisbow Sync Dehumidifier 36W
```
make-all: This should be in alphabetical order within the dehumidifier section.

make-all (changes requested):

PR #4968, opened 2026-04-26
Add support for Voldt 8-13A EV charger

fmckeogh:
Possibly WIP, as there are still some issue but I'm not sure if they're blockers.

1. Charge energy (17) should be user-writeable for use in the fixed charging mode.
2. Total energy (1) is always zero even after a long charge, but it's also zero even on the device debugging Tuya developer page.

Setting current, starting/stopping charges, viewing voltage/power/current/temperature/SW version all work.


custom_components/tuya_local/devices/voldt_ev_charger.yaml
```diff
@@ -0,0 +1,277 @@
...
+  # "Remaining available power"
+  - entity: sensor
+    name: Balance energy
+    dps:
+      - id: 15
+        type: integer
+        optional: true
+        name: sensor
+        unit: kWh
+        class: total_increasing
```
make-all: I don't think this is correct. You don't want this to keep increasing, in normal usage it should go down.

custom_components/tuya_local/devices/voldt_ev_charger.yaml
```diff
@@ -0,0 +1,277 @@
...
+        mapping:
+          - scale: 100
+  - entity: switch
+    icon: "mdi:ev-station"
+    dps:
+      - id: 18
+        type: boolean
+        name: switch
+  - entity: sensor
+    name: Hardware ID
```
make-all: This will be static, I don't think it needs a sensor, it can just appear as an attribute on another entity (status, perhaps)

custom_components/tuya_local/devices/voldt_ev_charger.yaml
```diff
@@ -0,0 +1,277 @@
...
+    category: diagnostic
+    dps:
+      - id: 22
+        type: string
+        optional: true
+        name: sensor
+  - entity: sensor
+    name: Software version
```
make-all: This will also be mostly static info

custom_components/tuya_local/devices/voldt_ev_charger.yaml
```diff
@@ -0,0 +1,277 @@
...
+    name: Last charge
+    dps:
+      - id: 25
+        type: integer
+        name: sensor
+        unit: kWh
+        class: measurement
+        mapping:
+          - scale: 100
+  - entity: binary_sensor
```
make-all: Often these dps are actually read-write, and used to control the update frequency from the device.

Generally when put into "online" mode, the device will start reporting its sensors frequently, and after a time (10 minutes, 30 minutes etc), it will automatically switch back to "offline". So making it a button with "online" mapped to true (the press action) allows an automation to "press" the button to get high frequency updates.

custom_components/tuya_local/devices/voldt_ev_charger.yaml
```diff
@@ -0,0 +1,277 @@
...
+  #       type: base64
+  #       name: sensor
+  #       optional: true
+  #       unit: kW
+  #       mask: "000000000000FFFF"
+  #       mapping:
+  #         - scale: 1000
+
+  - entity: sensor
+    name: Power
```
make-all: Should be `class: power`
"Single phase power" and the commented one above can be rolled in as attributes, as in a single phase device they should all read the same value. If you observe for a while, you will see which gives the most reliable result, and that one should be the primary "sensor" dp, with the others as attributes (optional if needed).

make-all (changes requested):

custom_components/tuya_local/devices/voldt_ev_charger.yaml
fmckeogh: I've added both software and hardware IDs to status, but I can't see them in Home Assistant?

custom_components/tuya_local/devices/voldt_ev_charger.yaml
fmckeogh: Changed to `class: measurement` :)

custom_components/tuya_local/devices/voldt_ev_charger.yaml
fmckeogh: Fixed, I removed the other one as it was always reporting zero anyway

custom_components/tuya_local/devices/voldt_ev_charger.yaml
fmckeogh: How do I do this? I couldn't find an example in other device YAMLs

custom_components/tuya_local/devices/voldt_ev_charger.yaml
make-all: They have buried extra attributes deeper in the UI a couple of releases ago. If you go into the entity more info dialog, it is still available under Details in the menu.

custom_components/tuya_local/devices/voldt_ev_charger.yaml
make-all: There is an example here:
https://github.com/make-all/tuya-local/blob/main/custom_components/tuya_local/devices/afyeev_16a_evcharger.yaml#L274

custom_components/tuya_local/devices/voldt_ev_charger.yaml
fmckeogh: Thanks!

custom_components/tuya_local/devices/voldt_ev_charger.yaml
fmckeogh: Fixed

fmckeogh:
I think there's a typo in the heading revert ("Battery" not "Bettery"), I changed it to EV because all the devices listed were for EVs, while I thought a battery charger would be like for rechargeable AAs, or power tool battery chargers?

make-all:
Not all of the devices are EV chargers, there are also tool chargers and general battery chargers listed.

fmckeogh:
Fair enough:) In that case I think that's all the suggestions fixed.

PR #4969, opened 2026-04-26
Add NovaDry 12 Smart Dehumidifier configuration

uhodav:
Add support for NovaDry 12 Smart dehumidifier

**Summary**

Adds a Tuya Local device configuration for the NovaDry 12 Smart dehumidifier.

**Tested device**

- Manufacturer: NovaDry
- Model: 12 Smart
- Product ID / productKey: edlv0juyoj8j0tdc
- Protocol: 3.4
- Setup: manually added using IP, device ID, and local key

**Known datapoints**

- DP 1 — Power switch (on/off)
- DP 3 — Target humidity (string enum: "30"–"80", step 5)
- DP 4 — Fan speed ("1" = low, "2" = high)
- DP 5 — Mode (dehumidify, drying, continuous)
- DP 6 — Current humidity (read-only integer)
- DP 10 — Ionizer (on/off)
- DP 16 — Child lock
- DP 17 — Timer (0h/1h/2h/3h)

**Usage notes**

- DP 3 uses string values ("30"–"80"), not integers
- DP 5 mode "drying" maps to the HA "laundry" mode

**Validation**

- Tested locally with Tuya Local
- Power on/off, target humidity, mode selection work correctly
- Fan speed, ionizer, child lock, and timer work correctly
- Current humidity reads correctly
- No private data (keys, IDs, IPs) included


custom_components/tuya_local/translations/en.json
```diff
@@ -381,6 +381,8 @@
                 "state_attributes": {
                     "mode": {
                         "state": {
+                            "drying": "Drying",
```
make-all: Isn't this just laundry mode under a different name?

custom_components/tuya_local/translations/en.json
```diff
@@ -381,6 +381,8 @@
...
+                            "dehumidify": "Dehumidify",
```
make-all, line 384: Isn't this just "normal" mode from the HA humidifier translations, under a different name?

custom_components/tuya_local/devices/novadry_12smart_dehumidifier.yaml
```diff
@@ -0,0 +1,101 @@
...
+            value: 80
+        range:
+          min: 30
+          max: 80
+      - id: 5
+        name: mode
+        type: string
+        mapping:
+          - dps_val: dehumidify
+            value: dehumidify
```
make-all, line 48: `normal` would fit here without defining a new translation

tests/devices/test_novadry_12smart_dehumidifier.py
make-all, line 1: Device specific tests are not needed, and do not generally work any more due to changes in HA's initialisation.

make-all (commented):
Best approach would be to separate Ukrainian translation improvements from the rest of the change, and undo the other translation file changes since they are not necessary.

tests/devices/test_novadry_12smart_dehumidifier.py
uhodav: tests have been removed

custom_components/tuya_local/devices/novadry_12smart_dehumidifier.yaml
uhodav: The mode is called Dehumidify
image

custom_components/tuya_local/translations/en.json
uhodav: image

custom_components/tuya_local/translations/en.json
uhodav: This word has different meanings in other languages.
1. Washing clothes (washing in a washing machine)
2. Drying clothes (removing moisture from clothes)

Therefore, I made a separate translation for the dehumidifier.

uhodav:
@make-all 
OK

1. Deleted the tests
2. Moved the translations to a [new pull request](https://github.com/make-all/tuya-local/pull/4981)
3. Left only one new property in the translations.

custom_components/tuya_local/translations/en.json
make-all: Since the translation is for dehumidifiers, there is no way to interpret this as wash clothes. Use the translation key that is available, and if the translation is unsuitable for a specific language, the translation should be changed.

custom_components/tuya_local/translations/en.json
make-all: Dehumidify is normal for dehumidifiers. Please use the existing translations where possible

make-all:
Merge with existing Vibrix Dryflow 12 Pro config

PR #4970, opened 2026-04-26
Add Olight Ostation 2 charger config

locoscientist:
Adds a device configuration YAML for the Olight Ostation 2 / Ostation 2 Pro battery charger. This is a config-only contribution using known Tuya datapoints DP101, DP102, DP108, DP110, and optional schedule controls DP111/DP112. No Python code changes are included.

### DIFF
--- a/custom_components/tuya_local/devices/olight_ostation2_charger.yaml
+++ b/custom_components/tuya_local/devices/olight_ostation2_charger.yaml
@@ -70,7 +70,7 @@ entities:
         type: string
         name: sensor
         optional: true
-        
+
   - entity: switch
     name: Pause schedule
     dps:
@@ -86,7 +86,7 @@ entities:
         type: string
         name: sensor
         optional: true
-        
+
   - entity: sensor
     name: DP113
     category: diagnostic
@@ -94,4 +94,3 @@ entities:
       - id: 113
         type: integer
         name: sensor
- 
--- a/custom_components/tuya_local/devices/olight_ostation2_charger.yaml
+++ b/custom_components/tuya_local/devices/olight_ostation2_charger.yaml
@@ -1,0 +1,97 @@
+name: Charger
+products:
+  - id: 1fnensm6
+    manufacturer: Olight
+    model: Ostation2
+  - id: v8znljza
+    manufacturer: Olight
+    model: Ostation2
+
+entities:
+  - entity: switch
+    name: Power
+    dps:
+      - id: 101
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/olight_ostation2_charger.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - id: 1fnensm6
+    manufacturer: Olight
+    model: Ostation2
+  - id: v8znljza
+    manufacturer: Olight
+    model: Ostation2
+
+entities:
+  - entity: switch
+    name: Power
```
make-all, line 12: name not needed

custom_components/tuya_local/devices/olight_ostation2_charger.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - entity: sensor
+    name: Fully charged count
+    category: diagnostic
+    dps:
+      - id: 102
+        type: integer
+        name: sensor
+
+  - entity: binary_sensor
+    name: DP103
```
make-all, line 27: Unknown dps should not be entities by themselves, make it an attribute on another entity so it can be observed.

custom_components/tuya_local/devices/olight_ostation2_charger.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+  - entity: sensor
+    name: Invalid battery count
+    category: diagnostic
+    dps:
+      - id: 108
+        type: integer
+        name: sensor
+  - entity: sensor
+    name: Raw Bay Telemetry
```
make-all, line 66: Follow HA naming convention of only first word capitalised.
Make this `hidden: true`, as it will not be useful without knowing how to decode it.
If the type is listed as raw in the Tuya docs, it is most likely base64 encoded, so a type of base64 prepares for any further decoding in future.

custom_components/tuya_local/devices/olight_ostation2_charger.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+  - entity: switch
+    name: Pause schedule
+    dps:
+      - id: 111
+        type: boolean
+        name: switch
+
+  - entity: sensor
+    name: Pause Window Raw
```
make-all, line 82: same comments as above

make-all (changes requested):

PR #4971, opened 2026-04-26
Add iPettie FW54 feeder config

locoscientist:
## Summary

Adds a Tuya Local device configuration for the iPettie FW54 food/water feeder.

## Tested device

- Manufacturer: iPettie
- Model tested: FW54
- Product ID / productKey: `tejdgn0tuyiq6vmt`
- Protocol/interface used: `3.5`

## Setup notes

The device uses protocol/interface `3.5`.

The device was added using protocol `3.5`.

## Known datapoints

- `3`: Manual feed / portion command
- `4`: Device status
- `13`: Problem / error indicator
- `111`: Water light switch

## Usage notes

- DP `111` controls the water light as a normal switch.
- DP `3` is used for manual feeding by writing the desired portion count.
- DP `4` reports feeder status. In my testing this is useful for detecting states such as feeding/idle.
- DP `13` reports a problem/error condition. In my testing, `0` means no problem and non-zero indicates a problem state.

In Home Assistant, I use an input number for the desired portion count and a script/button to send that value to the manual feed command.

## Validation

- Tested locally with Tuya Local
- Device manually added using IP, device ID, protocol `3.5`, and recovered local key
- Water light switch works
- Manual feed command works
- Status and problem entities report correctly
- No local keys, device IDs, IP addresses, or private account data are included


custom_components/tuya_local/devices/ipettie_fw54_petfeeder.yaml
```diff
@@ -0,0 +1,43 @@
+name: Pet feeder
+products:
+  - id: tejdgn0tuyiq6vmt
+    manufacturer: iPettie
+    model: FW54
+entities:
+  - entity: switch
+    name: Water light
```
make-all: Use a light entity.

custom_components/tuya_local/devices/ipettie_fw54_petfeeder.yaml
```diff
@@ -0,0 +1,43 @@
...
+    icon: "mdi:water-outline"
+    dps:
+      - id: 111
+        type: boolean
+        name: switch
+  - entity: number
+    name: Manual feed command
```
make-all: Use `translation_key: manual_feed` instead of this name and icon

custom_components/tuya_local/devices/ipettie_fw54_petfeeder.yaml
```diff
@@ -0,0 +1,43 @@
...
+    dps:
+      - id: 3
+        type: integer
+        name: value
+        range:
+          min: 1
+          max: 15
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
```
make-all, line 24: In addition to the true/false sensor, the original info should be retained in a `fault_code` attribute.

The type is most likely `bitfield` rather than `integer`, but the distinction only matters when you start pulling out non-zero values.

custom_components/tuya_local/devices/ipettie_fw54_petfeeder.yaml
```diff
@@ -0,0 +1,43 @@
...
+            value: false
+          - value: true
+  - entity: sensor
+    translation_key: status
+    icon: "mdi:paw"
+    category: diagnostic
+    dps:
+      - id: 4
+        type: string
+        name: sensor
```
make-all, line 43: It would be better to list all the possible values here and add `class: enum` to the sensor definition.

make-all (changes requested):

PR #4972, opened 2026-04-26
Add kettle preset selects

locoscientist:
## Summary

Adds optional DP106 and DP107 preset select entities to the existing Korex AX-WF306N kettle config.

## Why

The existing config already covers the core kettle functions.  The same product also exposes two additional preset controls:

- `106`: Boil preset
- `107`: Boil and cool preset

These are separate from the existing DP101 temperature control and appear to be app-style preset selections.

## Tested device

- Manufacturer/model: Korex AX-WF306N / Kogan smart kettle variant
- Product ID / productKey: `2vuvtw0cj8wmjrpc`
- Protocol/interface used locally: `3.3`

## Existing config

The existing Tuya Local config already supports this product ID as a kettle using DPs `1`, `5`, `10`, `101`, `102`, `103`, `104`, and `105`.

## Added datapoints

- `106`: Boil preset
- `107`: Boil and cool preset

## Validation

- Tested locally with LocalTuya-style DP mapping
- Existing power, current temperature, keep-warm, and status behavior work
- DP106 exposes boil preset options
- DP107 exposes boil-and-cool preset options
- No local keys, device IDs, IP addresses, or private account data are included

## Reference

Original Home Assistant community post documenting the LocalTuya setup and DP106/DP107 usage:

https://community.home-assistant.io/t/korex-ax-wf306n-smart-kettle-works-perfectly-with-localtuya/909674


custom_components/tuya_local/devices/korex_axwf306n_smart_kettle.yaml
```diff
@@ -204,3 +204,53 @@ entities:
       - id: 103
         type: boolean
         name: keep_warm
+  - entity: select
+    name: Boil preset
+    icon: "mdi:kettle"
```
make-all: Use `translation_key: kettle_mode`, and map the values below to kettle_mode values from the translation files: `infant_formula`, `green_tea`, `oolong_tea`, `coffee`, `boil`, `heat`, `keep_warm` etc.

If there is another setting to select between heating and boil and cooling, then they should be combined into a single select with that setting used as a condition to redirect. But if they are not, then I suggest that rather that "Boil preset", this would be better as "Heat preset", since the distinction between these is that the below one always boils the water first and lets it cool to the target temperature. So presumably this one just heats to the target temperature.

PR #4981, opened 2026-04-29
Uk translates

uhodav:
Corrections of translations for the Ukrainian language


DEVICES.md
make-all, line 1: Please remove this massive change from the changeset.

custom_components/tuya_local/icons.json
```diff
@@ -285,8 +285,10 @@
                 "state_attributes": {
                     "mode": {
                         "state": {
+                            "dehumidify": "mdi:water-off",
```
make-all: As per my comment on the other PR, I do not think this addition is necessary.

custom_components/tuya_local/translations/uk.json
```diff
@@ -364,23 +364,24 @@
...
-                "name": "Тревога"
+                "name": "Тривога"
             }
         },
         "humidifier": {
             "extended": {
                 "state_attributes": {
                     "mode": {
                         "state": {
+                            "dehumidify": "Осушити",
```
make-all: As with the comment on other PR, I do not think this addition is necessary. `normal` (part of standard HA translations) can be used instead.

make-all (changes requested):

custom_components/tuya_local/icons.json
uhodav: image

PR #4984, opened 2026-04-30
Document Smarter Pool auth helper

nicolas-albert:
## Summary
- Document smarterpool-auth as an option for Smarter Pool OEM app paired pool heat pumps
- Explain that it can retrieve Tuya device id, local key, DPS values and schema without re-pairing through Tuya Smart / Smart Life
- Add a warning not to share local key output publicly

## Validation
- git diff --check


make-all:
There are too many OEMs to be documenting the details of every single one here.

PR #4986, opened 2026-04-30
fix: multiple bugs for lights with packed/masked dps (e.g. dp 51)

### LINKED ISSUE #5034: [Bug] Ledvance Panel (ledvance_smart_plabl100x25b) becomes unavailable on light.turn_on in v2026.5.2
### Description

After updating from tuya-local v2026.1.2 to v2026.5.2, my Ledvance Smart Panels drop offline and become "unavailable" immediately when sending a turn_on command. 

In v1.2, I had to use a workaround: First turn on `switch.deckenlampe` (Master Switch DP 20), wait 500ms, then turn on `light.arbeitszimmer_hauptlicht`. This worked flawlessly. 
In v5.2, doing this (or just sending `light.turn_on` directly) causes the device connection to drop instantly.

**Device details:**
* **Type:** `ledvance_smart_plabl100x25b`
* **Protocol:** 3.3

**Suspected cause:**
Looking at the diagnostics, `force_dps: [51, 51, 51, 51, 51]` is heavily used. Could the new protocol handling in 5.x combined with `force_dps` overload the device's MCU when it wakes up from the DP 20 master switch? DP 20 is `false` when the light is off, which seems to turn off the actual light DPs completely.

### Steps to reproduce

1. Have a Ledvance Panel (`ledvance_smart_plabl100x25b`) configured.
2. Update tuya-local to version 2026.5.2.
3. Turn off the light via HA.
4. Try to turn the light back on via HA (`light.turn_on`).
5. Observe the device immediately dropping offline and becoming "unavailable".

### Expected behaviour

The light should turn on without the device dropping its connection or crashing the MCU, just as it was manageable in v2026.1.2.

### Additional context

I attached the JSON diagnostic file. I downgraded back to 2026.1.2 on my Home Assistant instance for now to keep the lights working.

[tuya_local-01KFBA16P65BBYQZNZEARDAGRT-Deckenlampe Arbeitszimmer-27b91ccd3c0db543bfc6f265a929d14c.json](https://github.com/user-attachments/files/27584861/tuya_local-01KFBA16P65BBYQZNZEARDAGRT-Deckenlampe.Arbeitszimmer-27b91ccd3c0db543bfc6f265a929d14c.json)

### Confirmation

### LINKED ISSUE #5054: Multiple bugs in masked/packed dp handling for lights (e.g. dp 51)
### Description

Four related bugs in the masked-dp code path, surfaced while setting up an Orison Chanfok Neo ceiling fan/light. The device exposes two HA light entities (main + RGB backlight) that share dp 51 — a packed base64 datapoint where each sub-function gets its own byte mask.

**1. Masked write fails when the dp has never been read.**
The first masked write to dp 51 raises `Cannot mask unknown current value`, because dp 51 isn't reported by the device until it's been written to. The masking code OR-s new bits onto the current value, but `None` has no bits.

**2. `light.turn_on(brightness=…)` from off leaves the light off.**
For packed dps that share an id across switch/brightness/color/effect, the brightness/color/effect branches populate `settings[dp_id]` first. The switch-on guard then skips writing because `id in settings` is already true. Brightness lands, the switch bit doesn't, the bulb stays dark.

**3. Scenes touching both sub-lights race on dp 51.**
`scene.turn_on` calls `async_reproduce_state` for each entity in parallel via `asyncio.gather`. Both entities do a read-modify-write of the shared dp; both reads see the same baseline before either's pending update lands, so the second write clobbers the first. One sub-light's bits are silently lost.

**4. `orison_chanfok_neo` template — `available` semantics on dp 20.**
The device template names dp 20 `available`, but dp 20 is also the device's master power switch and the device auto-toggles it off when both lights are off. That makes both light entities unavailable in HA, so HA won't even let you turn them back on.

### Steps to reproduce

- Bugs 1, 2, 4: add an Orison Chanfok Neo to HA and call `light.turn_on` with a brightness arg on either light entity from a cold-off state.
- Bug 3: create a scene that sets both `light.<device>_main` and `light.<device>_backlight`, then call `scene.turn_on`.

### Expected behaviour

- The first `light.turn_on` on a freshly added device should not raise.
- `light.turn_on(brightness=…)` from off should turn the light on at the requested brightness.
- Scene activation should reliably apply both sub-lights' state.
- The light entities should remain available regardless of dp 20's state.

### Additional context

Affected device config: `orison_chanfok_neo.yaml`. Bugs 1–3 are in the shared masked-dp code path and likely affect any device using packed dps for lights. Bug 4 is specific to this device template.

PR #4986 fixes all four (one commit per bug), with regression tests for #1 and #2 and end-to-end verification on real hardware.

### Confirmation

- [x] This report is for make-all/tuya-local, **not** for localtuya, a modified tuya-local, the Tuya app or cloud.
- [x] The device is externally powered (mains-wired ceiling fan/light).
- [x] I have checked the latest released version of tuya-local.
- [x] The device config filename (`orison_chanfok_neo`) is mentioned above.

williamoverton:
Fixes #5054.

Setting up an Orison Chanfok Neo ceiling fan/light surfaced four bugs in the masked-dp code path. The device exposes two HA light entities (main + RGB backlight) that share dp 51 — a packed base64 datapoint where each sub-function gets its own byte mask. One commit per fix.

1. **Masked write fails on a dp that's never been read.** dp 51 isn't reported until first write, so the initial write raises "Cannot mask unknown current value" — OR-ing onto None. Treat unknown as 0.

2. **`light.turn_on(brightness=…)` from off leaves the light off.** Packed dps share an id across switch/brightness/color/effect, so brightness lands in `settings[dp_id]` first. The switch-on branch then skips because `id in settings`, and the switch bit is never set. Allow the merge for masked dps; `get_values_to_set` OR-s the byte onto the pending value safely.

3. **Scenes touching both sub-lights race on dp 51.** `scene.turn_on` reproduces entities in parallel via `asyncio.gather`. Two read-modify-writes see the same baseline; the second clobbers the first. Per-device `asyncio.Lock` around the light set path.

4. **Orison template tweaks.** `force: true` on dp 51 so it polls from connect (makes #1's fallback a true edge case), and rename dp 20 `available` → `power`. dp 20 is the device's master switch, and the device auto-toggles it off when both lights are off — used as `available` that left both entities unavailable, so HA wouldn't let you turn them back on. The Lights master switch still works; the light's switch falls back to dp 20 via `value_mirror`. Happy to split this off if you'd rather keep the PR pure fix code.

Regression tests added for #1 and #2. Verified end-to-end on the device.


make-all:
The fix for #1 is problematic, as it assumes state which may not reflect reality. In your case it may not seem bad, but many other devices use multi-value  masked dps, and for some this may lead to critical bugs if the wrong state is assumed.

Since #4 negates the need for this for your device, I think it should be left out.

williamoverton:
Woops! Sorry for the headaches this caused! I can only say it works fine in my setup 😅

PR #4989, opened 2026-05-01
Add Honiture G20 and Q6 Pro Robot Vacuum and Mop device profiles 

starfire0606:
Adds device profiles for two Honiture robot vacuums with mopping capability.

**Honiture G20** (protocol 3.4):
Includes all three fan speeds (gentle, normal, strong), mopping water control, battery, brush/filter life sensors, carpet boost, carpet detection, seek robot, work mode, volume, and language. 

**Honiture Q6 Pro** (protocol 3.3):
Includes full robot_state status mapping, fan speeds (quiet, auto, strong, max), mopping water control, battery, brush life sensors, auto boost, mop detection, deep clean, room mode, do not disturb, and volume control.

Both tested on hardware.

### DIFF
--- a/custom_components/tuya_local/devices/honiture_g20_robot_vacuum_mop.yaml
+++ b/custom_components/tuya_local/devices/honiture_g20_robot_vacuum_mop.yaml
@@ -97,7 +97,6 @@ entities:
   - entity: sensor
     name: Edge brush life
     category: diagnostic
-    icon: "mdi:brush"
     dps:
       - id: 7
         type: integer
@@ -108,7 +107,6 @@ entities:
   - entity: sensor
     name: Roll brush life
     category: diagnostic
-    icon: "mdi:brush"
     dps:
       - id: 8
         type: integer
@@ -119,7 +117,6 @@ entities:
   - entity: sensor
     translation_key: filter_life
     category: diagnostic
-    icon: "mdi:air-filter"
     dps:
       - id: 9
         type: integer
@@ -130,7 +127,6 @@ entities:
   - entity: button
     name: Reset edge brush
     category: config
-    icon: "mdi:brush"
     dps:
       - id: 10
         type: boolean
@@ -139,7 +135,6 @@ entities:
   - entity: button
     name: Reset roll brush
     category: config
-    icon: "mdi:brush"
     dps:
       - id: 11
         type: boolean
@@ -148,7 +143,6 @@ entities:
   - entity: button
     name: Reset filter
     category: config
-    icon: "mdi:air-filter"
     dps:
       - id: 12
         type: boolean
@@ -166,19 +160,17 @@ entities:
   - entity: sensor
     name: Cleaning area
     category: diagnostic
-    icon: "mdi:floor-plan"
+    class: area
     dps:
       - id: 16
         type: integer
         name: sensor
         unit: "m²"
-        class: measurement
 
   - entity: sensor
     name: Cleaning time
     category: diagnostic
     class: duration
-    icon: "mdi:clock-outline"
     dps:
       - id: 17
         type: integer
@@ -246,16 +238,15 @@ entities:
         type: string
         mapping:
           - dps_val: only_sweep
-            value: sweep only
+            value: Sweep only
           - dps_val: only_mop
-            value: mop only
+            value: Mop only
           - dps_val: both_work
-            value: sweep and mop
+            value: Sweep and mop
 
   - entity: switch
     name: Carpet boost
     category: config
-    icon: "mdi:rug"
     dps:
       - id: 101
         type: boolean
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/honiture_g20_robot_vacuum_mop.yaml
```diff
@@ -0,0 +1,297 @@
+name: Honiture G20 Robot Vacuum and Mop
```
make-all: Top level name should be without branding, as other models may match the same config.

custom_components/tuya_local/devices/honiture_g20_robot_vacuum_mop.yaml
```diff
@@ -0,0 +1,297 @@
...
+          - dps_val: stop
+            value: stop
+      - id: 5
+        type: string
+        name: status
+        mapping:
+          - dps_val: standby
+            value: standby
+          - dps_val: smart_clean
+            value: cleaning
```
make-all: It is better to let all the different cleaning modes be differentiated here. This attribute affects both the `status` attribute of the vacuum entity, which is free to include any text, and the vacuum's state, which is constrained to only recognised states, but the vacuum implementation turns any unrecognised status into "cleaning" for that automatically.

custom_components/tuya_local/devices/honiture_g20_robot_vacuum_mop.yaml
```diff
@@ -0,0 +1,297 @@
...
+            value: charged
+          - dps_val: fault
+            value: error
+          - dps_val: sleep
+            value: sleep
+      - id: 14
+        type: string
+        name: fan_speed
+        mapping:
+          - dps_val: gentle
```
make-all, line 84: It is better to use the common values `low`, `medium`, `high` (all lower case) as the value here, as then they will be translated.

custom_components/tuya_local/devices/honiture_g20_robot_vacuum_mop.yaml
```diff
@@ -0,0 +1,297 @@
...
+  - entity: button
+    name: Reset roll brush
+    category: config
+    dps:
+      - id: 11
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Reset filter
```
make-all: `translation_key: filter_reset` exists for this

custom_components/tuya_local/devices/honiture_g20_robot_vacuum_mop.yaml
```diff
@@ -0,0 +1,297 @@
...
+    category: config
+    dps:
+      - id: 12
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Seek robot
```
make-all: This should be the `locate` attribute of the vacuum entity.

custom_components/tuya_local/devices/honiture_g20_robot_vacuum_mop.yaml
```diff
@@ -0,0 +1,297 @@
...
+        mapping:
+          - dps_val: low
+            value: low
+          - dps_val: middle
+            value: medium
+          - dps_val: high
+            value: high
+
+  - entity: select
+    name: Work mode
```
make-all: `translation_key: cleaning_mode` exists for this purpose (replacing the name and icon).
It supports values: `sweep`, `mop`, `sweep_and_mop` and others.

custom_components/tuya_local/devices/honiture_g20_robot_vacuum_mop.yaml
```diff
@@ -0,0 +1,297 @@
...
+  - entity: select
+    translation_key: language
+    category: config
+    dps:
+      - id: 105
+        name: option
+        type: string
+        mapping:
+          - dps_val: english
+            value: English
```
make-all: Language names should be lower case to match the translation keys.

custom_components/tuya_local/devices/honiture_q6_pro_robot_vacuum_mop.yaml
```diff
@@ -0,0 +1,244 @@
...
+  - entity: switch
+    name: Auto boost
+    category: config
+    dps:
+      - id: 137
+        type: boolean
+        name: switch
+
+  - entity: binary_sensor
+    name: Mop attached
```
make-all, line 224: This may be useful as an `available` attribute for marking other settings unavailable if they require the mop to function.

make-all:
Some of the comments also applied to the second config, I did not repeat them.

custom_components/tuya_local/devices/honiture_q6_pro_robot_vacuum_mop.yaml
starfire0606: Thanks for the suggestion! I've kept Mop attached as a standalone binary_sensor for now to keep the profile simple and readable. Users can build automations around it if they want to gate mop-related settings. Happy to revisit if you feel strongly about it.

PR #4994, opened 2026-05-03
Add support for 2 devices (LDH-7700 dehumidifier and uk002wl2 2gang switch)

minzoong:
Add support for 2 devices (LDH-7700 dehumidifier and uk002wl2 2gang switch)

### DIFF
--- a/custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
@@ -5,7 +5,7 @@ products:
 entities:
   - entity: humidifier
     class: dehumidifier
-    translation_key: extra
+    translation_key: extended
     dps:
       - id: 1
         type: boolean
--- a/custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
@@ -28,18 +28,29 @@ entities:
             value: sleep
           - dps_val: Auto
             value: auto
-  - entity: select
-    name: Fan speed
-    icon: "mdi:fan"
+      - id: 6
+        type: integer
+        name: current_humidity
+  - entity: fan
     dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 5
+        type: string
+        name: available
+        mapping:
+          - dps_val: Continuities
+            value: true
+          - value: false
       - id: 4
         type: string
-        name: option
+        name: speed
         mapping:
           - dps_val: low
-            value: Low
+            value: 50
           - dps_val: high
-            value: High
+            value: 100
   - entity: switch
     translation_key: ionizer
     dps:
@@ -52,14 +63,6 @@ entities:
       - id: 16
         type: boolean
         name: lock
-  - entity: sensor
-    class: humidity
-    dps:
-      - id: 6
-        type: integer
-        name: sensor
-        unit: "%"
-        class: measurement
   - entity: sensor
     class: temperature
     dps:
--- a/custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
@@ -1,0 +1,83 @@
+name: Dehumidifier
+products:
+  - id: msopwdnvklxbrnw9
+    model: LDH-7700
+entities:
+  - entity: humidifier
+    class: dehumidifier
+    translation_key: extra
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 2
+        type: integer
+        name: humidity
+        range:
+          min: 30
+          max: 80
+        mapping:
+          - step: 5
+      - id: 5
+        type: string
+        name: mode
+        mapping:
+          - dps_val: Continuities
+            value: continuous
+          - dps_val: Sleep
+            value: sleep
+          - dps_val: Auto
+            value: auto
+  - entity: select
+    name: Fan speed
+    icon: "mdi:fan"
+    dps:
+      - id: 4
+        type: string
+        name: option
+        mapping:
+          - dps_val: low
+            value: Low
+          - dps_val: high
+            value: High
+  - entity: switch
+    translation_key: ionizer
+    dps:
+      - id: 10
+        type: boolean
+        name: switch
+  - entity: lock
+    translation_key: child_lock
+    dps:
+      - id: 16
+        type: boolean
+        name: lock
+  - entity: sensor
+    class: humidity
+    dps:
+      - id: 6
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
```diff
@@ -0,0 +1,75 @@
+name: LDH-7700 Dehumidifier
```
make-all: Top level name should be without branding.
Manufacturer and model fields should be under products listing.

custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
```diff
@@ -0,0 +1,75 @@
...
+products:
+  - id: msopwdnvklxbrnw9
+entities:
+  - entity: humidifier
+    dps:
```
make-all, line 9: `class: dehumidifier`
also use `translation_key: extra` for extra mode translations

custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
```diff
@@ -0,0 +1,75 @@
...
+          min: 30
+          max: 80
+        mapping:
+          - step: 5
+      - id: 5
+        type: string
+        name: mode
+        mapping:
+          - dps_val: Continuities
+            value: Continuous
```
make-all: Use all lower case `continuous`, `sleep`, `auto` to match translation keys

custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
```diff
@@ -0,0 +1,75 @@
...
+      - id: 4
+        type: string
+        name: option
+        mapping:
+          - dps_val: low
+            value: Low
+          - dps_val: high
+            value: High
+  - entity: switch
+    name: Anion
```
make-all: Use `translation_key: ionizer` instead of this name and icon

custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
```diff
@@ -0,0 +1,75 @@
...
+        class: measurement
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 7
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+  - entity: sensor
```
make-all: Use binary_sensor without a name for class: problem.
The sensor needs to be mapped to true/false values (false for 0 and any other normal operation cases, true as a default)

Add an attribute called `fault_code` with the original raw value. See problem binary sensors in other configs if you cannot understand what I mean.

custom_components/tuya_local/devices/uk002wl2_2gang_switch.yaml
```diff
@@ -0,0 +1,60 @@
+name: 2-Gang Switch
```
make-all: Starting device names with a number is not advised. Use "Dual" instead of 2-gang to avoid it.

custom_components/tuya_local/devices/uk002wl2_2gang_switch.yaml
```diff
@@ -0,0 +1,60 @@
...
+products:
+  - id: uytmbjaogpksyu7i
```
make-all, line 3: Add model here, and manufacturer if known.

custom_components/tuya_local/devices/uk002wl2_2gang_switch.yaml
```diff
@@ -0,0 +1,60 @@
...
+entities:
+  - entity: switch
+    name: Switch 1
```
make-all: Use
```
    translation_key: switch_x
    translation_placeholders:
      x: "1"
```

custom_components/tuya_local/devices/uk002wl2_2gang_switch.yaml
```diff
@@ -0,0 +1,60 @@
...
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Switch 2
+    dps:
+      - id: 2
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Backlight
```
make-all: Use a light entity, and `translation_key: backlight`

custom_components/tuya_local/devices/uk002wl2_2gang_switch.yaml
```diff
@@ -0,0 +1,60 @@
...
+    category: config
+    dps:
+      - id: 16
+        type: boolean
+        name: switch
+  - entity: select
+    name: Relay Status
```
make-all: Use `translation_key: initial_state` instead of this name and icon.

For the mapping values, use "on", "off", memory (lower case to match translation keys). Do not forget to quote "on" and "off" to avoid YAML conversion to boolean

custom_components/tuya_local/devices/uk002wl2_2gang_switch.yaml
```diff
@@ -0,0 +1,60 @@
...
+        type: string
+        name: option
+        mapping:
+          - dps_val: off
+            value: Off
+          - dps_val: on
+            value: On
+          - dps_val: memory
+            value: Memory
+  - entity: number
```
make-all: Since this is a 24h timer, use a time entity with a `second` attribute. Use 
```
translation_key: timer_x
translation_placeholders:
  x: "1"
```

 instead of explicit name and icon

make-all:
Please stop force pushing, it loses all connection to previous history.

custom_components/tuya_local/devices/uk002wl2_2gang_switch.yaml
```diff
@@ -0,0 +1,91 @@
...
+          max: 86400
+  - entity: time
+    translation_key: timer_x
+    translation_placeholders:
+      x: "2"
+    dps:
+      - id: 8
+        type: integer
+        name: second
+          - dps_val: off
```
make-all: Seems to be missing `mapping:` line. Also, the type is `integer`, but the dps_val are strings (memory) and booleans (off / on), so maybe a larger section got cut by mistake from here.

make-all (dismissed):

custom_components/tuya_local/devices/uk002wl2_2gang_switch.yaml
make-all: translation_key was added, but name and icon were not removed.

custom_components/tuya_local/devices/uk002wl2_2gang_switch.yaml
minzoong: I'm sorry, I made a mistake with the copy&paste... T^T I tried to find the manufacturer info but couldn't, so I've just fixed the error for now. To be honest, I'm still quite new to the complex Git workflow and contributing to open source, so I'm a bit clumsy. Thank you so much for your guidance and patience!

custom_components/tuya_local/devices/ldh7700_dehumidifier.yaml
make-all: Sorry, it should be "extended", not "extra", my previous request was wrong.

PR #4998, opened 2026-05-04
feat(devices): update royal clima fresh AC entities

nkaliev:
Hi!
I made an update for existing device configuration.
Device itself has built-in fresh air intake fan (dp 102).
Current device config describes this fan as _select_ entity with 5 modes (options): "off", "auto", "low", "medium", "high".
I decided to convert it to the _fan_ entity, so it can be used more comfortable in automations, scripts, dashboard cards, etc.
Also I changed device model name as on the vendor's official site and in user manual. 

To be honest, I don't know what to do with old one _select_ entity, so I decided to mark it hidden (disabled by default for new users) for a backward compability to other users already using this integration with this device. If you already have some practices that solves that kind of...questions I will update branch accordingly.

### DIFF
--- a/custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
+++ b/custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
@@ -2,7 +2,7 @@ name: Air conditioner
 products:
   - id: mcnywzjlugvvvnjj
     manufacturer: Royal Clima
-    model: ROYAL FRESH FULL DC EU Inverter
+    model: Royal Fresh Full DC EU Inverter
 entities:
   - entity: climate
     translation_key: aircon_extra
@@ -139,7 +139,7 @@ entities:
             value: High
           - dps_val: auto
             value: Auto
-    hidden: true
+    deprecated: fan.fan_with_presets # 2026-05-05
   - entity: fan
     translation_key: fan_with_presets
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -2,7 +2,7 @@ name: Air conditioner
 products:
   - id: mcnywzjlugvvvnjj
     manufacturer: Royal Clima
-    model: Fresh full inverter
+    model: ROYAL FRESH FULL DC EU Inverter
```
make-all: all caps is unnecessary

custom_components/tuya_local/devices/royalclima_fresh_climate.yaml
```diff
@@ -139,6 +139,31 @@ entities:
             value: High
           - dps_val: auto
             value: Auto
+    hidden: true
```
make-all: Use `deprecated: fan_fan_with_presets  # 2026-05-05` instead of hidden, to hide for new installations, and warn for older ones that might be already using it. The datestamp comment is required for maintenance purposes.

make-all (changes requested):

nkaliev:
updated

PR #5004, opened 2026-05-06
feat: adding device thermostat EFAN230 by ENGO

vtashev:
adding device thermostat EFAN230 by ENGO

### DIFF
--- a/custom_components/tuya_local/devices/engo_efan230.yaml
+++ b/custom_components/tuya_local/devices/engo_efan230.yaml
@@ -1,0 +1,95 @@
+name: Fan Coil Thermostat
+products:
+  # Replace id with your device product id from Tuya IoT if needed.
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/engo_efan230.yaml
```diff
@@ -0,0 +1,95 @@
+name: Fan Coil Thermostat
+products:
+  # Replace id with your device product id from Tuya IoT if needed.
+  - id: "bf08c07c611b6b647dxast"
```
make-all: This is not a product id, it is a device id which is unique to your specific device.

vtashev: as my first device, please excuse my error
just corrected it

PR #5011, opened 2026-05-06
Add device amicismart as sm 63a

manic-mark:
Add device file for amiciSmart AS-SM-63A (Smart MCB + Energy Meter)
The device can be found here: https://www.amazon.in/dp/B0DJ5MNQFL
The device's Temperature reading tends to spike.
This file adds support for all readings and most settings
Despite the name AS-SM-63A, the device's highest trip current is 40A


custom_components/tuya_local/devices/amiciSmart_1P_JT.yaml
```diff
@@ -0,0 +1,294 @@
...
+      - id: 103
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+  # ==========================================
+  # CONFIG SETTINGS : TEMPERATURE
+  # ==========================================
+  - entity: number
+    name: Alrm Temp Max
```
make-all: Follow HA naming convention:
- only the first word in an entity or device name should be capitalised
- words should be spelt out in full, not arbitrarily abbreviated by leaving vowels out

custom_components/tuya_local/devices/amiciSmart_1P_JT.yaml
```diff
@@ -0,0 +1,294 @@
...
+        unit: A
+        mask: "0000FFFF0000000000000000"
+        range:
+          min: 10
+          max: 400
+        mapping:
+          - scale: 10
+            step: 5
+  - entity: switch
+    name: Alrm Current Max Shutoff
```
make-all: This is not related to the Alarm I think, but another limit for Shutdown

custom_components/tuya_local/devices/amiciSmart_1P_JT.yaml
```diff
@@ -0,0 +1,294 @@
...
+        name: value
+        unit: s
+        range:
+          min: 1
+          max: 100
+  # ==========================================
+  # CONFIG SETTINGS : MISC
+  # ==========================================
+  - entity: switch
+    name: Prepayment Feature Toggle
```
make-all: Entity names should be kept concise. Standard naming used by other configs is "Prepay"

make-all (changes requested):

custom_components/tuya_local/devices/amiciSmart_1P_JT.yaml
manic-mark: ty. will fix.

custom_components/tuya_local/devices/amiciSmart_1P_JT.yaml
manic-mark: When this switch is switched on, the breaker shuts power when the alarm is triggered. When it is switched off, the breaker maintains power to the load even when the alarm trips.

custom_components/tuya_local/devices/amiciSmart_1P_JT.yaml
manic-mark: thank you, i will amend this.

PR #5012, opened 2026-05-06
Decode Base64 timer and enable write support for Sanjun Steamer…

IsaacBreuer:
**update this part doesn't work well after HA restart, check my update below, that now part of the PR**
This PR updates the `sanjun_steamer.yaml` to convert the **DP 103** (Timer) from a read-only Base64 sensor into interactive `number` entities.

**Technical Details:**
- **Data Format:** Verified that DP 103 uses a 2-byte Base64 format where the first byte is Hours and the second byte is Minutes.
- **Write Support:** By using bitmasking (`FF00` for Hours and `00FF` for Minutes), users can now set the timer directly from Home Assistant.
- **Hardware Limits:** The Hour slider is capped at a maximum of 1, as confirmed by testing against the official SmartLife app (max timer is 60 minutes).
- **Error Handling:** Added `default: "AAA="` to DP 103 to prevent the "Cannot mask unknown current value" error on new installations.

**Tests Performed:**
- **Setting 1 Hour:** Moving the Hour slider to 1 successfully sets the physical steamer to 1:00.
- **Setting Minutes:** Moving the Minutes slider (e.g., to 30) correctly updates the physical display to 0:30.
- **Bi-directional Sync:** Verified that changes made on the physical device correctly update the sliders in Home Assistant.


IsaacBreuer:
### Refactor: Convert DP 103 to `select` entity for reliability

I have refactored the **DP 103** implementation from separate bitmasked `number` sliders to a single `select` entity.

**Why this change was necessary:**
* **Fixes Restart Errors:** During testing, using bitmasks on a 2-byte Base64 DP caused the error `Failed to perform the action number/set_value. Cannot mask unknown current value` after a Home Assistant restart. A `select` entity solves this by sending a complete Base64 payload, which does not require the integration to know the previous state.
* **Maintains Countdown Sync:** Because the sensor reports every minute as it counts down, I have mapped all 61 possible values (0 min through 1 hour). This prevents the entity from showing an "Unknown" state while the timer is active.
* **Verified Byte Logic:** The mapping correctly handles the transition where the first byte flips for the 60-minute mark (`AQA=`).

**Tests Performed:**
- Verified that the timer can be set immediately after a Home Assistant reboot without needing to "wake" the device via the Tuya/SmartLife app.
- Verified that the entity stays in sync with the physical device's display throughout the entire countdown.

custom_components/tuya_local/devices/sanjun_steamer.yaml
```diff
@@ -118,15 +118,136 @@ entities:
...
-    hidden: true
+  - entity: select
+    name: Timer Setting
+    icon: "mdi:timer-cog"
     dps:
       - id: 103
         type: base64
-        name: sensor
-        optional: true
+        name: option
```
make-all: This is not a good interface, 60 options is too many for a usable select interface.

I think it would be better to make a single number entity from 0-60, and make "AQA=" a special case mapping to 60 alongside a properly2 byte masked minute value.

IsaacBreuer:
We need someone who have steamers to know if the timers behave the same for them. 

The one I tested on is not labeled sanjun it's the steampsa brand.

https://steamspa.com/products/bkw-a-prx?pr_prod_strat=e5_desc&pr_rec_id=038f6f7a2&pr_rec_pid=7515375599703&pr_ref_pid=7515375698007&pr_seq=uniform

custom_components/tuya_local/devices/sanjun_steamer.yaml
IsaacBreuer: The issue with having a mapping just for the minutes  is not becuase I wanted to have the ability to be able to set it to 1 hour is we would be limited to 59 min is not a big deal. But if the value for setting only carriers the the second byte as a mask it won't be set when the first byte is unknown we need it to set the full value, there might be a way but I am new to this I am not sure how. Using the same dip for both bytes worked is its already set but not if 1 is still unknown 

See how was doing it first , worked until restart becuase 1st byte was unknown 
26018f9534247fd9d9d9efd45f6e1547ba88d171

make-all: This is why I suggest to set the mask to 2 bytes, so the 0 is conveyed.

```
  - entity: number
    translation_key: timer
    class: duration
    dps:
      - id: 103
        type: base64
        mask: FFFF
        optional: true
        name: value
        unit: min
        range:
           min: 0
           max: 60
        mapping:
          - dps_val: "AQA="
            value: 60
```

make-all: I'm not 100% sure of the above, the order of base64 decode and mask application here may mean you need to use `256` instead or "AQA=" in the mapping.

Issue #4859, opened 2026-04-09
Request support for Parkside Robot Mower PMRC 250 A1（BT）


Pigi-102:
### Log message

```text
2026-04-09 09:53:38.733 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches lscsmartconnect_videodoorbell with quality of 10%. LOCAL DPS: {"updated_at": 1775721201.6566672, "13": 100, "101": "STANDBY", "102":
0, "103": "MOWER_LEAN", "104": true, "105": 2, "106": 1234, "116": false, "118": true, "134": "V0.1.2 build Mar 11 2025 18:29:07"}
2026-04-09 09:53:38.733 WARNING (MainThread) [custom_components.tuya_local.config_flow] Include the previous log messages with any new device request to https://github.com/make-all/tuya-local/issues/
```

### Product ID

7yr5iwga

### Product Name

Parkside Robot Mower PMRC 250 A1（BT）

### DPS information

```text
From QueryThingDataModel:
{
  "result": {
    "model": "{\"modelId\":\"fjjhfw\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"默认服务\",\"properties\":[
{\"abilityId\":2,\"accessMode\":\"rw\",\"code\":\"switch_go\",\"description\":\"\",\"extensions\":{\"iconName\":\"icon-power\"},\"name\":\"工作开关\",\"typeSpec\":{\"type\":\"bool\"}},
{\"abilityId\":3,\"accessMode\":\"rw\",\"code\":\"mode\",\"description\":\"\\\"【必选】枚举值按需选择，不需要的可删除，可以调整排列顺序，但是单词必须一模一样，不可修改单词，例如不能把smart修改为AOTU。\\nstandby 待机\\nrandom 随机模式\\nsmart 自动模式\\nspot 定点模式\\ngoto_charge 回充模式\\\"\",\"extensions\":{\"iconName\":\"icon-dp_mode\"},\"name\":\"工作模式\",\"typeSpec\":{\"type\":\"enum\",\"range\":[\"standby\",\"random\",\"smart\",\"spot\",\"goto_charge\"]}},
[...]

PR #5018, opened 2026-05-08
feat(parkside_pmrc250_mower): dedicated config for PMRC 250 Ai

alexit84:
Splits the Parkside PMRC 250 Ai Bluetooth (product_id 7yr5iwga)
out of moebot_s_mower.yaml into its own device file.

The PMRC 250 Ai was added to moebot_s_mower.yaml in 50f4961d
(refs #4859) under the assumption that it shares behaviour with
the MoeBot S20. Field testing on my unit, with debug-level logs
from this integration, revealed two firmware divergences:

1. DPS 103 reports MOWER_LEAN as the default idle code on every
   poll, even when the robot is docked, on flat ground, with no
   error shown in the official Tuya app or on the robot's own UI.
   With the shared MoeBot config the Problem sensor in HA is
   therefore stuck permanently on "Tilted". Mapped to OK in the
   new file. All other DPS 103 codes are unchanged.

2. DPS 115 StartReturnStation is silently rejected when the
   mower is in MOWING state. The firmware requires the sequence
   PauseWork -> CancelWork -> StartReturnStation; the same flow
   is enforced by the official Tuya app (you cannot tap "return
   to base" while mowing without first pausing and cancelling).
   The YAML schema cannot express multi-step commands, so the
   mapping is left as-is and a short inline comment documents
   the constraint. End users can chain the commands in an HA
   script when needed.

The product entry has been removed from moebot_s_mower.yaml so
the two files do not both match 7yr5iwga. Existing PMRC 250 Ai
users may need to remove and re-add the device entry to pick up
the new config, since tuya-local caches the matched device type
in the config entry.

yamllint clean. Tested locally on my unit running firmware
V0.1.2 build Mar 11 2025 18:29:07.

### DIFF
--- a/custom_components/tuya_local/devices/parkside_pmrc250_mower.yaml
+++ b/custom_components/tuya_local/devices/parkside_pmrc250_mower.yaml
@@ -1,11 +1,8 @@
 name: Lawnmower
 products:
-  - id: mvt4l2evgq2l3nkn
-    manufacturer: MoeBot
-    model: S20
-  - id: icw5sal7xfcevsve
+  - id: 7yr5iwga
     manufacturer: Parkside
-    model: PMRDA 20-Li A1
+    model: PMRC 250 Ai Bluetooth
 entities:
   - entity: lawn_mower
     dps:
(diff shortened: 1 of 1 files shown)


make-all:
1. Seems to be a common issue based on previous reports for the WiFi versions of the device. It just means that sensor is useless, as it is always indicating an incorrect status (but the value that is indicated is still in line with the documented values for the device). I don't think this justifies a forked config.

2. This isn't really a problem with the config, it is a device behaviour, which can't really be coded within the limits of this integration. Forking the config also doesn't fix this.

PR #5025, opened 2026-05-09
Add support for Hiniso dehumidifier

DomJoke:
Adds a new dehumidifier device definition for tuya-local with updated DPS mappings based on live Tuya cloud data.

Key changes:
- Updated mode mapping to match actual device states (dehumidify, drying, sleep)
- Added support for fan speed (low / mid / high)
- Added UV and ionizer switches based on verified DPS points
- Added child lock support
- Fixed tank_full (remind, DP 25) handling by supporting both raw and quoted enum values ("on"/"off")
- Improved fault handling with bitmap-based sensor description mapping

This configuration is based on real device telemetry from Tuya cloud API responses and tested state transitions.

Notes:
- DPS 25 (remind) changes between "off" and "on" when tank state changes
- Enum values may be received inconsistently as raw or quoted strings depending on Tuya firmware
- Mapping is made defensive to ensure correct binary_sensor state evaluation in tuya-local

Tested behavior:
- All states and commands


custom_components/tuya_local/devices/hiniso_dehumidifier.yaml
```diff
@@ -0,0 +1,225 @@
...
+
+  - entity: binary_sensor
+    translation_key: tank_full
+    category: diagnostic
+    dps:
+      - id: 25
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: on
```
make-all: Remove these two options. The type is string, so there should not be boolean options in here

custom_components/tuya_local/devices/hiniso_dehumidifier.yaml
```diff
@@ -0,0 +1,225 @@
...
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+
+          - dps_val: null
+            value: false
+
+          - value: true
+
```
make-all: When the type is bitfield, the standard is to always include a `fault_code` attribute with the raw value. This is because the mapping to description can only cover one value, but a bitfield can indicate multiple errors at once.

custom_components/tuya_local/devices/hiniso_dehumidifier.yaml
```diff
@@ -0,0 +1,225 @@
...
+        type: boolean
+        name: switch
+
+
+  #
+  # UV
+  #
+
+  - entity: switch
+    name: UV
```
make-all: Use `translation_key: uv_sterilization` instead of this name and icon.

custom_components/tuya_local/devices/hiniso_dehumidifier.yaml
```diff
@@ -0,0 +1,225 @@
...
+      - id: 16
+        type: boolean
+        name: lock
+
+
+  #
+  # FAN SPEED
+  #
+
+  - entity: select
```
make-all: Using a fan entity with speed mapped to percentages is better, as it allows for voice commands "turn the fan up" and "down"

make-all (changes requested):

DomJoke:
Changes are done, open for re-review and pull request

DomJoke (commented):
Changes done

PR #5027, opened 2026-05-09
feat: add support for Ivation 50pt Dehumidifier

acenuke:


custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -79,7 +79,7 @@ entities:
         unit: F
         class: measurement
   - entity: sensor
-    name: Time remaining
+    name: Timer remaining
```
make-all: Rather than changing the name to avoid the warning, you should remove it to let the translation be used instead.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,128 @@
+name: MANUAL Ivation Dehumidifier
```
make-all: Top level name should be without branding, only the products listings should be branded.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,128 @@
...
+      - id: 6
+        type: string
+        name: mode
+        mapping:
+          - dps_val: "1"
+            value: normal
+          - dps_val: "3"
+            value: high
+  - entity: binary_sensor
+    name: Tank
```
make-all: Use `translation_key: tank_full` instead of the name and class here.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,128 @@
...
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 110
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
```
make-all, line 54: Use `translation_key: defrost` instead of the name and class here.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,128 @@
...
+    name: Compressor
+    class: running
+    category: diagnostic
+    dps:
+      - id: 115
+        type: boolean
+        name: sensor
+  - entity: sensor
```
make-all: Use an unnamed `binary_sensor` with `class: problem` here, and repeat dp 11 in the dps list, once with name `sensor` and `mapping` of `0` to `false`, and otherwise `true`, and the second time as `fault_code` with no mapping.

Optionally a third repeat with name `description` can be used if the meanings of the bits are known.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,128 @@
...
+    name: Fault Code
+    category: diagnostic
+    dps:
+      - id: 11
+        type: bitfield
+        name: sensor
+  - entity: sensor
+    name: Room temperature
```
make-all: This should be unnamed to allow the temperature class translations to be used. It should not be diagnostic.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,128 @@
...
+    translation_key: time_remaining
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 13
+        type: integer
+        name: sensor
+        unit: min
+  - entity: sensor
+    name: Actual Fan speed
```
make-all: Just "Fan speed" is adequate as a sensor name

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,128 @@
...
+      - id: 6
+        type: string
+        name: option
+        mapping:
+          - dps_val: "1"
+            value: low
+          - dps_val: "3"
+            value: high
+  - entity: switch
+    name: Continuous mode
```
make-all: This could be moved into the humidifier entity as `mode`.

If you add `translation_key: extended` this integration defines a `continuous` mode which can be used for the true case, and HA defines a `normal` mode that can be used for the false case.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,128 @@
...
+            value: "Off"
+          - dps_val: "1"
+            value: "Ultra-Low"
+          - dps_val: "2"
+            value: "Low"
+          - dps_val: "3"
+            value: "Medium"
+          - dps_val: "4"
+            value: "High"
+  - entity: select
```
make-all: Normally a fan entity is used with speed mapped to 50 and 100, as voice assistants work better with percentage fan speed than with either fan presets or select entities.

make-all (changes requested):

acenuke:
Sorry if I did not manage the pull requests right. I don't really use GitHub unless for my own needs in my own repositories, so not 100% sure that I did the right thing and followed your PR instructions

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,139 @@
...
+        name: humidity
+        range:
+          min: 35
+          max: 80
+      - id: 3
+        type: integer
+        name: current_humidity
+      - id: 6
+        type: string
+        name: mode
```
make-all: You cannot have 2 attributes with the same name.
Since this one is already separated into a fan speed below, remove this one.

make-all (changes requested):

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
acenuke: They are not the same. One is a preset, the other one is the actual fan speed (it can vary based on what it does). I will rename them

Based on your specific API JSON, you are controlling a Tuya-powered Dehumidifier or HVAC appliance.
In your device's exact hardware profile, fan_speed and windspeed have distinct operational roles. Here is how they differ right now:
⚙️ fan_speed (DP ID 116)
This controls the physical speed of the appliance's internal blower fan motor.
Current State: "value": "3" (an enum type, which typically maps to speed presets like Low/Med/High).
What it does: Changing this value directly increases or decreases the RPM of the physical motor inside the unit. A higher number forces more air volume through the internal coils to speed up dehumidification or cooling.
🍃 windspeed (DP ID 6)
On Tuya dehumidifiers, this controls the wind/airflow comfort logic or target operating mode, rather than the raw motor speed.
Current State: "value": "1" (also an enum type).
What it does: This typically controls a specific comfort mode (like 1 for Normal, 2 for Turbo/Dryer, or 3 for Quiet). When you change windspeed, the machine's internal processor automatically recalculates and shifts the underlying fan_speed and compressor output (comp_output) to match that environmental profile.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
make-all, line 1: I don't know why the filename has a v2 suffix. There is no other config with a similar name.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,136 @@
+name: Dehumidifier
+products:
+  - id: gywqtqpgiu74ozok
+    manufacturer: Ivation
+    model: IVADUWIFI50WP
+entities:
+  - entity: humidifier
+    class: dehumidifier
+    dps:
```
make-all, line 10: This needs `translation_key: extended` to pull in the `continuous` mode translation.

custom_components/tuya_local/devices/ivation_50pt_dehumidifier_v2.yaml
```diff
@@ -0,0 +1,136 @@
...
+        type: boolean
+        name: mode
+        mapping:
+          - dps_val: false
+            value: normal
+          - dps_val: true
+            value: continuous
+
+  - entity: fan
+    name: Fan
```
make-all: Name should not be needed

make-all (changes requested):

Issue #1038, opened 2023-08-28
Request support for Atorch at4pw


Adam-1974:
Here is a link to the json from Tuya: https://jsoneditoronline.org/#left=cloud.bcd999d58ddd4579ac00f762d849a1b2

Product ID: tjtigg991kvoiiqi

Link to instructions with description: http://en.atorch.cn/upload/20220629115640.pdf

make-all:
Useful information extracted from that dump:

The device seems to be an electricity meter from the manual.

```
"dps":{"1":true,"9":0,"17":0,"18":55,"19":0,"20":23092,"101":100,"102":0,"103":0,"104":2650,"105":1000,"106":26500,"107":"english","108":9,"109":3,"110":60,"111":true,"112":"controlled","113":false,"114":false,"115":false,"116":false,"117":"measurement","118":"safety_protection","119":750,"120":true,"123":30,"132":"off","133":0,"134":0,"135":0,"136":"single_rate","137":5,"138":"open","139":false,"140":50000,"141":false,"142":0,"143":30}
```
Seems to be a list of the dps, but not clear where this was captured from, as it is in the middle of a dump of other data that has obviously come from the cloud, so unsure whether this is representative of what will be received locally (it seems like a lot more data than is usual for this type of device)

```
  "1":{"code":"switch_1","iconname":"icon-dp_power2","id":"1","mode":"rw","name":"开关状态","property":{"type":"bool"},"schemaType":"bool","type":"obj"},
  "9":{"code":"countdown_1","iconname":"icon-dp_time2","id":"9","mode":"rw","name":"开关倒计时","property":{"max":360000,"min":0,"scale":0,"step":1,"type":"value","unit":"s"},"schemaType":"value","type":"obj"},
  "17":{"code":"add_ele","iconname":"icon-battery","id":"17","mode":"rw","name":"增加电量","property":{"max":5000000,"min":0,"scale":2,"step":1,"type":"value","unit":"kwh"},"schemaType":"value","type":"obj"},
  "18":{"code":"cur_current","iconname":"icon-Ele","id":"18","mode":"ro","name":"当前电流","property":{"max":120000,"min":0,"scale":3,"step":1,"type":"value","unit":"A"},"schemaType":"value","type":"obj"},
[...]

PR #5030, opened 2026-05-10
Add Atorch real-time refresh switch

Rkn21:
## Summary
- Add optional DPS 126 as a config switch for the Atorch AT4PBWP energy meter.
- Expose the firmware-dependent real time data refresh setting as `Real time data refresh` with the refresh icon.

## Validation
- Parsed custom_components/tuya_local/devices/atorch_at4pw_energymeter.yaml with PyYAML.
- Ran yamllint on custom_components/tuya_local/devices.
- Validated the modified config against custom_components/tuya_local/devices/device_config_schema.json.
- Ran git diff --check.

Note: uv is not available in my local shell, so the exact uv run commands from AGENTS.md were not run locally.

### DIFF
--- a/custom_components/tuya_local/devices/atorch_at4pw_energymeter.yaml
+++ b/custom_components/tuya_local/devices/atorch_at4pw_energymeter.yaml
@@ -230,13 +230,14 @@ entities:
         type: boolean
         name: switch
   - entity: switch
-    name: Real-time refresh (1s/60s)
+    name: Real time data refresh
     category: config
     icon: "mdi:refresh"
     dps:
       - id: 126
         type: boolean
         name: switch
+        optional: true
   - entity: sensor
     class: energy
     dps:
--- a/custom_components/tuya_local/devices/atorch_at4pw_energymeter.yaml
+++ b/custom_components/tuya_local/devices/atorch_at4pw_energymeter.yaml
@@ -229,6 +229,14 @@ entities:
       - id: 120
         type: boolean
         name: switch
+  - entity: switch
+    name: Real-time refresh (1s/60s)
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/atorch_at4pw_energymeter.yaml
```diff
@@ -229,6 +229,14 @@ entities:
       - id: 120
         type: boolean
         name: switch
+  - entity: switch
+    name: Real-time refresh (1s/60s)
+    category: config
+    icon: "mdi:refresh"
+    dps:
+      - id: 126
```
make-all, line 237: This dp is not present in the info at #1038, so I tihnk it might be a problem to add as a non-optional dp.

If the value resets automatically after a tmie, you could consider making it a button instead of a switch.

The entity name should be free of non-alphanumeric characters like hyphens, slashes and parentheses.

Rkn21: Thanks, that makes sense.

I have updated this in the latest commit:
- DP 126 is now marked as optional, since it appears to be firmware-dependent and is not present in the older device info from #1038.
- The entity name is now `Real time data refresh`, without hyphens, slashes or parentheses.

About switch vs button: in SmartLife this is exposed as an actual switch/toggle, labelled "Real time data refresh switch". When enabled, it switches the device to 1s refresh for about 60s, then the device automatically turns it off again. Since the state is visible and automatically returns to off, I think a switch still represents the Tuya behavior better than a stateless button.

Screenshot from SmartLife:

SmartLife real time data refresh switch

PR #5032, opened 2026-05-11
added Saiming 5050RGB02 lamp

Bananz0:
https://www.aliexpress.com/item/1005009321157078.html 
https://fcc.report/FCC-ID/2BDV5-5050RGB02
Support for Floor Lamp with scenes


make-all:
This device is the same as the existing mirabella_geniopixel_light. Due to the complexity of the scenes, it is better to implement on top of the existing number entity using input_select helpers.

PR #5038, opened 2026-05-11
Add NHZS fan speed switch device config

Daniel-Boluda:
Add NHZS fan speed switch device config

### DIFF
--- a/custom_components/tuya_local/devices/nhzs_fan_switch.yaml
+++ b/custom_components/tuya_local/devices/nhzs_fan_switch.yaml
@@ -2,7 +2,7 @@ name: Fan speed switch
 products:
   - id: cwwc6ga2gz8mvrqo
     manufacturer: NHZS
-    model: 风扇开关
+    model: cwwc6ga2gz8mvrqo
 entities:
   - entity: fan
     dps:
@@ -15,20 +15,19 @@ entities:
         range:
           min: 1
           max: 20
-  - entity: number
+  - entity: time
     translation_key: timer
     category: config
     dps:
       - id: 2
         type: integer
-        name: value
-        unit: min
+        name: minute
         optional: true
         range:
           min: 0
           max: 1440
   - entity: select
-    name: Relay status
+    translation_key: initial_state
     category: config
     dps:
       - id: 11
--- a/custom_components/tuya_local/devices/nhzs_fan_switch.yaml
+++ b/custom_components/tuya_local/devices/nhzs_fan_switch.yaml
@@ -43,7 +43,7 @@ entities:
           - dps_val: memory
             value: memory
   - entity: light
-    name: Backlight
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/nhzs_fan_switch.yaml
```diff
@@ -0,0 +1,51 @@
+name: Fan speed switch
+products:
+  - id: cwwc6ga2gz8mvrqo
+    manufacturer: NHZS
+    model: 风扇开关
```
make-all: This is not a model, it just says "Fan switch" in Chinese

custom_components/tuya_local/devices/nhzs_fan_switch.yaml
```diff
@@ -0,0 +1,51 @@
...
+    category: config
+    dps:
+      - id: 2
+        type: integer
+        name: value
+        unit: min
+        optional: true
+        range:
+          min: 0
+          max: 1440
```
make-all, line 28: Use a time entity for 24h timers.

custom_components/tuya_local/devices/nhzs_fan_switch.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - entity: select
+    name: Relay status
```
make-all: Use `translation_key: initial_state`

make-all (changes requested):

custom_components/tuya_local/devices/nhzs_fan_switch.yaml
Daniel-Boluda: From the device i only have     
    "product_id": "cwwc6ga2gz8mvrqo",
    "product_name": "NHZS-风扇开关",
    
   I will set the model as the product_id , is that ok?
   
   By the way it would be really usefull if we could search by product_id the existing devices from Hass UI, is this a feature you are planning?

Daniel-Boluda: Switch is similar to this:
https://www.amazon.es/Switch-Tempered-Compatible-Control-110-240V/dp/B09C8SXQMH/ref=pb_allspark_dp_sims_pao_desktop_session_based_d_sccl_3_6/525-2113708-3990258?pd_rd_w=cQfeS&content-id=amzn1.sym.89101b7d-ab12-43ec-9284-9bcc9bad8558&pf_rd_p=89101b7d-ab12-43ec-9284-9bcc9bad8558&pf_rd_r=S8P016V4V52MZK9WEDGF&pd_rd_wg=4yn4L&pd_rd_r=b474913e-280b-43de-83bf-76a4228daaaa&pd_rd_i=B09C8SXQMH&psc=1

PR #5039, opened 2026-05-11
Add smart litter box device config

Daniel-Boluda:
Add smart litter box device config

### DIFF
--- a/custom_components/tuya_local/devices/smart_litter_box.yaml
+++ b/custom_components/tuya_local/devices/smart_litter_box.yaml
@@ -198,11 +198,20 @@ entities:
         range:
           min: 0
           max: 1000
-  - entity: sensor
+  - entity: binary_sensor
     name: Fault
-    class: enum
+    class: problem
     category: diagnostic
     dps:
       - id: 22
         type: bitfield
         name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - dps_val: null
+            value: false
+          - value: true
+      - id: 22
+        type: bitfield
+        name: fault_code
--- a/custom_components/tuya_local/devices/smart_litter_box.yaml
+++ b/custom_components/tuya_local/devices/smart_litter_box.yaml
@@ -1,7 +1,7 @@
 name: Smart litter box
 products:
   - id: 5t7esmqqh92ssbe5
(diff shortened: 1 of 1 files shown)


make-all:
Some branding information is needed to differentiate this from other similar products. I am not comfortable adding it as the generic "smart_litter_box.yaml", as it misleads naive users into thinking their completely different device should work with it. Generic names are only suitable where there is an actual Tuya standard (or reference design) that many manufacturers follow, like for smartplugs and lightbulbs.

custom_components/tuya_local/devices/smart_litter_box.yaml
```diff
@@ -0,0 +1,214 @@
...
+    category: config
+    dps:
+      - id: 127
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 1000
+  - entity: binary_sensor
+    name: Fault
```
make-all, line 199: no name is required here., the class has translations that are better to allow to be used instead

custom_components/tuya_local/devices/smart_litter_box.yaml
```diff
@@ -0,0 +1,214 @@
...
+        name: switch
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 114
+        type: boolean
+        name: lock
+  - entity: number
+    name: Induction delay
```
make-all, line 124: I don't quite understand what is meant by "Induction" in the context of a cat litter tray. Maybe it is a mistranslation?

custom_components/tuya_local/devices/smart_litter_box.yaml
```diff
@@ -0,0 +1,214 @@
...
+    dps:
+      - id: 118
+        type: integer
+        name: value
+        unit: min
+        range:
+          min: 5
+          max: 60
+  - entity: binary_sensor
+    name: Store full
```
make-all, line 146: I'm not sure what is the difference between "Trash full" and "Store full"

make-all (changes requested):

Daniel-Boluda:
existing as tonepie_t1pro_catlitterbox.yaml

PR #5042, opened 2026-05-11
Update sanjun_steamer.yaml to fix colour changes

aicbaicb:
colour wheel not recognised by device. So I'm reverting to a select-box


make-all:
This isn't the way to fix things.

Issue #4526, opened 2026-02-17
Request support for Atorch GR2PWS


guffy1234:
### Device details

```json
{
  "result": {
    "active_time": 1771352166,
    "bind_space_id": "***",
    "category": "cz",
    "create_time": 1764165003,
    "custom_name": "GridHome2",
    "icon": "smart/icon/bay1598402984069otRu/4d5a974fa0f5e5678705f84ac6cfa2e1.jpg",
    "id": "***",
    "ip": "***",
    "is_online": true,
    "lat": "***",
    "local_key": "***",
    "lon": "***",
    "model": "GR2PWS",
    "name": "ATORCH Smart Energy Meter(GR2PWS) 2",
    "product_id": "sq1ahfxhgove4kgo",
    "product_name": "炬为WiFi联网远控电表(GR2PWS)",
    "sub": false,
    "time_zone": "+02:00",
    "update_time": 1771352174,
    "uuid": "8459d002b8acc33f"
  },
  "success": true,
  "t": 1771358169145,
  "tid": "b3e4b5190c3a11f1a0e39aa9e01ed45b"
}
```

### Ready to use definition atorch_gr2pws_energymeter.yaml

```yaml
name: Energy meter
products:
  - id: sq1ahfxhgove4kgo
    manufacturer: Atorch
    model: GR2PWS
entities:
  - entity: switch
    class: outlet
    dps:
      - id: 1
        type: boolean
        name: switch
  - entity: number
    category: config
    translation_key: timer
    class: duration
    dps:
      - id: 9
        type: integer
        name: value
        unit: min
        range:
          min: 0
          max: 360000
        mapping:
          - scale: 60
            step: 60
  - entity: sensor
    class: current
    dps:
      - id: 18
        name: sensor
        type: integer
        class: measurement
        unit: A
        mapping:
          - scale: 1000
  - entity: sensor
    class: power
    dps:
      - id: 19
        name: sensor
        type: integer
        class: measurement
        unit: W
        mapping:
          - scale: 100
  - entity: sensor
    class: voltage
    dps:
      - id: 20
        name: sensor
        type: integer
        class: measurement
        unit: V
        mapping:
          - scale: 100
  - entity: number
    category: config
    name: Price
    icon: "mdi:cash-multiple"
    dps:
      - id: 101
        type: integer
        name: value
        range:
          min: 0
          max: 99999
        mapping:
          - scale: 100
  - entity: sensor
    name: Cost
    icon: mdi:cash-clock
[...]

PR #5047, opened 2026-05-12
feat(atorch): add dedicated config for GR2PWS energy meter

guffy1234:
This splits GR2PWS out of the shared AT2PL config into a dedicated device file.

Although the devices are closely related, the GR2PWS data model and exposed behaviour differ enough to justify a separate config. In particular, the available DPs, writable/read-only behaviour, and some operating characteristics such as supported voltage/current thresholds and prepaid-related functions do not line up cleanly with the existing AT2PL mapping.

This config was aligned against the local Tuya data model and observed Smart Life behaviour. It also adds the prepaid/balance related entities exposed by this device.

DP 124 and DP 132 were intentionally kept as sensors even though the raw model marks them writable, because they appear to represent reported state rather than useful user-facing controls for this model.

### DIFF
--- a/custom_components/tuya_local/devices/atorch_gr2pws_energymeter.yaml
+++ b/custom_components/tuya_local/devices/atorch_gr2pws_energymeter.yaml
@@ -1,18 +1,32 @@
 name: Energy meter
 products:
-  - id: seqzqh45b0ggn9so
+  - id: sq1ahfxhgove4kgo
     manufacturer: Atorch
-    model: AT2PL
+    model: GR2PWS
 entities:
   - entity: switch
+    class: outlet
     dps:
       - id: 1
         type: boolean
         name: switch
+  # DP 17 is writable in the model, but its exact device-side purpose is unclear.
+  # It appears to be related to adding prepaid energy/credit.
+  - entity: number
+    name: Add energy
+    category: config
+    class: energy
+    dps:
       - id: 17
         type: integer
         optional: true
-        name: add_ele
+        name: value
+        unit: kWh
+        range:
+          min: 0
+          max: 5000000
+        mapping:
+          - scale: 100
   - entity: number
     translation_key: timer
     class: duration
@@ -30,7 +44,6 @@ entities:
             step: 60
   - entity: sensor
     class: current
-    category: diagnostic
     dps:
       - id: 18
         type: integer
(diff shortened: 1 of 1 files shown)


make-all:
I do not accept PRs to fork configs merely to cater to users' personal naming preferences.

If you can separate out the non-naming improvements they might be considered for the original config

guffy1234:
> I do not accept PRs to fork configs merely to cater to users' personal naming preferences.
> 
> If you can separate out the non-naming improvements they might be considered for the original config

hi @make-all 

Thank you for the review. I want to clarify that the goal of this PR is not to introduce a different file purely for naming preferences.

The reason for a separate definition is that the observed GR2PWS behaviour does not match the existing AT2PL definition in a few important ways.

- The electrical/protection ranges differ. From the existing AT2PL model in issue #1200, DP 104 max is 265.0 V, DP 106 max is 27500 W, and DP 119 max is 275.0 V. For GR2PWS, both the observed model in issue #4526 and the product information indicate a wider operating range, and the config here uses DP 104 max 320.0 V, DP 106 max 32000 W, and DP 119 max 320.0 V.
- Because these are protection-related settings, a shared file would be problematic in both directions: keeping the AT2PL limits would unnecessarily restrict GR2PWS users, while raising the shared limits to match GR2PWS would expose AT2PL users to values their device model does not advertise or support. That felt like a functional and safety-related mismatch rather than a naming difference.
- Some DPs also behave differently. In particular DP 138 differs between the models: AT2PL exposes open/close-style values in #1200, while GR2PWS exposes only memory in #4526. DP 124 and DP 132 also read more like reported state on GR2PWS, so in this config they were intentionally kept as sensors to avoid presenting ambiguous writable controls.

Based on those differences, I do not currently see a clean or safe way to support both models in a single file without either restricting one model incorrectly or exposing unsupported behaviour on the other.

So the intent here was to keep the common DP names and behaviour consistent where possible, but split the configuration where the device capabilities and semantics diverge. If you prefer a different structure for handling that model-specific behaviour, I am happy to rework it.

Characteristics:
AT2PL https://www.aliexpress.com/item/1005005863721278.html
GR2PWS https://www.aliexpress.com/item/1005006518681031.html

make-all:
I'm not convinced there is a need for higher voltage and power ranges for normal domestic installations.

PR #5048, opened 2026-05-12
feat: fix stale device cache handling and add support for another Tonepie T1Pro smart cat litter box 

giuseppe99barchetta:
Adds dedicated support for the Tonepie T1Pro smart cat litter box (`swrzomblcz1can76`) with a new device YAML.

Changes include:

* improved entity naming and semantics
* better separation between buttons, switches, sensors and diagnostic entities
* hidden diagnostic entities for uncertain/internal DPS values
* proper units and categories for several entities
* exact product_id matching for the new v3 device definition

Also fixes a runtime issue where stale cached `tinytuya` device objects could cause intermittent `914: Check device key or version` errors after failed setup retries, even when the config flow test succeeded.

All tests and lint checks pass.

### DIFF
--- a/custom_components/tuya_local/__init__.py
+++ b/custom_components/tuya_local/__init__.py
@@ -74,6 +74,20 @@ def get_device_unique_id(entry: ConfigEntry):
     )
 
 
+def cleanup_failed_device(hass: HomeAssistant, device_id: str):
+    """Drop cached device objects left behind by failed setup."""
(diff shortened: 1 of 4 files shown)


make-all:
Please do not mix multiple contributions up in a single PR

giuseppe99barchetta:
> Please do not mix multiple contributions up in a single PR

Ok thank you, I'll split that

PR #5049, opened 2026-05-12
feat: add support for another Tonepie T1Pro smart cat litter box

giuseppe99barchetta:
Adds dedicated support for the Tonepie T1Pro smart cat litter box (`swrzomblcz1can76`) with a new device YAML.

Changes include:

* improved entity naming and semantics
* better separation between buttons, switches, sensors and diagnostic entities
* hidden diagnostic entities for uncertain/internal DPS values
* proper units and categories for several entities
* exact product_id matching for the new v3 device definition

All tests and lint checks pass.


make-all:
A new config is not justified, as the dps are identical to the original. I have added the product id to the original config. Any further changes I noticed in the comment are based on assumptions rather than fact, so I have not included those.

PR #5050, opened 2026-05-12
Fix: stale device cache

giuseppe99barchetta:
Fixes a runtime issue where stale cached `tinytuya` device objects could be reused after a failed setup attempt.

In some cases this caused intermittent:
`914: Check device key or version`

even when:

* the local key was correct
* the protocol version was correct
* the config flow test succeeded

The fix clears the cached device object when `async_setup_entry()` fails, so retries create a fresh `tinytuya.Device` instance instead of reusing a potentially broken connection state.

Includes regression test coverage for the failed setup retry path.


custom_components/tuya_local/devices/tonepie_t1pro_catlitterbox_v3.yaml
make-all, line 1: This file should not be included in this PR

make-all (changes requested):

PR #5058, opened 2026-05-13
Add Tuya W-2839 water quality monitor config

tkoenig:
## Proposed change

Add device configuration for the Tuya W-2839 water quality monitor / WiFi Online Control Monitor.

This is a 3-in-1 pool/aquarium water quality monitor that reports:

- temperature on DP 8, scaled by 10
- pH on DP 106, scaled by 100
- oxidation reduction potential (ORP/redox) on DP 131 in mV

It also exposes the device warning thresholds as config number entities:

- high/low temperature warnings on DPs 102/103, scaled by 10
- high/low pH warnings on DPs 107/108, scaled by 100
- high/low ORP warnings on DPs 132/133 in mV

Tested with product ID `muwcszgae41ccndk` using protocol version 3.3.

The existing `phw218_waterquality_monitor` config does not match this device because the W-2839 only reports the three measurement datapoints above by default, while the PH-W218 config expects additional required datapoints.

## Device diagnostics

Initial Local Tuya diagnostics showed the following cached state:

```json
{
  "8": 116,
  "106": 704,
  "131": -3
}
```

After setting warning thresholds, the device reported:

```json
{
  "8": 116,
  "102": 350,
  "103": 210,
  "106": 701,
  "107": 1500,
  "108": 0,
  "131": 546,
  "132": 2000,
  "133": -2000
}
```

Home Assistant entities created and tested:

- temperature: `11.6 °C`
- pH: `7.04`
- ORP: `546 mV`
- high/low temperature warning numbers
- high/low pH warning numbers
- high/low ORP warning numbers

The threshold number writes were verified against the Tuya app.

## Type of change

- [x] Add new device configuration
- [x] Update supported devices list

## Checklist

- [x] The device configuration validates with `yamllint`
- [x] `pytest tests/test_device_config.py` passes

### DIFF
--- a/custom_components/tuya_local/devices/w2839_pool_monitor.yaml
+++ b/custom_components/tuya_local/devices/w2839_pool_monitor.yaml
@@ -1,0 +1,121 @@
+name: Water quality monitor
+products:
+  - id: muwcszgae41ccndk
+    model: W-2839
+entities:
+  - entity: sensor
+    class: temperature
+    icon: "mdi:water-thermometer"
+    dps:
+      - id: 8
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+        mapping:
+          - scale: 10
+  - entity: sensor
+    class: ph
+    dps:
+      - id: 106
+        type: integer
+        name: sensor
+        class: measurement
+        mapping:
+          - scale: 100
+          - dps_val: 1500
+            value: null
+  - entity: sensor
+    name: Oxidation reduction potential
+    icon: "mdi:virus-off"
+    dps:
+      - id: 131
+        type: integer
+        name: sensor
+        unit: mV
+        class: measurement
+        optional: true
+  - entity: number
+    translation_key: maximum_temperature
+    class: temperature
+    category: config
+    dps:
+      - id: 102
+        type: integer
+        name: value
+        unit: C
+        optional: true
+        range:
+          min: -100
+          max: 1100
+        mapping:
+          - scale: 10
+  - entity: number
+    translation_key: minimum_temperature
+    class: temperature
+    category: config
+    dps:
+      - id: 103
+        type: integer
+        name: value
+        unit: C
+        optional: true
+        range:
+          min: -100
+          max: 1100
+        mapping:
+          - scale: 10
+  - entity: number
+    name: Maximum pH
+    class: ph
+    category: config
+    dps:
+      - id: 107
+        type: integer
+        name: value
+        optional: true
+        range:
+          min: 0
+          max: 1500
+        mapping:
+          - scale: 100
+  - entity: number
+    name: Minimum pH
+    class: ph
+    category: config
+    dps:
+      - id: 108
+        type: integer
+        name: value
+        optional: true
+        range:
+          min: 0
+          max: 1500
+        mapping:
+          - scale: 100
+  - entity: number
+    name: Maximum ORP
+    category: config
+    icon: "mdi:virus-off"
+    dps:
+      - id: 132
+        type: integer
+        name: value
+        unit: mV
+        optional: true
+        range:
+          min: -2000
+          max: 2000
+  - entity: number
+    name: Minimum ORP
+    category: config
+    icon: "mdi:virus-off"
+    dps:
+      - id: 133
+        type: integer
+        name: value
+        unit: mV
+        optional: true
+        range:
+          min: -2000
+          max: 2000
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/w2839_pool_monitor.yaml
```diff
@@ -0,0 +1,129 @@
+name: W-2839 water quality monitor
```
make-all: Top level name should be without any branding. Keep the model and manufacturer info under products only.

custom_components/tuya_local/devices/w2839_pool_monitor.yaml
```diff
@@ -0,0 +1,129 @@
...
+products:
+  - id: muwcszgae41ccndk
+    manufacturer: Tuya
```
make-all: Tuya is not an end product manufacturer, just leave this off if unknown.

custom_components/tuya_local/devices/w2839_pool_monitor.yaml
```diff
@@ -0,0 +1,129 @@
...
+      - id: 8
+        type: integer
+        name: sensor
+        unit: C
+        class: measurement
+        mapping:
+          - scale: 10
+  - entity: sensor
+    class: ph
+    icon: "mdi:ph"
```
make-all: You shouldn't need an icon here, the class will define one (temperature above is OK, because the icon you specified is a more precise icon than there general thermometer icon the class would use).

custom_components/tuya_local/devices/w2839_pool_monitor.yaml
```diff
@@ -0,0 +1,129 @@
...
+    dps:
+      - id: 131
+        type: integer
+        name: sensor
+        unit: mV
+        class: measurement
+        optional: true
+  - entity: number
+    name: High temperature warning
+    category: config
```
make-all, line 41: Also add `class: temperature` here.

It may be better for consistency with other configs to use `translation_key: maximum_temperature` here instead of the explicit name and icon, and `minimum_temperature` below. Then the other settings should probably use the same "Maximum ..."/"Minimum ..." naming to match.

custom_components/tuya_local/devices/w2839_pool_monitor.yaml
```diff
@@ -0,0 +1,129 @@
...
+    icon: "mdi:thermometer-chevron-up"
+    dps:
+      - id: 102
+        type: integer
+        name: value
+        force: true
```
make-all: Is this `force: true` really effective? It can be problematic so if it is not having the effect you expected, it would be better to leave it out. Mostly it only works for some smartplug sensors, and rarely for config options like this.

make-all (changes requested):

tkoenig:
Thanks, for the review.

After further research,  I’d like to clarify one naming/safety detail.

For the W-2839 these config numbers are not measured max/min history values. The Tuya API names are:

- `temp_warn_max` / `temp_warn_min`
- `ph_warn_max` / `ph_warn_min`
- `orp_warn_max` / `orp_warn_min`

The Tuya app labels them as high/low alarm settings, e.g. “PH Hohe Alarmeinstellung” / “PH Niedrige Alarmeinstellung”.

Importantly, the W-2839 built-in power outlets/relays are not directly switchable; instead, they are triggered indirectly by these pH/ORP min/max alarm threshold values. So changing these numbers can affect physical outputs such as dosing pumps or chlorinators if connected.

So while I renamed them to `Maximum ...` / `Minimum ...` for consistency with the newer style, I’m concerned this could make them look like harmless readable statistics. Would you prefer keeping the clearer `High ... warning` / `Low ... warning` names here, matching the existing `phw218_waterquality_monitor`
config?

I also removed `force: true`. These threshold DPs are writable, but the device does not reliably report them after restart; `updatedps` did not make them consistently readable, so they may show `unknown` until written again.

References:

- HA Community W-2839 datapoints / ESPHome discussion:
  https://community.home-assistant.io/t/pool-ph-orp-tuya-w2839-modify-datapoint-how/596261
- Poolpowershop post saying the W-2839 sockets can only be switched indirectly via pH/ORP min/max values:
  https://www.poolpowershop-forum.de/forum/thread/1159265-w2839-neuer-3-in-1-wifi-orp-ph-und-temperatur-digital-monitor-detector-zum-aufba/?postID=1365263#post1365263

make-all:
They are configurable numbers, I don't see how they can be confused for "history values". They are setting minimum and maximum allowable values, beyond which the device will shutoff or alarm sound. Many devices have such settings.

PR #5063, opened 2026-05-14
Add Holtop CFA-500t ERV

gergely-sallai:
New device: Holtop CFA-500t energy recovery ventilator (product id 7788bnexpeq7j8or).

All 27 DPS mapped and confirmed via Tuya Smart Life app and local discovery.
Two entries are educated guesses:
- dp101 (heating active): inferred by symmetry with dp111 (heating function enable) — could not trigger during testing
- dp107 (factory reset): identified by elimination; same DPS ID used for factory reset in swimfun_inverboost_poolheatpump

The CFA-250 and CFA-350 models (both T and C mounting variants) are expected to share this configuration as part of the same product family, but could not be verified.

AI-assisted (Claude) with full understanding and hands-on verification of all confirmed DPS by the submitter.


make-all:
Many of these primitive entities could be combined into a climate entity. One thing that would be worth investigating is to check if the device supports setting two dps at once so the main switch and heating function switch could be combined into an hvac_mode supporting "off", "heat" and "fan_only". Apart from that, the indoor temperature could be the current_temperature, the electric heater temperature the temperature, and heating_active the hvac_action, and operating mode the preset_mode.

gergely-sallai:
I can try make these changes, but I'm not super convinced this is the right direction to take.
Do you have examples where an ERV is implemented as a Climate device?
I've took my inspiration roughly from: custom_components/tuya_local/devices/reventon_holtop_smart_erv.yaml
And what I'm doing is a custom card like this:
image


This logical joining of the ERV and a Climate device feels a bit "forced". The heating mode is more like a "enable heating to happen on demand" and not a "heat now" switch (although I haven't been able to validate this claim, due to warmer weather). The electric heater temperature is albeit very similar, but not a target temp to reach as in the sense of the Climate entity; it is only a "background" information as to until what point to heat try to heat up the incoming air.
So yes indeed, very similar to a climate entity, but still feels semantically different enough to have second thoughts.
Do you have an example where this approach is used for an ERV device?

And also should this be merged and we might experiment in a different PR?

gergely-sallai:
@make-all any input for my reasoning?

Also are the climate device changes a requirement or a suggestion for it to be merged?

make-all:
Ventilation is the V in HVAC, and climate certainly supports ventilation systems too. Normally for plain ventilation I would recommend a fan, but the heating feature looks like it would also fit into a climate entity, which is why I recommended that.
I don't understand how it feels forced to be honest. It seems you are stuck on designing your own custom UI for it, which perhaps means this is not the right integration for you.

PR #5075, opened 2026-05-16
Add mapping for code "0" and re-divine "class" to "state_class" for tonepie_t1pro_catlitterbox_max

djremco:
The connection to my cats litter-box was unstable because of the missing mapping, after adding the mapping of code "0" the connection was stable again. The "class: measurement"change to "state_class: measurement" is just wanted by Home Assistant.

### DIFF
--- a/custom_components/tuya_local/devices/tonepie_t1pro_catlitterbox_max.yaml
+++ b/custom_components/tuya_local/devices/tonepie_t1pro_catlitterbox_max.yaml
@@ -235,7 +235,7 @@ entities:
   - entity: sensor
     name: Total visits
     icon: "mdi:counter"
-    class: measurement
+    state_class: measurement
     category: diagnostic
     dps:
       - id: 7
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/tonepie_t1pro_catlitterbox_max.yaml
```diff
@@ -235,7 +235,7 @@ entities:
   - entity: sensor
     name: Total visits
     icon: "mdi:counter"
-    class: measurement
```
make-all, line 238: This should be removed. The class defined here is device class, not state class.

custom_components/tuya_local/devices/tonepie_t1pro_catlitterbox_max.yaml
```diff
@@ -253,7 +253,7 @@ entities:
         type: integer
         name: sensor
         unit: kg
-        class: measurement
+        state_class: measurement
```
make-all, line 256: This should remain as `class`, as the class defined at sensor dp level is the state class.

make-all (changes requested):

Issue #4491, opened 2026-02-12
Request support for Arlec Grid Connect smart pet feeder with camera 5l (PFC002HA Series 2)


kdog31:
### Log message

```text
This error originated from a custom integration.

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:435
integration: Tuya Local (documentation, issues)
First occurred: 10:25:17 PM (3 occurrences)
Last logged: 10:44:07 PM

Cloud device spec: [{"id": 103, "name": "basic_flip", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 106, "name": "motion_sensitivity", "type": "Enum", "format": "{\"range\":[\"0\",\"1\",\"2\"]}", "enumMap": {}}, {"id": 108, "name": "basic_nightvision", "type": "Enum", "format": "{\"range\":[\"0\",\"1\",\"2\"]}", "enumMap": {}}, {"id": 134, "name": "motion_switch", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 150, "name": "record_switch", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 151, "name": "record_mode", "type": "Enum", "format": "{\"range\":[\"1\",\"2\"]}", "enumMap": {}}]
```

### Product ID

cmubbmrw5ibt5gkl

### Product Name

Grid Connect Smart Pet feeder with camera 5l

### DPS information

```text
{
  "result": {
[...]

PR #5096, opened 2026-05-19
Adding support for PFC002HA Series 2

kdog31:
Adding support for PFC002HA Series 2 (Arlec grid connect 5L smart pet feeder with camera)

https://www.bunnings.com.au/arlec-5l-grid-connect-smart-pet-feeder-with-camera_p0397261

Potentially also adds support for Anko / Mirabella Genio 5L smart pet feeder with camera as they appear externally as the same device, however i do not have access to one to test

https://www.target.com.au/p/pet-feeder-with-camera-5l-anko/68955663

Initial support request #4491 ignored without much explanation, so did it myself. Devices builtin scheduler is trash and misses feeds so HA support is needed.

### DIFF
--- a/custom_components/tuya_local/devices/arlec_pfc002ha_petfeeder_camera.yaml
+++ b/custom_components/tuya_local/devices/arlec_pfc002ha_petfeeder_camera.yaml
@@ -69,14 +69,19 @@ entities:
   - entity: number
     translation_key: manual_feed
     dps:
-      - id: 203
+      - id: 201
         type: integer
         optional: true
         name: value
+        persist: false
         unit: portions
         range:
           min: 1
           max: 12
+        mapping:
+          - dps_val: null
+            value: 0
+            hidden: true
   - entity: sensor
     name: Feed result
     class: enum
--- a/custom_components/tuya_local/devices/arlec_pfc002ha_petfeeder_camera.yaml
+++ b/custom_components/tuya_local/devices/arlec_pfc002ha_petfeeder_camera.yaml
@@ -98,33 +98,6 @@ entities:
             value: 0
             hidden: true
   - entity: sensor
-    name: Feed result
-    class: enum
-    icon: "mdi:paw"
-    category: diagnostic
-    dps:
-      - id: 201
-        type: integer
-        optional: true
-        name: sensor
-        mapping:
-          - dps_val: -2001
-            value: failed
-          - dps_val: -2000
-            value: feeding
-          - dps_val: 2001
-            value: failed
-          - dps_val: 2000
-            value: feeding
-          - dps_val: 0
-            value: idle
-          - dps_val: null
-            value: idle
-      - id: 201
-        type: integer
-        optional: true
-        name: portions
-  - entity: sensor
     name: Portion size
     class: weight
     category: diagnostic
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/arlec_pfc002ha_petfeeder_camera.yaml
```diff
@@ -0,0 +1,127 @@
...
+          - dps_val: null
+            value: 0
+            hidden: true
+  - entity: sensor
+    name: Feed result
+    class: enum
+    icon: "mdi:paw"
+    category: diagnostic
+    dps:
+      - id: 201
```
make-all: This is the same dp id as used by manual feed above, but the range seems different. This may cause issues if not handled in the number entity above.

kdog31: Sensor has been removed from configuration entirely. Results for initial testing were good but have proven flakey at best over the past few days. Also added in mapping for manual_feed.

kdog31:
Removed faulty feed sensor, initial testing was good but longer term testing showed it to be hit and miss. Added some DPS mappings to manual_feed entity.

custom_components/tuya_local/devices/arlec_pfc002ha_petfeeder_camera.yaml
```diff
@@ -0,0 +1,122 @@
...
+    name: Portion size
+    class: weight
+    category: diagnostic
+    dps:
+      - id: 202
+        type: integer
+        name: sensor
+        unit: g
+        class: measurement
+        range:
```
make-all: Range not needed for a read-only sensor

custom_components/tuya_local/devices/arlec_pfc002ha_petfeeder_camera.yaml
```diff
@@ -0,0 +1,122 @@
...
+      - id: 151
+        type: string
+        name: option
+        mapping:
+          - dps_val: "1"
+            value: Event
+          - dps_val: "2"
+            value: Continuous
+  - entity: number
+    translation_key: manual_feed
```
make-all, line 70: Based on the info on #4491, I would expect dp 203 to be manual feed, and this to be a feedback sensor (despite being documented as rw, the description suggests it should be ro).

Is this actually working when using this config?

make-all (changes requested):

custom_components/tuya_local/devices/arlec_pfc002ha_petfeeder_camera.yaml
kdog31: Yes, 201 is working to trigger a feed on my device, this is my current running config. This confused me for a while as well as the mappings suggest it would be 203. I also cannot see any meaningful data coming out of 203, or 204, 206 or 208 for that matter. Maybe generic firmware that the grid connect device does not use? I cannot speak for the Mirabella Genio variant of what seems to be the same device however.

custom_components/tuya_local/devices/arlec_pfc002ha_petfeeder_camera.yaml
kdog31: range has been removed in [08a17e9](https://github.com/make-all/tuya-local/pull/5096/commits/08a17e99b23e77c011b2f3233276b61464474ad3)

PR #5099, opened 2026-05-19
Add Varpoolfaye 5kW Pool Heat Pump support

DawidSu:
Add device configuration for Varpoolfaye 5kW Pool Heat Pump


custom_components/tuya_local/devices/varpoolfaye_5kw_pool_heatpump.yaml
```diff
@@ -0,0 +1,71 @@
...
+        name: temperature
+        range:
+          min: 5
+          max: 40
+      - id: 3
+        type: integer
+        name: current_temperature
+      - id: 4
+        type: string
+        name: preset_mode
```
make-all: Since this only repeats the same settings that are available through hvac_mode, it does not need to be exposed as preset_mode. Rename to something else so that the hvac?mode mapping is the only way it is exposed to the HA climate API.

custom_components/tuya_local/devices/varpoolfaye_5kw_pool_heatpump.yaml
```diff
@@ -0,0 +1,71 @@
+name: Varpoolfaye 5kW Pool Heat Pump
```
make-all: Top level name should be generic, branding is already listed under products.

make-all (changes requested):

PR #5108, opened 2026-05-22
feat: add Local IP diagnostic sensor for all devices

pranjal-joshi:
Adds a diagnostic sensor entity that shows the LAN IP address of each Tuya device.

## What this does

- New TuyaLocalIPSensor (entity_category: diagnostic) is automatically created for every device
- Reads the IP from tinytuya's device object, so it shows the resolved address even when configured as 'auto'
- For sub-devices (gateway children), shows the parent/gateway IP
- Sensor platform is now always forwarded so this works even for devices that don't define any DPS-based sensors

## Changes

- sensor.py - new TuyaLocalIPSensor class + updated sync_setup_entry
- __init__.py - always include sensor platform in forwarded entities
- 	ranslations/en.json - added local_ip translation key
- icons.json - added mdi:ip-network icon

Error handling is in place so this won't break anything if the address isn't available for some reason.


make-all:
This has nothing to do with the tuya local protocol, and is just surfacing non-changing config info as a sensor.

pranjal-joshi:
@make-all 
This is simple yet useful when you're referring to IP address of certain devices, especially the once configured in "Auto" mode
Everytime looking up for the IP in file system is a cumbersome task and adding a diagnostic sensor can help in my opinion.
Please feel free to reopen this if you think this adds value and make life easier.

make-all:
The config step now resolves the IP address at configuration time if possible, so you can just open the config to see what the IP address was configured as.

pranjal-joshi:
@make-all I am already using the latest version and still it shows me Auto without any specific resolved IP, for all my devices that uses Auto behaves like this.
image

pranjal-joshi:
@make-all Is there a bug in the config flow method for resolving IP? Please check the screenshot posted above.

make-all:
No, I was mistaken, it is only the protocol that gets resolved, the IP address may need to remain as Auto in case it is not static

pranjal-joshi:
> No, I was mistaken, it is only the protocol that gets resolved, the IP address may need to remain as Auto in case it is not static

In that case, can we re-open the PR as it will add resolved IP as a diagnostic sensor for each device. Could help a lot while working with home networking.

PR #5112, opened 2026-05-23
Add Gadnic AC800 robot vacuum support

DEFALTUSER-24:
## Summary

Adds support for the Gadnic AC800 robot vacuum.
The profile was created manually using the tuya developer platform and by configuring the propierties based on examples from other robot vacuums. Tested with the tuya_local integration and real hardware, I can confirm that it works correctly.

## Device details

- Manufacturer / brand: Gadnic
- Model: AC800
- Device type: Vacuum

### Configuration
- Auto-boost mode
- Collect dust button (activates the base dust collection feature)
- Custom mode switch
- Dust collection frequency
- Edge clean mode switch (uses its own dp_id in Tuya)
- Mop water volume adjustment selector
- Setting to resume cleaning from where the robot stopped during the last cleaning job
- Switch to locate the vacuum
- Cleaning mode selector
- Do Not Disturb switch
- Buttons to reset the map, rolling brush, side brush, and filter
- Spiral cleaning mode switch (uses its own dp_id in Tuya)
- Volume settings
- Y-pattern mopping mode

### Diagnostics

Battery, cleaned area, cleaning time, mop installed status, current status (including whether it has an issue), remaining rolling brush lifetime, remaining side brush lifetime, total cleaned area, total cleaning time, total cleaning sessions, and filter lifetime.

## Notes

The device profile was validated with real hardware. Home Assistant logs are clean and the entities are exposed correctly.

### DIFF
--- a/custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
+++ b/custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
@@ -1,0 +1,371 @@
+name: Robot vacuum
+products:
+  - id: opn3uutjlhbeojab
+    manufacturer: Gadnic
+    model: AC800
+entities:
+  - entity: vacuum
+    dps:
+      - id: 1
+        type: boolean
+        name: activate
+      - id: 2
+        type: boolean
+        name: pause
+        optional: true
+      - id: 3
+        type: boolean
+        name: switch_charge
+        optional: true
+      - id: 4
+        type: string
+        name: command
+        optional: true
+        mapping:
+          - dps_val: goto_charge
+            value: return_to_base
+      - id: 5
+        type: string
+        name: status
+        mapping:
+          - dps_val: standby
+            value: idle
+          - dps_val: smart
+            value: cleaning
+          - dps_val: zone_clean
+            value: cleaning
+          - dps_val: part_clean
+            value: cleaning
+          - dps_val: cleaning
+            value: cleaning
+          - dps_val: paused
+            value: paused
+          - dps_val: goto_pos
+            value: cleaning
+          - dps_val: pos_arrived
+            value: idle
+          - dps_val: pos_unarrive
+            value: idle
+          - dps_val: goto_charge
+            value: returning
+          - dps_val: charging
+            value: docked
+          - dps_val: charge_done
+            value: docked
+          - dps_val: sleep
+            value: idle
+      - id: 9
+        type: string
+        name: fan_speed
+        mapping:
+          - dps_val: gentle
+            value: "off"
+          - dps_val: normal
+            value: medium
+          - dps_val: strong
+            value: high
+          - dps_val: max
+            value: maximum
+      - id: 12
+        type: string
+        name: direction_control
+        mapping:
+          - dps_val: forward
+            value: forward
+          - dps_val: backward
+            value: backward
+          - dps_val: turn_left
+            value: left
+          - dps_val: turn_right
+            value: right
+          - dps_val: stop
+            value: stop
+  - entity: switch
+    name: locate
+    icon: "mdi:map-marker-alert"
+    category: config
+    dps:
+      - id: 11
+        type: boolean
+        name: switch
+  - entity: select
+    translation_key: cleaning_mode
+    icon: "mdi:vacuum"
+    category: config
+    dps:
+      - id: 41
+        type: string
+        name: option
+        mapping:
+          - dps_val: both_work
+            value: sweep_and_mop
+          - dps_val: only_mop
+            value: mop
+          - dps_val: only_sweep
+            value: sweep
+  - entity: select
+    translation_key: mopping
+    category: config
+    dps:
+      - id: 10
+        type: string
+        name: option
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
```diff
@@ -0,0 +1,369 @@
...
+      - id: 1
+        type: boolean
+        name: activate
+      - id: 2
+        type: boolean
+        name: pause
+        optional: true
+      - id: 3
+        type: boolean
+        name: switch_charge
```
make-all: This is not part of the vacuum entity model, so will only act as a read-only extra attribute here. If it is a switch, then it should be its own entity.

custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
```diff
@@ -0,0 +1,369 @@
...
+        optional: true
+      - id: 4
+        type: string
+        name: command
+        optional: true
+        mapping:
+          - dps_val: goto_charge
```
make-all, line 23: Is this really the only command supported by the vacuum?

custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
```diff
@@ -0,0 +1,369 @@
...
+          - dps_val: backward
+            value: backward
+          - dps_val: turn_left
+            value: left
+          - dps_val: turn_right
+            value: right
+          - dps_val: stop
+            value: stop
+  - entity: switch
+    name: locate
```
make-all: The vacuum entity supports "locate" functionality natively, so this can be moved inside there.

custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
```diff
@@ -0,0 +1,369 @@
...
+      - id: 26
+        type: integer
+        name: value
+        unit: "%"
+        range:
+          min: 0
+          max: 100
+  - entity: switch
+    translation_key: do_not_disturb
+    icon: "mdi:bell-cancel"
```
make-all: In general icons should not need to be overridden for translated entities.

make-all (changes requested):

custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
DEFALTUSER-24: `switch_charge` is a setting that lets you “pause” the “returning to charging station” state. I removed it from the vacuum entity because you’re right, it doesn’t really make sense there. But if you agree, I think it could still be a good option to keep in the device configuration.

custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
DEFALTUSER-24: You were right about the commands. I checked the tuya dev platform and found the values I was missing so I added all of them. I should note that while DP 4 officially holds the `pose (clean_spot)` value, when testing on the actual hardware the vacuum completely ignores it because the manufacturer hardwired the Spiral Clean and Edge Clean behaviors to separate independent boolean switches instead (DP 102 and DP 101 respectively).

Because of this, the physical robot will not react to that specific native command. It requires triggering the independent `switch.spiral_clean` entity instead.

custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
DEFALTUSER-24: I understand your point about the locate button. I left it out of the vacuum entity because I thought the device would keep beeping indefinitely but it actually only plays the sound around 9 times before stopping automatically (I just found out about this). 

I think that this can still be annoying because if I find the vacuum before it finishes, I’d like have a way to stop the locate sound directly from the settings. So, if you agree, I think it would be good to add the Locate switch in the vacuum entity while also keeping an option in the configuration to disable it if the user wants.

custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
DEFALTUSER-24: Removed. Sorry about that.

custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
make-all: Yes, it should be in the config as its own switch entity, that was what I meant.

custom_components/tuya_local/devices/gadnic_ac800_vacuum.yaml
make-all: Nobody has ever raised that before, I think most vacuums will silence when picked up or the button pressed, which is more convenient than the HA UI when you are searching for the vacuum.

DEFALTUSER-24:
Done. Replaced the locate switch with the charge control switch. Thank you for your feedback.

PR #5114, opened 2026-05-23
feat(devices): add venta ah515 config

iakovmarkov:
Adds Venta AH515 support.

This config is based on the existing `venta_ah510_humidifier` device config. AH515 uses the same Tuya product/DPS layout, but needs its own model identifier and corrected mode semantics. 

It might be a good idea to use the same mode semantic in AH510 - I think the model doesn't have anything like "turbo" mode, it's the same thing as my device with only three modes - normal/manual, sleep and auto.

Changes:
- Added venta_ah515_humidifier
- Set model_id to AH515
- Kept the AH510 DPS layout
- Adjusted mode mapping based on tested AH515 behavior

Testing:
Tested locally with a physical Venta AH515. Before the change, Tuya Local failed with `Configuration file for venta_ah515_humidifier not found`. After the change, the device loads and the main controls work.

### DIFF
--- a/custom_components/tuya_local/devices/venta_ah515_humidifier.yaml
+++ b/custom_components/tuya_local/devices/venta_ah515_humidifier.yaml
@@ -1,0 +1,104 @@
+name: Humidifier
+products:
+  - id: ecihf2xeg7ghlfnh
+    manufacturer: Venta
+    model: Original Connect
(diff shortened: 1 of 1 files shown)


make-all:
Even the product id is the same as AH510, so there is no need to create a new config

PR #5120, opened 2026-05-24
C500 improvements

vemboy200:
Heres what i did

1. changed mapping for the air quality entity
2. added unhealthy icon for air quality entities
3. added icon for the timer switch entity
4. added select entity for fan speed (was in xtend tuya)(got help from ai adding this)

The mappings i changed better reflect what is being shown in the tuya/smartlife app, and yes the dps val of good is actually "poor" in the app, most of these dps vals are inaccurate to what is actually displayed in the app.

### DIFF
--- a/custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
+++ b/custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
@@ -116,6 +116,7 @@ entities:
           - dps_val: medium
             value: unhealthy
   - entity: switch
+    translation_key: timer
     name: Timer
     category: config
     dps:
--- a/custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
+++ b/custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
@@ -117,7 +117,7 @@ entities:
             value: severe
   - entity: switch
     name: Timer
-    icon: mdi:timer
+    icon: "mdi:timer"
     category: config
     dps:
       - id: 101
--- a/custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
+++ b/custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
@@ -116,7 +116,6 @@ entities:
           - dps_val: medium
             value: unhealthy
   - entity: switch
-    translation_key: timer
     name: Timer
     category: config
     dps:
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
```diff
@@ -32,6 +32,26 @@ entities:
             value: 80
           - dps_val: max
             value: 100
+  # The fan speeds are reffered to as sleep,1,2,3,4 in he app but idk how to convert number into string so the code doesnt freak out
+  #I think a translation key is also needed
+  - entity: select
```
make-all: This should be removed, as the speed is already exposed in the fan entity.

custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
```diff
@@ -108,15 +128,16 @@ entities:
...
-            value: excellent
+            value: good
           - dps_val: mild
             value: moderate
           - dps_val: good
-            value: good
-          - dps_val: medium
             value: poor
+          - dps_val: medium
+            value: unhealthy
```
make-all: New translated values should not be introduced when existing ones can be used.

custom_components/tuya_local/icons.json
```diff
@@ -789,6 +789,7 @@
                     "moderate": "mdi:emoticon-neutral",
                     "poor": "mdi:emoticon-sad",
                     "severe": "mdi:emoticon-sick",
+                    "unhealthy": "mdi:emoticon-sick",
```
make-all: Not required, as severe can be used instead

make-all (changes requested):

custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
```diff
@@ -108,15 +108,16 @@ entities:
...
   - entity: switch
     name: Timer
+    icon: mdi:timer
```
make-all: icon name needs quoting, as it contains embedded :

custom_components/tuya_local/devices/homemedics_c500_airpurifier.yaml
```diff
@@ -108,15 +108,16 @@ entities:
...
-            value: excellent
+            value: good
           - dps_val: mild
             value: moderate
           - dps_val: good
-            value: good
-          - dps_val: medium
             value: poor
+          - dps_val: medium
+            value: unhealthy
```
make-all: severe

PR #5121, opened 2026-05-24
DAYETTE Air Purifier AP 409

bapesupreme:
added control for DAYETTE Air Purifier AP 409

controls work and mirrors app, unsure about scheduling and if can be controlled in HA, unsure about sensors


bapesupreme:
@make-all 

Hello, should be fixed now.

bapesupreme:
hopefully this is better and passes

the test will tell me if optional: true is a valid DPS-level key in the schema

bapesupreme:
@make-all didn't realize I could run the workflows in my fork

should pass now

image

custom_components/tuya_local/devices/air_purifier_ap409.yaml
```diff
@@ -0,0 +1,109 @@
...
+        type: boolean
+        name: switch
+      - id: 3
+        type: string
+        name: preset_mode
+        optional: true
+        mapping:
+          - dps_val: auto
+            value: auto
+            default: true
```
make-all: default: true has no meaning here

custom_components/tuya_local/devices/air_purifier_ap409.yaml
```diff
@@ -0,0 +1,109 @@
...
+        mapping:
+          - dps_val: auto
+            value: Auto
+            default: true
+          - dps_val: sleep
+            value: Sleep
+          - dps_val: manual
+            value: Manual
+  - entity: select
+    name: Fan speed
```
make-all: This should be turned into a speed attribute on the fan entity.

custom_components/tuya_local/devices/air_purifier_ap409.yaml
```diff
@@ -0,0 +1,109 @@
...
+        mapping:
+          - dps_val: auto
+            value: auto
+            default: true
+          - dps_val: sleep
+            value: sleep
+          - dps_val: manual
+            value: manual
+  - entity: select
+    translation_key: mode
```
make-all: This is duplicating the preset mode above, so should be removed

custom_components/tuya_local/devices/air_purifier_ap409.yaml
```diff
@@ -0,0 +1,109 @@
...
+        unit: "%"
+  - entity: select
+    translation_key: timer
+    category: config
+    dps:
+      - id: 18
+        type: string
+        name: option
+        mapping:
+          - dps_val: cancel
```
make-all, line 60: The values should match the options under the timer translation key, which actually the same as the dps_val in this case

custom_components/tuya_local/devices/air_purifier_ap409.yaml
```diff
@@ -0,0 +1,109 @@
...
+          - dps_val: 2h
+            value: 2 Hours
+          - dps_val: 4h
+            value: 4 Hours
+          - dps_val: 6h
+            value: 6 Hours
+          - dps_val: 8h
+            value: 8 Hours
+  - entity: sensor
+    name: Countdown left
```
make-all: Use `translation_key: time_remaining` instead of this name

custom_components/tuya_local/devices/air_purifier_ap409.yaml
```diff
@@ -0,0 +1,109 @@
...
+        unit: min
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 7
+        type: boolean
+        name: lock
+  - entity: switch
+    name: Filter reset
```
make-all: This should be a button rather than a switch (it makes no sense to un-reset), in which case `translation_key: filter_reset` can be used instead of the name.

make-all (changes requested):

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: will implement

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: will implement

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: will implement

bapesupreme:
@make-all 

I'm not sure the best way to implement this.

The Smart Life app puts Pet mode under Speed.

Under Speed, you can choose Low, Medium, High, Pet.

When device is turned on, Auto is the default preset mode.

You can choose between Auto, Sleep, and Speed.

There is no "Manual mode" you can select, instead it displays Manual when you select a speed.

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: Since DPS 4 uses named string values rather than a numeric range, mapping them to percentages (25/50/75/100) works technically but results in a percentage slider in the HA UI, which doesn't reflect how the device actually works. A standalone select entity with named options (Low, Medium, High, Pet) is more intuitive and accurately represents the device's capabilities.

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: The device has two separate datapoints:

DPS 3 — mode: auto, sleep, manual
DPS 4 — fan speed: Low, Medium, High, Pet

Pet is a fan speed on DPS 4, not a mode on DPS 3. When any speed is selected, the device reports manual on DPS 3. There is no way to select Pet via DPS 3.

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: will implement

bapesupreme:
Here's the best I've come up with so far.

`name: Air purifier
products:
  - id: rufdceynizoy3e2p
    manufacturer: Dayette
    model: AP409
entities:
  - entity: fan
    translation_key: air_purifier
    dps:
      - id: 1
        type: boolean
        name: switch
      - id: 3
        type: string
        name: preset_mode
        mapping:
          - dps_val: auto
            value: auto
            default: true
          - dps_val: sleep
            value: sleep
          - dps_val: manual
            value: manual
  - entity: select
    translation_key: mode
    dps:
      - id: 3
        type: string
        name: option
        mapping:
          - dps_val: auto
            value: Auto
            default: true
          - dps_val: sleep
            value: Sleep
          - dps_val: manual
            value: Manual
  - entity: select
    name: Fan speed
    dps:
      - id: 4
        type: string
        name: option
        mapping:
          - dps_val: Low
            value: Low
          - dps_val: Medium
            value: Medium
          - dps_val: High
            value: High
          - dps_val: Pet
            value: Pet
  - entity: sensor
    class: pm25
    dps:
      - id: 2
        type: integer
        name: sensor
        class: measurement
        unit: ugm3
  - entity: sensor
    translation_key: filter_life
    category: diagnostic
    dps:
      - id: 5
        type: integer
        name: sensor
        unit: "%"
  - entity: select
    translation_key: timer
    category: config
    dps:
      - id: 18
        type: string
        name: option
        mapping:
          - dps_val: cancel
            value: cancel
          - dps_val: 2h
            value: 2h
          - dps_val: 4h
            value: 4h
          - dps_val: 6h
            value: 6h
          - dps_val: 8h
            value: 8h
  - entity: sensor
    category: diagnostic
    dps:
      - id: 19
        type: integer
        name: sensor
        unit: min
  - entity: lock
    translation_key: child_lock
    category: config
    dps:
      - id: 7
        type: boolean
        name: lock
  - entity: button
    translation_key: filter_reset
    category: config
    dps:
      - id: 11
        type: boolean
        name: button`

image
image

custom_components/tuya_local/devices/air_purifier_ap409.yaml
make-all: Any speed mapping has finite steps. But converting to numeric speed works better in voice assistants, where you can say "increase"/"decrease" which textual speeds generally require precise target names to be used.

custom_components/tuya_local/devices/air_purifier_ap409.yaml
make-all: What is this "pet" speed? If it is a higher speed than high, then high as 75, pet as 100 makes sense.

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: The air purifier doesn’t have any actual numeric speeds to increase or decrease. Only low, medium, high, and pet.

I personally would rather say those four “speeds” with a voice assistant than increase/decrease.

If you want them mapped to percentages, I could pick three numbers for low/med/high and leave pet out. It wouldn’t make sense to map pet to a number.

I tried mapping briefly and couldn’t get the slider to switch between the four.

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: Not quite sure, I’ll see if I can find more info.

there was no manual included and the website has sparse info.

bapesupreme: @make-all 

it is a higher speed but not sure if should be used regularly 

【𝐄𝐧𝐡𝐚𝐧𝐜𝐞𝐝 𝐏𝐞𝐭 𝐌𝐨𝐝𝐞】Our air purifiers for pet owners feature an Enhanced Pet Mode. Simply turn it on to instantly activate max fan speed—rapidly capturing pet dander, hair, lingering odors, fine particles, and dust. Delivering fresh, clean air, your family & furry pals get fresh, clean air in minutes, it’s perfect after playtime or near your pet’s favorite napping spots

custom_components/tuya_local/devices/air_purifier_ap409.yaml
make-all: still waiting for the implementation of these

bapesupreme: done

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: done

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: done

custom_components/tuya_local/devices/air_purifier_ap409.yaml
bapesupreme: done

PR #5122, opened 2026-05-24
Add support for Stahlmann Smart water leak control system

dext0r:
Add device config for [Stahlmann Smart](https://stahlmann-solutions.com/stahlmann-water-leakage-protection) water leak control system


custom_components/tuya_local/devices/stahlmann_smart_waterleak.yaml
```diff
@@ -0,0 +1,201 @@
...
+    icon: "mdi:attachment-plus"
+    category: config
+    dps:
+      - id: 108
+        type: boolean
+        name: button
+  - entity: sensor
+    name: Counter 1
+    class: water
+    hidden: true
```
make-all: Why hidden? Also "Counter" implies counting discrete items, but this has units of m³, so maybe "Meter", "Usage", or "Leakage" would be better (depending on what this device is measuring)

dext0r: > Why hidden? Also "Counter" implies counting discrete items, but this has units of m³, so maybe "Meter", "Usage", or "Leakage" would be better (depending on what this device is measuring)

Counter sensors require an additional card sold separately. Should I make them visible by default?

>  "Counter" implies counting discrete items, but this has units of m³, so maybe "Meter", "Usage

m³ unit and name "counter" straight from dps mapping:
```
"140": {
    "code": "counter6",
    "type": "Integer",
    "values": {
        "unit": "m\\u00B3",
        "min": 0,
        "max": 99999999,
        "scale": 0,
        "step": 1
    }
},
```

Agreed, I think "Meter" is the better name. It counts consumed water.

make-all: Since the dp is marked `optional: true`, I assume it will only be reported when it is available. In that case you can use `hidden: unavailable` above, and add another dp definition below:
```
      - id: 135
        type: integer
        optional: true
        name: available
        mapping:
          - dps_val: null
            value: false
          - value: true
```

PR #5126, opened 2026-05-25
Add support for NWT WDH-10EN dehumidifier

mmaarrkk02:


custom_components/tuya_local/translations/bg.json
```diff
@@ -418,7 +432,9 @@
                         "state": {
                             "laundry": "Сухо пране",
                             "purify": "Пречистете",
-                            "continuous": "Непрекъснато"
+                            "continuous": "Непрекъснато",
```
make-all: These changes are unnecessary.
`sleep` is already defined by [HA standard translations](https://github.com/home-assistant/core/blob/dev/homeassistant/components/humidifier/strings.json#L160).
For `dehumidify`, you can use `normal`, since dehumidifying is the normal operation for a dehumidifier.

custom_components/tuya_local/translations/bg.json
```diff
@@ -1121,6 +1137,16 @@
                     "12_hour": "12 часа",
                     "24_hour": "24 часа"
                 }
+            },
+            "fan_speed": {
```
make-all: This is unnecessary, because fan speed should be inside the fan entity.

custom_components/tuya_local/translations/bg.json
```diff
@@ -1424,6 +1450,9 @@
             },
             "snooze": {
                 "name": "Отлагане"
+            },
+            "swing_fan": {
```
make-all: This is unneccesary as a switch, as it should be inside the fan entity.

custom_components/tuya_local/devices/nwt_wdh10en_dehumidifier.yaml
```diff
@@ -0,0 +1,94 @@
...
+          min: 30
+          max: 90
+        mapping:
+          - step: 5
+      - id: 5
+        type: string
+        name: mode
+        mapping:
+          - dps_val: dehumidify
+            value: dehumidify
```
make-all: use normal here, as that is already defined as a standard humidifier mode by HA

custom_components/tuya_local/devices/nwt_wdh10en_dehumidifier.yaml
```diff
@@ -0,0 +1,94 @@
...
+          - dps_val: sleep
+            value: sleep
+          - dps_val: dry
+            value: laundry
+      - id: 6
+        type: integer
+        name: current_humidity
+  - entity: select
```
make-all: Use a fan entity for speed and oscillate (swing)

custom_components/tuya_local/devices/nwt_wdh10en_dehumidifier.yaml
```diff
@@ -0,0 +1,94 @@
...
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
+    translation_key: defrost
+    dps:
+      - id: 102
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
+    translation_key: filter_clean
```
make-all: Rather than define new translations in the same PR, as the instructions say you should separate these into two PRs with the translations only being defined after review. In this case, the binary sensor is most likely indicating a Dirty rather than clean filter, and in most cases it is just based on time not any sensor. Current practice is to make it a class: problem and just keep the name hardcoded, but if you feel strongly that this should be translated, we should consider the current similar binary_sensors in other devices that should be merged with this:

- filter (8 occurrences)
- filter change (4 occurrences)
- replace filter (3 occurrences)
- fresh air filter (2 occurrences - may be devices with multiple filters so cannot merge in general)
- dirty filter (1 occurrence)
- hepa filter replacement (1 occurrence - may be a device with multiple filters)
- filter blockage (1 occurrence - probably an actual sensor)
- filter clean required (1 occurrence)
- filter blocked (1 occurrence)
- filter clean due (1 occurrence)
- filter fault (1 occurrence)
- filter cleaning reminder (1 occurrence)

custom_components/tuya_local/devices/nwt_wdh10en_dehumidifier.yaml
```diff
@@ -0,0 +1,94 @@
...
+    dps:
+      - id: 103
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
+    translation_key: auto_close
```
make-all: What is this closing? Should there be a cover entity somewhere to represent that?

make-all (changes requested):

custom_components/tuya_local/devices/nwt_wdh10en_dehumidifier.yaml
```diff
@@ -0,0 +1,88 @@
...
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 19
+        type: bitfield
+        name: description
+        mapping:
+          - dps_val: "E1"
```
make-all: E1 and E2 are not valid values for a bitfield.

Most likely these are bits 1 and 2, so 1 and 2 should go here. You should also include a mapping for 0 to OK, for the no-error case.

make-all (changes requested):

custom_components/tuya_local/devices/nwt_wdh10en_dehumidifier.yaml
mmaarrkk02: `
{
          "abilityId": 19,
          "accessMode": "ro",
          "code": "fault",
          "description": "E1:温度sensor故障. E2:铜管sensor故障. ",
          "extensions": {
            "iconName": "icon-baojing",
            "scope": "fault",
            "attribute": "1024"
          },
          "name": "故障告警",
          "typeSpec": {
            "type": "bitmap",
            "label": [
              "E1",
              "E2"
            ],
            "maxlen": 2
          }
        }
`

make-all: The "E1" and "E2" are labels. The values will be bit values, most likely 1 and 2, but sometimes they skip bits, so not guaranteed.

PR #5127, opened 2026-05-25
Add support for S09 temperature and humidity sensor

mmaarrkk02:


make-all:
I'm not sure how important the calibration controls are, most people would not have the equipment at home to properly calibrate a sensor like this. The rest is already supported by `ir_remote_sensors.yaml`

make-all:
Modified ir_remote_sensors to include optional calibration

PR #5139, opened 2026-05-25
Add support for WS2812-based LED strip

n4dwxihyeyhes2:
New device definition for WS2812-based LED strip (Christmas tree lights) with Tuya support.
Example product: MBG Line, model CTree-001 (more technical details in Polish: https://www.elektroda.pl/news/news4094481.html).


custom_components/tuya_local/devices/ws2812_led_strip.yaml
```diff
@@ -0,0 +1,143 @@
+name: WS2812-based LED Strip
```
make-all: Top level name should be generic, as other device may match this config too.

custom_components/tuya_local/devices/ws2812_led_strip.yaml
```diff
@@ -0,0 +1,143 @@
...
+products:
+  - id: bf0fa4b46ce7b80518guh5
```
make-all: This is not a product id.

make-all: Product id for WiFi devices should be 16 characters

custom_components/tuya_local/devices/ws2812_led_strip.yaml
```diff
@@ -0,0 +1,143 @@
...
+              min: 0
+              max: 1000
+  - entity: number
+    name: LED count
+    category: config
+    hidden: true
+    icon: "mdi:counter"
+    dps:
+      - id: 103
+        readonly: true
```
make-all: If this is read-only, it should not be a number entity.

custom_components/tuya_local/devices/ws2812_led_strip.yaml
```diff
@@ -0,0 +1,143 @@
...
+        type: integer
+        name: value
+        range:
+          min: 1
+          max: 200
+  - entity: select
+    name: Color mode
```
make-all: There is no need to repeat this as a separate entity

custom_components/tuya_local/devices/ws2812_led_strip.yaml
```diff
@@ -0,0 +1,143 @@
...
+            value: Preset 9
+          - dps_val: 10
+            value: Preset 10
+          - dps_val: 11
+            value: Preset 11
+  - entity: text
+    name: Custom preset data
+    category: config
+    icon: "mdi:numeric-10-box"
+    dps:
```
make-all: Add `hidden: true`, as this is not possible for normal users to make use of without understanding the hex format.

make-all (changes requested):

custom_components/tuya_local/devices/ws2812_led_strip.yaml
n4dwxihyeyhes2: I thought this name is as generic as possible. But maybe something less technical would work better, like `Multicolor LED Christmas lights`?

custom_components/tuya_local/devices/ws2812_led_strip.yaml
n4dwxihyeyhes2: Is the ID reported by the device or some random 16-character alphanumeric string?

custom_components/tuya_local/devices/ws2812_led_strip.yaml
make-all: Yes, or just "Christmas lights", since the device name gets prepended on the front of entity names in some places in the HA UI, so having a long device name will cause overflow issues for entity naming and make it difficult to differentiate them.

custom_components/tuya_local/devices/ws2812_led_strip.yaml
make-all: It is reported by the device only during discovery, it can be found in the cloud (same API calls as local key) or by `tinytuya scan`

custom_components/tuya_local/devices/ws2812_led_strip.yaml
n4dwxihyeyhes2: The goal was to have a select box that can be placed on a dashboard (which is not possible for the `effect` attribute of a light entity). Worked around with a select helper and automations for syncing state.

custom_components/tuya_local/devices/ws2812_led_strip.yaml
n4dwxihyeyhes2: Removed `readonly`, this dp is updatable.

custom_components/tuya_local/devices/ws2812_led_strip.yaml
n4dwxihyeyhes2: Done.

custom_components/tuya_local/devices/ws2812_led_strip.yaml
n4dwxihyeyhes2: Done.

custom_components/tuya_local/devices/ws2812_led_strip.yaml
n4dwxihyeyhes2: Done.

PR #5141, opened 2026-05-25
Load custom device configs from the HA config directory

guybrushthre:
## Summary

Device configs bundled with the integration live under `custom_components/tuya_local/devices/` and are overwritten every time the integration is updated. This makes it awkward to run a device config that isn't merged yet (or a local tweak) — your file disappears on the next update.

This PR adds support for a **user-managed config directory** that lives outside `custom_components/` and therefore survives updates:

```
<config>/tuya_local/devices/my_device.yaml
```

(`<config>` is the Home Assistant config directory, i.e. where `configuration.yaml` lives.)

Files placed there are loaded **in addition to** the bundled configs and matched against devices in exactly the same way, making it the recommended place to test a config before submitting it as a PR.

## Behaviour

- Custom configs are discovered alongside bundled ones by `available_configs()`, `get_config()` and `TuyaDeviceConfig`.
- If a custom file has the **same filename** as a bundled one, the custom version takes precedence (lets you locally patch a shipped config). Duplicate filenames are de-duplicated in the listing.
- A missing custom directory is silently ignored — no behaviour change for users who don't create one.
- Configs are picked up on Home Assistant (re)start.

## Implementation

Directory resolution is centralised in four small helpers in `helpers/device_config.py` — `_bundled_config_dir()`, `_custom_config_dir()`, `_config_dirs()` and `_config_path()`. Because the existing `available_configs()` / `get_config()` / `TuyaDeviceConfig` were updated to use them, none of the ~25 call sites needed to change, and there's no new `hass` plumbing.

## Tests

Added three tests to `tests/test_device_config.py` (custom config adds a new device, custom overrides bundled, missing custom dir is ignored). Full suite was run locally on Python 3.14.5:

- `tests/test_device_config.py` — 35 passed
- `tests/test_config_flow.py` `tests/test_device.py` `tests/test_diagnostics.py` — 63 passed
- `ruff check` and `ruff format --check` clean

## Docs

Documented the custom directory and override behaviour in `custom_components/tuya_local/devices/README.md`.

### DIFF
--- a/custom_components/tuya_local/devices/README.md
+++ b/custom_components/tuya_local/devices/README.md
@@ -6,6 +6,30 @@ of Tuya DPs (Data Points) to HomeAssistant attributes.
 
 Each Tuya device may correspond to one or more entities in Home Assistant.
 
+## Custom device configurations
+
+The configurations in this directory are bundled with the integration and are
+replaced whenever the integration is updated. If you want to add or tweak a
+device configuration without it being overwritten on update, place your YAML
+files in a `tuya_local/devices` directory inside your Home Assistant config
+directory (the one containing `configuration.yaml`):
+
+```
+<config>/tuya_local/devices/my_device.yaml
+```
+
+Files found there are loaded in addition to the bundled configurations and are
+matched against your devices in exactly the same way. This is the recommended
+way to test a new configuration before submitting it as a PR.
+
+If a custom file has the same name as a bundled one (for example
+`deta_fan.yaml`), the custom version takes precedence, so you can also use this
+to locally patch a bundled configuration. Note that overriding a bundled config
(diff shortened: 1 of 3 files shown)


guybrushthre:
Would be great to get some feedback as to why this was closed.

make-all:
There has been ample prior discussion on this topic. You could have looked it up before submitting an unannounced architectural change to the project.

guybrushthre:
1. I didn't find any discussion on this topic.
2. Generally speaking, a PR is a means of discussion. At minimum, it's considered polite to respond to someone who made an effort to improve, even if you disagree with it.
3. This is hardly an architectural change, but OK.

Thanks for your contributions.

Issue #5071, opened 2026-05-16
Request support for Touchstone Sideline electric fireplace


mcguires5:
### Log message

```text
{
  "home_assistant": {
    "arch": "x86_64",
    "dev": false,
    "docker": true,
    "hassio": true,
    "installation_type": "Home Assistant OS",
    "os_name": "Linux",
    "os_version": "6.12.85-haos",
    "python_version": "3.14.2",
    "timezone": "America/New_York",
    "version": "2026.5.1",
    "virtualenv": false,
    "container_arch": "amd64",
    "supervisor": "2026.05.0",
    "host_os": "Home Assistant OS 17.3",
    "docker_version": "29.3.1",
    "chassis": "vm",
    "run_as_root": true
  },
  "custom_components": {
    "dreame_vacuum": {
      "documentation": "https://github.com/Tasshack/dreame-vacuum",
      "version": "v1.0.6",
      "requirements": [
        "pillow",
        "numpy",
        "pybase64",
        "requests",
        "pycryptodome",
        "python-miio",
        "mini-racer",
        "paho-mqtt"
      ]
    },
    "renpho": {
      "documentation": "https://github.com/neilzilla/hass-renpho",
      "version": "3.0.2",
      "requirements": [
        "pycryptodome",
        "requests",
        "aiohttp",
        "voluptuous",
        "pydantic",
        "aiohttp_socks"
      ]
    },
    "battery_notes": {
      "documentation": "https://andrew-codechimp.github.io/HA-Battery-Notes/",
      "version": "3.4.6",
      "requirements": []
    },
    "spotcast": {
      "documentation": "https://github.com/fondberg/spotcast",
      "version": "v6.0.0-a15",
      "requirements": [
        "spotipy>=2.23.0",
        "spotifyaio>=0.8.8",
        "RapidFuzz>=3.10.1"
      ]
    },
    "alexa_media": {
      "documentation": "https://github.com/alandtse/alexa_media_player/wiki",
      "version": "5.15.1",
      "requirements": [
        "alexapy==1.29.20",
        "packaging>=20.3",
        "wrapt>=1.14.0",
        "dictor>=0.1.12,<0.2"
      ]
    },
    "dwains_dashboard": {
      "documentation": "https://dwainscheeren.github.io/dwains-lovelace-dashboard/",
      "version": "3.8.0",
      "requirements": []
    },
    "alarmo": {
      "documentation": "https://github.com/nielsfaber/alarmo",
[...]

PR #5150, opened 2026-05-26
Update touchstone_sidelineredblue_fireplace.yaml

### LINKED ISSUE #4366: Request support for Touchstone Sideline 40inch 80027. 57 inch sideline fury works great
### Log message

```text
WARNING: Device matches brilliant_outdoor_pirswitch with quality of 43%. DPS: {"updated_at": 1769052024.86969, "1": false, "3": 21, "5": "0", "13": "f", "14": 67, "15": 70, "101": "1", "102": "4", "103": "3", "104": "9", "106": "cancel", "107": false, "108": false, "109": "L3"}
```

### Product ID

_No response_

### Product Name

Touchstone Sideline 40 inch 80027

### DPS information

```text
{
  "result": {
    "properties": [
      {
        "code": "switch",
        "custom_name": "",
        "dp_id": 1,
        "time": 1769052878989,
        "type": "bool",
        "value": false
      },
      {
        "code": "temp_set",
        "custom_name": "",
        "dp_id": 2,
        "time": 1769052875376,
        "type": "value",
        "value": 19
      },
      {
        "code": "temp_current",
        "custom_name": "",
        "dp_id": 3,
        "time": 1769052036391,
        "type": "value",
        "value": 20
      },
      {
        "code": "level",
        "custom_name": "",
        "dp_id": 5,
        "time": 1769052121805,
        "type": "enum",
        "value": "0"
      },
      {
        "code": "temp_unit_convert",
        "custom_name": "",
        "dp_id": 13,
        "time": 1769052875376,
        "type": "enum",
        "value": "f"
      },
      {
        "code": "temp_set_f",
        "custom_name": "",
        "dp_id": 14,
        "time": 1769052875376,
        "type": "value",
        "value": 67
      },
      {
        "code": "temp_current_f",
        "custom_name": "",
        "dp_id": 15,
        "time": 1769052036394,
        "type": "value",
        "value": 68
      },
      {
        "code": "FLAMECOLOUR",
        "custom_name": "",
        "dp_id": 101,
        "time": 1769026842184,
        "type": "enum",
        "value": "1"
      },
      {
        "code": "FLAMELIGHT",
        "custom_name": "",
        "dp_id": 102,
        "time": 1769026842170,
        "type": "enum",
        "value": "4"
      },
      {
        "code": "FLAMESPEED",
        "custom_name": "",
        "dp_id": 103,
        "time": 1769026842072,
        "type": "enum",
        "value": "3"
      },
      {
        "code": "COALBEDCOLOUR",
        "custom_name": "",
        "dp_id": 104,
        "time": 1769026842498,
        "type": "enum",
        "value": "9"
      },
      {
        "code": "COALBEDLIGHT",
        "custom_name": "",
        "dp_id": 105,
        "time": 1767129133367,
        "type": "enum",
        "value": "1"
      },
      {
        "code": "TIMER",
        "custom_name": "",
        "dp_id": 106,
        "time": 1769026842089,
        "type": "enum",
        "value": "cancel"
      },
      {
        "code": "HeatDisable",
        "custom_name": "",
        "dp_id": 107,
        "time": 1769026842085,
        "type": "bool",
        "value": false
      },
      {
        "code": "Lock",
        "custom_name": "",
        "dp_id": 108,
        "time": 1769026842497,
        "type": "bool",
        "value": false
      },
      {
        "code": "ember_bed_brightness",
        "custom_name": "",
        "dp_id": 109,
        "time": 1769026842500,
        "type": "enum",
        "value": "L3"
      }
    ]
  },
  "success": true,
  "t": 1769053233071,
  "tid": "1aef2183f74411f08807be2290680212"
}


LINE BREAK LINE BREAK LINE BREAK LINE BREAK
LINE BREAK LINE BREAK LINE BREAK LINE BREAK

switch	Boolean	
"{true,false}"
temp_set	Integer	
{
  "unit": "°C",
  "min": 19,
  "max": 30,
  "scale": 0,
  "step": 1
}
level	Enum	
{
  "range": [
    "0",
    "1",
    "2"
  ]
}
temp_unit_convert	Enum	
{
  "range": [
    "c",
    "f"
  ]
}
temp_set_f	Integer	
{
  "unit": "℉",
  "min": 67,
  "max": 86,
  "scale": 0,
  "step": 1
}
FLAMECOLOUR	Enum	
{
  "range": [
    "1",
    "2",
    "3"
  ]
}
FLAMELIGHT	Enum	
{
  "range": [
    "5",
    "4",
    "3",
    "2",
    "1",
    "0"
  ]
}
FLAMESPEED	Enum	
{
  "range": [
    "1",
    "2",
    "3",
    "4",
    "5"
  ]
}
COALBEDCOLOUR	Enum	
{
  "range": [
    "1",
    "2",
    "3",
    "4",
    "5",
    "6",
    "7",
    "8",
    "9",
    "10",
    "11",
    "12"
  ]
}
COALBEDLIGHT	Enum	
{
  "range": [
    "1",
    "2",
    "3",
    "4",
    "5",
    "6",
    "7",
    "8",
    "9",
    "10",
    "11",
    "12"
  ]
}
TIMER	Enum	
{
  "range": [
    "cancel",
    "30",
    "1H",
    "2H",
    "3H",
    "4H",
    "5H",
    "6H",
    "7H",
    "8H"
  ]
}
HeatDisable	Boolean	
"{true,false}"
Lock	Boolean	
"{true,false}"
ember_bed_brightness	Enum	
{
  "range": [
    "L4",
    "L3",
    "L2",
    "L1",
    "L0",
    "L5"
  ]
}
```

Its weird. I have a 40 inch and 57 inch touchstone. Both are "sideline" models though the 57 inch is called "sideline fury." They 57 inch works almost perfectly. Some color setting are wrong but the main fucntion WORKS awesome. The 40 inch is not identified at all and I cannot select the "touchstone_sideline" option when i have to pick a device type.

Cloud assisted setup doesnt help at all. The cloud assist doesnt even find the 57 inch. Setup manually for the 57 inch works great.

mcguires5:
I previously contributed an electric_fireplace.yaml update that turned out to conflate two distinct product ID variants
After testing, the merged version has these specific issues on my unit: 
1. DP 1 is master power, not HVAC mode -- The toggles for ember and flame do nothing because power is currently tied to HVAC mode in the new device config
2. DP 102 `"0"` mapping causes phantom "off" display
3. Flame color palette differs by product ID
4. Ember color names are invented for my variant
5. Ember brightness states aren't continuous brightness — there's a "Pulsating" effect
6. DP 108 is mapped as child_lock but there does not appear to be a child lock on this model
7. Timer enum values are unique to my variant

I'd like to **revert the merged changes to electric_fireplace.yaml back to its previous state** (matching the original product ID `qhwld7e4eqvu5fbp` Sideline Elite variant), and **add a separate `touchstone_sideline_fireplace.yaml`** profile tied specifically to product ID `pkgk72uonlqlzcil` with the verified mappings for my variant.

This avoids:
- Breaking the existing Sideline Elite profile users already rely on
- Misleading users with invented child_lock and color-name mappings
- Continuous-brightness misrepresentation of the Pulsating effect
- The DP 102 = "0" phantom display issue

And it correctly handles the genuine hardware differences between product IDs that can't be reconciled with conditional mappings (since tuya-local doesn't support per-product DPS conditions).

I have stepped-test diagnostics for every claim above and a fully working profile ready to submit. Happy to break this into smaller PRs if that's easier to review.

This PR adds a separate profile for product ID pkgk72uonlqlzcil while leaving the original Sideline Elite profile (qhwld7e4eqvu5fbp) untouched

Link to issue #5071
PR #5089

### DIFF
--- a/custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
+++ b/custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
@@ -1,4 +1,4 @@
-name: Touchstone Sideline electric fireplace
+name: Electric fireplace
 products:
   - id: pkgk72uonlqlzcil
     manufacturer: Touchstone
@@ -46,13 +46,15 @@ entities:
         name: preset_mode
         mapping:
           - dps_val: "0"
-            value: "none"
+            value: none
           - dps_val: "1"
-            value: "low"
+            value: low
+            hidden: true
           - dps_val: "2"
-            value: "high"
+            value: high
+            hidden: true
           - dps_val: "3"
-            value: "auto"
+            value: auto
       - id: 2
         name: temperature
         type: integer
@@ -63,7 +65,7 @@ entities:
         mapping:
           - constraint: temperature_unit
             conditions:
-              - dps_val: "f"
+              - dps_val: f
                 range:
                   min: 60
                   max: 88
@@ -74,15 +76,15 @@ entities:
         mapping:
           - constraint: temperature_unit
             conditions:
-              - dps_val: "f"
+              - dps_val: f
                 value_redirect: temp_current_f
       - id: 13
         name: temperature_unit
         type: string
         mapping:
-          - dps_val: "c"
+          - dps_val: c
             value: C
-          - dps_val: "f"
+          - dps_val: f
             value: F
       - id: 14
         name: temp_set_f
@@ -110,18 +112,32 @@ entities:
         type: string
         name: option
         mapping:
-          - dps_val: "c"
+          - dps_val: c
             value: celsius
-          - dps_val: "f"
+          - dps_val: f
             value: fahrenheit
-
-  - entity: select
-    name: Flame color
-    icon: "mdi:fire"
+            
+  - entity: light
+    translation_key: flame
+    category: config
     dps:
-      - id: 101
+      - id: 102
+        name: brightness
         type: string
-        name: option
+        mapping:
+          - dps_val: "1"
+            value: 51
+          - dps_val: "2"
+            value: 102
+          - dps_val: "3"
+            value: 153
+          - dps_val: "4"
+            value: 204
+          - dps_val: "5"
+            value: 255
+      - id: 101
+        name: effect
+        type: string
         mapping:
           - dps_val: "1"
             value: "Red + Blue"
@@ -129,26 +145,6 @@ entities:
             value: "Red"
           - dps_val: "3"
             value: "Blue"
-
-  - entity: select
-    name: Flame brightness
-    icon: "mdi:brightness-6"
-    category: config
-    dps:
-      - id: 102
-        type: string
-        name: option
-        mapping:
-          - dps_val: "1"
-            value: "20%"
-          - dps_val: "2"
-            value: "40%"
-          - dps_val: "3"
-            value: "60%"
-          - dps_val: "4"
-            value: "80%"
-          - dps_val: "5"
-            value: "100%"
 
   - entity: select
     name: Flame speed
@@ -169,61 +165,61 @@ entities:
             value: "4"
           - dps_val: "5"
             value: "5"
-
-  - entity: select
-    name: Ember color
-    icon: "mdi:campfire"
-    dps:
-      - id: 104
-        type: string
-        name: option
-        mapping:
-          - dps_val: "1"
-            value: "1"
-          - dps_val: "2"
-            value: "2"
-          - dps_val: "3"
-            value: "3"
-          - dps_val: "4"
-            value: "4"
-          - dps_val: "5"
-            value: "5"
-          - dps_val: "6"
-            value: "6"
-          - dps_val: "7"
-            value: "7"
-          - dps_val: "8"
-            value: "8"
-          - dps_val: "9"
-            value: "9"
-          - dps_val: "10"
-            value: "10"
-          - dps_val: "11"
-            value: "11"
-          - dps_val: "12"
-            value: "12"
-
-  - entity: select
-    name: Ember brightness
-    icon: "mdi:brightness-7"
+            
+  - entity: light
+    translation_key: embers
     category: config
     dps:
       - id: 109
+        name: brightness
         type: string
-        name: option
+        optional: true
         mapping:
-          - dps_val: "L5"
-            value: "Off"
-          - dps_val: "L0"
-            value: "25%"
-          - dps_val: "L1"
-            value: "50%"
-          - dps_val: "L2"
-            value: "75%"
-          - dps_val: "L3"
-            value: "100%"
-          - dps_val: "L4"
-            value: "Pulsating"
+          - dps_val: L5
+            value: 0
+          - dps_val: L0
+            value: 51
+          - dps_val: L1
+            value: 102
+          - dps_val: L2
+            value: 153
+          - dps_val: L3
+            value: 204
+          - dps_val: L4
+            value: 255
+          - dps_val: null
+            value: 0
+            hidden: true
+      - id: 104
+        name: effect
+        type: string
+        mapping:
+          - dps_val: "1"
+            value: "Orange"
+          - dps_val: "2"
+            value: "Red"
+          - dps_val: "3"
+            value: "Blue"
+          - dps_val: "4"
+            value: "Yellow"
+          - dps_val: "5"
+            value: "Green"
+          - dps_val: "6"
+            value: "Purple"
+          - dps_val: "7"
+            value: "Teal"
+          - dps_val: "8"
+            value: "Pink"
+          - dps_val: "9"
+            value: "White"
+          - dps_val: "10"
+            value: "PeachPuff"
+          - dps_val: "11"
+            value: "Color 11"
+            hidden: true
+          - dps_val: "12"
+            value: "Color 12"
+            hidden: true
 
   - entity: select
     translation_key: timer
@@ -257,3 +253,12 @@ entities:
             value: "7h"
           - dps_val: "8H"
             value: "8h"
+
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    hidden: true
+    dps:
+      - id: 108
+        type: boolean
+        name: lock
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
```diff
@@ -1,32 +1,69 @@
-name: Electric fireplace
+name: Touchstone Sideline electric fireplace
```
make-all: This change to the name is incorrect. Other brands or models of device may also match the config, so any branding needs to be kept under products matched to a specific product id

custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
```diff
@@ -1,32 +1,69 @@
...
         mapping:
-          - dps_val: false
+          - dps_val: "0"
             value: "off"
-          - dps_val: true
+          - dps_val: "1"
+            constraint: heat_disable
+            conditions:
+              - dps_val: false
+                value: "heat"
```
make-all: I would suggest to add `hidden: true` to the values under "1" and "2", so that the auto mode takes precedent.

Or maybe there is a better way to define this so that the preset_mode does not change when it is already not "off"

custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
```diff
@@ -84,40 +110,46 @@ entities:
...
-          - dps_val: c
+          - dps_val: "c"
             value: celsius
-          - dps_val: f
+          - dps_val: "f"
             value: fahrenheit
-  - entity: light
-    translation_key: flame
-    category: config
+
```
make-all: The rest of the changes below here are a step backwards

custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
```diff
@@ -232,10 +257,3 @@ entities:
             value: "7h"
           - dps_val: "8H"
             value: "8h"
-  - entity: lock
```
make-all, line 235: I suggest leaving this in but marking it as `hidden: true`. Other dps were identical to the red/orange/yellow version except for the possible values, so most likely this is too, even if it is only enabled in markets that require a child lock on heaters.

make-all (changes requested):

custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
```diff
@@ -46,13 +46,15 @@ entities:
         name: preset_mode
         mapping:
           - dps_val: "0"
-            value: "none"
+            value: none
           - dps_val: "1"
-            value: "low"
+            value: low
+            hidden: true
```
make-all: This was not what I meant above.

preset_mode should present all the options, but hvac_mode above has the same values for all options, so when the user changes to 'heat', which one should it use? It is better in that case to only expose one 'heat' as non-hidden. The others should show as heat when selected, so they should still be there, but marked `hidden: true` to hide them from the reverse mapping.

custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
mcguires5: I tried to mark it hidden true, but it still shows up in HA for me, I can disable the entity on my own HA instance so that is fine, but I just wanted to check and make sure I added the hidden marking correctly to the lock entity as it looks like it was ignored

make-all: Adding `hidden: true` only works when you are adding the device for the first time. Because you already configured the device, and added that later, it is already enabled in your config.

mcguires5:
Big update — I've confirmed DP 107 is actually the child lock on this variant, it is a "heat disable" flag as exposed in the tuya app, but it is actuated by holding the heat button on the touch panel for 10 seconds to engage a lock that prevents heat from being changed via the remote or touch panel. Verified that this combo toggles DP 107 from false → true, matching the same DP that the official Touchstone app's "Flame Only" toggle controls. So "Flame Only" in the app is really just a friendly label for engaging the child lock — both code paths write to the same DP.

This means your earlier suggestion to keep DP 108 as a (hidden) child_lock was correct in spirit, just on the wrong DP for this variant. Updated:

- DP 107 is now the visible `child_lock` lock entity (was a hidden heat_disable helper)
- DP 108 is unkown and through all testing has not changed state so I left it off now that we found out that the child lock is 107
- Removed the `fan_only` HVAC mode and `heat_disable` constraint from the climate entity — they were modeling the wrong feature. Climate is now just off/heat with low/high/auto presets. This makes sense the fireplace is not documented to have a fan only mode

Thanks for pointing toward the lock concept — turned out to be exactly the right framing, just on the other DP.

I have now fully tested all features and they all work including the newly found child lock! Thanks!

custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
mcguires5: Resolved anyway, found out 107 was the child lock thanks!

custom_components/tuya_local/devices/touchstone_sidelineredblue_fireplace.yaml
```diff
@@ -186,20 +194,10 @@ entities:
...
             value: peachpuff
           - dps_val: "11"
-            value: black
-          - dps_val: "12"
-            value: grey
-      - id: 104
-        name: effect
-        type: string
-        mapping:
+            value: peachpuff
```
make-all, line 197: My guess is that this will actually stay at the last color selected, so only peachpuff if you are trying them in order.

The original used black and grey to prevent them from being accidentally selected in the color wheel, since they are repeated as effects Mystery and Cycle. But if those effects are not present on this model, making them hidden or leaving them out altogether makes sense, and it doesn't really affect anything to change the value.

PR #5159, opened 2026-05-27
Add optional temperature and volume unit selections

rwalker777:

Fix pressure scale based on units.  Make temperature a selector like the app.  Make labels/values match the vendor app.

I couldn't figure out a way to hide/disable temperature for devices that always send 25.0 c / 77.0 f.  Since local API doesn't give model and users can't disable in UI due to no unique IDs.  Those devices will just have static values.

### DIFF
--- a/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
+++ b/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
@@ -34,7 +34,6 @@ entities:
 
   # Unit selects
   - entity: select
-    name: Temperature unit
     translation_key: temperature_unit
     category: config
     dps:
--- a/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
+++ b/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
@@ -18,6 +18,7 @@ entities:
     dps:
       - id: 12
         type: integer
+        optional: true
         name: sensor
         class: measurement
         mapping:
@@ -31,6 +32,38 @@ entities:
             value: F
           - value: C
 
+  # Unit selects
+  - entity: select
+    name: Temperature unit
+    translation_key: temperature_unit
+    category: config
+    dps:
+      - id: 20
+        type: string
+        optional: true
+        name: option
+        mapping:
+          - dps_val: f
+            value: Fahrenheit
+          - dps_val: c
+            value: Celsius
+
+  - entity: select
+    name: Volume units
+    icon: "mdi:ruler-square"
+    category: config
+    dps:
+      - id: 102
+        type: string
+        optional: true
+        name: option
+        mapping:
+          - dps_val: us_units
+            value: Gallon
+          - dps_val: metric_unit
+            value: Liter
+
+  # Optional pressure
   - entity: sensor
     class: pressure
     category: diagnostic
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
```diff
@@ -31,6 +32,38 @@ entities:
             value: F
           - value: C
 
+  # Unit selects
+  - entity: select
+    name: Temperature unit
```
make-all: This already exists in the config

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
```diff
@@ -31,6 +32,38 @@ entities:
...
+        optional: true
+        name: option
+        mapping:
+          - dps_val: f
+            value: Fahrenheit
+          - dps_val: c
+            value: Celsius
+
+  - entity: select
+    name: Volume units
```
make-all, line 51: This already exists as "Measurement units"

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
```diff
@@ -40,16 +73,42 @@ entities:
...
         name: unit
         mapping:
           - dps_val: us_units
             value: psi
-            scale: 10
           - value: bar
-            scale: 100
+
+  - entity: sensor
+    name: System Status
```
make-all, line 92: Seems like a candidate for a standard `class: problem` binary_sensor, with description attribute taken from below.
Are those actual values, using an integer, not a bitmap (bitfield)?

make-all (changes requested):

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
rwalker777: I don't know why, but the select was not working where it was before.  Maybe a bug?

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
rwalker777: The app and all the documentation calls it "Volume units".  Just updating to match.  Also makes it clear it's for volume and temperature is it's own setting.

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
rwalker777: They are actual integer values.  I had to figure this out via trial and error from what the app shows.  Given how often my unit loses Tuya cloud connection, it reinforces why Tuya-local is a must have.

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
make-all: It isn't just for volume though, it affected pressure as well

rwalker777: True.  In the app it's called "Volume units".  Your call.

make-all:
Manually merged required changes

PR #5161, opened 2026-05-27
Add foxgard fingerprint lock

itsiv:
Adds profile for the Foxgard Smart Fingerprint Door Lock (product 99gv5nmz).

https://www.amazon.com/dp/B0BNT6CR8C

Supports fingerprint/password/card/BLE/app/voice unlock, volume control, auto-lock, language selection, credential management, and remote code unlock via DP 61/73. Tested personally.


make-all:
Merged with primebras_athenas_lock.yaml, which was already a 93% match, missing only language and unlock_record_check

PR #5180, opened 2026-05-29
Fix remote.send_command never transmitting RF codes

### LINKED ISSUE #5179: remote.send_command silently drops RF codes (await never reached for rf: branch)
### Description

`remote.send_command` is a no-op for any RF code (commands learned with `command_type: rf`, or starting with `rf:`). The DP is built but never sent to the device, so the persiana/blind/whatever doesn't move. Nothing useful is logged at INFO either, so the failure is silent unless DEBUG is on. IR codes are unaffected.

### Steps to reproduce

1. Add a Tuya IR+RF blaster (e.g. Moes S11+ / NovaDigital RM6 Pro, profile `s11_rfir_remote`).
2. `remote.learn_command` with `command_type: rf` to capture a 433 MHz button — the code is saved into `tuya_local_remote_<id>_codes` as expected.
3. Call `remote.send_command` with that command name.

### Expected behaviour

DP 201 is set with `{"control":"rfstudy_send", ...}`, the device retransmits the RF code, and the persiana moves.

### Actual behaviour

Service returns 200 OK, no DP is sent, no INFO log appears, nothing happens.

### Additional context

In `custom_components/tuya_local/remote.py` (release 2026.5.4 and current `main`), inside `async_send_command`, the `_LOGGER.info(...)` and `await self._device.async_set_properties(dps_to_set)` calls are indented inside the `else` (IR) branch:

```python
if code.startswith("rf:"):
    dps_to_set = self._encode_send_code(code[3:], delay * 1000, is_rf=True)
else:
    dps_to_set = self._encode_send_code(code, delay * 1000)
    _LOGGER.info(...)                                # only runs for IR
    await self._device.async_set_properties(...)     # only runs for IR
```

For RF the payload is built into `dps_to_set` and then discarded. Dedenting the two calls one level fixes it; PR follows.

DEBUG log evidence (post-fix, RF now actually transmits):

```
INFO  remote sending command rf:eyJ... to persiana_udinese
DEBUG RM6 Pro sending dps update: {"201": "{\"control\":\"rfstudy_send\",\"rf_type\":\"sub_2g\",...,\"code\":\"eyJ...\"}}
```

### Confirmation

- [x] I am sure this is a bug, well enough described to implement.
- [x] This report is for make-all/tuya-local, not localtuya, a modified tuya-local, the Tuya app or cloud.
- [x] The device is externally powered (USB-mains).
- [x] Checked latest released version (2026.5.4) — bug is present.
- [x] Device config filename mentioned (`s11_rfir_remote`).
- [x] Log messages included.

rafaelreis-r:
Fixes #5179.

In `async_send_command`, `_LOGGER.info(...)` and `await self._device.async_set_properties(dps_to_set)` were indented inside the `else` (IR) branch, so RF codes built `dps_to_set` and then discarded it. Dedent both one level so they run for both branches.

I understand the change: same `dps_to_set` is built in either branch (RF or IR), and the single `async_set_properties` call below applies regardless. IR path behaviour is identical to before.

Tested on a NovaDigital RM6 Pro Tuya (`s11_rfir_remote`) — before the fix `remote.send_command` for an `rf:` code did nothing and logged nothing; after the fix the DP 201 `rfstudy_send` payload is transmitted and the persiana moves.

### DIFF
--- a/custom_components/tuya_local/remote.py
+++ b/custom_components/tuya_local/remote.py
@@ -276,13 +276,13 @@ class TuyaLocalRemote(TuyaLocalEntity, RemoteEntity):
                 dps_to_set = self._encode_send_code(code[3:], delay * 1000, is_rf=True)
             else:
                 dps_to_set = self._encode_send_code(code, delay * 1000)
-                _LOGGER.info(
-                    "%s sending command %s to %s",
-                    self._config.config_id,
-                    code,
-                    subdevice or "default device",
-                )
-                await self._device.async_set_properties(dps_to_set)
+            _LOGGER.info(
+                "%s sending command %s to %s",
+                self._config.config_id,
+                code,
+                subdevice or "default device",
+            )
+            await self._device.async_set_properties(dps_to_set)
 
             if len(codes) > 1:
                 self._flags[subdevice] ^= 1


make-all:
Changes applied in response to #5179

PR #5183, opened 2026-05-29
Add support for NBQ100-US Solar Microinverter

slickm0nty:
Adds support for NBQ100-US solar microinverter


custom_components/tuya_local/devices/nbq100_us_microinverter.yaml
```diff
@@ -0,0 +1,68 @@
+name: Solar Microinverter
+products:
+  - id: ndkye5yafr9dzv2a
+    manufacturer: Sukflow
+    model: NBQ800-US
+    model_id: KW42300C100027
+entities:
+  - entity: switch
+    name: Status
```
make-all: How can status be a switch? This does not make sense

custom_components/tuya_local/devices/nbq100_us_microinverter.yaml
```diff
@@ -0,0 +1,68 @@
+name: Solar Microinverter
```
make-all: Follow HA naming convention: only the first word of device and entity names should be capitalised.

make-all (changes requested):

custom_components/tuya_local/devices/nbq100_us_microinverter.yaml
slickm0nty: This would be the on/off switch. In my head status = on/off made sense since naming it power would be ambiguous as it also refers to generated power. Would "On/off" be a more appropriate entity name?

make-all: No name is better

custom_components/tuya_local/devices/nbq100_us_microinverter.yaml
slickm0nty: updated

custom_components/tuya_local/devices/nbq100_us_microinverter.yaml
slickm0nty: updated

custom_components/tuya_local/devices/nbq100_us_microinverter.yaml
```diff
@@ -0,0 +1,67 @@
+name: Solar microinverter
+products:
+  - id: ndkye5yafr9dzv2a
+    manufacturer: Sukflow
+    model: NBQ800-US
+    model_id: KW42300C100027
```
make-all: This model id does not seem necessary (the model is already id type, so this is more like a stocking code or something not intended to be user facing)

custom_components/tuya_local/devices/nbq100_us_microinverter.yaml
```diff
@@ -0,0 +1,67 @@
...
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 107
+        type: integer
+        name: sensor
+        unit: C
+  - entity: sensor
+    class: frequency
+    name: Grid Hz
```
make-all: Just leave this unnamed so it picks up standard HA translations is better I think. It is not ambiguous, because only AC has a frequency.

make-all (changes requested):

custom_components/tuya_local/devices/nbq100_us_microinverter.yaml
slickm0nty: agreed

custom_components/tuya_local/devices/nbq100_us_microinverter.yaml
slickm0nty: agreed

PR #5189, opened 2026-05-31
feat(devices): add WiFi+BLE 1 Gang relay switch with remote

mmaarrkk02:


make-all:
There is nothing useful here that is not already covered by simple_switch_timerv2.yaml

PR #5190, opened 2026-05-31
feat(devices): add W-W601, W-W602, W-W603 simple switch variants

mmaarrkk02:
- simple_switch_w601.yaml: single switch with timer (DP 1, 9)
- simple_dual_switch_w602.yaml: dual switch with timers (DP 1, 2, 9, 10)
- simple_triple_switch_w603.yaml: triple switch with timers (DP 1, 2, 3, 9, 10, 11)


make-all:
simple_switch_timerv2, simple_dual_switch_timer_v2, simple_triple_switch_timer already match these configs, so this PR is not needed.

PR #5196, opened 2026-05-31
Add support for cat litter box Red Solution 30S

alexeykin1984:
Add support for cat litter box Red Solution 30S


make-all:
Unfortunately this device appears to be made exclusively for the Russian market.
I am not accepting such devices until Putin ends his invasion of neighboring countries.

PR #5200, opened 2026-06-01
Add Aquatech X8 Fusion water heater support

pranksta02:
Adds support for the Aquatech X8 Fusion heat pump water heater.

Verified:
- Eco mode
- Standard mode
- Hybrid mode
- Heating Element mode
- Tank temperature
- Ambient temperature
- Coil temperature
- Exhaust temperature
- Return air temperature
- Return water temperature
- Compressor telemetry
- Fan telemetry

Product tested:
Aquatech X8 Fusion 320L

DP mapping verified from Tuya app and device telemetry.


make-all:
File is in the wrong location. Are you sure you have tested this, and not an already existing compatible config?

pranksta02:
Ah crap, yes I am definitely running it on my setup, I just placed it in the wrong place here. This is my first time doing a pull request on git. There are similar units existing but they have different DP mapping and different modes.

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
```diff
@@ -0,0 +1,183 @@
+name: Aquatech X8 Fusion
+products:
+  - id: "*"
```
make-all: This is not a valid product id

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
```diff
@@ -0,0 +1,183 @@
+name: Aquatech X8 Fusion
```
make-all: Top level name should be generic in case of other products matching.

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
```diff
@@ -0,0 +1,183 @@
...
+        type: integer
+        name: current_temperature
+        unit: C
+
+      - id: 4
+        type: string
+        name: work_mode
+        hidden: true
+
+  - entity: select
```
make-all: Duplicate entities are not desirable

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
```diff
@@ -0,0 +1,183 @@
...
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 21
+        type: bitfield
+        name: fault_code
+
+  - entity: sensor
+    name: Return Air Temperature
```
make-all: HA naming convention is to only capitalise the first word of a device or entity name.

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
```diff
@@ -0,0 +1,183 @@
...
+  - entity: sensor
+    name: Compressor
+    category: diagnostic
+    dps:
+      - id: 108
+        type: integer
+        name: sensor
+
+  - entity: binary_sensor
+    name: Internal Flag DP109
```
make-all: Unknown dps do not need entities, just add them as an attribute on another entity to be able to monitor them and use for device matching

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
```diff
@@ -0,0 +1,183 @@
...
+    category: diagnostic
+    dps:
+      - id: 109
+        type: boolean
+        name: sensor
+
+  - entity: sensor
+    name: Fan
```
make-all: Fan speed? unit: rpm?

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
pranksta02: Updated with device id

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
pranksta02: Updated

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
pranksta02: Updated

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
pranksta02: updated

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
pranksta02: updated

custom_components/tuya_local/aquatech_x8_fusion_water_heater.yaml
pranksta02: updated, unit is hz

pranksta02:
Thanks for the review feedback. I have reworked the device configuration and tested the updated version on my Aquatech Fusion/X8 320L.

Changes made:

* Moved the YAML into `custom_components/tuya_local/devices/`.
* Replaced the wildcard with the confirmed product ID `4zrn3rqjibkude2z`.
* Changed the top-level name to the generic `Heat pump water heater`.
* Added Aquatech as the manufacturer and Fusion/X8 as the model.
* Removed the duplicate Mode select; mode control is now handled by the water-heater entity.
* Updated entity names to Home Assistant sentence case.
* Corrected DP101 to coil temperature and DP103 to return-air temperature.
* Confirmed through Tuya’s Things Data Model that DP108 and DP110 report compressor and fan frequency in Hz.
* Confirmed DP109 is the four-way valve and DP113 is the electric heating element, rather than leaving them as unknown internal flags.
* Retained the device’s raw `boot` enum and mapped it to the Hybrid/high-demand mode.

I tested power on/off, target-temperature control, Standard, Eco, Hybrid and Heating Element modes, along with the available temperature and diagnostic entities. Everything appears to be working correctly.

PR #5206, opened 2026-06-01
Add support for joekol dual pump watering system

ffais:
Added joekol dual pump watering system. I've tested locally all the functionality.


custom_components/tuya_local/devices/joekol_wd01ade_watering_pump.yaml
```diff
@@ -0,0 +1,73 @@
+name: Dual pump watering system
+products:
+  - id: 7lanttzn7fuel3zb
+    manufacturer: JOEKOL
+    model: WD-01ADE
+    model_id: e1n14ca8
+entities:
+  - entity: switch
```
make-all: These are maybe better as `valve` entities (class: water).

custom_components/tuya_local/devices/joekol_wd01ade_watering_pump.yaml
```diff
@@ -0,0 +1,73 @@
...
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Pump B
+    class: switch
+    dps:
+      - id: 2
+        type: boolean
+        name: switch
+  - entity: number
```
make-all: The timers should be time entities. Although you need to set your user settings to 24h time to get the UI to display properly, it is a much better representation than up to 86400 seconds.

custom_components/tuya_local/devices/joekol_wd01ade_wateringpump.yaml
```diff
@@ -0,0 +1,73 @@
...
+        name: value
+        unit: s
+        optional: true
+        range:
+          min: 1
+          max: 86400
+  - entity: text
+    name: Pump A schedule
+    category: config
+    icon: "mdi:calendar-clock"
```
make-all, line 49: The reset below here should be marked `hidden: true`, as they are advanced entities that will require some knowledge of the format to make use of.

make-all (changes requested):

custom_components/tuya_local/devices/joekol_wd01ade_watering_pump.yaml
ffais: I change it with time, but is not 100% clear what you mean with "Although you need to set your user settings to 24h time to get the UI to display properly".

make-all: If you click on your user image in bottom left of HA, it will open your user profile.
About half way down the page is the Time format setting.

ffais: Okay, now I understand what you're referring to. Are any further changes needed for the PR?

Issue #5197, opened 2026-05-31
Clarify brightness range lower-bound semantics


Zuz666:
## Description

This issue is based on the investigation done while refining #5195, which extends the reviewer fix from c33c67b2da0046e5ec92230a470c515bc8f4481c. PR #5195 fixes a narrower write-path bug where `brightness_to_value()` can return values below a configured non-zero brightness range minimum. During that work, a broader ambiguity became clear: brightness range lower bounds do not currently describe whether the lower bound is a brightness value or a special boundary/off value.

A configured brightness range only tells tuya-local the accepted numeric interval, for example:

```yaml
range:
  min: 0
  max: 100
```

or:

```yaml
range:
  min: 1
  max: 6
```

It does not explicitly say whether the lower bound is:

- the first visible brightness level,
- an off state,
- a valid but visually negligible brightness level,
- or a device-specific boundary value that is normalised internally by the device.

The current read path implicitly treats a reported brightness DP value of `0` as HA brightness `0`, because conversion only happens for truthy values. See the current read path in `light.py`:

https://github.com/make-all/tuya-local/blob/3c8577ef853e32ff6779639b411b916e5d744ffe/custom_components/tuya_local/light.py#L181-L186

So `DP 0 -> HA brightness 0` today.

The write path, however, passes the configured range directly to Home Assistant's `brightness_to_value()`. For some zero-based ranges, this can map low non-zero HA brightness values back to DP `0`:

```text
brightness_to_value((0, 100), 1) = -0.6039
round/clamp -> DP 0
```

That gives a possible round trip of:

```text
HA brightness 1 -> DP 0 -> HA brightness 0
```

For wider zero-based ranges the behaviour is different:

```text
brightness_to_value((0, 1000), 1) = 2.925
round -> DP 3
value_to_brightness((0, 1000), 3) -> HA brightness 1
```

In that case the HA round trip is preserved, but the first physical DP values (`1`, `2`) are not reachable from minimum HA brightness.

A similar ambiguity can theoretically exist for non-zero ranges too. For example, with `range: 1..6`, tuya-local currently treats DP `1` as a valid brightness value because the read path maps it to a non-zero HA brightness:

```text
value_to_brightness((1, 6), 1) = 42
```

But the `range` declaration alone does not prove that DP `1` is physically a visible brightness level rather than a special boundary/off value for a particular device.

## Steps to reproduce

This is a modelling issue rather than a report for one specific device.

[...]

Issue #2589, opened 2024-12-01
 PlantsIO Ivy Smart Planter display won't toggle off


thewade:
### Description

Continued from the conversation at https://github.com/make-all/tuya-local/discussions/2168.

When I try to toggle the display off, it briefly toggles off but then toggles back on.

The datapoint is correctly updated and the screen turns off. However in Home Assistant it shows as on with a brightness of 1%.

### Steps to reproduce

From the device config I press the slider to turn the device off.

### Expected behaviour

The device state should be 'off' and the toggle should be in the off state.

### Additional context

The entity:
```
  - entity: light
    translation_key: display
    category: config
    dps:
      - id: 101
        type: integer
        name: brightness
        range:
          min: 0
          max: 100
      - id: 135
        type: boolean
        name: available
        mapping:
          - dps_val: false
            value: true
          - dps_val: true
            value: false
```

A value of 0 is off, with 1-100 being brightness values.

When toggled off, in the device diagnostics with see the dps value is 0 which is correct.

```
  "data": {
    "name": "Smart Planter",
    "type": "plantsio_ivy_smart_planter",
    "device_id": "**REDACTED**",
    "device_cid": "",
    "local_key": "**REDACTED**",
    "host": "**REDACTED**",
    "protocol_version": 3.4,
    "tinytuya_version": "1.15.1",
    "api_version_set": 3.4,
    "api_version_used": 3.4,
    "api_working": true,
    "status": {},
    "cached_state": {
      "updated_at": 1733082474.188266,
      "101": 0,
```
However under entities we see the state is 'on' and the brightness is 3.

```
        {
          "disabled": false,
          "disabled_by": null,
          "entity_category": "config",
          "device_class": null,
          "original_device_class": null,
          "icon": null,
          "unit_of_measurement": null,
          "state": {
            "entity_id": "light.smart_planter_display",
            "state": "on",
            "attributes": {
              "supported_color_modes": [
                "brightness"
              ],
              "color_mode": "brightness"
              "brightness": 3,
              "available": true,
              "friendly_name": "Smart Planter Display",
              "supported_features": 0
            },
            "last_changed": "2024-12-01T19:47:31.928245+00:00",
            "last_reported": "2024-12-01T19:47:54.194543+00:00",
            "last_updated": "2024-12-01T19:47:52.130242+00:00"
          }
```
Files:
[...]

Issue #1173, opened 2023-10-15
Request support for Petrust TK-WF002 (GN_PW01)


pot-noble:
# YAML
```
name: Petrust Smart Fountain
products:
  - id: 5jdws4vgfzhkq7vs
primary_entity:
  entity: switch
  icon: "mdi:power"
  dps:
    - id: 1
      type: boolean
      name: switch
secondary_entities:
  - entity: select
    name: Mode
    class: enum
    icon: "mdi:auto-mode"
    dps:
      - id: 2
        type: string
        name: option
        mapping:
          - dps_val: "normal"
            value: "Normal"
          - dps_val: "smart"
            value: "Smart"
  - entity: select
    name: Water Control
    class: enum
    icon: "mdi:water-circle"
    dps:
      - id: 11
        type: string
        name: option
        mapping:
          - dps_val: "close"
            icon: "mdi:water-off"
            value: "Close"
          - dps_val: "low"
            value: "Low"
          - dps_val: "middle"
            value: "Middle"
          - dps_val: "high"
            value: "High"
  - entity: number
    name: Status Brightness
    icon: "mdi:brightness-6"
    category: config
    dps:
      - id: 12
        type: integer
        name: value
        range:
          min: 0
          max: 4
  - entity: number
    name: Filter lifetime
    icon: "mdi:air-filter"
    category: config
    dps:
      - id: 7
        type: integer
        name: value
        range:
          min: 0
          max: 30
  - entity: sensor
    name: Replace filter in
    icon: "mdi:air-filter"
    category: diagnostic
    dps:
      - id: 3
        type: integer
        name: sensor
        unit: days
  - entity: button
    name: Replace filter
    icon: "mdi:air-filter"
    category: config
    dps:
      - id: 5
        type: boolean
        name: button
  - entity: number
    name: Clean life
    icon: "mdi:pump"
    category: config
    dps:
      - id: 8
        type: integer
        name: value
        range:
          min: 0
          max: 20
  - entity: sensor
    name: Clean in
    icon: "mdi:pump"
    category: diagnostic
    dps:
      - id: 4
        type: integer
        name: sensor
        unit: days
  - entity: button
    name: Pump cleaned
    icon: "mdi:pump"
    category: config
    dps:
      - id: 6
        type: boolean
        name: button
  - entity: sensor
    name: Water level
    class: enum
    category: diagnostic
    dps:
      - id: 9
        name: sensor
        type: string
        mapping:
          - dps_val: "low"
[...]

PR #5218, opened 2026-06-02
fix(light): reserve zero brightness as off

Zuz666:
## Problem
Brightness-only light entities use the brightness DP as their on/off state. For zero-based ranges, DP 0 is the off value. Very low non-zero Home Assistant brightness can otherwise be converted or rounded to DP 0, so `light.turn_on(brightness=1)` can turn the light off instead of setting the first visible level (related analysis: #5197).

## Example
Kogan KAWHTNOSLPA `kogan_noise_machine.yaml` defines the light entity as DP 104 brightness with range `0..255` and no switch DP on the light entity. In this configured entity shape, tuya-local uses DP 104 for brightness and infers the on/off state from the same value.

This is a code-path example derived from the config only. I found neither a test on actual Kogan hardware nor a device-owner report confirming the physical behavior of DP 104.

## Change
For brightness-only lights with zero-based ranges, non-zero HA brightness is converted over the first-on value to max range, keeping DP 0 reserved for off. This also accounts for configured `step`, so low brightness does not round back to 0.

Regression tests were added for low-brightness writes and for reading the first on value.

## Impact / risk
A config scan found 36 light entities across 29 device configs matching the condition changed here: no `switch` DP on the light entity and a brightness range starting at 0.

For this entity shape, tuya-local already treats DP 0 as off: `async_turn_off()` writes 0, and `is_on` requires brightness greater than 0. Device-specific evidence for that convention exists for other matching devices (#2589, #2767, #1173), but it does not establish the behavior of Kogan DP 104. Some devices with the same config shape may still use DP 0 differently, so side effects are possible.

## Test plan
The integration code path was exercised with Home Assistant 2026.5.4 using a local fake Tuya device matching Kogan KAWHTNOSLPA / `kogan_noise_machine.yaml` ([fake_kogan_noise_machine.py](https://github.com/user-attachments/files/28507550/fake_kogan_noise_machine.py)), where DP 104 controlled the simulated light entity state, and with the regression tests.

The fake-device result validates the integration code path only; it is not hardware validation for the Kogan device.

### DIFF
--- a/custom_components/tuya_local/light.py
+++ b/custom_components/tuya_local/light.py
@@ -541,7 +541,8 @@ class TuyaLocalLight(TuyaLocalEntity, LightEntity):
             bright = 255
             r = self._brightness_dps.range(self._device)
             if r:
-                r = self._brightness_conversion_range(r, bright)
+                # Full HA brightness always maps to the DP max, so the zero-off
+                # range adjustment is only needed for non-max brightness values.
                 bright = color_util.brightness_to_value(r, bright)
             _LOGGER.info(
                 "%s turning light on to brightness %d",
--- a/custom_components/tuya_local/light.py
+++ b/custom_components/tuya_local/light.py
@@ -191,8 +191,17 @@ class TuyaLocalLight(TuyaLocalEntity, LightEntity):
             r = self._brightness_dps.range(self._device)
             val = self._brightness_dps.get_value(self._device)
             if r and val:
+                r = self._brightness_conversion_range(r, val)
                 val = color_util.value_to_brightness(r, val)
             return val
+
+    def _brightness_conversion_range(self, dp_range, value):
+        """Return the range used to convert non-off brightness values."""
+        if not self._switch_dps and dp_range[0] == 0 and dp_range[1] > 0 and value:
+            first_on_value = max(1, self._brightness_dps.step(self._device))
+            if first_on_value <= dp_range[1]:
+                return (first_on_value, dp_range[1])
+        return dp_range
 
     @property
     def _unpacked_rgbhsv(self):
@@ -304,6 +313,7 @@ class TuyaLocalLight(TuyaLocalEntity, LightEntity):
                 bright = params.get(ATTR_WHITE)
                 r = self._brightness_dps.range(self._device)
                 if r:
+                    r = self._brightness_conversion_range(r, bright)
                     bright = _ha_brightness_to_dp_value(bright, r)
 
                 _LOGGER.info(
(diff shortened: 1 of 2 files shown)


make-all:
I'm still not sure what you are actually trying to fix here, as you seem to be testing with a mock device based on the above explanation, and there have never been reports from users of actual devices.

Zuz666:
You're right that my original Kogan test used a mock, not real hardware. I should have separated that from hardware evidence more clearly, so I investigated the affected configurations.

The issue is a mismatch between two existing contracts:

- Without a `switch` DP, tuya-local already treats brightness DP `0` as off: `async_turn_off()` writes `0`, and `is_on` requires brightness greater than `0`.
- Home Assistant's [`brightness_to_value()`](https://github.com/home-assistant/core/blob/2026.5.2/homeassistant/util/color.py#L710-L721) says: “Do not include 0 if the light is off for value 0.”

We currently pass the full zero-based range to that helper. With Kogan's configured `0..255` range:

`brightness_to_value((0, 255), 1) == 0.003921...`

tuya-local's integer conversion then writes:

`int(round(0.003921...)) == 0`

Therefore `turn_on(brightness=1)` can write the same DP value as `turn_off`. This PR excludes the reserved off value from non-zero brightness conversion.

There is real-device evidence for the underlying `0 = off, positive = on` contract. In [#2589](https://github.com/make-all/tuya-local/issues/2589), the PlantsIO Ivy owner reported: “A value of 0 is off, with 1-100 being brightness values” and “The datapoint is correctly updated and the screen turns off.” HA nevertheless displayed brightness `3` (about 1%) and reported the physically-off display as on. [c6c0a76a](https://github.com/make-all/tuya-local/commit/c6c0a76adc434ab886bdff841f1aa949c5918ef8) fixed that read-side zero-boundary collision by not scaling the off value. This PR fixes the symmetric write path. The reverse service call was not captured in #2589; its result follows from the conversion above, not from another hardware observation.

The [Rohnson R-9530 owner](https://github.com/make-all/tuya-local/issues/2767#issuecomment-2618351585) also stated “0 = turned off, 1 = low, 2 = high,” and the [Petrust TK-WF002 data model](https://github.com/make-all/tuya-local/issues/1173) describes brightness `0` as turning the indicator off.

Across 36 matching entities in 29 configs, I found direct evidence for six entities in five device families and no confirmed counterexample where DP `0` is a visible minimum. Kogan itself remains unverified, so I have updated the PR description to make clear that its fake-device test only reproduces the code path and is not hardware validation.

The PR does not introduce `0 = off`; it makes conversion consistent with the existing `is_on` and `turn_off` behavior. Would you consider this evidence sufficient for the shared behavior, or would you prefer it to be opt-in per config?

Thank you for maintaining this integration, and for your patience while I investigated this properly.

make-all:
I committed a slightly different version that leaves open the possibility of a light that uses 0 as a non-off discrete brightness value with separate on/off switch, as this is intended to be supported.

PR #5220, opened 2026-06-02
Add generic water tank level monitor support

WGrobler:
Adds support for a generic water tank level indicator/monitor (product ID: h8lvyoahr6s6aybf).

Exposes water level as a percentage (DPS 22), liquid depth in mm (DPS 2, diagnostic), and liquid state as a string sensor (DPS 1, diagnostic).


make-all:
This is already a match for ept_level_sensor and me201w_level_sensor, so can be merged with whichever of those is closer to the actual device. The characterization of this as "generic" is misleading, as there is not a lot of products on the market matching this in the same way as there are with smartswitches and lightbulbs.
The difference between those two is just the scaling on the calibration controls, so if this model really has no calibration control at all, then it is arbitrary which of those to match to.

PR #5227, opened 2026-06-02
Add support for Comfort Zone Tower Fan (CZTF423S)

cmoates:
Adds device configuration for the Comfort Zone CZTF423S tower fan with:
- Primary fan entity with speed control (1-5) and oscillation
- Sleep timer select entity with 1-12 hour options
- Support for normal, nature, and sleep preset modes

All gathered using the tinytuya monitor and tested on my device.

### DIFF
--- a/custom_components/tuya_local/devices/ciztf423s_tower_fan.yaml
+++ b/custom_components/tuya_local/devices/ciztf423s_tower_fan.yaml
@@ -1,64 +1,63 @@
-name: Tallboy Tower Fan
+name: Comfort Zone Tower Fan
 products:
   - id: eb414680be46558e014mtb
-    name: TOWER FAN（CZTF423S）
-primary_entity:
-  entity: fan
-  dps:
-    - dp: 1
-      name: switch
-      type: boolean
-    - dp: 2
-      name: preset_mode
-      type: string
-      mapping:
-        - dpsval: normal
-          value: normal
-        - dpsval: nature
-          value: nature
-        - dpsval: sleep
-          value: sleep
-    - dp: 3
-      name: speed
-      type: integer
-      range:
-        min: 1
-        max: 5
-    - dp: 5
-      name: oscillate
-      type: boolean
-secondary_entities:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/ciztf423s_tower_fan.yaml
```diff
@@ -0,0 +1,63 @@
+name: Comfort Zone Tower Fan
```
make-all: Top level name should be unbranded in case other fans match this config (likely, as the dp layout matches 3 other already exsiting configs of various brands with the only difference being the number of speeds (3, 12 and 15).

custom_components/tuya_local/devices/ciztf423s_tower_fan.yaml
```diff
@@ -0,0 +1,63 @@
...
+products:
+  - id: eb414680be46558e014mtb
+    name: Tower Fan（CZTF423S）
```
make-all: Here should be `manufacturer` and `model` instead of name.

custom_components/tuya_local/devices/ciztf423s_tower_fan.yaml
```diff
@@ -0,0 +1,63 @@
...
+entities:
+  - entity: fan
+    dps:
```
make-all, line 9: Use `translation_key: fan_with_presets`, as HA does not provide any predefined presets for fans any more.

custom_components/tuya_local/devices/ciztf423s_tower_fan.yaml
```diff
@@ -0,0 +1,63 @@
...
+        name: speed
+        type: integer
+        range:
+          min: 1
+          max: 5
+      - id: 5
+        name: oscillate
+        type: boolean
+  - entity: select
+    name: Sleep Timer
```
make-all: Use `translation_key: timer` instead of this name and icon

make-all (changes requested):

PR #5228, opened 2026-06-03
Add Sakura B1 / HTW IX75B

Flechaa:
Adds a config for the Giatsu Sakura B1 (HTW IX75B) Wi-Fi monosplit air conditioner (Tuya category ``kt``, product id ``luacvecdj9biucd1``).
This is the same OEM unit sold under multiple brands; the GIApp (Tuya rebrand) app reports it as Sakura B1, and it is also retailed as the HTW IX75B. Both brands are listed under products.

This model is based off the vivion airconditioner schema with one difference: it exposes two extra fan speeds (``MediumLow``, ``MediumHigh``), mapped here as ``medium_low`` and ``medium_high``.

## Schema
```
[
  {
    "code": "switch",
    "iconname": "icon-dp_power3",
    "id": 1,
    "mode": "rw",
    "name": "开关",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "temp_set",
    "iconname": "icon-dp_temp",
    "id": 2,
    "mode": "rw",
    "name": "温度设置",
    "property": {
      "unit": "℃",
      "min": 16,
      "max": 90,
      "scale": 0,
      "step": 1,
      "type": "value"
    },
    "type": "obj"
  },
  {
    "code": "temp_current",
    "iconname": "icon-dp_temp",
    "id": 3,
    "mode": "ro",
    "name": "当前温度",
    "property": {
      "unit": "℃",
      "min": -50,
      "max": 50,
      "scale": 0,
      "step": 1,
      "type": "value"
    },
    "type": "obj"
  },
  {
    "code": "mode",
    "iconname": "icon-dp_mode",
    "id": 4,
    "mode": "rw",
    "name": "工作模式",
    "property": {
      "range": [
        "auto",
        "cold",
        "hot",
        "wet",
        "wind"
      ],
      "type": "enum"
    },
    "type": "obj"
  },
  {
    "code": "fan_speed_enum",
    "iconname": "icon-a_mode_fan",
    "id": 5,
    "mode": "rw",
    "name": "风速",
    "property": {
      "range": [
        "strong",
        "auto",
        "low",
        "middle",
        "high",
        "mute",
        "MediumLow",
        "MediumHigh"
      ],
      "type": "enum"
    },
    "type": "obj"
  },
  {
    "code": "anion",
    "iconname": "icon-a_nav_mode",
    "id": 11,
    "mode": "rw",
    "name": "空清",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "heat",
    "iconname": "icon-dp_flower",
    "id": 12,
    "mode": "rw",
    "name": "电加热",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "fault",
    "iconname": "icon-Panic",
    "id": 22,
    "mode": "ro",
    "name": "故障告警",
    "property": {
      "label": [
        "E0",
        "E1",
        "E2",
        "E3",
        "E4",
        "E5",
        "E6",
        "E7",
        "P8",
        "P9",
        "P10",
        "P11",
        "P12",
        "P13",
        "P14"
      ],
      "type": "bitmap",
      "maxlen": 15
    },
    "scope": "fault",
    "type": "obj"
  },
  {
    "code": "sleep",
    "id": 25,
    "mode": "rw",
    "name": "睡眠",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "gear_vertical",
    "id": 31,
    "mode": "rw",
    "name": "上下摆风",
    "property": {
      "range": [
        "off",
        "same",
        "vane_1",
        "vane_2",
        "vane_3",
        "vane_4",
        "vane_5"
      ],
      "type": "enum"
    },
    "type": "obj"
  },
  {
    "code": "switch_horizontal",
    "id": 33,
    "mode": "rw",
    "name": "左右摆风",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "display",
    "id": 36,
    "mode": "rw",
    "name": "屏显",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "timer_switch_on_en",
    "id": 101,
    "mode": "rw",
    "name": "定时开机开关",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "timer_switch_off_en",
    "id": 102,
    "mode": "rw",
    "name": "定时关机开关",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "timer_switch_on",
    "id": 103,
    "mode": "rw",
    "name": "定时开机",
    "property": {
      "type": "string",
      "maxlen": 255
    },
    "type": "obj"
  },
  {
    "code": "timer_switch_off",
    "id": 104,
    "mode": "rw",
    "name": "定时关机",
    "property": {
      "type": "string",
      "maxlen": 255
    },
    "type": "obj"
  },
  {
    "code": "model_sign_1",
    "id": 105,
    "mode": "ro",
    "name": "单冷冷热标识位",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "model_sign_2",
    "id": 106,
    "mode": "ro",
    "name": "定频变频标识位",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "model_sign_3",
    "id": 107,
    "mode": "ro",
    "name": "上下摆叶是否支持位置调节标识位",
    "property": {
      "type": "bool"
    },
    "type": "obj"
  },
  {
    "code": "fault_code_bp",
    "id": 108,
    "mode": "ro",
    "name": "室内机错误代码（变频机故障及保护代码）",
    "property": {
      "unit": "",
      "min": 0,
      "max": 29,
      "scale": 0,
      "step": 1,
      "type": "value"
    },
    "type": "obj"
  }
]
```

### DIFF
--- a/custom_components/tuya_local/devices/htw_ix75b_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/htw_ix75b_airconditioner.yaml
@@ -18,10 +18,12 @@ entities:
             conditions:
               - dps_val: auto
                 value: heat_cool
+                available: heat_model
               - dps_val: cold
                 value: cool
               - dps_val: hot
                 value: heat
+                available: heat_model
               - dps_val: wet
                 value: dry
               - dps_val: wind
@@ -51,11 +53,11 @@ entities:
           - dps_val: low
             value: low
           - dps_val: MediumLow
-            value: medium_low
+            value: medlow
           - dps_val: middle
             value: medium
           - dps_val: MediumHigh
-            value: medium_high
+            value: medhigh
           - dps_val: high
             value: high
           - dps_val: mute
@@ -78,6 +80,29 @@ entities:
             value: "on"
           - value: "off"
             hidden: true
+      - id: 31
+        type: string
+        name: option
+        mapping:
+          - dps_val: "off"
+            value: "off"
+          - dps_val: same
+            value: "on"
+          - dps_val: vane_1
+            value: Position 1
+            available: swing_positions
+          - dps_val: vane_2
+            value: Position 2
+            available: swing_positions
+          - dps_val: vane_3
+            value: Position 3
+            available: swing_positions
+          - dps_val: vane_4
+            value: Position 4
+            available: swing_positions
+          - dps_val: vane_5
+            value: Position 5
+            available: swing_positions
       - id: 33
         type: boolean
         name: swing_horizontal_mode
@@ -86,14 +111,24 @@ entities:
             value: "off"
           - dps_val: true
             value: "on"
-      - id: 103
-        type: string
-        name: timer_on_time
+      - id: 105
+        type: boolean
+        name: heat_model
         optional: true
-      - id: 104
-        type: string
-        name: timer_off_time
+        hidden: true
+        mapping:
+          - dps_val: true
+            value: false
+          - value: true
+      - id: 106
+        type: boolean
+        name: invertor
         optional: true
+      - id: 107
+        type: boolean
+        name: swing_positions
+        optional: true
+        hidden: true
   - entity: switch
     translation_key: ionizer
     category: config
@@ -109,6 +144,15 @@ entities:
       - id: 12
         type: boolean
         name: switch
+      - id: 105
+        type: boolean
+        name: available
+        optional: true
+        hidden: true
+        mapping:
+          - dps_val: true
+            value: false
+          - value: true
   - entity: binary_sensor
     class: problem
     category: diagnostic
@@ -127,29 +171,6 @@ entities:
         type: bitfield
         name: fault_code
         optional: true
-  - entity: select
-    name: Vertical swing
-    icon: "mdi:angle-acute"
-    category: config
-    dps:
-      - id: 31
-        type: string
-        name: option
-        mapping:
-          - dps_val: "off"
-            value: "Off"
-          - dps_val: same
-            value: Swing
-          - dps_val: vane_1
-            value: Position 1
-          - dps_val: vane_2
-            value: Position 2
-          - dps_val: vane_3
-            value: Position 3
-          - dps_val: vane_4
-            value: Position 4
-          - dps_val: vane_5
-            value: Position 5
   - entity: light
     translation_key: display
     category: config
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/htw_ix75b_airconditioner.yaml
```diff
@@ -0,0 +1,215 @@
+name: Air conditioner
+products:
+  - id: luacvecdj9biucd1
+    manufacturer: Giatsu
+    model: Sakura B1
+  - id: luacvecdj9biucd1
```
make-all, line 3: There should not be duplicate entries in the product list. If the same ID is known to be used for different brands, then the first one encountered normally is left as the entry, though in some cases we have listed both. But only if this has been confirmed from actual devices, do not make such entries based on an assumption because products look the same from pictures.

Flechaa: My HTW IX75B is detected on the app as a Sakura B1 hence why I made this assumption, though I cannot tell if the same is for a real Sakura B1.
In this case is it better to remove the HTW IX75B from the list?

make-all: It is better to remove the Sakura B1 from the products list, because it may have a different product id despite being based on the same firmware. Unless we get access to one, we will not know for sure.

custom_components/tuya_local/devices/htw_ix75b_airconditioner.yaml
```diff
@@ -0,0 +1,215 @@
...
+            hidden: true
+      - id: 33
+        type: boolean
+        name: swing_horizontal_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: "on"
+      - id: 103
```
make-all: Depending on the format of the string, it is probably possible to expose these two timers as `time` entities.

custom_components/tuya_local/devices/htw_ix75b_airconditioner.yaml
```diff
@@ -0,0 +1,215 @@
...
+    category: diagnostic
+    dps:
+      - id: 108
+        type: integer
+        name: sensor
+        optional: true
+        range:
+          min: 0
+          max: 29
+  - entity: binary_sensor
```
make-all: This should be kept as an attribute under the climate entity, and used to filter the hvac_mode mapping (using `available: heating_model`).

custom_components/tuya_local/devices/htw_ix75b_airconditioner.yaml
```diff
@@ -0,0 +1,215 @@
...
+    name: Cooling and heating model
+    category: diagnostic
+    hidden: true
+    dps:
+      - id: 105
+        type: boolean
+        name: sensor
+        optional: true
+  - entity: binary_sensor
+    name: Inverter model
```
make-all: This is static info, so can just be an attribute under the climate entity.

make-all (changes requested):

custom_components/tuya_local/devices/htw_ix75b_airconditioner.yaml
```diff
@@ -0,0 +1,215 @@
...
+    category: diagnostic
+    hidden: true
+    dps:
+      - id: 106
+        type: boolean
+        name: sensor
+        optional: true
+  - entity: binary_sensor
+    name: Vane position control
```
make-all: This should be used as an available attribute on the Vertical swing entity above.

custom_components/tuya_local/devices/htw_ix75b_airconditioner.yaml
Flechaa: The string is stored in 4 characters, the first 2 are the hours on the timer and the others are the minutes (e.g. 2 hours and 3 minutes would be "0203"), I've looked and it doesn't seem like tuya-local can do it without a patch as it tries to split the value by ":".

make-all: It should work with:
```
  - entity: time
    name: "On timer"
    category: config
    dps:
      - id: 103
        type: string
        optional: true
        name: hms
```

custom_components/tuya_local/devices/htw_ix75b_airconditioner.yaml
Flechaa: I'm not sure what should be filtered out there, this field seems to always be false.

PR #5232, opened 2026-06-03
Add Inkbird ISC-028-BW and CRC16/MODBUS checksum support

bjursten:
## Summary

Adds support for the **Inkbird ISC-028-BW** WiFi smoker controller and introduces
`crc16_modbus` for device configs that use masked writes into CRC-protected
base64/hex blobs.

The ISC-028-BW stores all user settings in DP `102` (`setting_para`, 92 bytes).
Bytes 90–91 hold a **CRC-16/MODBUS** checksum over bytes 0–89. Standard masked
writes merge the changed field but leave the old checksum, which the device
silently rejects. DP `109` (boolean OTA) works without this because it is not
a blob DP.

Verified against Tuya IoT **Device Logs** and live Home Assistant testing
(protocol **3.5** required).

## Device: Inkbird ISC-028-BW

| Field | Value |
|-------|-------|
| Product ID | `8l5th5vqvszbuvlm` |
| Model ID | `e1ku12z0` |
| Protocol | 3.5 (3.3/auto failed in testing) |

### DPs

| DP | Code | Access | Notes |
|----|------|--------|-------|
| 101 | `real_time_data` | ro | 11 bytes: 5 temps (uint16 LE ÷10 °F) + fan PID % (byte 10) |
| 102 | `setting_para` | rw | 92 bytes; CRC at bytes 90–91 |
| 103 | `alarm` | ro | Bitmap (probe lost, grill open, etc.) |
| 104–108 | presets | rw | Not decoded yet |
| 109 | `ota_update` | rw | Boolean |

### DP 102 layout (verified)

| Offset | Field |
|--------|-------|
| 0 | Fan on/off (`0`/`1`) |
| 1 | LCD unit (`0`=°C, `1`=°F) |
| 7–8 | Hold target (uint16 LE ÷10, stored as °F internally) |
| 80 | Sound (`0`/`1`) |
| 90–91 | CRC-16/MODBUS over bytes 0–89 (uint16 LE) |

Disconnected probe sentinel on DP 101: `6000` (`0x1770`).

### Home Assistant entities

- **Climate** — grill temp (read), fan on/off + hold target (write via DP 102)
- **Fan output** sensor — PID fan % (read-only, DP 101 byte 10)
- **Temperature unit** select — DP 102 byte 1
- **Sound** switch — DP 102 byte 80
- **Initialize settings** button — seeds DP 102 if not yet cached locally
- Temperature sensors for grill + 4 meat probes
- Alarm binary sensors (grill open, probe lost)

Fan on/off is controlled via **Climate** (Heat/Off), not a separate fan entity,
because byte 10 of DP 101 is PID output only.

## Core change: `crc16_modbus`

Optional integer on a DP config. When set (byte offset of checksum), any write
to that DP recomputes CRC-16/MODBUS over `payload[0:offset]` and stores the
result as uint16 little-endian at `payload[offset:offset+2]`.

This is generic and may help other Inkbird/raw-blob devices; ISC-028-BW is
the first consumer.

## Test plan

- [x] `pytest tests/test_device_config.py -k crc16_modbus`
- [ ] Add device in Tuya Local with protocol **3.5**, product `8l5th5vqvszbuvlm`
- [ ] Reads: grill/meat temps, fan PID %, hold target match hardware LCD
- [ ] Write **Sound** off → device stops beeping
- [ ] Write **Temperature unit** °C → LCD switches to Celsius
- [ ] **Climate Off** → fan stops; **Heat** + target temp → fan runs toward hold
- [ ] Diagnostics: cached DP 102 changes after writes; bytes 90–91 match new CRC

## Notes

- Different protocol from ISC-007-BW (integer DPs) and IBBQ-4BW (integer probe DPs).
- Similar raw/base64 style to INT-12-BW, but ISC-028 uses one CRC-protected settings blob.
- Presets (DP 104–108) not exposed yet.

### DIFF
--- a/custom_components/tuya_local/devices/device_config_schema.json
+++ b/custom_components/tuya_local/devices/device_config_schema.json
@@ -227,6 +227,11 @@
                                     "enum": ["big", "little"],
                                     "description": "The endianness of masked data in the base64 or hex DP. Default is 'big', so only 'little' needs to be explicitly specified."
                                 },
+                                "crc16_modbus": {
+                                    "type": "integer",
+                                    "minimum": 2,
+                                    "description": "Byte offset of a CRC-16/MODBUS checksum stored as uint16 little-endian. Recomputed over bytes 0 to offset-1 whenever this DP is written."
+                                },
                                 "unit": {
                                     "type": "string",
                                     "description": "The unit of measurement for a sensor DP (e.g., °C, %, W). Some units can use ASCII replacements like C for °C, F for °F, and ugm3 for μg/m³ to simplify input and avoid accidental incorrect Unicode character usage."
--- a/custom_components/tuya_local/devices/inkbird_isc028bw_smokercontrol.yaml
+++ b/custom_components/tuya_local/devices/inkbird_isc028bw_smokercontrol.yaml
@@ -1,0 +1,351 @@
+# Inkbird ISC-028-BW WiFi smoker (5 probes, fan PID, hold target)
+# Product id: 8l5th5vqvszbuvlm | Tuya model id: e1ku12z0
+# DP 101 real_time_data: 11 bytes — 5x uint16 LE (÷10 °F) + fan % (byte 10, read-only)
+# DP 102 setting_para: 92 bytes — byte 0 fan; byte 1 °C/°F; bytes 7-8 hold (÷10 °F); byte 80 sound
+# DP 102 bytes 90-91: CRC-16/MODBUS over bytes 0-89 (crc16_modbus: 90 on writes)
+# Masks for endianness: little — FFFF at LSB end (see INT-12-BW pattern)
+# Requires Tuya protocol 3.5 locally
+
+name: BBQ smoker
+products:
+  - id: 8l5th5vqvszbuvlm
+    manufacturer: Inkbird
+    model: ISC-028-BW
+
+entities:
+  - entity: climate
+    dps:
+      - id: 102
+        type: base64
+        optional: true
+        force: true
+        crc16_modbus: 90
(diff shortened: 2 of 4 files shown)


bjursten:
**Tested successfully** on Inkbird ISC-028-BW (product `8l5th5vqvszbuvlm`), Home Assistant OS 2026.5.4, Tuya Local 2026.6.0 with this PR, protocol **3.5**.

Verified working:
- All DP 102 writes (sound, °C/°F, climate fan on/off, hold target)
- Reads: grill + 4 meat probes, fan PID %
- DP 102 CRC-16/MODBUS valid after writes (bytes 90–91 match computed checksum)

Example live state from diagnostics: grill 71.4 °F, hold 77.3 °F (25.2 °C), fan PID 26%, sound off, LCD °C, climate heat. Device type `inkbird_isc028bw_smokercontrol`, `api_working`: true, no integration issues reported.

make-all:
I'm not sure I want extra protocols on top of protocols in this integration. It is already overloaded.

Can you not layer this on top with a separate integration?

bjursten:
Thanks for reviewing this and for the honest feedback.

I understand not wanting to grow the core further. For this device, the CRC step is required for DP 102 writes to be accepted — without it, climate/unit/sound changes are silently rejected — so a YAML-only contribution isn't viable upstream.

I won't pursue a separate integration. I'll maintain the device config and the small `crc16_modbus` patch locally for my own setup instead.

Thanks again for tuya-local — it's been a great base for exploring this hardware. I'll close this PR shortly.

bjursten:
Closing — maintaining locally.

PR #5239, opened 2026-06-04
add: Support of Afyeev 32A EV Charger

vanyasvl:
Changes:
- Added a new device configuration for the Afyeev 32A 7kW EV Charger.
- Restricted the Charging Current selection to only the values supported by the device, so users can no longer select an out-of-range current.

Reference: [Charger manual](https://manuals.plus/ae/1005007587878624)

### DIFF
--- a/custom_components/tuya_local/devices/afyeev_32a_evcharger.yaml
+++ b/custom_components/tuya_local/devices/afyeev_32a_evcharger.yaml
@@ -33,13 +33,6 @@ entities:
         type: string
         name: system_version
   - entity: sensor
-    name: Firmware version
-    category: diagnostic
-    dps:
-      - id: 23
-        type: string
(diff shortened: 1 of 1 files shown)


make-all:
I'm not sure I need multiple configs for the same device, as there are over 1500 device configs already and I am starting to get complaints about performance (albiet from a user who insists on using remote network based storage over an ssh link). I get frequent requests from users for setting out of range input values for chargers, in particular allowing them to be set lower than the minimum advertised current so they can optimise the  charging for solar output so it seems that in general chargers can accept a wider range of values than advertised.

vanyasvl:
This particular charger doesn't accept to set currents except fixed values form the list, that's why I limit it. I did many tests with this charger to filter what values for what input values and modes are accepted and works.

PR #5240, opened 2026-06-04
Add Intex PureSpa inflatable spa (salt water model)

### LINKED ISSUE #5273: Request support for SPA PRODUCT WITH SALT
### Log message

```text
Device matches None with quality of 0%. LOCAL DPS: {"updated_at": 1780862961.3775747, "103": true, "104": true, "106": true, "107": false, "108": false, "109": 90, "110": 68, "114": 0, "117": "off"}
```

### Product ID

chsaskllmust5d7a

### Product Name

SPA PRODUCT WITH SALT

### DPS information

```text
name: SPA PRODUCT WITH SALT
products:
  - id: chsaskllmust5d7a
entities:
  - entity: switch
    name: Disinfection
    dps:
      - id: 103
        type: boolean
        name: switch
  - entity: switch
    name: Main switch
    dps:
      - id: 104
        type: boolean
        name: switch
  - entity: switch
    name: Filtration
    dps:
      - id: 106
        type: boolean
        name: switch
  - entity: switch
    name: Bubbles
    dps:
      - id: 107
        type: boolean
        name: switch
  - entity: switch
    name: Water heater
    dps:
      - id: 108
        type: boolean
        name: switch
  - entity: number
    name: Set water temperature
    dps:
      - id: 109
        type: integer
        name: value
        unit: F
        range:
          min: 50
          max: 104
  - entity: sensor
    name: Water temperature
    class: temperature
    dps:
      - id: 110
        type: integer
        name: sensor
        unit: F
        class: measurement
```

The device is an Intex whirlpool.

EnO33:
Adds support for the Intex PureSpa inflatable spa (salt water model, controller SB-HSWF20-2, product key chsaskllmust5d7a)

Climate entity (heat/off via DP 108, setpoint DP 109, current temp DP 110). The device reports temperatures in Fahrenheit, converted to the user's unit via unit: F
Switches: power (104), filtration (106), bubbles (107), salt sanitizer (103). All switch mappings verified against the physical control panel states
Diagnostic timer sensor (114)
Preset mode (117) observed only with value warm; other values not confirmed

Tested on a single device

### DIFF
--- a/custom_components/tuya_local/devices/intex_purespa_spa.yaml
+++ b/custom_components/tuya_local/devices/intex_purespa_spa.yaml
@@ -4,17 +4,16 @@ products:
     manufacturer: Intex
     model: PureSpa
 entities:
-  - entity: climate
-    translation_key: thermostat
+  - entity: water_heater
     dps:
       - id: 108
         type: boolean
-        name: hvac_mode
+        name: operation_mode
         mapping:
           - dps_val: false
             value: "off"
           - dps_val: true
-            value: heat
+            value: electric
       - id: 109
         type: integer
         name: temperature
@@ -26,12 +25,6 @@ entities:
         type: integer
         name: current_temperature
         unit: F
-      - id: 117
-        type: string
-        name: preset_mode
-        mapping:
-          - dps_val: warm
-            value: warm
   - entity: switch
     name: Power
     dps:
@@ -53,17 +46,15 @@ entities:
         type: boolean
         name: switch
   - entity: switch
-    name: Sanitizer
-    icon: "mdi:shaker-outline"
+    translation_key: electrolytic_sterilization
     dps:
       - id: 103
         type: boolean
         name: switch
   - entity: sensor
-    name: Timer
-    category: diagnostic
+    translation_key: time_remaining
     class: duration
-    icon: "mdi:timer-outline"
+    category: diagnostic
     dps:
       - id: 114
         type: integer
--- a/custom_components/tuya_local/devices/intex_purespa_spa.yaml
+++ b/custom_components/tuya_local/devices/intex_purespa_spa.yaml
@@ -1,0 +1,71 @@
+name: Spa
+products:
+  - id: chsaskllmust5d7a
+    manufacturer: Intex
+    model: PureSpa
+entities:
+  - entity: climate
+    translation_key: thermostat
+    dps:
+      - id: 108
+        type: boolean
+        name: hvac_mode
+        mapping:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/intex_purespa_spa.yaml
```diff
@@ -0,0 +1,71 @@
+name: Spa
+products:
+  - id: chsaskllmust5d7a
+    manufacturer: Intex
+    model: PureSpa
+entities:
+  - entity: climate
```
make-all: water_heater might be a better match if it only does heating.

custom_components/tuya_local/devices/intex_purespa_spa.yaml
```diff
@@ -0,0 +1,71 @@
...
+        name: switch
+  - entity: switch
+    name: Sanitizer
+    icon: "mdi:shaker-outline"
+    dps:
+      - id: 103
+        type: boolean
+        name: switch
+  - entity: sensor
+    name: Timer
```
make-all: Use `translation_key: time_remaining` for timer sensors (instead of explicit name and icon).

But a method for setting the timer appears to be missing, so perhaps this is not just a sensor.

custom_components/tuya_local/devices/intex_purespa_spa.yaml
```diff
@@ -0,0 +1,71 @@
...
+        name: switch
+  - entity: switch
+    name: Bubbles
+    icon: "mdi:chart-bubble"
+    dps:
+      - id: 107
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Sanitizer
```
make-all: There are translation keys for `uv_sterilization` and `electrolytic_sterilization` which might be appropriate here if this is one of those.

make-all (changes requested):

EnO33:
water_heater, converted to water_heater with operation_mode mapping electric/off, since the spa only heats. Also dropped DP 117 (preset_mode) as water_heater has no preset and it only ever reported warm

timer, switched to translation_key: time_remaining. On the setting method: I couldn't find a writable DP for the timer in the local data; DP 114 looks like a read-only countdown (stayed at 0 during capture). The filtration runtime timer is set on the physical panel (2h/4h/6h/off) but I didn't observe a corresponding settable DP, so it's exposed read-only for now. Happy to make it a number/select if the right DP turns up

sanitizer, it's a salt water system, so used translation_key: electrolytic_sterilization

PR #5257, opened 2026-06-05
Add Klamer IR Heater 2600 W.

reitermarkus:
Add support for Klamer Infrared Heater 2600 W (https://klamer-home.com/products/infrarotheizung).

### DIFF
--- a/custom_components/tuya_local/devices/klamer_ir_heater_2600w.yaml
+++ b/custom_components/tuya_local/devices/klamer_ir_heater_2600w.yaml
@@ -33,6 +33,14 @@ entities:
             conditions:
               - dps_val: f
                 value_redirect: temp_current_f
+      - id: 4
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: Low
+            value: Low
+          - dps_val: High
+            value: High
       - id: 13
         type: string
         name: temperature_unit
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/klamer_ir_heater_2600w.yaml
```diff
@@ -0,0 +1,173 @@
+name: KLAMER Infrared Heater 2600 W
```
make-all: Top level name should be without branding.
Brands belong under the products section connected to a product id.

custom_components/tuya_local/devices/klamer_ir_heater_2600w.yaml
```diff
@@ -0,0 +1,173 @@
...
+        range:
+          min: 41
+          max: 113
+        hidden: true
+      - id: 15
+        type: integer
+        name: temp_current_f
+        hidden: true
+  - entity: switch
+    name: Open Window Detection
```
make-all: HA naming convention is to only capitalise the first word of an entity or device name.

custom_components/tuya_local/devices/klamer_ir_heater_2600w.yaml
```diff
@@ -0,0 +1,173 @@
...
+    category: config
+    icon: "mdi:window-open-variant"
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+  - entity: select
+    translation_key: mode
```
make-all: This can be preset_mode of the climate entity above

custom_components/tuya_local/devices/klamer_ir_heater_2600w.yaml
```diff
@@ -0,0 +1,173 @@
...
+  - entity: sensor
+    translation_key: time_remaining
+    class: duration
+    category: diagnostic
+    icon: "mdi:timer"
+    dps:
+      - id: 20
+        type: integer
+        name: sensor
+        range:
```
make-all: range is not needed here, as it is a read-only sensor which will just show whatever the device is reporting

make-all (changes requested):

PR #5262, opened 2026-06-06
add support for Martin Jerry dimmers US-STD-KN01 and US-SD-TC01

wjcarpenter:
add support for Martin Jerry dimmers [US-STD-KN01](https://www.martinjerry.com/us-std-kn01-support) and [US-SD-TC01](https://www.martinjerry.com/us-sd-tc01)

These devices have fairly simple controls and only expose 2 DPs, so I couldn't modify an existing devices file.


make-all:
simple_lightbulb.yaml matches these.

You have put your device ids under products instead of product id, so there is nothing here that can be added to the existing config.

PR #5263, opened 2026-06-06
  New device: Reqteq Deck Boss 800 grill

hartmms:
tested to work on my new pellet gril. 
I added the grill to my wifi using the tuya app, then used the Tuya cloud API access to verify all the device items were correct for this specific device model.


custom_components/tuya_local/devices/recteq_rt800db_grill.yaml
```diff
@@ -0,0 +1,88 @@
+name: Grill
+# products:
```
make-all: Not sure why this is commented out.

custom_components/tuya_local/devices/recteq_rt800db_grill.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - id: 105
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 255
+        mapping:
+          - step: 5
+  - entity: number
+    name: Temperature adjust
```
make-all: Use `translation_key: temperature_calibration`

Is there a scale factor, or is it adjusting +/- 127 F?

make-all (changes requested):

hartmms:
uncommented name and added a translation key. 

I'm not 100% sure why the range is 127 to -127. It's the same as other recte(c|q)* devices here. The control panel on the device shows.  It's an RTD temperature probe, and I see other conversations online stating that 127 and -127 is a degrees offset.  Not sure of the units, but I suspect C.

make-all:
Changes mentioned in last comment does not seem to have been pushed.

hartmms:
New push with requested changes.

PR #5264, opened 2026-06-06
support for modern flames orion device lights/flame only mode

hartmms:
The original device configuration only supported 3 modes:  off, heat and fan only.   This specific fireplace supports a "flame/lights" only mode where neither the heater or fan runs.  However, the current device configuration doesn't support setting that mode.  

I added a "cool" mode for the HVAC device that will enable only the flame/lights mode.


make-all:
I think it would be better to separate the device power from the climate entity. "off" for the climate entity would then be "flame only" mode, with a separate switch for device power. This is what was done with the recently added touchstone_sidelineredblue_fireplace.yaml which seems to operate the same way.

hartmms:
move climate from boolean switch to pure hvac_mode with preset
rename eco to low heat, boost to high heat
added boolean switch for power control

PR #5267, opened 2026-06-07
Add EuropAce EDH3122A Dehumidifier

c1em3ntchua:
Added europace_edh3122a_dehumidifier.yaml to support EuropAce EDH3122A Dehumidifier. Have been using the config for a few months and has been stable.

### DIFF
--- a/custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
@@ -1,0 +1,186 @@
+name: Dehumidifier
+products:
+  - id: fj86ko
+    manufacturer: EuropAce
+    model: EDH3122A 12L Dehumidifier
+entities:
+  - entity: humidifier
+    class: dehumidifier
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 2
+        type: string
+        name: mode
+        mapping:
+          - dps_val: "0"
+            value: Auto
+          - dps_val: "1"
+            value: Continuous
+      - id: 3
+        type: integer
+        name: current_humidity
+      - id: 4
+        type: integer
+        name: humidity
+        range:
+          min: 30
+          max: 80
+        mapping:
+          - step: 5
+  - entity: fan
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 6
+        type: string
+        name: speed
+        mapping:
+          - dps_val: "0"
+            value: 100
+          - dps_val: "1"
+            value: 50
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 7
+        type: boolean
+        name: lock
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 11
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 11
+        type: bitfield
+        name: fault_code
+      - id: 11
+        name: description
+        type: bitfield
+        mapping:
+          - dps_val: 0
+            value: OK
+          - dps_val: 1
+            value: "E2: Room temperature sensor fault"
+          - dps_val: 2
+            value: "E1: Coil sensor fault"
+          - dps_val: 4
+            value: "P1: Defrost"
+          - dps_val: 8
+            value: "FL: Tank full"
+          - dps_val: 16
+            value: "LO: Low temperature"
+          - dps_val: 32
+            value: "HI: High temperature"
+  - entity: binary_sensor
+    translation_key: tank_full
+    category: diagnostic
+    dps:
+      - id: 11
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 8
+            value: true
+          - value: false
+  - entity: binary_sensor
+    translation_key: defrost
+    category: diagnostic
+    dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
```diff
@@ -0,0 +1,186 @@
+name: Dehumidifier
+products:
+  - id: fj86ko
```
make-all: This does not seem to be the correct product id. Product id for WiFi devices should be 16 characters in length.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
```diff
@@ -0,0 +1,186 @@
...
+    manufacturer: EuropAce
+    model: EDH3122A 12L Dehumidifier
+entities:
+  - entity: humidifier
+    class: dehumidifier
+    dps:
```
make-all, line 10: Add `translation_key: extended` to extend translations to cover continuous mode.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
```diff
@@ -0,0 +1,186 @@
...
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 2
+        type: string
+        name: mode
+        mapping:
+          - dps_val: "0"
+            value: Auto
```
make-all: Use lower case for modes to match translation keys.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
```diff
@@ -0,0 +1,186 @@
...
+    translation_key: timer
+    category: config
+    dps:
+      - id: 12
+        type: string
+        name: option
+        optional: true
+        mapping:
+          - dps_val: "0"
+            value: "Off"
```
make-all: Use `cancel` to match the translation key.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
```diff
@@ -0,0 +1,186 @@
...
+    translation_key: time_remaining
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 13
+        type: integer
+        name: sensor
+        unit: min
+  - entity: switch
+    name: Inside drying
```
make-all: "Anti mould" would be a more understandable name.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
```diff
@@ -0,0 +1,186 @@
...
+    category: config
+    icon: "mdi:hair-dryer"
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+        optional: true
+        persist: false
```
make-all: Persist: false is a bit strange here. Why would you need this to not persist?

make-all (dismissed):

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
c1em3ntchua: Found the product id, will make requested change.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
c1em3ntchua: Will make requested change.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
c1em3ntchua: Will make requested change.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
c1em3ntchua: Will make requested change.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
c1em3ntchua: I'd like to suggest to retain this, as both the manual and the wording on the device label the switch as 'Inside Drying'. Let me know what you think.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
c1em3ntchua: I adapted this from `shinco_30d_dehumidifier.yaml`. I can make the requested change if it is not necessary.

custom_components/tuya_local/devices/europace_edh3122a_dehumidifier.yaml
make-all: Currently there is only one device using this name, and the name to me sounds too literally technical, describing technically what it is doing without conveying why you would want that.  There are 9 devices using the name "Anti-mould" for same function, and another 5 using "Anti-mildew". I would prefer to go with the majority now rather than adding to the work that will need to be done to migrate the variations when a translation is eventually added.

c1em3ntchua: Thanks for the explanation. Respect the decision.

PR #5269, opened 2026-06-07
Add temperature relay controller device

ma7moodz:
Adds support for the Tuya temperature relay controller.

Product ID: ipomh65jy8nuiwwm
Product name: 温度电量控制器

Exposes:

* Relay switch (DP 2)
* Temperature sensor (DP 27, scale 10)
* Voltage sensor (DP 13, scale 10)

Tested locally with Tuya Local.

Before this addition, the device was automatically identified as a thermostat. However, it only exposes an on/off relay and temperature measurement and does not provide any temperature control functionality. As a result, the thermostat configuration did not correctly represent the device.

After adding this device definition, the relay, temperature, and voltage entities were created and working correctly.


ma7moodz:
After removing the explicit Temperature and Voltage entity names to satisfy the YAML tests, I retested the device locally. The relay, temperature, and voltage entities are still created correctly and functioning as expected.

make-all:
I will not add odd devices like this without some branding info, as it is confusing to users to see generic names for a device which is anything but generic.

custom_components/tuya_local/devices/temperature_relay_controller.yaml
```diff
@@ -0,0 +1,35 @@
+name: Temperature relay controller
+products:
+  - id: ipomh65jy8nuiwwm
+    name: 温度电量控制器
```
make-all, line 4: manufacturer and model should be here rather than name.

ma7moodz: Thanks for taking a look.

I bought this device on AliExpress, but it appears to be completely unbranded. The enclosure is labeled only "Smart Switch", and I couldn't find a manufacturer or model number on the device, packaging, or listing, I attached an image and link of the exact listing I bought it from.

In Home Assistant it shows up as:

* Manufacturer: Tuya
* Model: 温度电量控制器 (ipomh65jy8nuiwwm)

Would you like me to use those values in the device definition 
manufacturer: Tuya
model: 温度电量控制器 (ipomh65jy8nuiwwm)
Or is there another naming convention you would prefer for unbranded devices?
Please let me know what I can do from my side to have this up to the standards of this project. 

AliExpress listing link:
https://a.aliexpress.com/_c34fu92p

Relay_TempProbe_Tuya

make-all:
I suggest to use SciMagic-RC as the manufacturer, and scimagicrc_temperature_relay.yaml as the filename. The model can be left out, as it does not seem to have one that is obvious in that listing.

make-all:
Oh wait, TYTE-D1 is the model.  There is already a TYTE-D1 configuration file, but it has many more datapoints than you have here, but dps 2, 27 and 13 do match the functions they are given here.

It has separate control for the relay as well as climate control for automatic control.

make-all:
Why is this needed instead of the existing TYTE-D1 config?

make-all:
There does not appear to be a genuine need for this config

PR #5278, opened 2026-06-08
kolanky_evcharger: support GOODCELL 32A variant charge current

jaypo82:
## Why

The `7bvgooyjhiua1yyq` product id ships as both a Kolanky 3.6kW (16A) and a GOODCELL 7.2kW 32A charger, but the charge current was hardcoded to dp 114 (`Set16A`, 6–16A). On a 32A unit that dp clamps at 16, so anything above 16A just snaps back — there's no way to set the real current from HA.

## What

- Charge current now follows `DeviceMaxSetA` (dp 113): writes `Set32A` (dp 115, 8–32A) by default, and redirects to `Set16A` (dp 114, 6–16A) when the charger reports `Max16A`.
- Added GOODCELL 7.2kW 32A to the product list (same product id as the existing Kolanky 3.6kW entry).

Same `constraint`/`value_redirect` approach already used in `goodcell_ev_charger.yaml`.

## How to Test

- 32A unit (`DeviceMaxSetA` = `Max32A`): charge current exposes 8–32A and writes dp 115; the setpoint now holds instead of reverting to 16. Verified on a GOODCELL 7.2kW 32A unit.
- 16A unit (`DeviceMaxSetA` = `Max16A`): the redirect sends writes to dp 114 with the original 6–16A range — i.e. unchanged from current behaviour. (Not tested on physical 16A hardware; behaviour mirrors the pre-change config and the goodcell pattern.)
- `uv run pytest tests/test_device_config.py` and `uv run yamllint custom_components/tuya_local/devices`.

### DIFF
--- a/custom_components/tuya_local/devices/kolanky_evcharger.yaml
+++ b/custom_components/tuya_local/devices/kolanky_evcharger.yaml
@@ -1,11 +1,7 @@
 name: EV charger
 products:
   - id: 7bvgooyjhiua1yyq
-    manufacturer: Kolanky
-    model: "3.6kW"
-  - id: 7bvgooyjhiua1yyq
-    manufacturer: GOODCELL
-    model: "7.2kW 32A"
+    manufacturer: Kolanky / Goodcell
 entities:
   - entity: sensor
     translation_key: status
--- a/custom_components/tuya_local/devices/kolanky_evcharger.yaml
+++ b/custom_components/tuya_local/devices/kolanky_evcharger.yaml
@@ -54,29 +54,29 @@ entities:
     category: config
     class: current
     dps:
-      - id: 115
+      - id: 114
         type: integer
         name: value
         unit: A
         range:
-          min: 8
-          max: 32
+          min: 6
+          max: 16
         mapping:
           - constraint: max_current
             conditions:
-              - dps_val: Max16A
+              - dps_val: Max32A
                 value_redirect: value_alt
                 range:
-                  min: 6
-                  max: 16
-      - id: 114
+                  min: 8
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/kolanky_evcharger.yaml
```diff
@@ -51,13 +54,37 @@ entities:
...
-      - id: 114
+      - id: 115
         type: integer
         name: value
         unit: A
+        range:
+          min: 8
+          max: 32
+        mapping:
+          - constraint: max_current
```
make-all, line 61: This should be the other way around to avoid the case where an overload current can be set if the device does not return any max_current (ie the lowest current model should be default).
Although they use different dps, I don't know how the different models react, and what protection is in the apparently shared firmware against issues like this.

make-all (changes requested):

custom_components/tuya_local/devices/kolanky_evcharger.yaml
jaypo82: Good call — flipped it. dp 114 (Set16A, 6–16A) is now the default, so a device that doesn't report `max_current` stays on the safe low range; it only escalates to dp 115 (Set32A, 8–32A) when `DeviceMaxSetA` reports `Max32A`.

On the 32A unit I have, `DeviceMaxSetA` reads `Max32A`, dp 115 holds the setpoint across 8–32A, and dp 114 was clamped to 16 — which is what prompted this. I don't have a 16A unit to confirm that side, but the default path is unchanged from the original config.

PR #5279, opened 2026-06-08
Add BW salt chlorinator device config

phooper-ic:
Add support for BW Series Salt Chlorinator, also sold as Thermotec Salt Chlorinator in the UK.

Product ID: ypdzzmgqbq3dqhqu
Model ID: fxalqw

Confirmed DPS:
101 power
102 child lock
103 work mode/output level
104 water temperature
105 fault bitmap
106 fault status
107 salinity status
108 fault clear
109 self-clean time
110 self-clean remaining
111 salinity
112 salinity clear
113 work countdown / remaining time
114 work time

Notes:
- F-code descriptions are taken from the BW Series user manual troubleshooting table.
- DPs 113 and 114 are marked rw in the Tuya model, but in cloud app they are read only, so they are exposed as diagnostic sensors only. May do further testing to see if they are safe to write to.
- Manual available here: https://manuals.plus/m/eddc529b18f8f768901ed1aa3c9a6035249842e99aa8044ef6688fde0a7daab2_optim.pdf

### DIFF
--- a/custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
+++ b/custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
@@ -53,6 +53,15 @@ entities:
         class: measurement
 
   - entity: sensor
+    name: Fault Bitmap
+    category: diagnostic
+    dps:
+      - id: 105
+        name: sensor
+        type: integer
+        optional: true
+
+  - entity: sensor
     name: Fault Status
     class: enum
     category: diagnostic
@@ -65,25 +74,27 @@ entities:
           - dps_val: "work"
             value: "working"
           - dps_val: "F1"
-            value: "F1"
-          - dps_val: "F1_L"
-            value: "F1_L"
+            value: "F1 - Water temperature outside normal range"
           - dps_val: "F2"
-            value: "F2"
+            value: "F2 - Water flow error"
           - dps_val: "F3"
-            value: "F3"
+            value: "F3 - Water temperature sensor breakdown"
           - dps_val: "F4"
-            value: "F4"
+            value: "F4 - Electrode abnormal"
+          - dps_val: "F5"
+            value: "F5 - Low salinity"
+          - dps_val: "F6"
+            value: "F6 - High salinity"
           - dps_val: "F8"
-            value: "F8"
+            value: "F8 - Low input voltage"
           - dps_val: "F9"
-            value: "F9"
+            value: "F9 - High output current"
           - dps_val: "FC"
-            value: "FC"
+            value: "FC - Circuit abnormal"
           - dps_val: "FP"
-            value: "FP"
+            value: "FP - Power off abnormal"
           - dps_val: "Fb"
-            value: "Fb"
+            value: "Fb - Communication failure"
 
   - entity: sensor
     name: Salinity Status
@@ -176,13 +187,3 @@ entities:
         type: integer
         unit: h
         class: measurement
-
-  - entity: sensor
-    name: Fault Bitmap
-    class: enum
-    category: diagnostic
-    dps:
-      - id: 105
-        name: sensor
-        type: integer
-        optional: true
--- a/custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
+++ b/custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
@@ -55,7 +55,7 @@ entities:
   - entity: sensor
     name: Fault Bitmap
     category: diagnostic
-    disabled: true
+    hidden: true
     dps:
       - id: 105
         name: sensor
@@ -151,7 +151,7 @@ entities:
   - entity: sensor
     name: Salinity
     category: diagnostic
-    disabled: true
+    hidden: true
     dps:
       - id: 111
         name: sensor
--- a/custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
+++ b/custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
@@ -5,7 +5,7 @@ products:
 
 entities:
   - entity: switch
-    name: 24-hour cycle
+    name: Daily cycle
     dps:
       - id: 101
         name: switch
@@ -42,6 +42,7 @@ entities:
   - entity: sensor
     name: Water temperature
     class: temperature
+    icon: "mdi:pool-thermometer"
     dps:
       - id: 104
         name: sensor
--- a/custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
+++ b/custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
@@ -1,28 +1,25 @@
-name: Salt Chlorinator
+name: Salt chlorinator
 products:
   - id: ypdzzmgqbq3dqhqu
-    name: BW Salt Chlorinator
-    model_id: fxalqw
     manufacturer: BW
 
 entities:
   - entity: switch
-    name: Power
+    name: 24-hour cycle
     dps:
       - id: 101
         name: switch
         type: boolean
 
-  - entity: switch
-    name: Child Lock
-    category: config
+  - entity: lock
+    translation_key: child_lock
     dps:
       - id: 102
-        name: switch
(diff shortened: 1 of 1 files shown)


phooper-ic:
Further local testing showed DP105 Fault Bitmap and DP111 Salinity remain unknown on the tested device because they do not appear to be reported through the local DPS interface. I have therefore made them disabled by default while keeping them available as optional diagnostic entities.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
+name: Salt Chlorinator
+products:
+  - id: ypdzzmgqbq3dqhqu
+    name: BW Salt Chlorinator
+    model_id: fxalqw
+    manufacturer: BW
+
+entities:
+  - entity: switch
+    name: Power
```
make-all: Generally a name is not needed for the main switch of a device. Unnamed switches will appear with the device name in the UI.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
...
+    dps:
+      - id: 101
+        name: switch
+        type: boolean
+
+  - entity: switch
+    name: Child Lock
```
make-all: Standard pattern for child locks is to use a lock entity with `translation_key: child_lock`. Dp name needs to be `lock`

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
...
+    category: config
+    dps:
+      - id: 102
+        name: switch
+        type: boolean
+
+  - entity: select
+    name: Output Level
```
make-all: HA naming convention is to only capitalise the first word of an entity or device name.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
+name: Salt Chlorinator
```
make-all: HA naming convention is to only capitalise the first word of an entity or device name.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
...
+products:
+  - id: ypdzzmgqbq3dqhqu
+    name: BW Salt Chlorinator
```
make-all: Name not needed here. model should be used before model_id (model_id is only needed if you want to put both a readable model name and a model id). But the model should be from a label on the device or packaging, not from Tuya cloud API data (these are just database ids and have no relevance to this integration).

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
...
+    class: temperature
+    dps:
+      - id: 104
+        name: sensor
+        type: integer
+        unit: C
+        class: measurement
+
+  - entity: sensor
+    name: Fault Bitmap
```
make-all: Use a binary_sensor with `class: problem` as follows:
```yaml
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 105
        type: bitfield
        optional: true
        name: sensor
        mapping:
          - dps_val: null
            value: false
          - dps_val: 0
            value: false
          - value: true
      - id: 105
        type: bitfield
        optional: true
        name: fault_code
```

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
...
+    category: diagnostic
+    hidden: true
+    dps:
+      - id: 105
+        name: sensor
+        type: integer
+        optional: true
+
+  - entity: sensor
```
make-all: I think you can just add this as an attribute to the binary_sensor above:

```
      - id: 106
        type: string
        optional: true
        name: description
```
With the mapping below to expand to more detailed description.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
...
+            value: "FC - Circuit abnormal"
+          - dps_val: "FP"
+            value: "FP - Power off abnormal"
+          - dps_val: "Fb"
+            value: "Fb - Communication failure"
+
+  - entity: sensor
+    name: Salinity Status
+    class: enum
+    category: diagnostic
```
make-all, line 99: `translation_key: status` will allow the low/normal/high to be translated.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
...
+    name: Clear Fault
+    category: config
+    dps:
+      - id: 108
+        name: button
+        type: boolean
+        optional: true
+
+  - entity: number
+    name: Self-Clean Time
```
make-all: "Self clean interval" is maybe clearer about what kind of "time" this is.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
...
+  - entity: button
+    name: Clear Salinity Fault
+    category: config
+    dps:
+      - id: 112
+        name: button
+        type: boolean
+
+  - entity: sensor
+    name: Work Countdown
```
make-all: Not sure the meaning of this. The Tuya naming is sometimes not very clear, so an improvement on the name is probably possible.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
```diff
@@ -0,0 +1,191 @@
...
+    category: diagnostic
+    dps:
+      - id: 113
+        name: sensor
+        type: integer
+        unit: h
+        class: measurement
+
+  - entity: sensor
+    name: Work Time
```
make-all: This is maybe better named as "Runtime"

make-all (changes requested):

phooper-ic:
Thanks for the review and suggestions. I've updated the device definition accordingly, including the naming, child lock entity, and configuration layout changes.

While testing the revised fault handling, I induced an F2 water flow fault and found that DP106 correctly reports the fault state (`F2`), however DP105 is not present in the local DPS updates on the tested device. As a result, I have removed DP105 and based the problem sensor on DP106 instead. The fault condition and description now update correctly during testing.

I also removed DP111, which similarly did not report any values locally during testing. The remaining entities are all based on DPS values that have been observed and verified on the device.

Additionally, I renamed several entities to better reflect the device behaviour discovered during testing. For example, the chlorination output setting controls runtime within the 24-hour operating cycle rather than modulating the cell output directly, and the runtime/self-clean sensors have been named to reflect their actual function.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
phooper-ic: Testing the device showed this is not a main switch in the traditional sense, but instead enables/disables the device's repeating 24-hour operating cycle. I'll implement an explicit "24-hour cycle" name to better reflect the device behaviour.

custom_components/tuya_local/devices/bw_salt_chlorinator.yaml
phooper-ic: I tested the fault handling by inducing an F2 water flow fault. During the fault, DP106 correctly reports F2, however DP105 was not present in the local DPS updates. Using DP105 as the primary fault indicator therefore resulted in the sensor remaining OK while the fault description showed an active error. The problem sensor has therefore been based on DP106, which correctly reflects the fault state during testing. DP105 has been removed entirely (similarly for DP111).

PR #5290, opened 2026-06-09
feat: Atomi coffee maker device config

btoddb:
## Device

Atomi Smart Coffee Maker
- Product Category = kfj
- product_name = Coffee Maker
- product_id = gmbbmrpkqecjoeoa
- Model = RE70X-TC02
- modelId = 000002cuuc

## Testing
- HA Core production deployment = 2026.6.1
- Made lots of notes in the device config
- Some of the reported fields don't seem to be used - noted in config as comments

attached are diagnostics from iot.tuya.com, and 000002cuuc-kfj-dashboard-card.yaml is a sample dashboard tile using stock components

[000002cuuc-kfj-category-instruction-set.json](https://github.com/user-attachments/files/28769416/000002cuuc-kfj-category-instruction-set.json)
[000002cuuc-kfj-dashboard-card.yaml](https://github.com/user-attachments/files/28769417/000002cuuc-kfj-dashboard-card.yaml)
[000002cuuc-query-device-details.json](https://github.com/user-attachments/files/28769418/000002cuuc-query-device-details.json)
[000002cuuc-query-properties.json](https://github.com/user-attachments/files/28769419/000002cuuc-query-properties.json)
[000002cuuc-query-things-data-model.json](https://github.com/user-attachments/files/28769420/000002cuuc-query-things-data-model.json)

### DIFF
--- a/custom_components/tuya_local/devices/atomi_coffee_maker.yaml
+++ b/custom_components/tuya_local/devices/atomi_coffee_maker.yaml
@@ -36,11 +36,12 @@ entities:
   - entity: sensor
     # read only
     # status of the machine
-    # Manual Schedule - tapping the "timer/clock" icon on the device sets this status.
-    #   However, I can't figure out how to set up the device such that it reports a Minutes Until Brew
-    #   Press and hold the timer button and then use the hour/minute buttons to set a time, presumably
+    # Manual Schedule - tapping the "timer/clock" icon on the device sets
+    #   this status. However, I can't figure out how to set up the device
+    #   such that it reports a Minutes Until Brew Press and hold the timer
+    #   button and then use the hour/minute buttons to set a time, presumably
     #   when the device should start brewing
-    name: Status
+    translation_key: status
     class: enum
     category: diagnostic
     dps:
@@ -76,8 +77,9 @@ entities:
 
   - entity: sensor
     # read only
-    # value of fault - this is for development, and can be removed when figure out what other bits mean.
-    #  When (if) the other bits are determined, setup a binary_sensor like Carafe not on base
+    # value of fault - this is for development, and can be removed when
+    #   figure out what other bits mean. When (if) the other bits are
+    #   determined, setup a binary_sensor like Carafe not on base
     #  (bit 2 is carafe not on base)
     name: Raw fault status
     category: diagnostic
@@ -113,9 +115,11 @@ entities:
     # read only
     # marking as optional and hidden.  I've tested a bunch and
     #   I can't figure out to render a value in this field.
-    # I've setup a manual schedule, which worked, but didn't see any changes in the data.
-    # The SmartLife app has a schedule but it doesn't seem to translate to a value for this field.
-    #   My guess is the schedule is a cloud only based feature
+    # I've setup a manual schedule, which worked, but didn't see any changes
+    #   in the data.
+    # The SmartLife app has a schedule but it doesn't seem to translate to a
+    #   value for this field. My guess is the schedule is a cloud only based
+    #   feature
     name: Minutes Until Brew
     hidden: true
     category: diagnostic
@@ -130,8 +134,8 @@ entities:
 
   - entity: button
     name: Start Brewing
-    # The switch may or may not turn on the machine.  For instance, if there isn't
-    #   enough water or carafe is missing
+    # The switch may or may not turn on the machine.
+    #   For instance, if there isn't enough water or carafe is missing
     dps:
       - id: 103
         type: boolean
@@ -173,8 +177,8 @@ entities:
     name: Set Time
     # write only
     # calibration - sets the time when called
-    #   marked as optional because it isn't reported when adding a device and throws off
-    #   the matching algorithm
+    #   marked as optional because it isn't reported when adding a device
+    #   and throws off the matching algorithm
     dps:
       - id: 105
         optional: true
--- a/custom_components/tuya_local/devices/atomi_coffee_maker.yaml
+++ b/custom_components/tuya_local/devices/atomi_coffee_maker.yaml
@@ -1,11 +1,9 @@
-name: Atomi Coffee Maker
+name: Coffee maker
 products:
-  - id: 000002cuuc
-    manufacturer: Atomi
-    name: Atomi Smart Coffee Maker
   - id: gmbbmrpkqecjoeoa
     manufacturer: Atomi
-    name: Atomi Smart Coffee Maker
+    model: RE70X-TC02
+    model_id: 000002cuuc
 
 entities:
   - entity: binary_sensor
@@ -54,16 +52,16 @@ entities:
           - dps_val: status2
             value: Brewing
           - dps_val: status3
-            value: Warming
+            value: Keeping warm
           - dps_val: status4
-            value: Manual Schedule
+            value: Scheduled on device
 
   - entity: binary_sensor
     # read only
     # bit 0 (value = 0), OK (carafe on base)
     # bit 2 (value = 4), carafe not on base
     # not sure what other bits can be set or what they mean
-    class: problem
+    class: occupancy
     name: Carafe
     category: diagnostic
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/atomi_coffee_maker.yaml
```diff
@@ -0,0 +1,189 @@
+name: Atomi Coffee Maker
```
make-all: HA naming convention is to only capitalise the first word of device and entity names.

Convention in this integration is to make the top level name generic, as often there are many different brands sold globally using the same basic hardware. Branding is kept under products only

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
```diff
@@ -0,0 +1,189 @@
...
+products:
+  - id: 000002cuuc
```
make-all: This is not a product id, so this entry should be removed, and only the one below used

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
```diff
@@ -0,0 +1,189 @@
...
+    manufacturer: Atomi
+    name: Atomi Smart Coffee Maker
+  - id: gmbbmrpkqecjoeoa
+    manufacturer: Atomi
+    name: Atomi Smart Coffee Maker
```
make-all: name is not needed under products.
If the model is known, that can be added, but it is OK without

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
```diff
@@ -0,0 +1,189 @@
...
+        name: sensor
+        mapping:
+          - dps_val: status1
+            value: Idle
+          - dps_val: status2
+            value: Brewing
+          - dps_val: status3
+            value: Warming
+          - dps_val: status4
+            value: Manual Schedule
```
make-all: `manual` or `scheduled`. Both together don't really make sense, and I don't know what you mean by that.

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
```diff
@@ -0,0 +1,189 @@
...
+      - id: 3
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: status1
+            value: Idle
+          - dps_val: status2
+            value: Brewing
+          - dps_val: status3
+            value: Warming
```
make-all: `keeping_warm`

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
```diff
@@ -0,0 +1,189 @@
...
+          - dps_val: status4
+            value: Manual Schedule
+
+  - entity: binary_sensor
+    # read only
+    # bit 0 (value = 0), OK (carafe on base)
+    # bit 2 (value = 4), carafe not on base
+    # not sure what other bits can be set or what they mean
+    class: problem
```
make-all: Maybe `occupancy` is a better class to use here.

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
```diff
@@ -0,0 +1,189 @@
...
+    # Read only
+    # water in the tank - this sensor is only evaluated when
+    #   brewing starts, so can't be used to warn ahead of brewing.
+    #   Furthermore, the sensor will change to "Ready" when brewing starts,
+    #   and then change to "Needs water" if the water is low.
+    #   If the coffee maker is "cold" it will take almost a minute
+    #   to become "Needs water"
+    #   Also, even though it has units of ML, seems
+    #   to only be a 1 or 0 (needs water or ready)
+    name: Water Tank
```
make-all: Use `translation_key: tank_empty`. No class is needed as translation key overrides all the same things.

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
```diff
@@ -0,0 +1,189 @@
...
+            value: true
+          - value: false
+
+  - entity: sensor
+    # read only
+    # value of fault - this is for development, and can be removed when
+    #   figure out what other bits mean. When (if) the other bits are
+    #   determined, setup a binary_sensor like Carafe not on base
+    #  (bit 2 is carafe not on base)
+    name: Raw fault status
```
make-all: Add a general binary_sensor with `class: problem` to catch all the other errors without specific sensor. Then add fault_code as an attribute.

make-all (changes requested):

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
btoddb: Thx, still navigating all the information

modelId: 000002cuuc, I found in 000002cuuc-query-things-data-model.json, which is from the Tuya IoT API
model: RE70X-TC02, I found in 000002cuuc-query-device-details.json, which is f

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
btoddb: TL;DR I have no idea where I got Manual Schedule, but it is sorta fitting.  There are 2 ways to schedule with this device, in the app and on the device.  "Manual" is supposed to signify scheduling was done on the device.  I could remove it, call it Scheduled on Device, or something else??

More info:

Through the app you get all the normal/expected scheduling features like, "every day at 6am".  Scheduling on the device only allows setting a time to start brewing - no date, no repeat, etc.  Scheduling through the app never sends a status chaneg, so the user doesn't have any indication a schedule is pending and brewing just starts at the desired time - all in the cloud.  However, pressing the button on the device sends a status = status4, which means "next brew time" according to their manual.

It might be best to remove this since it offers little value, but I left it in for documentation. I suppose the status could be useful to the user, seeing the device ack'ing the next brew time, but I could never find a field with the scheduled time display.

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
btoddb: Not exactly sure what you're looking for, but something like:
- change this entity to be named Fault Code to preserve the raw fault code
- add a new entity like this:
```
  - entity: binary_sensor
    class: problem
    category: diagnostic
    hidden: true
    dps:
      - id: 6
        type: bitfield
        name: sensor
```

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
btoddb: I changed the mapping to flip false and true to make the entity more meaningful.  I understand how the carafe could be like an person "occupying" a space, but not sure I like the values "Detected" and "Clear".  In the end this is just a fancy boolean, "on the base" or "not on the base" and the UI can be designed around whatever the values are.

But what I really don't like is the occupancy class has a default icon of a "house" .. lol

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
btoddb: I personally like the wording "Water tank", but I see it gives nice default water drop icons

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
make-all:
```
  - entity: binary_sensor
    class: problem
    category: diagnostic
    dps:
      - id: 6
        type: bitfield
        name: sensor
        mapping:
          - dps_val: 0
            value: false
          - dps_val: 4
            value: false
          - value: true
      - id: 6
        type: bitfield
        name: fault_code
```

custom_components/tuya_local/devices/atomi_coffee_maker.yaml
make-all: The suggestions I made were based on another comment which I forgot to submit, which was to use values that match existing translations where possible. So `idle`, `keeping_warm` and `scheduled` for this one would be closest I think. Brewing does not really have a match, so I will deal with that separately after merge.

PR #5303, opened 2026-06-11
fix(yuhome_laundry_rack): correct DP ids to match actual device

lweikit:
## Summary

I'm the original author of this config (#4643). After reading the **live DPS** off my device (product id `ttawi9yohqflpt4a`, protocol 3.4), the original DP ids were wrong — several pointed at DPs the device doesn't expose. This corrects them and adds a few capabilities the device actually has.

## Live DPS evidence

| Function | Original (#4643) | Corrected | Live device |
|---|---|---|---|
| action/state | dp 8 | dp **9** | dp9 = `stop` |
| timers | dp 9,10,11 | dp **10,11,13** | dp10/11/13 = `cancel` |
| duration sensors | dp 12,13,14 | dp **15,16,17** | dp15/16/17 = `0` |
| fault | — | dp **18** | dp18 = `0` |
| light effect | — | dp **102** | dp102 = `Cool_White` |
| voice control | — | dp **104** | dp104 = `False` |

The device exposes **no dp 8/12/14** at all, so the original entities on those ids never worked.

## Changes

- Correct action/timer/duration DP ids to match the hardware.
- Add light `effect` (dp 102: Warm/Cool/Dual white).
- Add **Voice control** switch (dp 104).
- Add **Fault** `problem` binary_sensor (dp 18, sensor + fault_code bitfield).
- Add `stop` → idle cover-action mapping.

## Testing

- Verified against my own unit by polling live DPS over the local protocol (v3.4).
- `yamllint` clean.

### DIFF
--- a/custom_components/tuya_local/devices/yuhome_laundry_rack.yaml
+++ b/custom_components/tuya_local/devices/yuhome_laundry_rack.yaml
@@ -5,6 +5,7 @@ products:
     model: Yu Turbo
 entities:
   - entity: switch
+    name: Power
     dps:
       - id: 1
         type: boolean
@@ -28,7 +29,7 @@ entities:
         range:
           min: 0
           max: 100
-      - id: 8
+      - id: 9
         type: string
         name: action
         mapping:
@@ -36,12 +37,25 @@ entities:
             value: closing
           - dps_val: down
             value: opening
+          - dps_val: stop
+            value: idle
           - value: null
   - entity: light
+    name: Light
     dps:
       - id: 4
         type: boolean
         name: switch
+      - id: 102
+        type: string
+        name: effect
+        mapping:
+          - dps_val: Warm_White
+            value: Warm white
+          - dps_val: Cool_White
+            value: Cool white
+          - dps_val: Dual_White
+            value: Dual white
   - entity: switch
     translation_key: uv_sterilization
     dps:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/yuhome_laundry_rack.yaml
```diff
@@ -5,6 +5,7 @@ products:
     model: Yu Turbo
 entities:
   - entity: switch
+    name: Power
```
make-all: This is better without the explicit name

custom_components/tuya_local/devices/yuhome_laundry_rack.yaml
```diff
@@ -28,20 +29,33 @@ entities:
...
         mapping:
           - dps_val: up
             value: closing
           - dps_val: down
             value: opening
+          - dps_val: stop
+            value: idle
           - value: null
   - entity: light
+    name: Light
```
make-all: This is better without the explicit name

custom_components/tuya_local/devices/yuhome_laundry_rack.yaml
```diff
@@ -192,8 +214,23 @@ entities:
...
     category: diagnostic
     dps:
-      - id: 14
+      - id: 17
         type: integer
         optional: true
         name: sensor
         unit: min
+  - entity: binary_sensor
+    name: Fault
```
make-all: This would be better without the explicit name

make-all (changes requested):

PR #4620, opened 2026-03-03
Add Aquark Mr. Silence pool heat pump

sashahilton00:
Adds support for the [Aquark Mr. Silence](https://www.aquark.com/mr-silence/) pool heat pump.

I've written the configuration file based on the query things data model response, reading through the translated descriptions and adjusting accordingly. Have had some issues when attempting to test it locally, so a once over by someone more experienced with writing these files would be great. Error from HA logs is as follows:

```
This error originated from a custom integration.

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:445
integration: Tuya Local (documentation, issues)
First occurred: 01:41:35 (1 occurrence)
Last logged: 01:41:35

Device matches gardenpac_heatpump with quality of 100%. LOCAL DPS: {"updated_at": 1772502080.3085387, "1": false, "102": 12, "103": true, "104": 0, "105": "warm", "106": 26, "107": 18, "108": 40, "115": 0, "116": 0, "117": false}
```

Below is the query things data model response:

```json
{
  "result": {
[...]

PR #5312, opened 2026-06-11
Add Aquark Heat Pump.

reitermarkus:
Add Aquark Heat Pump.

I found that this matches `fairland_iphcr15_heatpump.yaml` and `gardenpac_heatpump.yaml`, except for the temperature range and some naming differences. I have formatted them so the diff between them is minimal.

However, after adding this, adding a new device doesn't suggest any of the three device types. Not sure what is wrong.

### DIFF
--- a/custom_components/tuya_local/devices/fairland_iphcr15_heatpump.yaml
+++ b/custom_components/tuya_local/devices/fairland_iphcr15_heatpump.yaml
@@ -94,7 +94,7 @@ entities:
     category: config
     dps:
       - id: 103
-        type: string
+        type: boolean
         name: option
         mapping:
           - dps_val: false
--- a/custom_components/tuya_local/devices/fairland_iphcr15_heatpump.yaml
+++ b/custom_components/tuya_local/devices/fairland_iphcr15_heatpump.yaml
@@ -17,12 +17,12 @@ entities:
           - dps_val: false
             value: "off"
           - dps_val: true
-            constraint: work_mode
+            constraint: mode
             conditions:
-              - dps_val: cool
-                value: cool
               - dps_val: warm
                 value: heat
+              - dps_val: cool
+                value: cool
               - dps_val: smart
                 value: heat_cool
       - id: 102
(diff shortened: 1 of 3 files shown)


custom_components/tuya_local/devices/fairland_iphcr15_heatpump.yaml
```diff
@@ -89,6 +89,18 @@ entities:
           - dps_val: 4
             value: true
           - value: false
+  - entity: select
+    translation_key: temperature_unit
+    category: config
+    dps:
+      - id: 103
+        type: string
```
make-all: This is inconsistent with the two definitions above.

custom_components/tuya_local/devices/fairland_iphcr15_heatpump.yaml
```diff
@@ -17,12 +17,12 @@ entities:
           - dps_val: false
             value: "off"
           - dps_val: true
-            constraint: mode
+            constraint: work_mode
```
make-all: These naming and ordering changes are not needed, please revert them.

make-all (changes requested):

make-all:
Duplicate of #4620

custom_components/tuya_local/devices/fairland_iphcr15_heatpump.yaml
reitermarkus: Thank you.

Seems like this is what caused none of the three to be suggested.

custom_components/tuya_local/devices/fairland_iphcr15_heatpump.yaml
reitermarkus: I wrote in the PR description:

> I have formatted them so the diff between them is minimal.

So yes, the changes not technically needed, but make comparing the three very similar device types more easy.

make-all: So reformat the new one if that is the goal

reitermarkus: Okay, I reverted the changes in the existing files, except where they were already different.

PR #5320, opened 2026-06-12
Add Poiema Fan Support

Zankio:
Add Poiema Fan Support (https://www.poiema.com.tw/products/fan)


custom_components/tuya_local/devices/poiema_fan.yaml
```diff
@@ -0,0 +1,110 @@
+name: Fan
+products:
+  - id: yrgy1hmepnqysltk
+    manufacturer: Poiema
+    name: Fan
```
make-all: name is not needed here

custom_components/tuya_local/devices/poiema_fan.yaml
```diff
@@ -0,0 +1,110 @@
...
+entities:
+  - entity: fan
+    icon: "mdi:fan"
```
make-all: Icon is not needed here

custom_components/tuya_local/devices/poiema_fan.yaml
```diff
@@ -0,0 +1,110 @@
...
+            value: "90°"
+          - dps_val: "120"
+            value: "120°"
+          - dps_val: "60L"
+            value: "60° slow"
+          - dps_val: "90L"
+            value: "90° slow"
+          - dps_val: "120L"
+            value: "120° slow"
+  - entity: switch
```
make-all: A lock entity with `translation_key: child_lock` is the standard pattern here.

custom_components/tuya_local/devices/poiema_fan.yaml
```diff
@@ -0,0 +1,110 @@
...
+    name: Child lock
+    icon: "mdi:lock"
+    dps:
+      - id: 14
+        type: boolean
+        name: switch
+  - entity: switch
```
make-all: light entity with `translation_key: indicator` is the standard pattern here if this is for an indicator type LED.

custom_components/tuya_local/devices/poiema_fan.yaml
```diff
@@ -0,0 +1,110 @@
...
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 21
+        type: integer
+        name: sensor
+        unit: "°C"
+  - entity: select
+    name: Auto off timer
+    icon: "mdi:timer-outline"
```
make-all: Use `translation_key: timer` to get translations for the select options. In that case, the original dps_val should be used for the value as those are matching the translation keys.

make-all (changes requested):

custom_components/tuya_local/devices/poiema_fan.yaml
Zankio: Adopt translation_key: timer and use the original dps_val as the value.

custom_components/tuya_local/devices/poiema_fan.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - id: 3
+        type: integer
+        name: speed
+        range:
+          min: 1
+          max: 12
+
+  - entity: select
+    name: Oscillation angle
+    icon: mdi:arrow-oscillating
```
make-all: icon needs to be enclosed in quotes, due to the embedded `:`

custom_components/tuya_local/devices/poiema_fan.yaml
```diff
@@ -0,0 +1,108 @@
...
+  - entity: sensor
+    class: temperature
+    dps:
+      - id: 21
+        type: integer
+        name: sensor
+        unit: "°C"
+  - entity: select
+    translation_key: timer
+    icon: "mdi:timer-outline"
```
make-all: Icon not needed here

PR #5324, opened 2026-06-14
Honor target_range in number min/max; fix Frizzlife LP365P volume scaling

### LINKED ISSUE #5323: Number entities ignore target_range for min/max (wrong slider bounds; values un-writable)
### The problem

`DpsConfig.range()` (in `helpers/device_config.py`) applies `scale` but not
`target_range`. A `number` entity derives `native_min_value` /
`native_max_value` from `range()`, so any number whose displayed value is
produced by a `target_range` remap gets its **min/max reported in raw dps
units instead of display units**.

Two consequences:

1. **Wrong slider bounds.** A number mapping raw `0–9` → `0.01–0.1` via
   `target_range` shows a slider of `0–9` instead of `0.01–0.1`.
2. **Values become un-writable when the display range is wider than the raw
   range.** If raw `0–64` maps to `20–1300`, `native_max_value` is `64`, so
   core rejects any set > 64 before the (correct) `target_range` write
   conversion runs.

### Minimal reproduction

```yaml
- id: 119
  type: integer
  name: value
  range: { min: 0, max: 9 }
  mapping:
    - target_range: { min: 0.01, max: 0.1 }
```
Expected slider: `0.01–0.1`. Actual: `0–9`.

### Real-world case

`frizzlife_lp365p_watermonitorandshutdoff` — the **Allowed flow volume**
setting is offset-encoded by the firmware: `gallons = raw*20 + 20`
(raw `0–64` ⇄ `20–1300` gal). Verified by reading raw dps directly while
changing the value in the Smart Life app:

| App (gal) | dp 115 raw |
|---|---|
| 20  | 0 |
| 160 | 7 |
| 200 | 9 |
| 1300 | 64 |

The shipped profile uses `scale: 0.01` with a `{dps_val: 0, value: 50}`
special case, so HA displays ~900 for a real 200 gal and cannot write the
value. The app writes it fine, so the firmware is correct — the integration
just can't represent an offset with a wider-than-raw display range in a
writable number.

### Fix

Make `range()` report the `target_range` (display units) when configured.
`_find_map_for_value()` already returns the `target_range` default mapping
for out-of-raw-range inputs, and the write conversion reads the raw `range`
from config directly, so no other change is needed for direct mappings.

PR attached (also corrects the Frizzlife volume mapping to a `target_range`).

nalditopr:
Fixes #5323

### What

1. **`DpsConfig.range()` now honors `target_range`.** A `number`'s
   `native_min_value`/`native_max_value` come from `range()`, which applied
   `scale` but ignored `target_range`, so numbers with a `target_range` remap
   exposed their bounds in raw dps units. That made sliders show raw units
   (e.g. a 0.01–0.1 min-flow showing 0–9) and rejected values outside the raw
   range before the `target_range` conversion could run. `range()` now returns
   the `target_range` bounds when configured (guarded by `scaled`; the write
   path reads the raw range from config directly, so it is unaffected).

2. **Frizzlife LP365P "Allowed flow volume" fixed (scale + writability).**
   The firmware offset-encodes this value: `gallons = raw*20 + 20`
   (raw `0–64` ⇄ `20–1300` gal). The old mapping used `scale: 0.01` plus a
   `{dps_val: 0, value: 50}` special case, which mis-read (~900 for a real
   200 gal). It also routed the value through `value_redirect`
   (mode4/5/6 → modeN_val), which does **not** apply the leaf `target_range`
   on write — so sets never reached the device. Replaced with a direct
   `target_range` mapping on the value DP.

### Verified on a live LP365P (protocol 3.5)

- Min-flow slider bounds correct (0.01–0.1). ✅
- Allowed-flow-volume **reads** correctly, matching the Smart Life app
  (raw 0/7/9/64 ⇄ 20/160/200/1300 gal). ✅
- Allowed-flow-volume **writes** correctly from HA — set 200 gal → dp115
  raw 9, confirmed by direct device read. ✅ (Through `value_redirect` the
  same set left dp115 unchanged; direct binding fixes it.)

### Note on multi-mode

The direct mapping tracks the Mode 4 register (dp115). The previous
`value_redirect` form auto-followed the active advanced mode but was
non-writable for this offset encoding. If preserving auto-follow is
preferred, the `value_redirect` write path would need to apply the
redirected dp's `target_range` (it currently honors `scale` but not
`target_range`); happy to take it that direction instead.

### DIFF
--- a/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
+++ b/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
@@ -309,43 +309,10 @@ entities:
           - dps_val: us_units
             value: gal
           - value: L
-      - id: 111
-        type: string
-        name: value
-        unit: gal
-        range:
-          min: 20
-          max: 1300
-        mapping:
-          - dps_val: mode4
-            value_redirect: mode4_val
-          - dps_val: mode5
-            value_redirect: mode5_val
-          - dps_val: mode6
-            value_redirect: mode6_val
       - id: 115
         type: integer
-        name: mode4_val
-        range:
-          min: 0
-          max: 64
-        mapping:
-          - target_range:
-              min: 20
-              max: 1300
-      - id: 116
-        type: integer
-        name: mode5_val
-        range:
-          min: 0
-          max: 64
-        mapping:
-          - target_range:
-              min: 20
-              max: 1300
-      - id: 117
-        type: integer
-        name: mode6_val
+        name: value
+        unit: gal
         range:
           min: 0
           max: 64
--- a/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
+++ b/custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
@@ -312,49 +312,47 @@ entities:
       - id: 111
         type: string
         name: value
+        unit: gal
         range:
-          min: 0
-          max: 5000
+          min: 20
+          max: 1300
         mapping:
           - dps_val: mode4
-            step: 100
             value_redirect: mode4_val
           - dps_val: mode5
-            step: 100
             value_redirect: mode5_val
           - dps_val: mode6
-            step: 100
             value_redirect: mode6_val
       - id: 115
         type: integer
         name: mode4_val
         range:
           min: 0
-          max: 50
+          max: 64
         mapping:
-          - dps_val: 0
-            value: 50
-          - scale: 0.01
+          - target_range:
+              min: 20
+              max: 1300
       - id: 116
         type: integer
         name: mode5_val
         range:
           min: 0
-          max: 50
+          max: 64
         mapping:
-          - dps_val: 0
-            value: 50
-          - scale: 0.01
+          - target_range:
+              min: 20
+              max: 1300
       - id: 117
         type: integer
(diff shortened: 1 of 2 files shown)


nalditopr:
Root cause of the volume write failure isolated on a live LP365P: it is the `value_redirect` write path, not a read-only DP. With the value modeled via `value_redirect` (mode4/5/6 -> modeN_val) a set does **not** reach the device. Re-modeling the same DP (115) as the entity's direct value with the identical `target_range` makes the write land correctly (set 200 gal -> dp115 raw 9, verified by direct device read). So the leaf `target_range` is honored on direct reads/writes but lost when reached through `value_redirect` on write. Options: (a) fix the value_redirect write path to apply the redirected dp's target_range, or (b) model these per-mode values directly. Happy to adjust this PR whichever way you prefer.

custom_components/tuya_local/devices/frizzlife_lp365p_watermonitorandshutdoff.yaml
```diff
@@ -309,52 +309,17 @@ entities:
           - dps_val: us_units
             value: gal
           - value: L
-      - id: 111
```
make-all, line 312: The problem with simplifying this is it will now only work in one mode.

Did you try putting the target_range in each mapping here where it will be picked up by the logic to report the min and max to HA, as well as on the target dp of the value_redirect to use when calculating actual changes?

make-all (changes requested):

PR #5333, opened 2026-06-15
Add GoDrone 01B smart cat litter box

adyachenko:
Adds support for the GoDrone 01B smart cat litter box (product id qyvypzich5aywr9q).

I captured the DPs locally from my own device over the LAN (protocol 3.5). It uses the same Tuya dp layout as the existing Sailesi litter box, so the mapping is based on that config: cat weight, time in box, fault reporting, clean/empty/deodorize/calibrate buttons, presence event, child lock, induction clean delay/interval, capacity and litter calibration, detection sensitivity, etc.

DP 131/132/133 are present on this device but not on the Sailesi one. They did not respond to any app setting (scheduled cleaning, sterilization and do-not-disturb turned out to be cloud timers, not local dps), so their purpose is unknown and they are exposed hidden for now.

### DIFF
--- a/custom_components/tuya_local/devices/smart_01b_litterbox.yaml
+++ b/custom_components/tuya_local/devices/smart_01b_litterbox.yaml
@@ -1,0 +1,247 @@
+name: Litter box
+products:
+  - id: qyvypzich5aywr9q
+    manufacturer: GoDrone
+    model: 01B
+entities:
+  - entity: sensor
+    class: weight
+    dps:
+      - id: 6
+        type: integer
+        optional: true
+        name: sensor
+        unit: g
+        class: measurement
+      - id: 129
+        type: integer
+        optional: true
+        name: lbs
+  - entity: sensor
+    name: Daily time spent
+    class: duration
+    category: diagnostic
+    dps:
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/smart_01b_litterbox.yaml
```diff
@@ -0,0 +1,247 @@
...
+    name: DP 131
+    category: diagnostic
+    icon: "mdi:help-circle"
+    hidden: true
+    dps:
+      - id: 131
+        type: integer
+        optional: true
+        name: sensor
+  - entity: binary_sensor
```
make-all: The unknown dps can just be attributes on another entity, there is no good reason to make them entities in their own right. Convention is to name them `unknown_nnn` where `nnn` is the dp id.

make-all (changes requested):
Filename should be godrone_01b_litterbox.yaml. Words like "smart" are not useful for differentiating devices within this integration, as all supported devices are "smart"

make-all:
I don't think this needs a new config actually. With minor changes the Sailesi config can be used instead

PR #5334, opened 2026-06-15
feat(devices): add Petzengo PF28-C camera pet feeder

adyachenko:
Adds support for the Petzengo PF28-C, a Wi-Fi camera pet feeder (product id lnhpwq8ydzwbxc9s).

I captured the DPs locally from my own device over the LAN (protocol 3.4) and confirmed them by operating the feeder while monitoring DP changes:

- 104/105/106/108/134 – camera controls (watermark, privacy, motion sensitivity, night vision, motion detection)
- 203 – manual feed (portions, write-only)
- 204 – feed state (standby/feeding/done)
- 210 – battery
- 219 – fault/problem (bitfield)
- 130 – feeding schedule, 131 – feed record, 236 – factory reset

DPs 210 and 219 are mapped the same way as the existing Frienhund ACF180W-A config, which uses the same Tuya template.

### DIFF
--- a/custom_components/tuya_local/devices/pf28c_camera_petfeeder.yaml
+++ b/custom_components/tuya_local/devices/pf28c_camera_petfeeder.yaml
@@ -1,0 +1,179 @@
+name: Pet feeder
+products:
+  - id: lnhpwq8ydzwbxc9s
+    manufacturer: Petzengo
+    model: PF28-C
+entities:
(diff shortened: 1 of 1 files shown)


make-all:
Unfortunately this product appears to be exclusive to the Russian market. I am not currently supporting such devices, until Putin stops his senseless war against Ukraine.

PR #5344, opened 2026-06-16
Eta noble smart humidifier (device, icons)

jiriks74:
Adds a configuration and icons for the ETA Noble Smart humidifier.


custom_components/tuya_local/icons.json
```diff
@@ -367,7 +367,8 @@
                         "state": {
                             "laundry": "mdi:tshirt-crew",
                             "purify": "mdi:air-filter",
-                            "continuous": "mdi:fan"
+                            "continuous": "mdi:fan",
+                            "normal": "mdi:water-percent"
```
make-all: Not needed, it is already defined at
https://github.com/home-assistant/core/blob/dev/homeassistant/components/humidifier/icons.json#L48

custom_components/tuya_local/icons.json
```diff
@@ -411,7 +412,10 @@
                 "default": "mdi:lightbulb"
             },
             "ambient_light": {
-                "default": "mdi:lamps"
+                "default": "mdi:led-on",
```
make-all: led-on is not a good fit for ambient light.
I don't want this change.

custom_components/tuya_local/icons.json
```diff
@@ -481,6 +485,9 @@
             },
             "snooze_duration": {
                 "default": "mdi:alarm-snooze"
+            },
```
make-all: This is normally covered by a fan entity, so not needed as a number.

custom_components/tuya_local/devices/eta_noble_humidifier.yaml
```diff
@@ -0,0 +1,186 @@
...
+        name: humidity
+        range:
+          min: 30
+          max: 75
+        mapping:
+          - step: 5
+      - id: 14
+        type: integer
+        name: current_humidity
+  - entity: number
```
make-all: Other similar devices all use a fan entity to define the spray level, it is better to be consistent and follow that.

jiriks74: Ok. I tried to be consistent as I saw the `number` entity used in other humidifiers too and it directly maps to the device values.

https://github.com/make-all/tuya-local/blob/8eedf7ff4021122d5f8cd5873488d8c12463aa80/custom_components/tuya_local/devices/klarta_humea_humidifier.yaml#L68-L87

Will change.

PR #5350, opened 2026-06-17
Add AtmosC A Series air purifier

ralphxtrem:
Add support for AtmosC A Series air purifier.

Product ID: acgosool
7 entities: fan (8 speeds), mode select (Manual/Auto/Sleep), PM2.5 sensor, anion switch, UV switch, child lock, filter sensor.
DP mapping confirmed physically on my own device.


atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
+name: AtmosC A Series Air Purifier
+products:
+  - id: acgosool
```
make-all: WiFi product ids are always 16 characters in length. I think you have given a different id here.

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
...
+    name: AtmosC A Series
+entities:
+  - entity: fan
+    icon: "mdi:air-purifier"
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+      - id: 4
+        name: preset_mode
```
make-all: Use speed rather than preset_mode here.

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
...
+          - dps_val: "5"
+            value: "5"
+          - dps_val: "6"
+            value: "6"
+          - dps_val: "7"
+            value: "7"
+          - dps_val: "8"
+            value: "8"
+  - entity: select
+    translation_key: mode
```
make-all: This can be preset_mode of the fan entity. Make the values all lowercase to match the translation_key.

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
...
+        type: string
+        mapping:
+          - dps_val: "Manual"
+            value: "Manual"
+          - dps_val: "Auto"
+            value: "Auto"
+          - dps_val: "Sleep"
+            value: "Sleep"
+  - entity: sensor
+    name: PM2.5
```
make-all: name not required here, it hides the translation for pm25 class.

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
...
+    class: pm25
+    dps:
+      - id: 2
+        name: sensor
+        class: measurement
+        type: integer
+        unit: ugm3
+  - entity: switch
+    name: Anion
```
make-all: Use `translation_key: ionizer` instead of the explicit name and icon here.

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
...
+    category: config
+    icon: "mdi:atom"
+    dps:
+      - id: 6
+        name: switch
+        type: boolean
+  - entity: switch
+    name: UV
```
make-all: Use `translation_key: uv_sterilization` here.

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
...
+    category: config
+    icon: "mdi:lightbulb-on"
+    dps:
+      - id: 9
+        name: switch
+        type: boolean
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    icon: "mdi:hand-back-right-off"
```
make-all: Icon not needed here, as the translation_key already defines it.

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
...
+    dps:
+      - id: 7
+        name: lock
+        type: boolean
+  - entity: sensor
+    name: Filter
```
make-all: Use `translation_key: filter_life`

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
+name: AtmosC A Series Air Purifier
```
make-all: Top level name should be generic. Branding belongs under products.

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
...
+products:
+  - id: acgosool
+    name: AtmosC A Series
```
make-all: This should be split into manufacturer and model

atmosc_a_series.yaml
```diff
@@ -0,0 +1,87 @@
...
+entities:
+  - entity: fan
+    icon: "mdi:air-purifier"
```
make-all: Use `translation_key: air_purifier` instead of this icon.

make-all (changes requested):

ralphxtrem:
Thanks for the review! All comments addressed: real 16-char product_id from the Tuya IoT platform, translation keys throughout, mode merged into the fan's preset_mode with lowercase values, manufacturer/model split, and file moved to the devices directory.

PR #5351, opened 2026-06-17
Add ACIQ EX-150 pool heat pump

jdholmes1978:
## Summary
- Add a device profile for the ACIQ EX-150 pool heat pump
- Map heat/cool/auto HVAC modes, fan modes, Fahrenheit temperature setpoints, current temperature, and diagnostic sensors

## Validation
- .venv/bin/python -m pytest tests/test_device_config.py
- .venv/bin/yamllint custom_components/tuya_local/devices/aciq_ex150_pool_heatpump.yaml
- .venv/bin/ruff check .
- .venv/bin/ruff check --select I .
- .venv/bin/ruff format --check .
- .venv/bin/yamllint custom_components/tuya_local/devices

### DIFF
--- a/custom_components/tuya_local/devices/aciq_ex150_pool_heatpump.yaml
+++ b/custom_components/tuya_local/devices/aciq_ex150_pool_heatpump.yaml
@@ -1,0 +1,140 @@
+name: Pool heat pump
+products:
+  - id: yrjx6kq77vapja8a
+    manufacturer: ACIQ
+    model: EX-150
+entities:
+  - entity: climate
+    translation_key: pool_heatpump
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: work_mode
+            conditions:
+              - dps_val: smart
+                value: auto
+              - dps_val: warm
+                value: heat
+              - dps_val: cool
+                value: cool
+      - id: 2
+        type: integer
+        name: cooling_setpoint
+        unit: F
+        hidden: true
+        range:
+          min: 45
+          max: 95
+      - id: 3
+        type: integer
+        name: current_temperature
+        unit: F
+      - id: 4
+        type: string
+        name: work_mode
+        hidden: true
+      - id: 5
+        type: string
+        name: fan_mode
+        mapping:
+          - dps_val: silence
+            value: low
+          - dps_val: smart
+            value: auto
+          - dps_val: booster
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/aciq_ex150_pool_heatpump.yaml
```diff
@@ -0,0 +1,140 @@
...
+              - dps_val: cool
+                value_redirect: cooling_setpoint
+                range:
+                  min: 45
+                  max: 95
+      - id: 108
+        type: integer
+        name: min_temperature
+        unit: F
+        readonly: true
```
make-all, line 78: `readonly:true` is not needed on inherently read-only attributes.
`unit` is also not needed on all the dps, as climate only supports one unit, and that is defined by the temperature_unit dp.

custom_components/tuya_local/devices/aciq_ex150_pool_heatpump.yaml
```diff
@@ -0,0 +1,140 @@
...
+              - dps_val: warm
+                value: heat
+              - dps_val: cool
+                value: cool
+      - id: 2
+        type: integer
+        name: cooling_setpoint
+        unit: F
+        hidden: true
+        range:
```
make-all, line 30: A different range will be needed when temperature_unit is `c`

custom_components/tuya_local/devices/aciq_ex150_pool_heatpump.yaml
```diff
@@ -0,0 +1,140 @@
...
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 21
+        type: integer
+        name: fault_code
+
+  - entity: sensor
```
make-all, line 134: The meaning of this is unclear, I don't think it needs to be an entity, probably better as an attribute on the the problem sensor

custom_components/tuya_local/devices/aciq_ex150_pool_heatpump.yaml
```diff
@@ -0,0 +1,140 @@
...
+        type: string
+        name: option
+        mapping:
+          - dps_val: c
+            value: celsius
+          - dps_val: f
+            value: fahrenheit
+
+  - entity: sensor
+    name: Display water temperature
```
make-all, line 99: This sensor is probably better without a name, and without the diagnostic category

make-all (changes requested):

PR #5358, opened 2026-06-18
Add ezAIoT Lite Socket Thermostat (product id: lzfux5zela3cbpnd)

brickronny:
New device config for the ezAIoT Socket Thermostat Lite (product id: lzfux5zela3cbpnd).

Maps DPs for climate control (heat/cool/auto/cycle timer/off via plug mode), relay switch, countdown timer dropdown, display brightness/timeout, high/low temperature alarms, cooling compressor delay, frost protection, temperature hysteresis, and calibration.

Adds translations to all 23 language files; German fully translated, all others use English fallback.

### DIFF
--- a/custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
+++ b/custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
@@ -1,0 +1,338 @@
+name: Socket thermostat
+products:
+  - id: lzfux5zela3cbpnd
+    manufacturer: ezAIoT
+    model: Lite
+entities:
+  - entity: climate
+    translation_key: thermostat
+    dps:
+      - id: 110
+        type: string
+        name: hvac_mode
+        mapping:
+          - dps_val: "plug"
+            value: "off"
+          - dps_val: "heat"
+            value: "heat"
+          - dps_val: "cool"
+            value: "cool"
+          - dps_val: "prg"
+            value: "auto"
+          - dps_val: "cycle"
+            value: "fan_only"
+      - id: 126
+        type: integer
+        name: current_temperature
+        mapping:
+          - scale: 10
+      - id: 135
+        type: integer
+        name: temperature
+        unit: C
+        mapping:
+          - scale: 10
+        range:
+          min: -500
+          max: 3000
+      - id: 127
+        type: integer
+        name: max_temperature
+        optional: true
+        mapping:
+          - scale: 10
+      - id: 130
+        type: integer
+        name: min_temperature
+        optional: true
+        mapping:
+          - scale: 10
+      - id: 114
+        type: integer
+        name: hvac_action
+        mapping:
+          - dps_val: 0
+            value: idle
+          - dps_val: 1
+            constraint: hvac_mode
+            conditions:
+              - dps_val: heat
+                value: heating
+              - dps_val: cool
+                value: cooling
+              - dps_val: auto
+                value: heating
+              - dps_val: fan_only
+                value: heating
+              - dps_val: "off"
+                value: idle
+
+  # Direct relay switch (used in plug mode)
+  - entity: switch
+    translation_key: manual_socket
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+
+  # Countdown timer as dropdown (minutes)
+  - entity: select
+    translation_key: manual_socket_countdown
+    icon: "mdi:timer"
+    dps:
+      - id: 9
+        type: integer
+        name: option
+        optional: true
+        mapping:
+          - dps_val: 0
+            value: "0"
+          - dps_val: 60
+            value: "1"
+          - dps_val: 120
+            value: "2"
+          - dps_val: 180
+            value: "3"
+          - dps_val: 240
+            value: "4"
+          - dps_val: 300
+            value: "5"
+          - dps_val: 600
+            value: "10"
+          - dps_val: 900
+            value: "15"
+          - dps_val: 1800
+            value: "30"
+          - dps_val: 2700
+            value: "45"
+          - dps_val: 3600
+            value: "60"
+
+  # Display on button
+  - entity: button
+    translation_key: display_on
+    dps:
+      - id: 118
+        name: button
+        optional: true
+        type: boolean
+
+  # Operating mode select
+  - entity: select
+    translation_key: work_mode
+    dps:
+      - id: 110
+        type: string
+        name: option
+        mapping:
+          - dps_val: "plug"
+            value: "plug"
+          - dps_val: "heat"
+            value: "heat"
+          - dps_val: "cool"
+            value: "cool"
+          - dps_val: "prg"
+            value: "prg"
+          - dps_val: "cycle"
+            value: "cycle"
+
+  # Diagnostics: relay state
+  - entity: binary_sensor
+    class: power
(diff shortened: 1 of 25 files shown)


custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+            constraint: hvac_mode
+            conditions:
+              - dps_val: heat
+                value: heating
+              - dps_val: cool
+                value: cooling
+              - dps_val: auto
+                value: heating
+              - dps_val: fan_only
+                value: heating
```
make-all, line 66: Why heating and not fan?

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+              - dps_val: "off"
+                value: idle
+
+  # Direct relay switch (used in plug mode)
+  - entity: switch
+    translation_key: manual_socket
```
make-all, line 72: I think this is better without a translation_key, and with an available attribute based on dp 110.

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+
+  # Countdown timer as dropdown (minutes)
+  - entity: select
+    translation_key: manual_socket_countdown
```
make-all, line 80: Reusing the existing `timer` translation key would be better than introducing this very specific new duplicate key.

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+    translation_key: display_on
+    dps:
+      - id: 118
+        name: button
+        optional: true
+        type: boolean
+
+  # Operating mode select
+  - entity: select
+    translation_key: work_mode
```
make-all, line 122: This is just duplicating the hvac_mode that already exists above.  In any case, I do not like this translation key, the existing `heat_pump_mode` could be used instead with fewer translation changes.

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+          - dps_val: "cool"
+            value: "cool"
+          - dps_val: "prg"
+            value: "prg"
+          - dps_val: "cycle"
+            value: "cycle"
+
+  # Diagnostics: relay state
+  - entity: binary_sensor
+    class: power
```
make-all, line 141: This is already exposed as hvac_action on the climate entity.

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 26
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
```
make-all, line 164: It is better to add a `fault_code` attribute here with the original value

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+    category: config
+    dps:
+      - id: 40
+        name: lock
+        type: boolean
+
+  # Frost protection
+  - entity: switch
+    category: config
+    translation_key: protection_frost
```
make-all, line 178: The existing `anti_frost` translation key should be used instead of introducing a new one unnecessarily.

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+    dps:
+      - id: 113
+        name: switch
+        type: boolean
+
+  # Cooling delay enable
+  - entity: switch
+    category: config
+    translation_key: protection_cooling_delay
```
make-all, line 187: This is pretty obscure, and I'm not convinced a translation key is worth the effort

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+    icon: "mdi:pump"
+    dps:
+      - id: 133
+        name: switch
+        optional: true
+        type: boolean
+
+  # Cooling delay time
+  - entity: number
+    translation_key: protection_cooling_delay_time
```
make-all, line 197: Likewise

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+        optional: true
+        unit: min
+        range:
+          min: 0
+          max: 10
+
+  # Upper temperature alarm enable + threshold
+  - entity: switch
+    category: config
+    translation_key: high_temp_alarm_limit
```
make-all, line 213: This seems more suited to the entity below.
In general, words should be spelled out in full in translation keys, as using abbreviations can make it difficult for translators.

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+        unit: C
+        mapping:
+          - scale: 10
+        range:
+          min: -500
+          max: 3000
+
+  # Temperature hysteresis
+  - entity: number
+    translation_key: temp_differential
```
make-all, line 266: Use `name: Temperature hysteresis` to be consistent with other devices.

custom_components/tuya_local/devices/ezaiot_lite_socket_thermostat.yaml
```diff
@@ -0,0 +1,338 @@
...
+        optional: true
+        mapping:
+          - scale: 10
+        unit: C
+        range:
+          min: -150
+          max: 150
+
+  # Display brightness
+  - entity: select
```
make-all, line 301: Use a light entity, with `translation_key: display` , a `brightness` dp and map the values into the range 0-255. (85, 170, 255)

make-all (changes requested):
It is better to avoid introducing translation keys at the same time as a new config.
Merging of this will be delayed until time can be found to fill the gaps in the translation, and based on the review there are now a number of translation changes that need to be undone.

custom_components/tuya_local/translations/bg.json
```diff
@@ -236,6 +236,9 @@
             },
             "radio_stop": {
                 "name": "Радио стоп"
+            },
+            "display_on": {
+                "name": "Display on"
```
make-all, line 241: This is not Bulgarian.

Either drop the "translations" from this PR, or do them properly.

Issue #5353, opened 2026-06-18
New device support: RGBIC+CCTIC Wi-Fi LED strip uses DP61 raw payload for white/CCT


Puoskar1:
## Device information

**Device name printed on Finnish user guide:**  
Älykäs LED-valonauha RGBIC+CCTIC, 5M, Wifi

**English description:**  
Smart LED strip RGBIC+CCTIC, 5 m, Wi-Fi

**Device class:**  
RGBIC / addressable RGB + CCT Wi-Fi LED strip

## Summary

This device was initially detected only as `enbrighten_curtainlights`. That gave partial RGB control, but no working native white / colour-temperature control.

After testing custom Tuya Local profiles, the working local control paths appear to be:

```text
DP20 = power
DP21 = work mode: white, colour, scene, music
DP24 = RGB/HSV colour data only
DP61 = working white/CCT control path using raw/base64 payloads
```

The Tuya IoT data model lists DP22 as `bright_value` and DP23 as `temp_value`, but those did not appear to work locally on this controller.

## Environment

- Home Assistant: 2026.2.0
- Tuya Local: 2026.5.0
- tinytuya: 1.18.0
- Protocol version: 3.5
- Installation type: Home Assistant Core
- Device connection: Wi-Fi
- Device type: RGBIC+CCTIC LED strip

## Tuya IoT Query Things Data Model

Relevant datapoints:

```text
20 switch_led bool rw
21 work_mode enum rw: white, colour, scene, music
22 bright_value value rw: min 10, max 1000
23 temp_value value rw: min 0, max 1000
24 colour_data string rw:
   HHHH SSSS VVVV
   H 0-360
   S 0-1000
   V 0-1000
26 countdown value rw: min 0, max 86400, unit s
27 music_data string wr
28 control_data string wr
47 light_pixel value ro: min 1, max 1024
51 dreamlight_scene_mode raw rw
52 dreamlightmic_music_data raw rw
53 lightpixel_number_set value rw: min 1, max 1000
61 paint_colour_data raw rw
```

DP61 description from the Tuya data model says it is used for Dreamlight/addressable-strip dimming/colour control. It mentions white mode and HSV/combined mode payloads.

### RGB / colour mode

RGB colour works through DP24.

Example observed state:

```text
21 = colour
24 = 0030001403e8
61 = AAEAFAAAMAJk
```

DP24 decodes as HSV-like data:

```text
0030 0014 03e8
H = 48
S = 20
V = 1000
```

[...]

PR #5361, opened 2026-06-19
Add mask_base support and RGBIC CCTIC DP61 LED strip profile

Puoskar1:
## Summary

This PR adds support for a Tuya RGBIC+CCTIC Wi-Fi LED strip that uses packed DP61 raw/base64 data for white brightness and colour temperature.

It also adds a small generic `mask_base` option for masked packed DP writes. This is needed when a masked DP write has to preserve unmasked bytes, but the current packed DP value is either missing or incompatible with the mask shape.

Fixes / follows up: #5353

## What is supported

* On/off via DP20
* HS colour via DP24
* White brightness via DP61
* Colour temperature via DP61
* Switching between colour mode and white/CCT mode
* Safe masked writes when DP61 is missing at startup
* Safe masked writes when DP61 currently contains the RGB-mode 9-byte payload instead of the white-mode 7-byte payload

## What is intentionally not included

This does not attempt full Tuya app parity. It does not implement scene editing, per-segment painting, music/mic configuration, or the full Dreamlight protocol.

## Device details

Device name from Finnish user guide:

`Älykäs LED-valonauha RGBIC+CCTIC, 5M, Wifi`

Observed relevant DPs:

```text
20 switch_led
21 work_mode: white, colour, scene, music
24 colour_data / HSV colour
61 paint_colour_data / packed raw white+CCT payload
```

Observed white-mode DP61 shape:

```text
00 00 00 14 00 BB TT
```

Where:

* `BB` is white brightness, usable range 3–100 on this device
* `TT` is colour temperature, 0–100 mapped to 2700–6500 K

## Implementation notes

`mask_base` is only used in the masked write path. It is not used for reads, so Home Assistant does not report fabricated brightness or colour-temperature state before the device has actually reported the DP.

The fallback order for masked writes remains:

1. pending value for the same DP
2. current device value, if compatible with the mask width
3. `mask_base`, if configured

## Tested

Live tested on my own device with Home Assistant and Tuya Local:

* Turn on/off
* Set HS colour
* Set white brightness
* Set colour temperature
* Switch RGB/HS → white/CCT
* Low white brightness floor fixed at 3%

Local validation:

```text
uv run pytest
uv run pytest tests/test_device_config.py
uv run ruff check .
uv run ruff check --select I .
uv run ruff format --check .
uv run yamllint custom_components/tuya_local/devices
```

### DIFF
--- a/custom_components/tuya_local/devices/device_config_schema.json
+++ b/custom_components/tuya_local/devices/device_config_schema.json
@@ -218,6 +218,10 @@
                                     "pattern": "^[0-9a-fA-F]+$",
                                     "description": "A hexadecimal mask to extract an integer from a base64 or hex DP value."
                                 },
+                                "mask_base": {
+                                    "type": ["integer", "string"],
+                                    "description": "Raw DP value to use as the base only when writing a masked value and the current DP value is unavailable."
+                                },
                                 "mask_signed": {
                                     "type": "boolean",
                                     "description": "true if the masked integer is signed. Default is false, which should not be specified explicitly to avoid clutter."
--- a/custom_components/tuya_local/devices/tuya_rgbic_cctic_led_strip_dp61.yaml
+++ b/custom_components/tuya_local/devices/tuya_rgbic_cctic_led_strip_dp61.yaml
@@ -43,7 +43,7 @@ entities:
         mask_base: "AAAAFABQMg=="
         mask: "0000000000FF00"
         range:
-          min: 0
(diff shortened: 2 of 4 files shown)


make-all:
Guessing state is not something I want to start doing, as different users will have different opinions on how settings should default, which will only lead to bug reports.

This integration is focused on making the data provided by devices available, not on trying to workaround problems with the devices themselves.

PR #5366, opened 2026-06-19
Add Ledvance SMART+ WiFi Wide Brick RGB

DawidSu:
## Summary
- Add device config for Ledvance SMART+ WiFi Wide Brick RGB (RGBCW lightbulb)
- Based on `rgbcw_lightbulb_xld_cl002` with `force: true` on DP 20 for reliable status polling and do_not_disturb support (DP 34)
- Update DEVICES.md with new entry

## Device details
- **Manufacturer:** Ledvance
- **Model:** SMART+ WiFi Wide Brick RGB
- **Protocol version:** 3.5
- **DPs used:** 20 (switch), 21 (color_mode), 22 (brightness), 23 (color_temp), 24 (rgbhsv), 25 (scene_data), 26 (timer), 27 (music_data), 30 (rhythm), 31 (sleep), 32 (wakeup), 34 (do_not_disturb), 51 (scene)

## Test plan
- [x] Tested with 4 physical devices
- [x] Status polling confirmed working with `force: true` on DP 20
- [x] On/off, brightness, color temp, RGB all functional


custom_components/tuya_local/devices/ledvance_smart_plus_wifi_wide_brick_rgb.yaml
```diff
@@ -1,6 +1,6 @@
 name: RGBCW lightbulb
 products:
-  - id: bf0b923764d77ccbfcyixv
+  - id: bf0b923764d77ccb
```
make-all, line 3: You can't just shorten the id to pass the test. The product id is a completely different id than the device id.

make-all:
Merged with rgbcw_lightbulb_xld_cl002.yaml, which already had most of the differences here.

Force: true is known to cause problems in contexts other than read-only sensors on smartplugs, usually the real solution is to use a different protocol version.

PR #5375, opened 2026-06-20
Add TH09B temp/humidity sensor

decompil3d:
Hi folks - this is a new device config for the TH09B temperature and humidity sensor. It seems to be an amalgam of the TH08 and TH05 in terms of DPs. I've tested the config locally and it works with the device.

### DIFF
--- a/custom_components/tuya_local/devices/th09b_temp_humidity_sensor.yaml
+++ b/custom_components/tuya_local/devices/th09b_temp_humidity_sensor.yaml
@@ -1,0 +1,80 @@
+name: Temperature and humidity sensor
+products:
+  - id: ykdqgxukxt7rjyft
+    model: TH09B
+entities:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/th09b_temp_humidity_sensor.yaml
```diff
@@ -0,0 +1,80 @@
...
+    class: humidity
+    dps:
+      - id: 2
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+  - entity: sensor
+    name: Power state
+    class: enum
```
make-all, line 26: Is this for battery level?

make-all:
Appears to be identical to TH08 (additional dps for calibration also show in the data model for a variant that was reported before the original PR that created this config)

Issue #3824, opened 2025-10-08
Request support for VANCOO TGM50-WIFI-EP


aleks31031992:
### Log message

```text
Регистратор: custom_components.tuya_local.config_flow
Источник: custom_components/tuya_local/config_flow.py:433
интеграция: Tuya Local (документация, проблемы)
Зафиксировано первый раз: 18:20:54 (7 случаев)
Зафиксировано последний раз: 19:29:47

Device matches beok_tgr81_thermostat_c with quality of 101%. DPS: {"updated_at": 1759937182.8683605, "1": true, "2": "manual", "16": 35, "24": 190, "36": "open", "45": 0}
Device matches woods_milan_airconditioner with quality of 32%. DPS: {"updated_at": 1759937270.2152152, "1": true, "2": 280, "3": 271, "4": "home", "5": "1", "9": false, "11": 0, "15": 60, "19": -10, "101": 5, "102": 60, "103": false, "104": false, "106": "1", "107": 12, "108": false, "109": true, "110": "1", "111": false}
Device matches woods_milan_airconditioner with quality of 32%. DPS: {"updated_at": 1759937479.6931794, "1": true, "2": 280, "3": 272, "4": "home", "5": "1", "9": false, "11": 0, "15": 60, "19": -10, "101": 5, "102": 60, "103": false, "104": false, "106": "1", "107": 12, "108": false, "109": true, "110": "1", "111": false}
Device matches woods_milan_airconditioner with quality of 32%. DPS: {"updated_at": 1759937718.1527946, "1": true, "2": 265, "3": 276, "4": "home", "5": "0", "9": false, "11": 0, "15": 60, "19": -10, "101": 5, "102": 60, "103": false, "104": false, "106": "1", "107": 12, "108": false, "109": true, "110": "1", "111": false}
Device matches woods_milan_airconditioner with quality of 32%. DPS: {"updated_at": 1759940985.4300272, "1": true, "2": 300, "3": 293, "4": "home", "5": "1", "9": false, "11": 0, "15": 60, "19": -10, "101": 5, "102": 60, "103": false, "104": false, "106": "1", "107": 12, "108": false, "109": true, "110": "1", "111": false}
```

### Product ID

bf10c60ff83bd2a388zf3n

### Product Name

Thermostat VANCOO TGM50-WIFI-EP

### DPS information

```text
{
  "result": {
[...]

PR #5381, opened 2026-06-20
Fix unload KeyError for failed device setup

geekypilot:
## Summary

This fixes an unload crash that can happen when a Tuya device is offline during Home Assistant startup. In that state Home Assistant may leave the config entry in `setup_retry`, while `tuya-local` never finishes creating the usual cached device data under `hass.data`.

Later, when Home Assistant reloads, updates, or shuts down that entry, the integration still runs its unload path. Before this change, that path assumed the cached device entry and all of its internal keys existed, so it could raise a `KeyError` instead of completing the unload.

This PR makes that cleanup path defensive for setup-failed entries:

- `async_unload_entry` now tolerates a missing `hass.data[DOMAIN][device_id]` entry and treats it as already unloaded.
- `async_delete_device` now tolerates missing or partially initialized cached handles.
- Device cleanup still awaits `async_stop()` before removing the cached `device` object, so a stop failure does not silently discard the device reference first.

The intent is to keep normal unload behavior unchanged for devices that completed setup, while preventing failed setup or `setup_retry` entries from crashing later reload/update/shutdown flows.

## Reported traceback

      File "/config/custom_components/tuya_local/__init__.py", line 1001, in async_update_entry
        await async_unload_entry(hass, entry)
      File "/config/custom_components/tuya_local/__init__.py", line 993, in async_unload_entry
        await async_delete_device(hass, config)
      File "/config/custom_components/tuya_local/device.py", line 822, in async_delete_device
        del hass.data[DOMAIN][device_id]["device"]
    KeyError: '<device_id>'

## Device context

- Device: Venta AH550 humidifier, detected as type `venta_ah510_humidifier`.
- Setup state: offline during Home Assistant boot, leaving the entry in `setup_retry`.
- Entry reason: `tuya-local device offline`.
- Related issues: #3824, #3174.

## Scope notes

This PR is intentionally narrow. It addresses unload cleanup for setup-failed or partially initialized entries, but does not try to redesign broader cache recovery behavior. In particular, it does not add new handling for unrelated cases where platform entities were already forwarded and the device cache is later externally corrupted.

## Testing

- `uv run pytest tests/test_config_flow.py::test_async_unload_entry_ignores_missing_device_data -q`
- `uv run pytest tests/test_device.py::test_delete_keeps_device_entry_when_stop_fails -q`
- `uv run pytest tests/test_device.py tests/test_config_flow.py -q`
- `uv run pytest`
- `uv run ruff check .`
- `uv run ruff check --select I .`
- `uv run ruff format --check .`
- `uv run yamllint custom_components/tuya_local/devices`

### DIFF
--- a/custom_components/tuya_local/device.py
+++ b/custom_components/tuya_local/device.py
@@ -813,7 +813,16 @@ def setup_device(hass: HomeAssistant, config: dict):
 async def async_delete_device(hass: HomeAssistant, config: dict):
     device_id = get_device_id(config)
     _LOGGER.info("Deleting device: %s", device_id)
-    await hass.data[DOMAIN][device_id]["device"].async_stop()
-    del hass.data[DOMAIN][device_id]["device"]
-    del hass.data[DOMAIN][device_id]["tuyadevice"]
-    del hass.data[DOMAIN][device_id]["tuyadevicelock"]
+    domain_data = hass.data.get(DOMAIN, {})
+    device_entry = domain_data.get(device_id)
+    if device_entry is None:
+        return
+
+    device = device_entry.get("device")
+    if device is not None:
+        await device.async_stop()
+        device_entry.pop("device", None)
+    device_entry.pop("tuyadevice", None)
+    device_entry.pop("tuyadevicelock", None)
+    if not device_entry:
+        domain_data.pop(device_id, None)
--- a/custom_components/tuya_local/device.py
+++ b/custom_components/tuya_local/device.py
@@ -824,5 +824,8 @@ async def async_delete_device(hass: HomeAssistant, config: dict):
         device_entry.pop("device", None)
     device_entry.pop("tuyadevice", None)
     device_entry.pop("tuyadevicelock", None)
+    # Platform setup may cache entity instances in this bucket by config_id.
+    # Only drop empty buckets here; async_unload_entry removes the whole bucket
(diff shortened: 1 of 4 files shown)


custom_components/tuya_local/device.py
```diff
@@ -813,7 +813,16 @@ def setup_device(hass: HomeAssistant, config: dict):
...
+    if device_entry is None:
+        return
+
+    device = device_entry.get("device")
+    if device is not None:
+        await device.async_stop()
+        device_entry.pop("device", None)
+    device_entry.pop("tuyadevice", None)
+    device_entry.pop("tuyadevicelock", None)
+    if not device_entry:
```
make-all, line 830: Why `not` here?

geekypilot: That condition is intentional, though I can see why it is not obvious locally. `device_entry` can contain more than just the core device handles. Platform setup may also cache entity instances in the same dict by `config_id`. My testing confirms that after removing `device`, `tuyadevice`, and `tuyadevicelock`, the bucket can still contain those entity entries.

So `not device_entry` means “only remove the per-device bucket if this helper actually emptied it.” In the normal unload path, 'async_unload_entry()' removes the whole bucket after forwarded platform unloads complete.

I’ll add a short code comment here to make that clearer.

PR #5385, opened 2026-06-22
Add Airwoods fresh air heat pump

Ranzig93:
Adds a device config for an Airwoods AV-HTPF fresh air heat pump / ventilation heat pump.

The device was successfully tested by manual selection in tuya-local.

Product ID/productKey is currently unknown, so the products block is omitted.

Tested functions:

* Power on/off via DP101
* HVAC modes via DP115
* Target temperature via DP106
* Current/supply air temperature via DP130
* Supply and exhaust air speed via DP102/DP103
* Diagnostic sensors via DP109, DP116, DP119, DP122, DP123, DP124, DP126

### DIFF
--- a/custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
+++ b/custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
@@ -1,0 +1,259 @@
+name: Fresh air heat pump
+
+entities:
+  - entity: climate
+    name: Fresh air heat pump
+    dps:
+      - id: 101
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: "3"
+                value: auto
+              - dps_val: "5"
+                value: cool
+              - dps_val: "6"
+                value: heat
+              - dps_val: "8"
+                value: dry
+              - dps_val: "10"
+                value: fan_only
+
+      - id: 115
+        type: string
+        name: mode
+        hidden: true
+
+      - id: 106
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 10
+          max: 40
+        mapping:
+          - step: 1
+
+      - id: 130
+        type: integer
+        name: current_temperature
+        unit: C
+
+  - entity: number
+    name: Supply air speed
+    icon: "mdi:fan-chevron-up"
+    dps:
+      - id: 102
+        type: string
+        name: value
+        mapping:
+          - dps_val: "0"
+            value: 1
+          - dps_val: "1"
+            value: 2
+          - dps_val: "2"
+            value: 3
+          - dps_val: "3"
+            value: 4
+          - dps_val: "4"
+            value: 5
+          - dps_val: "5"
+            value: 6
+          - dps_val: "6"
+            value: 7
+          - dps_val: "7"
+            value: 8
+          - dps_val: "8"
+            value: 9
+          - dps_val: "9"
+            value: 10
+        range:
+          min: 1
+          max: 10
+
+  - entity: number
+    name: Exhaust air speed
+    icon: "mdi:fan-chevron-down"
+    dps:
+      - id: 103
+        type: string
+        name: value
+        mapping:
+          - dps_val: "0"
+            value: 1
+          - dps_val: "1"
+            value: 2
+          - dps_val: "2"
+            value: 3
+          - dps_val: "3"
+            value: 4
+          - dps_val: "4"
+            value: 5
+          - dps_val: "5"
+            value: 6
+          - dps_val: "6"
+            value: 7
+          - dps_val: "7"
+            value: 8
+          - dps_val: "8"
+            value: 9
+          - dps_val: "9"
+            value: 10
+        range:
+          min: 1
+          max: 10
+
+  - entity: number
+    name: Bypass lower limit temperature
+    category: config
+    icon: "mdi:valve"
+    dps:
+      - id: 107
+        type: integer
+        name: value
+        unit: C
+        range:
+          min: 5
+          max: 30
+        mapping:
+          - step: 1
+
+  - entity: number
+    name: Bypass temperature range
+    category: config
+    icon: "mdi:delta"
+    dps:
+      - id: 108
+        type: integer
+        name: value
+        unit: C
+        range:
+          min: 2
+          max: 15
+        mapping:
+          - step: 1
+
+  - entity: sensor
+    name: Indoor temperature
+    class: temperature
+    dps:
+      - id: 117
+        type: integer
+        name: sensor
+        unit: C
+
+  - entity: sensor
+    name: Outdoor temperature
+    class: temperature
+    dps:
+      - id: 118
+        type: integer
+        name: sensor
+        unit: C
+
+  - entity: sensor
+    name: Supply air temperature
+    class: temperature
+    dps:
+      - id: 130
+        type: integer
+        name: sensor
+        unit: C
+
+  - entity: sensor
+    name: Exhaust air temperature
+    class: temperature
+    dps:
+      - id: 131
+        type: integer
+        name: sensor
+        unit: C
+
+  - entity: sensor
+    name: Stop code
+    category: diagnostic
+    icon: "mdi:alert-octagon-outline"
+    dps:
+      - id: 109
+        type: integer
+        name: sensor
+
+  - entity: binary_sensor
+    name: Alarm
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 119
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - dps_val: null
+            value: false
+          - value: true
+      - id: 119
+        type: bitfield
+        name: fault_code
+
+  - entity: binary_sensor
+    name: Filter clean
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 123
+        type: boolean
+        name: sensor
+
+  - entity: binary_sensor
+    name: Fan switch
+    category: diagnostic
+    icon: "mdi:fan"
+    dps:
+      - id: 125
+        type: boolean
+        name: sensor
+
+  - entity: sensor
+    name: Expansion valve pulse
+    category: diagnostic
+    icon: "mdi:valve"
+    dps:
+      - id: 124
+        type: integer
+        name: sensor
+
+  - entity: sensor
+    name: Filter left time
+    class: duration
+    category: diagnostic
+    icon: "mdi:air-filter"
+    dps:
+      - id: 126
+        type: integer
+        name: sensor
+        unit: d
+
+  - entity: sensor
+    name: Compressor frequency
+    category: diagnostic
+    icon: "mdi:sine-wave"
+    dps:
+      - id: 116
+        type: integer
+        name: sensor
+        unit: Hz
+
+  - entity: binary_sensor
+    name: Defrost
+    category: diagnostic
+    icon: "mdi:snowflake-melt"
+    dps:
+      - id: 122
+        type: boolean
+        name: sensor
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,262 @@
+name: Fresh air heat pump
+
+#     manufacturer: Airwoods
+#     model: AV-HTPF35 / AV-HTPF60 / AV-HTPF90
+
+entities:
+  - entity: climate
+    name: Fresh air heat pump
```
make-all: There is no need to repeat the device name for the entity

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,262 @@
...
+    name: Stop code
+    category: diagnostic
+    icon: "mdi:alert-octagon-outline"
+    dps:
+      - id: 109
+        type: integer
+        name: sensor
+
+  - entity: binary_sensor
+    name: Alarm
```
make-all: Name not needed

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,262 @@
...
+        name: value
+        unit: C
+        range:
+          min: 2
+          max: 15
+        mapping:
+          - step: 1
+
+  - entity: sensor
+    name: Indoor temperature
```
make-all: This should be the current_temperature in the climate entity, not the Supply air temperature.

A separate sensor is not needed

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,262 @@
...
+    name: Exhaust air temperature
+    class: temperature
+    dps:
+      - id: 131
+        type: integer
+        name: sensor
+        unit: C
+
+  - entity: sensor
+    name: Stop code
```
make-all: What is this in real terms? It may be better as an attribute on the problem sensor below, or some other entity if it is connected with one in particular.

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,262 @@
...
+    name: Expansion valve pulse
+    category: diagnostic
+    icon: "mdi:valve"
+    dps:
+      - id: 124
+        type: integer
+        name: sensor
+
+  - entity: sensor
+    name: Filter left time
```
make-all: Use `translation_key: filter_life` instead of name and icon

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,262 @@
...
+    dps:
+      - id: 116
+        type: integer
+        name: sensor
+        unit: Hz
+
+  - entity: binary_sensor
+    translation_key: defrost
+    category: diagnostic
+    icon: "mdi:snowflake-melt"
```
make-all: Icon not needed when translation_key is used

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,262 @@
...
+    name: Filter clean
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 123
+        type: boolean
+        name: sensor
+
+  - entity: binary_sensor
+    name: Fan switch
```
make-all: Using switch in the name is misleading when it is only a sensor

make-all (changes requested):

Ranzig93:
Thanks, I updated the config based on the review comments and the Tuya API device model.

Changes made:
- Changed climate current_temperature to DP117, as the Tuya API identifies it as temp_indoor / indoor temperature.
- Kept DP130 as a separate supply air temperature sensor, as the Tuya API identifies it as sa_temp.
- Moved DP109 to a stop_code attribute on the problem sensor. The Tuya API identifies it as st_code / stop code, but neither the manual nor the Tuya API describe what the numeric values mean. The device currently reports 47 while DP119 is 0 and no fault is active.
- Renamed DP125 to 10 speed fan mode, as the Tuya API describes it as selecting 3-speed vs 10-speed fan control.
- Added additional read-only diagnostic sensors discovered through the Tuya API, including CO2, VOC level, humidity, PM2.5, compressor temperatures, evaporator temperature, condenser temperature, and radiator temperature.

I can provide the full raw Tuya API model output if needed.

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+        hidden: true
+
+      - id: 106
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 10
+          max: 40
+        mapping:
```
make-all: Mapping is unnecessary, as step of 1 is the default

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+          - step: 1
+
+      - id: 117
+        type: integer
+        name: current_temperature
+        unit: C
```
make-all: Unit is unnecessary, as climate only supports one unit, and the temperature above already defines it.

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+
+  - entity: number
+    name: Supply air speed
```
make-all: This is probably better as a fan entity

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+          - dps_val: "8"
+            value: 9
+          - dps_val: "9"
+            value: 10
+        range:
+          min: 1
+          max: 10
+
+  - entity: number
+    name: Exhaust air speed
```
make-all: likewise

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+          - dps_val: "9"
+            value: 10
+        range:
+          min: 1
+          max: 10
+
+  - entity: number
+    name: Bypass lower limit temperature
+    category: config
+    icon: "mdi:valve"
```
make-all: class: temperature

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+        range:
+          min: 5
+          max: 30
+        mapping:
+          - step: 1
+
+  - entity: number
+    name: Bypass temperature range
+    category: config
+    icon: "mdi:delta"
```
make-all: class: temperature_delta

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+        name: value
+        unit: C
+        range:
+          min: 2
+          max: 15
+        mapping:
+          - step: 1
+
+  - entity: sensor
+    category: diagnostic
```
make-all: diagnostic not generally needed on sensors, as HA now separates them into their own group anyway. Only add it for sensors that are connected to config entities, or are really for diagnostic/debug purposes.

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+      - id: 111
+        type: integer
+        name: sensor
+        unit: ppm
+        mapping:
+          - dps_val: 0
+            value: null
+
+  - entity: sensor
+    name: VOC level
```
make-all: Use `translation_key: air_quality` instead

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+    name: Filter clean
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 123
+        type: boolean
+        name: sensor
+
+  - entity: binary_sensor
+    name: 10 speed fan mode
```
make-all: This seems like it should be used in the supply and exhaust fan speed entities to restrict the available options. I don't think it needs to be a separate sensor.

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,349 @@
...
+    dps:
+      - id: 126
+        type: integer
+        name: sensor
+        unit: d
+
+  - entity: sensor
+    name: Compressor frequency
+    category: diagnostic
+    icon: "mdi:sine-wave"
```
make-all: use class: frequency instead of setting the icon explicitly

make-all (changes requested):

Ranzig93:
Thanks for the review. I addressed the comments and removed the separate 10 speed fan mode sensor.

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
Ranzig93: Thanks

However, the device exposes DPs 102 and 103 as string enum values from "0" to "9" for the 10 discrete fan speed levels. Because Home Assistant fan entities use percentage values, the displayed values are not ideal. Depending on the mapping/range, the UI shows values like 1% 12%, 23%, 34%, etc., or the write-back behavior does not work correctly.

I left them as fan entities for now to match the review suggestion, but I am not fully sure what the best solution is for this device. Would you prefer to keep them as fan entities with the percentage display limitation, or should these be changed back to number entities because they better represent the actual Tuya enum values?

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,327 @@
...
+            value: 60
+          - dps_val: "6"
+            value: 70
+          - dps_val: "7"
+            value: 80
+          - dps_val: "8"
+            value: 90
+          - dps_val: "9"
+            value: 100
+        range:
```
make-all: Range not needed when there are mapped values already in the HA expected 0-100 range

custom_components/tuya_local/devices/airwoods_av_htpf_fresh_air_heat_pump.yaml
```diff
@@ -0,0 +1,327 @@
...
+        name: sensor
+        mapping:
+          - dps_val: "0"
+            value: null
+          - dps_val: "1"
+            value: excellent
+          - dps_val: "2"
+            value: good
+          - dps_val: "3"
+            value: medium
```
make-all: `moderate` is covered by the translations (not medium)

make-all (changes requested):

PR #5386, opened 2026-06-22
Update Funai ERW-150 damper labels

dankarization:
Follow-up to #4074.

This updates the Funai ERW-150 DP 102 damper select labels to match the device UI:

- `Auto_1` -> `+16 °C`
- `Auto_2` -> `+20 °C`
- `Manual` remains `Manual`

The raw DP values are unchanged, only the Home Assistant select labels are adjusted.

Validation:
- Parsed `custom_components/tuya_local/devices/funai_erw150_ventilation.yaml` with PyYAML and confirmed the DP 102 mapping.
- Full pytest suite was not run locally because this checkout does not have pytest/uv installed.

### DIFF
--- a/custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
+++ b/custom_components/tuya_local/devices/funai_erw150_ventilation.yaml
@@ -39,9 +39,9 @@ entities:
           - dps_val: "Manual"
             value: "Manual"
           - dps_val: "Auto_1"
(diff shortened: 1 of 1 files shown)


make-all:
Since this config has existed for some time already, there may be a large number of users who would be inconvenienced by a sudden change in the values of this select.

Backward compatibility is more important than matching the Tuya App for me.

PR #5389, opened 2026-06-22
Fix iHome AutoVac Nova stop/pause commands and status mapping

backcountrymountains:
## Summary

The existing config for the iHome AutoVac Nova (`iHRV6`, product id `gejo2t2mgux3avdq`) has three bugs that prevent stop, pause, and cleaning state from working correctly. All three were identified and verified by testing against a real device.

### Bug 1: DPS 1 (`power`) causes vacuum to permanently show as IDLE while cleaning

The `power` DP has a `dps_val: null -> value: false, hidden: true` mapping. Because this device never actually reports a value for DPS 1, `power_dps.get_value()` permanently returns Python `False` via that hidden null mapping.

In `vacuum.py`, the `activity` property has this check after the specific status matches but before the `CLEANING` fallback:

```python
elif self._power_dps and self._power_dps.get_value(self._device) is False:
    return VacuumActivity.IDLE
```

This means any cleaning status (e.g. `smart_clean`) that doesn't match an earlier branch falls through to this check, sees `False`, and returns `IDLE` instead of `CLEANING`. The vacuum card shows idle while the robot is actively cleaning, and the stop/pause buttons are suppressed.

**Fix:** Remove DPS 1 entirely. The device does not use it.

### Bug 2: Stop and pause commands have no effect

The device ignores `standby` written to DPS 3. Stop and pause must be sent as boolean `false` to DPS 2.

The previous config used DPS 2 as a `command` boolean with complex conditional mappings referencing DPS 3 as a hidden `mode` DP. In practice, when `vacuum.stop` is called, `async_stop` writes `standby` to DPS 3, which the device ignores entirely.

**Fix:** Rename DPS 2 to `activate` and DPS 3 to `command` (string). Add `value_redirect: activate` to the `stop` mapping on the command DP, so that commanding stop sends `{2: false}` instead of `{3: "standby"}`. Add a `{dps_val: false, value: stop}` mapping on `activate` so the redirect's boolean type-conversion produces `false` rather than `true` (a non-empty string like `"stop"` would otherwise coerce to `True` in `_correct_type`).

Pause is handled automatically: since `"pause"` is absent from the command DP's values, `async_pause` falls through to the `elif self._activate_dps` branch and sends `false` to DPS 2 directly.

### Bug 3: `paused` status maps to `"pause"` instead of `"paused"`

The existing mapping has `dps_val: paused -> value: pause`. The `activity` property in `vacuum.py` checks `elif status == "paused"` (with a `d`), so the device's paused state was never recognised and fell through to `CLEANING`.

**Fix:** Change `value: pause` to `value: paused`.

## Test plan

- [ ] `vacuum.start` sends `{3: "smart"}` and device begins cleaning
- [ ] `vacuum.stop` sends `{2: false}` (not `{3: "standby"}`) and device stops
- [ ] `vacuum.pause` sends `{2: false}` and device pauses
- [ ] `vacuum.return_to_base` sends `{3: "chargego"}` and device docks
- [ ] Vacuum shows `cleaning` activity while running (not `idle`)
- [ ] Vacuum shows `paused` activity when paused
- [ ] Vacuum shows `docked` activity when charging/charged
- [ ] Fan speed, locate, and error sensor continue to function

### DIFF
--- a/custom_components/tuya_local/devices/ihome_autovac_nova.yaml
+++ b/custom_components/tuya_local/devices/ihome_autovac_nova.yaml
@@ -1,0 +1,101 @@
+name: Vacuum
+products:
+  - id: gejo2t2mgux3avdq
+    manufacturer: iHome
+    model: AutoVac Nova
+    model_id: iHRV6
+entities:
+  - entity: vacuum
+    dps:
+      - id: 2
+        type: boolean
+        name: activate
+        mapping:
+          - dps_val: false
+            value: stop
+
+      - id: 3
+        type: string
+        name: command
+        mapping:
+          - dps_val: smart
+            value: start
+          - dps_val: chargego
+            value: return_to_base
+          - dps_val: standby
+            value: stop
+            value_redirect: activate
+
+      - id: 5
+        type: string
+        name: status
+        mapping:
+          - dps_val: standby
+            value: standby
+          - dps_val: smart_clean
+            value: cleaning
+          - dps_val: wall_clean
+            value: cleaning
+          - dps_val: spot_clean
+            value: cleaning
+          - dps_val: mop_clean
+            value: mopping
+          - dps_val: goto_charge
+            value: returning
+          - dps_val: charging
+            value: charging
+          - dps_val: charge_done
+            value: charged
+          - dps_val: paused
+            value: paused
+          - dps_val: cleaning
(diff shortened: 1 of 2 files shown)


make-all:
I cannot accept a PR made against a 3 year out of date copy of the respository.

PR #5390, opened 2026-06-23
Added new device configuration for Omni X Duplex Water Heater in the …

spokemon:
…Tuya Local custom component.

### DIFF
--- a/custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
+++ b/custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
@@ -13,14 +13,10 @@ entities:
           - dps_val: false
             value: "off"
           - dps_val: true
-            constraint: mode
-            conditions:
-              - dps_val: hot
-                value: heat_pump
+            value: heat_pump
       - id: 2
         type: string
-        name: mode
-        hidden: true
+        name: unused_mode  # should be hot always for this model
       - id: 4
         type: integer
         name: temperature
--- a/custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
+++ b/custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
@@ -1,0 +1,34 @@
+name: Water heater
+products:
+  - id: 5qklaasyaemcifen
+    manufacturer: Omni X
+    model: Duplex
+entities:
+  - entity: water_heater
+    translation_only_key: hot_water
+    dps:
+      - id: 1
+        type: boolean
+        name: operation_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: "on"
+      - id: 2
+        type: string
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
```diff
@@ -0,0 +1,34 @@
+name: Water heater
+products:
+  - id: 5qklaasyaemcifen
+    manufacturer: Omni X
+    model: Duplex
+entities:
+  - entity: water_heater
+    translation_only_key: hot_water
```
make-all: This translation key is only intended for benchtop water dispensors, for general domestic water heaters, no translation key is necessary.

custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
```diff
@@ -0,0 +1,34 @@
...
+    dps:
+      - id: 1
+        type: boolean
+        name: operation_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: "on"
```
make-all: "on" is not a valid operation mode for a water_heater. You should use the type of heater it is ("heat_pump", "electric", "gas")

custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - id: 2
+        type: string
+        name: mode
+        hidden: true
```
make-all: mode is hidden, but not referenced by any other dp. Maybe you had an earlier version that combined it with operation_mode above?

custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - id: 4
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 30
+          max: 60
+      - id: 15
+        type: bitfield
+        name: fault_code
```
make-all: fault_code should be separated out into its own binary_sensor following the `class: problem` pattern used by most other configs.

make-all (dismissed):

custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
spokemon: it is now referenced by dp1. the heat pump had other modes (cooling, auto) in the app, but the manufacturer says must not use those other modes.

custom_components/tuya_local/devices/omni_x_duplex_waterheater.yaml
spokemon: changes made. will reference dp2 and set value 'heat_pump'

PR #5391, opened 2026-06-23
added Djive ARC Portable Fan

smo87:
added support for Djive ARC Portable Fan

### Command:
- fan (speed, oscillate, preset mode)
- sound
- UV Sterilization

### Sensor:
- PM2.5

### Diagnostic:
- battery
- Filter life
- battery charging
- problem

### Configuration:
- filter reset

DP mapping check with Tuya Developer Platform and confirmed on my own device.

### DIFF
--- a/custom_components/tuya_local/devices/djive_arc_portable_fan.yaml
+++ b/custom_components/tuya_local/devices/djive_arc_portable_fan.yaml
@@ -47,14 +47,14 @@ entities:
           - value: true
 
   - entity: switch
-    name: UV Sterilization
+    translation_key: uv_sterilization
     dps:
       - id: 102
         type: boolean
         name: switch
 
   - entity: switch
-    name: Sound
+    translation_key: sound
     dps:
       - id: 13
         type: boolean
@@ -86,7 +86,7 @@ entities:
         name: sensor
 
   - entity: sensor
-    name: Filter life
+    translation_key: filter_life
     category: diagnostic
     dps:
       - id: 103
@@ -94,11 +94,11 @@ entities:
         name: sensor
         unit: "%"
 
-#  - entity: button
-#    name: Reset filter
-#    class: restart
-#    category: config
-#    dps:
-#      - id: 104
-#        type: boolean
-#        name: button
+  - entity: button
+    translation_key: filter_reset
+    category: config
+    dps:
+      - id: 104
+        type: boolean
+        optional: true
+        name: button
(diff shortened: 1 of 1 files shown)


smo87:
@make-all I fixed the **products.0.id** error

smo87:
@make-all I fixed also translation_key errors and added filter reset configuration

custom_components/tuya_local/devices/djive_arc_portable_fan.yaml
```diff
@@ -0,0 +1,104 @@
+name: ARC Portable fan
```
make-all: Top level name should be without branding.

custom_components/tuya_local/devices/djive_arc_portable_fan.yaml
```diff
@@ -0,0 +1,104 @@
...
+products:
+  - id: 4srsoiozpa7wd9ha
+    manufacturer: djive
+    model: ARC Portable
+entities:
+  - entity: fan
+    dps:
```
make-all, line 9: Use `translation_key: fan_with_presets`, as HA by default does not provide any translations for preset_mode.

custom_components/tuya_local/devices/djive_arc_portable_fan.yaml
```diff
@@ -0,0 +1,104 @@
...
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 24
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
```
make-all, line 48: Add `fault_code` as an attribute here with the original value, so it is not lost.

make-all (changes requested):

custom_components/tuya_local/devices/djive_arc_portable_fan.yaml
smo87: fixed

custom_components/tuya_local/devices/djive_arc_portable_fan.yaml
smo87: fixed

custom_components/tuya_local/devices/djive_arc_portable_fan.yaml
smo87: fixed

PR #5401, opened 2026-06-24
Fix mypin_6l_videopetfeeder, added various compatible models

pdwonline:
### Fixes
- DP 245 (feed_publish): renamed from "Log offline feed" to "Manual feed"
- DP 245: added `persist: false` (device does not echo DP back after feeding)
- DP 245: corrected range (min: 1, max: 10) and added hidden mapping for idle value -2000

### New entities
- DP 237: added separate `text` entity "Meal plan" for mealplan-card compatibility

### Device info
- Tested on Pretty Paws PP003 / Petrust PP003 / Lexmart PP003
- Same hardware as Mypin 6L (product id: 8zrzjldq07tlxnlv)
- DP 237 uses hex encoding (not base64)
- Cached state example: {"237": "7f1400011000000"} = all days, 14:00, 1 portion, enabled

### Notes
- `persist: false` on DP 245 is essential: the device does not echo the value
  back after dispensing food, causing HA to wait indefinitely without it

### DIFF
--- a/custom_components/tuya_local/devices/mypin_6l_videopetfeeder.yaml
+++ b/custom_components/tuya_local/devices/mypin_6l_videopetfeeder.yaml
@@ -3,15 +3,6 @@ products:
   - id: 8zrzjldq07tlxnlv
     manufacturer: Mypin
     model: 6L
-  - id: 8zrzjldq07tlxnlv
-    manufacturer: Pretty Paws
-    model: PP003
-  - id: 8zrzjldq07tlxnlv
-    manufacturer: Petrust
-    model: PP003
-  - id: 8zrzjldq07tlxnlv
-    manufacturer: Lexmart
-    model: PP003
 entities:
   - entity: switch
     name: Camera record
@@ -209,7 +200,7 @@ entities:
         optional: true
         name: backup_log
   - entity: text
-    name: Meal plan
+    translation_key: Meal plan
     icon: "mdi:calendar-clock"
     dps:
       - id: 237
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/mypin_6l_videopetfeeder.yaml
```diff
@@ -3,6 +3,15 @@ products:
   - id: 8zrzjldq07tlxnlv
     manufacturer: Mypin
     model: 6L
+  - id: 8zrzjldq07tlxnlv
```
make-all: This id is already above, you should not duplicate ids in the list

custom_components/tuya_local/devices/mypin_6l_videopetfeeder.yaml
```diff
@@ -195,14 +204,18 @@ entities:
...
-      - id: 237
-        type: string
-        optional: true
-        name: schedule
       - id: 247
         type: string
         optional: true
         name: backup_log
+  - entity: text
+    name: Meal plan
```
make-all: Use `translation_key: meal_plan` instead of explicit name and icon

custom_components/tuya_local/devices/mypin_6l_videopetfeeder.yaml
pdwonline: ok, removed them

custom_components/tuya_local/devices/mypin_6l_videopetfeeder.yaml
pdwonline: done

pdwonline:
- `translation_key: manual_feed` applied as requested.
- For the `text` entity on DP 237: `translation_key: meal_plan` does not exist 
  in the tuya-local translations, so used `name: Meal plan` instead. 
  Happy to add a translation key if you can point me to where text entity 
  translation keys are defined.

make-all:
`meal_plan` exists as a translation key. You did not try that, you left it as "Meal plan"

pdwonline:
Sorry. Fixed that. into
translation_key: meal_plan

PR #5403, opened 2026-06-24
Add support for DUUX Blade Fan/DXCF23JP

nttntt:
Add support for DUUX Blade Fan/DXCF23JP
Note: I couldn't retrieve the Product ID, so it is left blank. The device has been tested and works with this configuration.


make-all:
Apart from product details, this is identical to Duux Whisper Flex Ultimate, so merged with that

PR #5405, opened 2026-06-24
Ovlaim 809 ceiling fan: add sleep preset mode, make timer a read-only sensor

### LINKED ISSUE #5406: Ovlaim 809 ceiling fan: add hidden sleep mode; timer datapoint is read-only (follow-up to #5373)
## Background

The Ovlaim 809 ceiling fan with CCT light (Tuya WiFi canopy, product `ftqsaibnb3insvp9`, category `fsd`) was added in #5373 (which resolved #5374). That config got brightness, color temperature, fan speed, the `fresh`/`nature` modes, direction, and the timer entity working.

## Findings (follow-up to #5373)

Re-checking the device's Tuya cloud specification and testing on real hardware surfaced two things:

1. **`fan_mode` (dp61) has a third value: `sleep`.** The device enum is `["fresh", "nature", "sleep"]`, but the config maps only `fresh` and `nature`, so `sleep` cannot be selected from Home Assistant. The Smart Life app exposes only two mode buttons, which is why this was missed initially.

2. **The fan timer (`countdown_left_fan`, dp64) is effectively read-only.** It was exposed as a writable `number`. In testing, the device accepts and stores whatever value is written to it, but it never starts a countdown or turns the fan off, so writing it does nothing useful. (Working out how the app actually arms the timer would take deeper investigation that isn't worth the effort right now.)

## Solution

PR #5405 adds the `sleep` preset and changes `countdown_left_fan` from a writable number to a read-only `sensor` (`time_remaining`, duration, diagnostic). yamllint clean, `tests/test_device_config.py` passes.

agigante80:
Resolves #5406

Follow-up to #5373. Two corrections to `ovlaim_809_ceiling_fanlight.yaml`, from the device's Tuya cloud specification (product `ftqsaibnb3insvp9`):

### 1. Add the `sleep` fan preset mode (dp61)
The device `fan_mode` enum is `[fresh, nature, sleep]`, but the config mapped only `fresh` and `nature`. This adds the missing `sleep` value so it can be selected from Home Assistant.

### 2. Make the fan timer a read-only sensor (dp64)
`countdown_left_fan` was exposed as a writable `number`. We ran several tests: the device accepts and stores whatever value is written to it, but it never starts a countdown or turns the fan off, so in practice it behaves as read-only. (Working out exactly how the app arms the timer would take deeper investigation that isn't worth the effort right now.) Changed from a writable number to a read-only `sensor` (`time_remaining`, duration, diagnostic).

### Verification
- Real hardware: the sensor reflects the value the device reports; `sleep` is selectable and accepted by the device.
- `uv run yamllint custom_components/tuya_local/devices/ovlaim_809_ceiling_fanlight.yaml` clean
- `uv run pytest tests/test_device_config.py` 32 passed

### DIFF
--- a/custom_components/tuya_local/devices/ovlaim_809_ceiling_fanlight.yaml
+++ b/custom_components/tuya_local/devices/ovlaim_809_ceiling_fanlight.yaml
@@ -65,5 +65,6 @@ entities:
     dps:
       - id: 64
         type: integer
+        optional: true
         name: sensor
         unit: min
--- a/custom_components/tuya_local/devices/ovlaim_809_ceiling_fanlight.yaml
+++ b/custom_components/tuya_local/devices/ovlaim_809_ceiling_fanlight.yaml
@@ -45,6 +45,8 @@ entities:
             value: fresh
           - dps_val: nature
             value: nature
+          - dps_val: sleep
+            value: sleep
       - id: 62
         name: speed
         type: integer
@@ -56,16 +58,12 @@ entities:
         name: direction
         type: string
         optional: true
-  - entity: number
-    translation_key: timer
+  - entity: sensor
+    translation_key: time_remaining
     class: duration
-    category: config
+    category: diagnostic
     dps:
       - id: 64
-        name: value
         type: integer
-        optional: true
+        name: sensor
         unit: min
-        range:
-          min: 0
-          max: 540


make-all:
The fact it is labelled countdown_left suggests it is intended to be read only. Normally when such a dp is present, there will be another enum type one for setting the timer.

agigante80:
Thanks, you were right to flag this. We ran several tests: the device accepts and stores whatever value is written to `countdown_left_fan`, but we couldn't get it to actually start a countdown or turn the fan off, so in practice it behaves as read-only. Working out exactly how the app arms the timer would take deeper investigation, which isn't really worth the effort right now.

So I've set `countdown_left_fan` (dp64) as a read-only `sensor` (`time_remaining`, duration, diagnostic) instead of a writable number, plus added the missing `sleep` fan preset (the device's `fan_mode` enum is `fresh/nature/sleep`). yamllint clean, `tests/test_device_config.py` passes.

PR #5410, opened 2026-06-25
Fix inverted direction mapping for Fanco Eco Silent Deluxe

### LINKED ISSUE #5409: Fix Fanco Eco Silent Deluxe direction mapping
### Description

The direction mapping for the Fanco Eco Silent Deluxe ceiling fan appears to be reversed.

Device file:

`custom_components/tuya_local/devices/fanco_ecosilentdeluxe.yaml`

Device details:

* Product ID: `agyqzsfpafsvkrl9`
* Manufacturer: `Fanco`
* Model: `Eco Silent Deluxe`
* Model ID: `MZ-DCFAN`
* Direction DP: `8`

The current config exposes DP `8` directly as the fan direction:

```yaml
- id: 8
  name: direction
  type: string
```

From diagnostics and local testing, the raw Tuya values for `forward` and `reverse` are physically inverted when exposed in Home Assistant. Selecting `forward` in Home Assistant causes the fan to run in the reverse direction, and selecting `reverse` causes it to run in the forward direction.

I believe this can be fixed by adding an explicit mapping to DP `8`:

```yaml
- id: 8
  name: direction
  type: string
  mapping:
    - dps_val: forward
      value: reverse
    - dps_val: reverse
      value: forward
```

I have tested this locally by overriding the device YAML in my Home Assistant instance, and the Home Assistant `forward` and `reverse` options then match the actual physical fan direction.

I have verified this across four separate fans.

I am ready to submit a PR with this one-file change.

### Steps to reproduce

_No response_

### Expected behaviour

_No response_

### Additional context

_No response_

### Confirmation

- [ ] I am sure this is a bug or improvement, that is well enough described that it can be implemented.

corneels:
Fixes the reversed direction mapping for the Fanco Eco Silent Deluxe ceiling fan `fanco_ecosilentdeluxe.yaml`

The existing config exposes DP `8` directly as `direction`, but on this device the raw Tuya `forward` and `reverse` values are physically inverted in Home Assistant.

This PR adds an explicit mapping so that Home Assistant’s `forward` and `reverse` options match the actual fan direction.

Tested locally on 4 Fanco Eco Silent Deluxe fans.

This PR fixes #5409

### DIFF
--- a/custom_components/tuya_local/devices/fanco_ecosilentdeluxe.yaml
+++ b/custom_components/tuya_local/devices/fanco_ecosilentdeluxe.yaml
@@ -28,6 +28,11 @@ entities:
       - id: 8
         name: direction
         type: string
+        mapping:
+          - dps_val: forward
+            value: reverse
+          - dps_val: reverse
+            value: forward
   - entity: light
     dps:
       - id: 15


make-all:
Sorry, but when one report like this is received after 3.5 years, it is more likely that you have an installation or understanding problem than the device has these switched.

PR #5413, opened 2026-06-25
(feat) New Device Support - Hiniso 10L Smart Dehumidifier (Model RM10S)

malhar-c:
## Description
Adds support for the Hiniso 10L Smart Dehumidifier (Model RM10S) 
The existing hiniso dehumidifier integration (hiniso_dehumidifier.yaml) didn't work probably becasue of a new firmware, as most of the DP IDs are different in this model, few are noted below.

Protocol Version: 3.5

### Mapping Notes:
* **Modes (DP 5):** 
  * `dehumidify` -> `Normal` (Compressor reacts to manual RH target)
  * `auto` -> (AU) `comfort` (Hardware ignores manual target in this mode and sets internal compressor logic based on ambient room temperature).
  * `continuous` -> (CO) `boost` (Hardware ignores humidity sensors and runs compressor 24/7).
  * *Note on DP 2:* When the unit enters `auto` mode, the Tuya Wi-Fi chip overrides DP 2 and hardcodes it to `30`. This is hardware-level behavior, but it's not eactly right, the compressor logic is internal it depends on the ambient temerature and follows a hardware level logic, which is this: temp>27 = humidity 50, 20<temp<27 = humidity 55, and 5<temp<20 = humidity 60%. temp<5 = compressor turns off. But DP-2 is set to 30% for whatever reason, this is a known hardware limitation.
* **Child Lock (DP 16):** This unit doesn't have that feature, but the Tuya chip broadcasts and accepts the command. Mapped as a hidden `config` entity as without this the Stage 2 of pairing fails (device doesn't show up).
* **Timer:** Moved to DP 101

## Device Signature
```
"cached_state": {
      "updated_at": 1782400067.400447,
      "1": true,
      "2": 50,
      "4": "high",
      "5": "dehumidify",
      "6": 59,
      "16": false,
      "25": "off",
      "101": "None"
    }
```

### DIFF
--- a/custom_components/tuya_local/devices/hiniso_10L_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/hiniso_10L_dehumidifier.yaml
@@ -130,23 +130,3 @@ entities:
       - id: 16
         type: boolean
         name: switch
-
-  # SWING (Hidden)
-  - entity: switch
-    name: Swing
-    category: config
-    hidden: true
-    dps:
-      - id: 8
-        type: boolean
-        name: switch
-
-  # SECONDARY HUMIDITY ENUM (Hidden)
-  - entity: sensor
-    name: Secondary Humidity Enum
-    category: diagnostic
-    hidden: true
-    dps:
-      - id: 3
-        type: string
-        name: sensor
--- a/custom_components/tuya_local/devices/hiniso_10L_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/hiniso_10L_dehumidifier.yaml
@@ -1,4 +1,4 @@
-name: Hiniso 10L Dehumidifier
+name: Hiniso 10L Smart Dehumidifier
 
 entities:
   # CORE HUMIDIFIER CONTROLS
@@ -21,11 +21,11 @@ entities:
         type: string
         mapping:
           - dps_val: dehumidify
-            value: normal
+            value: manual # Dehumidify mode, runs the compressor based on the set target humidity.
           - dps_val: auto
-            value: auto
+            value: comfort # Comfort mode (AU) as per the manual, relies on internal logic to turn on/off compressor based on ambient temperature
           - dps_val: continuous
-            value: continuous
+            value: boost # Continous mode (CO) or the drying mode, runs the compressor continuously
       - id: 6
         name: current_humidity
         type: integer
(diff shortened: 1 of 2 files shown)


malhar-c:
Hi @make-all can we squash all these into one commit? I was playing around with this, hence multiple commits.

make-all:
It will be squashed on merge. Do not worry about the multiple commits while in review

custom_components/tuya_local/devices/hiniso_10L_dehumidifier.yaml
```diff
@@ -0,0 +1,151 @@
+# HINISO Model: RM10S (10L Smart Dehumidifier)
+# Device ID: d7696cfb645eb04673gobl
+
+name: Hiniso 10L Smart Dehumidifier
```
make-all: Top level name should be generic, in case other products also match the config.

If you find the product id for the device, you can add a products section with the branding info (manufacturer and model).

custom_components/tuya_local/devices/hiniso_10L_dehumidifier.yaml
```diff
@@ -0,0 +1,151 @@
...
+          # Dehumidify mode, runs the compressor based on set target humidity
+          - dps_val: dehumidify
+            value: normal
+          # Comfort mode (AU) as per the manual, relies on internal logic.
+          # it controls the compressor based on ambient temperature.
+          - dps_val: auto
+            value: comfort
+          # Continuous mode (CO) runs the compressor continuously
+          - dps_val: continuous
+            value: boost
```
make-all: If you add `translation_key: extended` to the entity definition, `continuous` is available as an additional translated mode name.

custom_components/tuya_local/devices/hiniso_10L_dehumidifier.yaml
```diff
@@ -0,0 +1,151 @@
...
+      - id: 6
+        name: current_humidity
+        type: integer
+
+  # HUMIDITY SENSOR (Making this Sensor available for use in automations)
+  - entity: sensor
+    name: Relative Humidity
```
make-all: Instead of name and icon, use `class: humidity`

custom_components/tuya_local/devices/hiniso_10L_dehumidifier.yaml
```diff
@@ -0,0 +1,151 @@
...
+      - id: 6
+        type: integer
+        name: sensor
+        unit: "%"
+        class: measurement
+
+  # WATER TANK FULL
+  - entity: binary_sensor
+    translation_key: tank_full
+    icon: mdi:water-alert-outline
```
make-all: Icon not needed as the translation key already defines it

custom_components/tuya_local/devices/hiniso_10L_dehumidifier.yaml
```diff
@@ -0,0 +1,151 @@
...
+        name: sensor
+        mapping:
+          - dps_val: "on"
+            value: true
+          - dps_val: "off"
+            value: false
+
+  # FAN SPEED
+  - entity: select
+    name: Fan Speed
```
make-all: Use a fan entity, with speed 50 and 100.

make-all (changes requested):

PR #5420, opened 2026-06-26
feat(mcs350): add Teploluxe MCS350 thermostat

eldarsudden:
Add new thermostat
products:
  - id: "tfmfbr341u0tx3fb" manufacturer: Teplolux model: Teploluxe MCS350 Thermostat

Successfully tested using real devices. All described entities are working fine.  

image

### DIFF
--- a/custom_components/tuya_local/devices/teploluxe_mcs350.yaml
+++ b/custom_components/tuya_local/devices/teploluxe_mcs350.yaml
@@ -75,8 +75,7 @@ entities:
 
   - entity: sensor
     name: Power consumption
-    device_class: power
-    state_class: measurement
+    class: power
     category: diagnostic
     dps:
       - id: 22
@@ -86,8 +85,7 @@ entities:
 
   - entity: sensor
     name: Air temperature
-    device_class: temperature
-    state_class: measurement
+    class: temperature
     category: diagnostic
     dps:
       - id: 102
--- a/custom_components/tuya_local/devices/teploluxe_mcs350.yaml
+++ b/custom_components/tuya_local/devices/teploluxe_mcs350.yaml
@@ -1,0 +1,137 @@
+name: Teploluxe MCS350 Thermostat
+products:
+  - id: "tfmfbr341u0tx3fb"
+    manufacturer: Teplolux
+    model: Teploluxe MCS350 Thermostat
+entities:
+  - entity: climate
+    translation_key: thermostat
+    dps:
+      - id: 111
+        name: hvac_mode
+        type: string
+        mapping:
+          - dps_val: "OFF" 
+            value: "off"
+          - dps_val: "ON"
+            constraint: preset_mode
+            conditions:
+              - dps_val: manual
+                # manual
+                value: heat
+              - dps_val: program
+                # program
+                value: auto
+      - id: 2
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: manual
+            value: manual
+          - dps_val: program
+            value: program
+
+      - id: 16
+        type: integer
+        name: temperature
+        range:
+          min: 5
+          max: 45
+
+      - id: 24
+        type: integer
+        name: current_temperature
+
+      - id: 106
+        type: boolean
+        name: hvac_action
+        mapping:
+          - dps_val: true
+            value: heating
+          - dps_val: false
+            value: idle
+
+  - entity: number
+    name: Temp calibration
+    category: config
+    dps:
+      - id: 27
+        type: integer
+        name: value
+        range:
+          min: -9
+          max: 9
+
+  - entity: number
+    name: Max Air Temp
+    category: config
+    dps:
+      - id: 103
+        type: integer
+        name: value
+        range:
+          min: 5
+          max: 45
+
+  - entity: sensor
+    name: Power consumption
+    device_class: power
+    state_class: measurement
+    category: diagnostic
+    dps:
+      - id: 22
+        type: integer
+        name: sensor
+        unit: kW
+
+  - entity: sensor
+    name: Air temperature
+    device_class: temperature
+    state_class: measurement
+    category: diagnostic
+    dps:
+      - id: 102
+        type: integer
+        name: sensor
+        unit: °C
+
+  - entity: switch
+    name: Window check
+    category: config
+    dps:
+      - id: 110
+        type: string
+        name: switch
+        mapping:
+          - dps_val: "OFF"
+            value: false
+          - dps_val: "ON"
+            value: true
+
+  - entity: select
+    name: Sensor Mode
+    category: config
+    dps:
+      - id: 43
+        type: string
+        name: option
+        mapping:
+          - dps_val: flor
+            value: Floor Sensor
+          - dps_val: air
+            value: Air Sensor
+          - dps_val: booth
+            value: Both Sensors
+
+  - entity: switch
+    name: Self training
+    category: config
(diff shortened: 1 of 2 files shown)


eldarsudden:
@make-all Hi, fixed issues. Please, review once again. Thanks!

custom_components/tuya_local/devices/teploluxe_mcs350.yaml
```diff
@@ -0,0 +1,135 @@
+name: Teploluxe MCS350 Thermostat
```
make-all: Top level name should be generic without branding, in case other products also match this config

custom_components/tuya_local/devices/teploluxe_mcs350.yaml
```diff
@@ -0,0 +1,135 @@
...
+products:
+  - id: "tfmfbr341u0tx3fb"
+    manufacturer: Teploluxe
+    model: Teploluxe MCS350 Thermostat
```
make-all: manufacturer should not be repeated in the model

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
```diff
@@ -0,0 +1,135 @@
...
+    translation_key: thermostat
+    dps:
+      - id: 111
+        name: hvac_mode
+        type: string
+        mapping:
+          - dps_val: "OFF"
+            value: "off"
+          - dps_val: "ON"
+            constraint: preset_mode
```
make-all, line 17: Since this is already available as preset_mode, it is not necessary to complicate the config by including it here. "ON" can be mapped to `heat`, and the presets used to control program/manual, since those actually match the Tuya naming also.

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
```diff
@@ -0,0 +1,135 @@
...
+        type: boolean
+        name: hvac_action
+        mapping:
+          - dps_val: true
+            value: heating
+          - dps_val: false
+            value: idle
+
+  - entity: number
+    name: Temp calibration
```
make-all: Use `translation_key: temperature_calibration` instead of this name

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
```diff
@@ -0,0 +1,135 @@
...
+    dps:
+      - id: 27
+        type: integer
+        name: value
+        range:
+          min: -9
+          max: 9
+
+  - entity: number
+    name: Max Air Temp
```
make-all: Use `translation_key: maximum_temperature` and `class: temperature` instead of this name

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
```diff
@@ -0,0 +1,135 @@
...
+    dps:
+      - id: 103
+        type: integer
+        name: value
+        range:
+          min: 5
+          max: 45
+
+  - entity: sensor
+    name: Power consumption
```
make-all: name not necessary

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
```diff
@@ -0,0 +1,135 @@
...
+    class: power
+    category: diagnostic
+    dps:
+      - id: 22
+        type: integer
+        name: sensor
+        unit: kW
```
make-all, line 101: Does this not need scaling?

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
```diff
@@ -0,0 +1,135 @@
...
+    class: temperature
+    category: diagnostic
+    dps:
+      - id: 102
+        type: integer
+        name: sensor
+        unit: °C
+
+  - entity: switch
+    name: Window check
```
make-all: Use "Open window detection" to be consistent with other configs.

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
```diff
@@ -0,0 +1,135 @@
...
+        type: string
+        name: switch
+        mapping:
+          - dps_val: "OFF"
+            value: false
+          - dps_val: "ON"
+            value: true
+
+  - entity: select
+    name: Sensor Mode
```
make-all: Follow HA naming convention of only capitalising the first word of an entity or device name

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
```diff
@@ -0,0 +1,135 @@
...
+        mapping:
+          - dps_val: flor
+            value: Floor Sensor
+          - dps_val: air
+            value: Air Sensor
+          - dps_val: booth
+            value: Both Sensors
+
+  - entity: switch
+    name: Self training
```
make-all: "Learning mode" is used by other configs

make-all (dismissed):
Filename should end with "_thermostat.yaml" to make it easier for users to see that this config might match their thermostat.

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
eldarsudden: Screenshot_2026-07-03-03-49-46-19_a3b3f7ff182fbca1933b088238f0f0da
It is exactly the same as in Smart Life app

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
eldarsudden: Screenshot_2026-07-03-03-54-22-57_a3b3f7ff182fbca1933b088238f0f0da
Names taken from original application. Looks like simple translation from chinese.

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
eldarsudden: It seems not as max temperature overall, but max air sensor temperature. I don't know what exactly is it.

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
make-all: How can it read 5.5kW without scaling? The value is an integer

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
eldarsudden: Since program editing is not available at HA,  thermostat modes that are really _ necessary "heat" and "auto".  It's complicated, but works and looks pretty good. In my opinion preset mode  is unnecessary because of useless

custom_components/tuya_local/devices/teploluxe_mcs350_thermostat.yaml
eldarsudden: I'm sorry. You are right. Maybe I was little bit tired tonight.
Screenshot_2026-07-03-07-34-05-80_c3a231c25ed346e59462e84656a70e50

PR #5421, opened 2026-06-27
feat(laresar_l6_nex_vacuum): add Laresar L6 NEX robot vacuum and mop

h0me5k1n:
## Summary

- Adds device config for the Laresar L6 NEX robot vacuum and mop (product ID `pi5tljdjcchcn0mr`)
- Tested locally against a real device running in Home Assistant

## Entities

- **Vacuum** — status and battery level (docked/cleaning/paused/returning/idle/error)
- **Buttons** — start cleaning, pause, return to dock, find robot, collect dust, reset edge brush, reset roll brush, reset filter, reset map
- **Selects** — cleaning mode, suction power, water tank level, work mode, carpet clean preference, dust collection frequency, language
- **Switches** — do not disturb, resume cleaning, customize mode, auto boost
- **Number** — volume (0–100%)
- **Sensors** — battery, clean time, clean area, edge brush life, roll brush life, filter life, mop life, total clean area, total clean count, total clean time, mop state
- **Binary sensor** — fault status (note: also shows as active when do not disturb mode is enabled)

## Notes

- Map/room selection features (DPs 14, 15, 16) are not supported locally; those require the cloud app
- DPS values confirmed against the Tuya IoT "Query Things Data Model" API

### DIFF
--- a/custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
+++ b/custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
@@ -63,36 +63,8 @@ entities:
       - id: 1
         type: boolean
         name: activate
-        optional: true
-        mapping:
-          - dps_val: false
-            constraint: pause
-            conditions:
-              - dps_val: true
-                value: false
-              - dps_val: false
-                value: false
-                hidden: true
-              - dps_val: null
-                value: false
-                hidden: true
-          - dps_val: true
-            constraint: pause
-            conditions:
-              - dps_val: false
-                value: true
-              - dps_val: true
-                value: true
-                hidden: true
-              - dps_val: null
-                value: true
-                hidden: true
-          - dps_val: null
-            value: false
-            hidden: true
       - id: 2
         type: boolean
-        optional: true
         name: pause
         hidden: true
 
--- a/custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
+++ b/custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
@@ -126,9 +126,8 @@ entities:
         name: button
 
   - entity: select
-    name: Cleaning mode
+    translation_key: cleaning_mode
     category: config
-    icon: "mdi:robot-vacuum"
     dps:
       - id: 4
         type: string
@@ -314,7 +313,6 @@ entities:
           max: 100
 
   - entity: sensor
-    name: Battery
     class: battery
     category: diagnostic
     dps:
--- a/custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
+++ b/custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
@@ -1,0 +1,450 @@
+name: Robot vacuum and mop
+products:
+  - id: pi5tljdjcchcn0mr
+    name: L6 Nex
+    manufacturer: Laresar
+    model: L6 NEX
+
+entities:
+  - entity: vacuum
+    dps:
+      - id: 5
+        type: string
+        name: status
+        readonly: true
+        mapping:
+          - dps_val: "charge_done"
+            value: "docked"
+          - dps_val: "charging"
+            value: "docked"
+          - dps_val: "standby"
+            value: "idle"
+          - dps_val: "smart"
+            value: "cleaning"
+          - dps_val: "zone_clean"
+            value: "cleaning"
+          - dps_val: "part_clean"
+            value: "cleaning"
+          - dps_val: "cleaning"
+            value: "cleaning"
+          - dps_val: "paused"
+            value: "paused"
+          - dps_val: "goto_pos"
+            value: "returning"
+          - dps_val: "goto_charge"
+            value: "returning"
+          - dps_val: "pos_arrived"
+            value: "idle"
+          - dps_val: "sleep"
+            value: "idle"
+          - dps_val: "pos_unarrive"
+            value: "error"
+      - id: 8
+        type: integer
+        name: battery_level
+        readonly: true
+        unit: "%"
+
+  - entity: button
+    name: Start cleaning
+    category: config
+    icon: "mdi:play"
+    dps:
+      - id: 1
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Pause
+    category: config
+    icon: "mdi:pause"
+    dps:
+      - id: 2
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Return to dock
+    category: config
+    icon: "mdi:home"
+    dps:
+      - id: 3
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Find robot
+    category: config
+    icon: "mdi:map-marker"
+    dps:
+      - id: 11
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Collect dust now
+    category: config
+    icon: "mdi:vacuum"
+    dps:
+      - id: 38
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Reset edge brush
+    category: config
+    icon: "mdi:brush"
+    dps:
+      - id: 18
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Reset roll brush
+    category: config
+    icon: "mdi:brush"
+    dps:
+      - id: 20
+        type: boolean
+        name: button
+
+  - entity: button
+    translation_key: filter_reset
+    category: config
+    dps:
+      - id: 22
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Reset map
+    category: config
+    icon: "mdi:map-outline"
+    dps:
+      - id: 13
+        type: boolean
+        name: button
+
+  - entity: select
+    name: Cleaning mode
+    category: config
+    icon: "mdi:robot-vacuum"
+    dps:
+      - id: 4
+        type: string
+        name: option
+        mapping:
+          - dps_val: "smart"
+            value: "Auto"
+          - dps_val: "chargego"
+            value: "Return to charge"
+          - dps_val: "zone"
+            value: "Zone clean"
+          - dps_val: "pose"
+            value: "Spot clean"
+          - dps_val: "part"
+            value: "Room clean"
+
+  - entity: select
+    name: Suction power
+    category: config
+    icon: "mdi:fan"
+    dps:
+      - id: 9
+        type: string
+        name: option
+        mapping:
+          - dps_val: "Close"
+            value: "Off"
+          - dps_val: "gentle"
+            value: "Eco"
+          - dps_val: "normal"
+            value: "Standard"
+          - dps_val: "strong"
+            value: "Strong"
+
+  - entity: select
+    name: Water tank level
+    category: config
+    icon: "mdi:water"
+    dps:
+      - id: 10
+        type: string
+        name: option
+        mapping:
+          - dps_val: "low"
+            value: "Low"
+          - dps_val: "middle"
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+    manufacturer: Laresar
+    model: L6 NEX
+
+entities:
+  - entity: vacuum
+    dps:
+      - id: 5
+        type: string
+        name: status
+        readonly: true
```
make-all, line 14: readonly not needed for attributes that are read-only by definition in the HA API

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+      - id: 3
+        type: boolean
+        name: button
+
+  - entity: button
+    name: Find robot
+    category: config
+    icon: "mdi:map-marker"
+    dps:
+      - id: 11
```
make-all: Would be better to use the `locate` function of vacuum.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
+name: Robot vacuum and mop
+products:
+  - id: pi5tljdjcchcn0mr
+    name: L6 Nex
+    manufacturer: Laresar
+    model: L6 NEX
+
+entities:
+  - entity: vacuum
```
make-all, line 9: vacuum should generally have a "command" dp defined, with the basic vacuum commands, to allow vacuum specific cards to work.

Since this device uses button style booleans for start, pause and dock, it may require some clever logic to make this work.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+            value: "returning"
+          - dps_val: "pos_arrived"
+            value: "idle"
+          - dps_val: "sleep"
+            value: "idle"
+          - dps_val: "pos_unarrive"
+            value: "error"
+      - id: 8
+        type: integer
+        name: battery_level
```
make-all: Battery level was deprecated in [HA 2023.8](https://developers.home-assistant.io/blog/2023/07/10/vacuum-updates/), then forgotten about and deprecated again in [HA 2025.8](https://developers.home-assistant.io/blog/2025/07/02/vacuum-battery-properties-deprecated/) with a removal date of August 2026.

A separate sensor should be used instead.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+            value: "Return to charge"
+          - dps_val: "zone"
+            value: "Zone clean"
+          - dps_val: "pose"
+            value: "Spot clean"
+          - dps_val: "part"
+            value: "Room clean"
+
+  - entity: select
+    name: Suction power
```
make-all: This would be better as the `fan_speed` feature of the vacuum entity

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+            value: "Off"
+          - dps_val: "gentle"
+            value: "Eco"
+          - dps_val: "normal"
+            value: "Standard"
+          - dps_val: "strong"
+            value: "Strong"
+
+  - entity: select
+    name: Water tank level
```
make-all: This should be a sensor rather than a select if the name is accurate.

Or it should be `translation_key: mopping` if it is a select to control the rate of water dispensing, with values `"off"`, `low`, `medium`, `high` (lower case to match translation keys)

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+            value: "Low"
+          - dps_val: "middle"
+            value: "Medium"
+          - dps_val: "high"
+            value: "High"
+          - dps_val: "closed"
+            value: "Closed"
+
+  - entity: select
+    name: Work mode
```
make-all: Use `translation_key: cleaning_mode` instead of name and icon.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+    category: config
+    icon: "mdi:cog"
+    dps:
+      - id: 41
+        type: string
+        name: option
+        mapping:
+          - dps_val: "both_work"
```
make-all, line 168: Values for `cleaning_mode` should be `sweep`, `mop`, `sweep_and_mop`

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+          max: 100
+
+  - entity: sensor
+    class: battery
+    category: diagnostic
+    dps:
+      - id: 8
+        type: integer
+        name: sensor
+        readonly: true
```
make-all: readonly not required for sensors, as they are inplicitly read-only

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+    category: diagnostic
+    dps:
+      - id: 31
+        type: integer
+        name: sensor
+        unit: "min"
+        readonly: true
+
+  - entity: sensor
+    name: Mop state
```
make-all, line 396: Can be used as an `available` attribute on other mopping related entities to disable them dynamically, or as a filter for select entities to dynamically remove mopping functions from the options. Does not need to be a sensor on its own, as it is only useful for changing function availability.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+        name: sensor
+        readonly: true
+        mapping:
+          - dps_val: "installed"
+            value: "Installed"
+          - dps_val: "none"
+            value: "Not installed"
+
+  - entity: binary_sensor
+    name: Fault status
```
make-all: Name not needed

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
```diff
@@ -0,0 +1,448 @@
...
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 28
```
make-all, line 414: A copy of the raw value should be kept as a `fault_code` attribute. See other devices for examples, 90% of devices have a problem sensor.

make-all (changes requested):

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — removed `readonly: true` from the status dp.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — `locate` is now a dp within the vacuum entity (DP11).

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — removed `battery_level` from the vacuum entity. Battery is now only exposed as a standalone `sensor` with `class: battery`.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — moved suction power to `fan_speed` dp within the vacuum entity (DP9).

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — using `translation_key: mopping` (DP10). Values mapped to `off`, `low`, `medium`, `high`.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — `translation_key: cleaning_mode` applied to DP41 (work mode), not DP4 (cleaning pattern).

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — mapped to `sweep`, `mop`, and `sweep_and_mop`.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — removed `readonly: true` from all sensor entities.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — removed the `name:` field from the binary_sensor entity.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — added a second reference to DP28 with `name: fault_code` to expose the raw bitfield value as an attribute. Confirmed working in HA after restart.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed. This device uses separate boolean DPs (DP1=start, DP2=pause, DP3=return to dock) rather than a single string command DP.

Solution:
- Added DP4 (`chargego`) as the `command` dp with `return_to_base` mapping — DP4 is already referenced by the Cleaning pattern select, so both coexist on the same physical DP.
- Added DP1 as `activate` (boolean) and DP2 as `pause` with `hidden: true`.
- The Pattern C constraint mapping from ilife_v20 was tried but the L6 NEX reports its paused state directly via DP5 status, so the constraint disambiguation is unnecessary. Simple boolean activate/pause works correctly.
- Removed the now-redundant Start cleaning, Pause, and Return to dock button entities.

Validated in HA: vacuum card Start, Pause, and Return to base buttons all work. `supported_features` includes START, PAUSE, and RETURN_HOME.

custom_components/tuya_local/devices/l6_nex_robot_vacuum.yaml
h0me5k1n: Fixed — added DP40 (mop installed/none) as an `available` dp to both the mopping select (DP10) and the mop life sensor (DP23). Both entities now disable dynamically when no mop pad is attached. The mop state sensor is retained as a standalone diagnostic entity.

Validated in HA: mopping select shows as available when mop is installed.

PR #5422, opened 2026-06-27
feat(devices): add Rojeco PAF-186B 2L single pet feeder

eigger:
## Summary
- Add device config for the Rojeco PAF-186B 2L single-bowl pet feeder (product id: `xatwi1ep0ntk551r`)
- Supports portion feed buttons (1, 2, 3, 5 portions), custom feed amount (1–40 portions), status sensor, and feed report

## Test plan
- [x] `uv run yamllint custom_components/tuya_local/devices/rojeco_pet_feeder_paf186b_2l_single.yaml`
- [x] `uv run pytest tests/test_device_config.py`
- [x] Verify feed buttons and status sensor with physical device

Made with [Cursor](https://cursor.com)

### DIFF
--- a/custom_components/tuya_local/devices/rojeco_pet_feeder_paf186b_2l_single.yaml
+++ b/custom_components/tuya_local/devices/rojeco_pet_feeder_paf186b_2l_single.yaml
@@ -4,6 +4,18 @@ products:
     manufacturer: Rojeco
     model: PAF-186B
 entities:
+  - entity: number
+    translation_key: manual_feed
+    category: config
+    dps:
+      - id: 3
+        type: integer
+        optional: true
+        name: value
+        unit: portions
+        range:
+          min: 1
+          max: 40
   - entity: text
     translation_key: meal_plan
     category: config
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/rojeco_pet_feeder_paf186b_2l_single.yaml
```diff
@@ -0,0 +1,112 @@
...
+  - entity: text
+    translation_key: meal_plan
+    category: config
+    hidden: true
+    dps:
+      - id: 1
+        type: base64
+        name: value
+        optional: true
+  - entity: button
```
make-all: This should use the "Manual feed" number entity like every other pet feeder.

If you want to add individual buttons for different feed amounts, you can use helpers to do that, but it adds clutter to force that onto everyone.

custom_components/tuya_local/devices/rojeco_pet_feeder_paf186b_2l_single.yaml
```diff
@@ -0,0 +1,112 @@
...
+      - id: 3
+        type: integer
+        name: button
+        optional: true
+        persist: false
+        mapping:
+          - dps_val: 5
+            value: true
+  - entity: number
+    name: Custom feed amount
```
make-all: Use `translation_key: manual_feed` instead of this non-standard name

make-all (changes requested):

custom_components/tuya_local/devices/rojeco_pet_feeder_paf186b_2l_single.yaml
eigger: Does the number entity, which used to trigger only when the quantity changed, now trigger immediately when using a helper?

make-all: I think it triggers if you set it via a script or automation. It is only the UI that needs it to change before sending anything.

PR #5428, opened 2026-06-27
Add Pioneer ECOasis 150 ERV

ptparker:
Adds support for the Pioneer ECOasis 150 energy recovery ventilator (Tuya product id `jqjrmutjj0zfhjzr`). No existing config matched its DPS, so this is a new device.

Entities: fan (on/off + preset modes), supply/exhaust fan speed numbers (0-8), PM2.5 / eCO2 / TVOC / indoor humidity / indoor temperature sensors, filter reset button, and fault-code / filter-status diagnostics.

Notes:
- The `manua` preset value is the device firmware's actual raw value (confirmed from the raw Tuya IoT portal JSON), not a typo — noted in a comment.
- Filter status codes 0-3 are undocumented, so they are exposed as raw enum values.

Mapped against my own unit and in daily use. Validated locally: `yamllint` clean and `pytest tests/test_device_config.py` passes.

### DIFF
--- a/custom_components/tuya_local/devices/pioneer_ecoasis_150_erv.yaml
+++ b/custom_components/tuya_local/devices/pioneer_ecoasis_150_erv.yaml
@@ -12,7 +12,7 @@ entities:
   #################################################################
 
   - entity: fan
-    name: ERV
+    translation_only_key: fan_with_presets
     dps:
       - id: 1
         type: boolean
@@ -25,17 +25,17 @@ entities:
           # Device firmware reports the truncated value "manua" (confirmed
           # from the raw Tuya IoT portal JSON), not "manual".
           - dps_val: manua
-            value: Manual
+            value: manual
           - dps_val: auto
-            value: Auto
+            value: auto
           - dps_val: sleep
-            value: Sleep
+            value: sleep
           - dps_val: "4"
-            value: Pure L
+            value: low
           - dps_val: "5"
-            value: Pure M
+            value: medium
           - dps_val: "6"
-            value: Pure H
+            value: high
 
   - entity: number
     name: Supply fan speed
@@ -66,9 +66,7 @@ entities:
   #################################################################
 
   - entity: sensor
-    name: PM2.5
     class: pm25
-    icon: "mdi:molecule"
     dps:
       - id: 3
         type: integer
@@ -77,9 +75,7 @@ entities:
         class: measurement
 
   - entity: sensor
-    name: eCO2
     class: carbon_dioxide
-    icon: "mdi:molecule-co2"
     dps:
       - id: 6
         type: integer
@@ -88,9 +84,7 @@ entities:
         class: measurement
 
   - entity: sensor
-    name: TVOC
     class: volatile_organic_compounds_parts
-    icon: "mdi:molecule"
     dps:
       - id: 7
         type: integer
@@ -99,9 +93,7 @@ entities:
         class: measurement
 
   - entity: sensor
-    name: Indoor humidity
     class: humidity
-    icon: "mdi:water-percent"
     dps:
       - id: 8
         type: integer
@@ -110,9 +102,7 @@ entities:
         class: measurement
 
   - entity: sensor
-    name: Indoor temperature
     class: temperature
-    icon: "mdi:thermometer"
     dps:
       - id: 9
         type: integer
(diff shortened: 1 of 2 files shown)


ptparker:
updated it, the test passes locally.

custom_components/tuya_local/devices/pioneer_ecoasis_150_erv.yaml
```diff
@@ -0,0 +1,163 @@
...
+    model: ECOasis 150
+    name: Pioneer ECOasis 150 ERV
+
+entities:
+  #################################################################
+  # Controls
+  #################################################################
+
+  - entity: fan
+    name: ERV
```
make-all: Use `translation_key: fan_with_presets` to use translated preset modes `manual`, `auto`, `sleep`, `low`, `medium`, `high` below.

custom_components/tuya_local/devices/pioneer_ecoasis_150_erv.yaml
```diff
@@ -0,0 +1,163 @@
...
+        range:
+          min: 0
+          max: 8
+
+  #################################################################
+  # Sensors
+  #################################################################
+
+  - entity: sensor
+    name: PM2.5
```
make-all: name and icon not required, default HA class translations should be used instead.

custom_components/tuya_local/devices/pioneer_ecoasis_150_erv.yaml
```diff
@@ -0,0 +1,163 @@
...
+    icon: "mdi:molecule"
+    dps:
+      - id: 3
+        type: integer
+        name: sensor
+        unit: "µg/m³"
+        class: measurement
+
+  - entity: sensor
+    name: eCO2
```
make-all: Likewise

custom_components/tuya_local/devices/pioneer_ecoasis_150_erv.yaml
```diff
@@ -0,0 +1,163 @@
...
+    icon: "mdi:molecule-co2"
+    dps:
+      - id: 6
+        type: integer
+        name: sensor
+        unit: ppm
+        class: measurement
+
+  - entity: sensor
+    name: TVOC
```
make-all: Likewise

custom_components/tuya_local/devices/pioneer_ecoasis_150_erv.yaml
```diff
@@ -0,0 +1,163 @@
...
+    translation_key: filter_reset
+    category: diagnostic
+    icon: "mdi:filter-remove"
+    dps:
+      - id: 13
+        type: boolean
+        name: button
+
+  - entity: sensor
+    name: Fault code
```
make-all: Standard pattern for fault sensors is to use a binary_sensor (class: problem) to indicate any fault, with an attribute `fault_code` to retain the original raw value.

custom_components/tuya_local/devices/pioneer_ecoasis_150_erv.yaml
```diff
@@ -0,0 +1,163 @@
...
+  - entity: sensor
+    name: Filter status
+    class: enum
+    category: diagnostic
+    icon: "mdi:filter-check"
+    dps:
+      - id: 105
+        type: string
+        name: sensor
+        # Status codes 0-3 are undocumented; exposed as raw values.
```
make-all: In that case, this should probably just be an attribute `filter_status` on the problem binary_sensor, as displaying raw data in an entity like this will only lead to future incompatible changes as the meaning of these values is worked out.

custom_components/tuya_local/devices/pioneer_ecoasis_150_erv.yaml
```diff
@@ -0,0 +1,163 @@
...
+          - dps_val: sleep
+            value: Sleep
+          - dps_val: "4"
+            value: Pure L
+          - dps_val: "5"
+            value: Pure M
+          - dps_val: "6"
+            value: Pure H
+
+  - entity: number
```
make-all: Probably two fans is a better representation here, though you need to decide which the preset_mode is more naturally associated with, or duplicate it along with the switch on both.

make-all (changes requested):

ptparker:
Thanks for the review — pushed an update addressing the comments:

- **Presets**: switched to the `fan_with_presets` translation key with translated preset values (`manual`/`auto`/`sleep`/`low`/`medium`/`high`).
- **Sensors**: dropped the redundant `name`/`icon` from the device-class sensors so the default HA class translations are used.
- **Fault + filter status**: replaced both with a single `binary_sensor` (`class: problem`) that indicates any fault, keeping the raw values as `fault_code` and `filter_status` attributes. The filter status codes (0–3) are undocumented, so exposing them only as a raw attribute avoids baking in a meaning that may change.

On the two-fans suggestion, I'd argue for keeping one `fan` + two `number`s. This is an ERV, not two independent fans: the meaningful value is the supply/exhaust **delta**, which is how you set intentional positive/negative house pressure. Also, DP101/102 are **only writable in manual mode** — read-only in auto/sleep/pure — so `fan` entities would imply always-settable speeds the device doesn't allow. And there's one power switch (DP1) and one mode (DP2), so two fans would have to fake-share them. Open to it if you feel strongly, but I think this maps the hardware more honestly.

make-all:
The device is not exposing its settings as a delta, it is exposing them as two fans with settable speeds. There is no issue with putting the power on both, or with chosing one as primary (supply probably) and only having a speed for exhaust.

Even as numbers, the fact that they are only value in manual mode is not obvious from the config, so I don't think making the fans into fan entities makes any difference there.

Issue #239, opened 2022-09-24
Air Conditioner Recognized as Switch


Aptul9:
Hi, I recently tried setting up my air conditioner to use locally. The problem is that it gets recognized as a switch after I provide all the details. I have a Sendo air conditioner, I will attach all the device information below:

> {
>   "result": {
>     "category": "kt",
>     "functions": [
>       {
>         "code": "switch",
>         "desc": "{}",
>         "name": "开关",
>         "type": "Boolean",
>         "values": "{}"
>       },
>       {
>         "code": "temp_set",
>         "desc": "{\"unit\":\"℃/F\",\"min\":160,\"max\":900,\"scale\":1,\"step\":10}",
>         "name": "设置温度",
>         "type": "Integer",
>         "values": "{\"unit\":\"℃/F\",\"min\":160,\"max\":900,\"scale\":1,\"step\":10}"
>       },
>       {
>         "code": "mode",
>         "desc": "{\"range\":[\"cold\",\"hot\",\"wet\",\"wind\",\"auto\"]}",
>         "name": "工作模式",
>         "type": "Enum",
>         "values": "{\"range\":[\"cold\",\"hot\",\"wet\",\"wind\",\"auto\"]}"
>       },
>       {
>         "code": "mode_eco",
>         "desc": "{}",
>         "name": "ECO模式",
>         "type": "Boolean",
>         "values": "{}"
>       },
>       {
>         "code": "heat",
>         "desc": "{}",
>         "name": "辅热",
>         "type": "Boolean",
>         "values": "{}"
>       },
>       {
>         "code": "light",
>         "desc": "{}",
>         "name": "灯光",
>         "type": "Boolean",
>         "values": "{}"
>       },
>       {
>         "code": "lock",
>         "desc": "{}",
>         "name": "童锁",
>         "type": "Boolean",
>         "values": "{}"
>       },
>       {
>         "code": "switch_horizontal",
>         "desc": "{}",
>         "name": "左右摆风开关",
>         "type": "Boolean",
>         "values": "{}"
>       },
>       {
>         "code": "sleep",
>         "desc": "{}",
>         "name": "睡眠功能",
>         "type": "Boolean",
>         "values": "{}"
>       },
>       {
>         "code": "health",
>         "desc": "{}",
>         "name": "健康",
>         "type": "Boolean",
>         "values": "{}"
>       }
>     ],
>     "status": [
>       {
>         "code": "switch",
>         "name": "开关",
>         "type": "Boolean",
[...]

PR #5434, opened 2026-06-28
Add fan_mode and vertical swing to Sendo air conditioner

Chrispikaan:
The existing sendo_airconditioner_c.yaml was missing two features that are available in the SmartLife app:

Fan speed control (DP 5, windspeed): added fan_mode with values mute, low, mid, high and auto
Vertical swing (DP 105, up_down_wind): added swing_mode as a boolean (on/off)
Also renamed the existing DP 33 from swing_mode to swing_horizontal_mode to avoid conflict

Tested on a Sendo Aion model.

### DIFF
--- a/custom_components/tuya_local/devices/sendo_airconditioner_c.yaml
+++ b/custom_components/tuya_local/devices/sendo_airconditioner_c.yaml
@@ -30,7 +30,6 @@ entities:
         range:
           min: 160
           max: 320
-          # for F, 600 to 900
         unit: C
       - id: 3
         name: current_temperature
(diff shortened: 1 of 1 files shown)


make-all:
I'm not sure these are the same devices, since these dps do not appear in the original request #239

Chrispikaan:
@make-all Thanks for the clarification. Would it be possible to add the Sendo Aion as a separate device? I have a working YAML configuration that I tested on my own unit.

make-all:
Yes, if no other device is matching, then adding sendo_aion_airconditioner.yaml would be the solution

Chrispikaan:
@make-all Great, I already have the working yaml that I created and tested on my Sendo Aion added on my own HomeAssistant server. Would you be able to add it to the integration so other users can benefit from it as well?

PR #5442, opened 2026-06-29
Add Olight Ostation 2

ProZsolt:
Adds a device configuration YAML for the Olight Ostation 2 / Ostation 2 Pro battery charger.


custom_components/tuya_local/devices/olight_ostation_2_battery_charger.yaml
```diff
@@ -0,0 +1,489 @@
...
+      - id: 110
+        type: base64
+        name: sensor
+        optional: true
+        mask: "00FF000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
+        mapping:
+          - dps_val: 0
+            value: Detecting
+          - dps_val: 1
+            value: Ni-MH
```
ProZsolt, line 127: The mapping is not working. Without mapping it returns `0` and `1`, but with this mapping it returns `Unknown` in both cases.

Is it a bug or I'm doing something wrong?

custom_components/tuya_local/devices/olight_ostation_2_battery_charger.yaml
```diff
@@ -0,0 +1,489 @@
...
+    name: AA slot 1 charge time min
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 110
+        type: base64
+        name: sensor
+        unit: min
+        optional: true
+        mask: "00000000FF000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
ProZsolt: I want to combine these two into one duration sensor. The first byte contains the hours the second byte contains the minutes. I want to sum it up. Can we do some clever thing with mapping or do I need to develop something to handle this.

custom_components/tuya_local/devices/olight_ostation_2_battery_charger.yaml
ProZsolt: Changing the sensor to an enum fixed it

ProZsolt:
The tests should pass now

make-all:
Supercedes the stalled #4970

custom_components/tuya_local/devices/olight_ostation_2_battery_charger.yaml
```diff
@@ -0,0 +1,588 @@
...
+products:
+  - id: 1fnensm6
+    manufacturer: Olight
+    model: Ostation 2
+  - id: v8znljza
+    manufacturer: Olight
+    model: Ostation 2
+entities:
+  - entity: switch
+    name: Power
```
make-all: Name not needed here

custom_components/tuya_local/devices/olight_ostation_2_battery_charger.yaml
```diff
@@ -0,0 +1,588 @@
...
+        name: sensor
+  - entity: sensor
+    name: Invalid battery count
+    category: diagnostic
+    dps:
+      - id: 108
+        type: integer
+        name: sensor
+  - entity: sensor
+    name: DP113
```
make-all: Unknown ids should not be top level entities, just make them attributes of another entity so they are available for observation, but not needing migration when the real purpose is discovered.

make-all (changes requested):

PR #5444, opened 2026-06-29
Add support for Point PODH20 dehumidifier

viljasenville:
This PR adds a new device configuration file for Point PODH20 dehumidifier. Datapoints are captured from integration logs and Tuya developer platform.

### DIFF
--- a/custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
@@ -5,7 +5,6 @@ products:
     model: PODH20
 entities:
   - entity: humidifier
-    translation_key: dehumidifier
     class: dehumidifier
     dps:
       - id: 1
@@ -49,6 +48,9 @@ entities:
             value: "medium"
           - dps_val: "High"
             value: "high"
+      - id: 102
+        name: oscillate
+        type: boolean
   - entity: binary_sensor
     translation_key: tank_full
     class: problem
@@ -75,15 +77,7 @@ entities:
       - id: 101
         type: boolean
         name: switch
-  - entity: switch
-    name: Oscillate
-    category: config
-    dps:
-      - id: 102
-        type: boolean
-        name: switch
   - entity: sensor
-    translation_key: current_temperature
     class: temperature
     category: diagnostic
     dps:
@@ -91,6 +85,7 @@ entities:
         type: integer
         name: sensor
         unit: °C
+        class: measurement
   - entity: switch
     translation_key: sleep
     category: config
--- a/custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
@@ -84,7 +84,7 @@ entities:
       - id: 103
         type: integer
         name: sensor
-        unit: °C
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
```diff
@@ -0,0 +1,95 @@
+name: Dehumidifier
+products:
+  - id: l42hyrkieezkgysd
+    manufacturer: Point
+    model: PODH20
+entities:
+  - entity: humidifier
+    class: dehumidifier
```
make-all, line 8: Add `translation_key: extended` to get additional modes

custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
```diff
@@ -0,0 +1,95 @@
...
+        name: current_humidity
+        readonly: true
+      - id: 5
+        type: string
+        name: mode
+        mapping:
+          - dps_val: "Dry"
+            value: "normal"
+          - dps_val: "Fan"
+            value: "Fan"
```
make-all: I normally use `purify` for fan only modes, as normally the reason to use such a mode on a dehumidifier is to run the air through the filter.

custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
```diff
@@ -0,0 +1,95 @@
...
+            value: "auto"
+  - entity: fan
+    translation_key: fan_with_presets
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 4
+        type: string
+        name: preset_mode
```
make-all: Preference is to use speed before using preset_mode, where the modes are clearly a progression from low to high (no auto or others complicating things)

custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
```diff
@@ -0,0 +1,95 @@
...
+          - value: false
+  - entity: lock
+    translation_key: child_lock
+    category: config
+    dps:
+      - id: 16
+        type: boolean
+        name: lock
+  - entity: switch
+    name: Light
```
make-all: Use a light entity rather than a switch called light.

This may be a candidate for `translation_key: indicator` or similar.

custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
```diff
@@ -0,0 +1,95 @@
...
+    category: config
+    dps:
+      - id: 101
+        type: boolean
+        name: switch
+  - entity: sensor
+    class: temperature
+    category: diagnostic
```
make-all: No need for diagnostic here, sensors are already in their own category in recent versions of HA.

make-all (changes requested):

custom_components/tuya_local/devices/point_podh20_dehumidifier.yaml
viljasenville: If I understood correctly, this should be fixed now.

viljasenville:
@make-all I updated the PR for another review round. Thanks for comments!

PR #5452, opened 2026-06-30
Add Kibernetik 12K portable split air conditioner

mco85:
## Summary

Adds a device config for the **Kibernetik 12K BTU portable split air conditioner** (Tuya product id `deedvplnx2yvge08`, category `kt`, product name "Portable Split Air Conditioner"). Sold in Switzerland via Fust (art. 111235), controlled through the Smart Life app.

## Why a new config

This device was being matched (by dps fallback, no product-id match) to the existing `kogan_kawfpac12ya` config, which maps the fan (dps 5) as a **3-speed** device (low/medium/high). The Kibernetik is a **5-speed** unit — the Tuya cloud schema reports dps 5 enum range `['1','2','3','4','5']`, and on the 3-speed config steps 4/5 showed up as raw values instead of named fan modes.

## Fan steps verified on real hardware

Selected each level in the Smart Life app and read back the raw dps 5 value via Home Assistant:

| Smart Life | dps 5 | fan_mode |
|------------|:-----:|----------|
| Silent | `1` | `silent` |
| Low    | `2` | `low` |
| Mid    | `3` | `medium` |
| High   | `4` | `high` |
| Turbo  | `5` | `turbo` |

Remaining datapoints mirror the Kogan PAC_Y12 config, on which this device operates correctly (hvac off/cool/fan_only/dry, target/current temp, preset sleep/comfort, vertical swing, °C/°F select).

## Note on the Kogan Y12

The Kogan Y12 appears to be the same PAC_Y12 hardware and may also be a 5-speed unit with the same fan mislabel, but I could not verify that on a Kogan device, so I left its config unchanged rather than risk breaking it.

## Tests

- `uv run yamllint custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml` — clean

### DIFF
--- a/custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
@@ -22,6 +22,8 @@ entities:
                 value: fan_only
               - dps_val: DRY
                 value: dry
+              - dps_val: HEAT
+                value: heat
       - id: 2
         name: temperature
         type: integer
@@ -68,9 +70,6 @@ entities:
             value: sleep
           - dps_val: false
             value: comfort
-      - id: 102
-        name: unknown_102
-        type: boolean
       - id: 104
         name: swing_mode
         type: boolean
@@ -79,15 +78,9 @@ entities:
             value: "vertical"
           - dps_val: false
             value: "off"
-      - id: 106
-        name: unknown_106
-        type: integer
-      - id: 109
-        name: unknown_109
-        type: boolean
       - id: 110
         name: unknown_110
-        type: integer
+        type: string
   - entity: select
     translation_key: temperature_unit
     category: config
@@ -100,6 +93,13 @@ entities:
             value: celsius
           - dps_val: F
             value: fahrenheit
+  - entity: switch
+    translation_key: auto_dry
+    category: config
+    dps:
+      - id: 109
+        type: boolean
+        name: switch
   - entity: number
     translation_key: timer
     category: config
@@ -121,3 +121,21 @@ entities:
         type: integer
         name: sensor
         unit: min
+  - entity: binary_sensor
+    translation_key: defrost
+    category: diagnostic
+    dps:
+      - id: 102
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 106
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
--- a/custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
@@ -20,10 +20,10 @@ entities:
                 value: cool
               - dps_val: FAN
                 value: fan_only
+              # dp4 also lists HEAT, but this is a cool-only model (see dp110).
+              # Left out until a heat-capable model can confirm it.
               - dps_val: DRY
                 value: dry
-              - dps_val: HEAT
-                value: heat
       - id: 2
         name: temperature
         type: integer
@@ -79,7 +79,10 @@ entities:
           - dps_val: false
             value: "off"
       - id: 110
-        name: unknown_110
+        # "model" per Tuya data model; meaning of values unconfirmed.
+        # This cool-only 5-speed unit reports "0" - a heat-capable or
+        # 3-speed model may report differently (may gate heat/fan speeds).
+        name: model
         type: string
   - entity: select
     translation_key: temperature_unit
--- a/custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
@@ -1,0 +1,123 @@
+name: Air conditioner
+products:
+  - id: deedvplnx2yvge08
+    manufacturer: Kibernetik
+    model: 12K portable split
+entities:
+  - entity: climate
+    translation_only_key: aircon_extra
+    dps:
+      - id: 1
+        name: hvac_mode
+        type: boolean
+        mapping:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
```diff
@@ -0,0 +1,107 @@
...
+      - id: 4
+        name: mode
+        type: string
+        hidden: true
+      - id: 5
+        name: fan_mode
+        type: string
+        mapping:
+          - dps_val: "1"
+            value: silent
```
make-all: silent and turbo are not standard translations.

Suggest using `translation_key: aircon_extra` on the  entity, and use `quiet` and `strong` from the additional translations.

custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
```diff
@@ -0,0 +1,107 @@
...
+      - id: 104
+        name: swing_mode
+        type: boolean
+        mapping:
+          - dps_val: true
+            value: "vertical"
+          - dps_val: false
+            value: "off"
+      - id: 105
+        name: countdown
```
make-all: Normally should be split to a `timer` (time entity if 24h range using second or minute resolution, or number entity otherwise)

make-all (changes requested):

mco85:
Thanks for the review — both addressed in the latest push:

**Fan modes:** switched the climate entity to `translation_only_key: aircon_extra` and mapped dps 5 to the standard extras `quiet` (1) and `strong` (5), keeping low/medium/high.

**Timer/countdown:** verified the semantics on real hardware — set a 1 h timer in the Smart Life app and read back the raw dps:
- `dps 103` ("timer") → `1` — the setting, in **hours**. Modelled as a `number` entity (`translation_key: timer`, `class: duration`, unit `h`, range 0–24), matching e.g. `kogan_vostok`/`duux_north`.
- `dps 105` ("countdown") → `60` — the **remaining** time, in **minutes**, read-only as it counts down. Modelled as a `sensor` entity (`translation_key: time_remaining`, `class: duration`, unit `min`, diagnostic), matching e.g. `igenix`/`fral`.

Both moved off the climate entity. `yamllint` clean and `tests/test_device_config.py` passes (32).

make-all:
I don't think the Kogan config is a mistake, as the manual mentions only 3 fan speeds. But some of the unknown dps may be selecting options like this (and also the presence of heat mode, which would allow merging the 12YA and 16YA configs).

If you can download the Query Things Data Model result for your device from platform.tuya.com, then it might help shed some light on the rest of the dps.

mco85:
Good call — here's the **Query Things Data Model** (`/v2.0/cloud/thing/{id}/model`) for the device, which decodes the rest of the dps:

| dp | code | access | type | range |
|----|------|--------|------|-------|
| 1 | Power | rw | bool | |
| 2 | temp_set | rw | value | 16–86 (raw, spans °C+°F) |
| 3 | temp_current | ro | value | −7–110 |
| **4** | **mode** | rw | enum | `COOL, FAN, DRY, HEAT` |
| 5 | windspeed | rw | enum | `1..5` |
| 19 | temp_unit_convert | rw | enum | `c, f` |
| 101 | SLEEP | rw | bool | |
| 102 | Defrost | ro | bool | |
| 103 | Timer | rw | value | 0–24 (h) |
| 104 | Shake | rw | bool | (swing) |
| 105 | countdown | ro | value | 0–1440 (min) |
| 106 | Alarm | ro | bitmap | |
| 109 | innerdry | rw | bool | |
| 110 | model | ro | enum | `0, 1` |

So **HEAT is in the dp4 enum** — I've added `HEAT → heat` to the hvac mapping. That confirms the merge potential with the 16YA; dp110 (`model`, ro enum `0/1`) also looks like it might distinguish the cool-only vs heat-pump variant, though I can't confirm the mapping. And you're right that the Kogan 3-speed config isn't wrong — this unit is just a 5-speed variant of the same platform.

I've also named and modelled the previously-unknown dps:
- **102 Defrost** → `binary_sensor` (translation_key `defrost`, diagnostic) — same dp/pattern as `ballu_aura`
- **109 innerdry** → `switch` (translation_key `auto_dry`, config)
- **106 Alarm** → `binary_sensor` class `problem` (0 → clear, else fault); I kept it a plain boolean since I don't have the bit definitions

Fair warning on **HEAT**: the schema exposes it, but my unit is cool-focused and I haven't bench-tested heating — happy to drop it if you'd rather keep this config cool-only. `yamllint` clean, `tests/test_device_config.py` passes (32).

make-all:
Since you have identifed the name of dp 110, you should change the attribute name from unknown_110 to model to reflect that, even if you have not identified what the values mean. I suspect it may be selecting whether heat should be available or not if you previously thought it is not available based on your device, but is showing in Tuya info.

Unfortunately the two Kogan airconditioners were both added in the same PR and there were no logs supplied, so we have no datapoints for model to compare with. It will either be differentiating the heat vs cool only models, or it may be differentiating the 5-speed vs 3-speed models. But since we don't know, a new config seems to be the best solution here, with the heat mode left out for now, but a comment showing what your cool-only 5 speed model is reporting for model  would be useful, so that in future if someone has a heat model they can compare and see if the model is distinct.

mco85:
Done — all three:

- **Renamed** dp110 `unknown_110` → `model`.
- **Dropped HEAT** from the dp4 mapping for now (left an inline comment noting the schema lists it, to be confirmed by a heat-capable unit).
- Recorded the observed value: this **cool-only 5-speed** unit reports **`model` = `"0"`** (comment in the config). So if someone later has a heat model or a 3-speed model, they can compare and we'll know whether dp110 distinguishes heat-vs-cool or 5-speed-vs-3-speed.

`yamllint` clean, `tests/test_device_config.py` passes (32).

make-all:
#5715 seems to confirm that model is NOT indicating the presence of heat mode, as it is also "0" there, with the same product id, on a device supporting heat mode.

PR #5453, opened 2026-06-30
feat(devices): add configuration for Globmatic gate opener

clinis:
Adds a device config for the Globmatic Automatic sliding gate opener - Castor.

Entities:
- button: `dp 101`, triggers the motor to toggle state (similar to the behaviour of a remote IR button), by setting the position to end states (0 or 100).
- cover: `dp 101`, allows to see buttons to toggle state and the state of the motor in a nice UI.
- sensor: `dp 101`, exposes the gate position, only values [0, 50, 100], it is a limitation of the physical sensor.
- sensor: `dp 10`, exposes the state the sensor reports.

Notes:
- physical sensor reports position reversed, hence the reversed mapping and the reversed icons

Tested working on real hardware through this config.
Ran locally before submitting: yamllint, pytest tests/test_device_config.py, all pass.

### DIFF
--- a/custom_components/tuya_local/devices/globmatic_gate_opener.yaml
+++ b/custom_components/tuya_local/devices/globmatic_gate_opener.yaml
@@ -1,0 +1,76 @@
+name: Gate opener
+
+products:
+  - id: 4iispvj6a6vsavxl
+    manufacturer: Globmatic
+    model: Automatic sliding gate opener - Castor
+    model_id: bf6b6de074afba4988esep
+
+entities:
+  - entity: button
+    name: "Motor toggle"
+    dps:
+      - id: 101
+        name: button
+        type: integer
+        mapping:
+          - dps_val: 0
+            value: true
+            icon: "mdi:gate-open"
+          - dps_val: 50
+            value: true
+            icon: "mdi:gate-arrow-left"
+          - dps_val: 100
+            value: true
+            icon: "mdi:gate"
+
+  - entity: cover
+    name: "Motor control"
+    class: gate
+    dps:
+      - id: 101
+        name: control
+        type: integer
+        mapping:
+          - dps_val: 0
+            value: open
+          - dps_val: 100
+            value: close
+      - id: 101
+        name: action
+        type: integer
+        mapping:
+          - dps_val: 0
+            value: opened
+          - dps_val: 50
+            value: opening
+          - dps_val: 100
+            value: closed
+      - id: 101
+        name: position
+        type: integer
+        range:
+          min: 0
+          max: 100
+        mapping:
+          - invert: true
+
+  - entity: sensor
+    name: "Gate position"
+    dps:
+      - id: 101
+        name: sensor
+        type: integer
+        range:
+          min: 0
+          max: 100
+        mapping:
+          - invert: true
+
+  - entity: sensor
+    name: "Gate state"
+    dps:
+      - id: 10
+        name: sensor
+        type: string
+        optional: true
--- a/custom_components/tuya_local/devices/globmatic_gate_opener.yaml
+++ b/custom_components/tuya_local/devices/globmatic_gate_opener.yaml
@@ -7,23 +7,6 @@ products:
     model_id: bf6b6de074afba4988esep
 
 entities:
-  - entity: button
-    name: "Motor toggle"
-    dps:
-      - id: 101
-        name: button
-        type: integer
-        mapping:
-          - dps_val: 0
-            value: true
-            icon: "mdi:gate-open"
-          - dps_val: 50
-            value: true
-            icon: "mdi:gate-arrow-left"
-          - dps_val: 100
-            value: true
-            icon: "mdi:gate"
-
   - entity: cover
     name: "Motor control"
     class: gate
@@ -48,18 +31,6 @@ entities:
             value: closed
       - id: 101
         name: position
-        type: integer
-        range:
-          min: 0
-          max: 100
-        mapping:
-          - invert: true
-
-  - entity: sensor
-    name: "Gate position"
-    dps:
-      - id: 101
-        name: sensor
         type: integer
         range:
           min: 0
--- a/custom_components/tuya_local/devices/globmatic_gate_opener.yaml
+++ b/custom_components/tuya_local/devices/globmatic_gate_opener.yaml
@@ -7,7 +7,6 @@ products:
 
 entities:
   - entity: cover
-    name: "Motor control"
     class: gate
     dps:
       - id: 101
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/globmatic_gate_opener.yaml
```diff
@@ -0,0 +1,76 @@
...
+    model_id: bf6b6de074afba4988esep
+
+entities:
+  - entity: button
+    name: "Motor toggle"
+    dps:
+      - id: 101
+        name: button
+        type: integer
+        mapping:
```
make-all: This mapping will not lead to deterministic behaviour. This button should be removed. The cover entity should be enough to control the device

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
```diff
@@ -0,0 +1,76 @@
...
+          - dps_val: 100
+            value: close
+      - id: 101
+        name: action
+        type: integer
+        mapping:
+          - dps_val: 0
+            value: opened
+          - dps_val: 50
+            value: opening
```
make-all: Does this really show only when the gate is opening? Not when it is closing, or when it is stopped partially open?

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
```diff
@@ -0,0 +1,76 @@
...
+        name: position
+        type: integer
+        range:
+          min: 0
+          max: 100
+        mapping:
+          - invert: true
+
+  - entity: sensor
+    name: "Gate position"
```
make-all: This is not necessary as a sensor. The position attribute of the cover can be used instead.

make-all (changes requested):

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
clinis: Yes, this was a sensor added more for debugging, can be removed

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
clinis: The sensor reports value 50 (and state `opening_closing`) when it is opening, closing, or stopped mid-way.

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
clinis: The goal of this button was to mimic the physical button on a RF command. It triggers the motor, does not have direction, is just a button that toggles direction. It is simpler than the HA cover UI, because it shows two buttons if I want to change direction. 

But I can remove it

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
```diff
@@ -0,0 +1,47 @@
+name: Gate opener
+
+products:
+  - id: 4iispvj6a6vsavxl
+    manufacturer: Globmatic
+    model: Automatic sliding gate opener - Castor
+    model_id: bf6b6de074afba4988esep
```
make-all: This is your device id, unique to your device. There is no use to include it in a config file that will be distributed to all users.

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+entities:
+  - entity: cover
+    name: "Motor control"
```
make-all: I suggest to drop the name and let HA translations for gate device class be used

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
make-all: So you should map it to null, as the state "opening" is misleading in 2 of 3 cases.

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
clinis: Removed

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
clinis: Ok with removing the custom entity names. Updated PR

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
clinis: Removed the misleading mapping of the `50` value

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - id: 101
+        name: position
+        type: integer
+        range:
+          min: 0
+          max: 100
+        mapping:
+          - invert: true
+
+  - entity: sensor
```
make-all, line 29: What is being sensed here? A sensor needs a class, translation_key or name, really. Unlike some other entity types it does not make sense to have an unnamed sensor.

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
```diff
@@ -0,0 +1,44 @@
...
+        name: action
+        type: integer
+        mapping:
+          - dps_val: 0
+            value: opened
+          - dps_val: 50
+            value: null
+          - dps_val: 100
+            value: closed
+      - id: 101
```
make-all, line 20: I think only the position is needed here. The other attributes are just derived from position, and the cover should be able to do that without explicitly defining them.

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
make-all: I see where the confusion with the unnamed sensor may have come from. My suggestion was specific to the cover entity, which has a class already, clarifying that it is a gate.

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
clinis: This maps the status reported by the sensor, which has values `closed` | `opening_closing` | `opened`.
The change to be `translation_key: status` is ok

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
clinis: The new mapping works. But we lose the ability to use the "button shortcuts" that HA provides.
Image
I can open a follow-up PR to add the missing mapping that enables this functionality

custom_components/tuya_local/devices/globmatic_gate_opener.yaml
make-all: Yes, I realised it was in the history from the first commit you made, and it seemed close to the existing translation key `status`

PR #5454, opened 2026-06-30
Add configuration for CAST LB500C Litter Box

Thenas:
#### Adds device config for the CAST LB500C self-cleaning cat litter box.
Product ID: frho2xmkraxldtwo
Entities:
- Cat weight sensor (kg)
- Daily toilet count sensor
- Toilet duration sensor
- Today's toilet count sensor
- Status sensor (standby/leveling/cleaning)
- Notification binary sensor (no weight detected)
- Fault binary sensor (drum not installed, overload)
- Auto clean switch
- Clean now / Cancel buttons
- Cleaning delay, valid toilet time, drum rotations, min cat weight selects

### DIFF
--- a/custom_components/tuya_local/devices/cast_lb500c_litterbox.yaml
+++ b/custom_components/tuya_local/devices/cast_lb500c_litterbox.yaml
@@ -1,0 +1,206 @@
+name: Litter box
+products:
+  - id: frho2xmkraxldtwo
+    manufacturer: CAST
+    model: LB500C
+entities:
+  - entity: sensor
+    class: weight
+    dps:
+      - id: 6
+        type: integer
+        optional: true
+        name: sensor
+        unit: kg
+        class: measurement
+        precision: 1
+        mapping:
+          - scale: 0.453592
+  - entity: sensor
+    name: Daily toilet count
+    icon: "mdi:emoticon-poop"
+    dps:
+      - id: 7
+        type: integer
+        optional: true
+        name: sensor
+        unit: times
+        class: measurement
+  - entity: sensor
+    name: Toilet duration
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 110
+        type: integer
+        optional: true
+        name: sensor
+        unit: s
+        class: measurement
+  - entity: sensor
+    name: Today's toilet count
+    category: diagnostic
+    dps:
+      - id: 108
+        type: integer
+        optional: true
+        name: sensor
+        unit: times
+        class: measurement
+  - entity: sensor
+    translation_key: status
+    class: enum
+    category: diagnostic
+    dps:
+      - id: 101
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: isidle
+            value: standby
+          - dps_val: idlevelling
+            value: caking
+          - dps_val: isclean
+            value: cleaning
+  - entity: binary_sensor
+    name: Notification
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 21
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 21
+        type: bitfield
+        name: fault_code
+      - id: 21
+        type: bitfield
+        name: description
+        mapping:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/cast_lb500c_litterbox.yaml
```diff
@@ -0,0 +1,206 @@
...
+    dps:
+      - id: 6
+        type: integer
+        optional: true
+        name: sensor
+        unit: kg
+        class: measurement
+        precision: 1
+        mapping:
+          - scale: 0.453592
```
make-all: It seems the unit is lb. Probably it is better to declare that and let HA do the scaling. Otherwise depending on user settings it may lead to double conversion with rounding error introduced.

custom_components/tuya_local/devices/cast_lb500c_litterbox.yaml
```diff
@@ -0,0 +1,206 @@
...
+    category: diagnostic
+    dps:
+      - id: 110
+        type: integer
+        optional: true
+        name: sensor
+        unit: s
+        class: measurement
+  - entity: sensor
+    name: Today's toilet count
```
make-all: Try to avoid punctuation in entity names, as it can lead to problems if users export their entities to other systems.

custom_components/tuya_local/devices/cast_lb500c_litterbox.yaml
```diff
@@ -0,0 +1,206 @@
...
+        name: sensor
+        mapping:
+          - dps_val: isidle
+            value: standby
+          - dps_val: idlevelling
+            value: caking
+          - dps_val: isclean
+            value: cleaning
+  - entity: binary_sensor
+    name: Notification
```
make-all: May be better as an event if it is a notification

custom_components/tuya_local/devices/cast_lb500c_litterbox.yaml
```diff
@@ -0,0 +1,206 @@
...
+      - id: 21
+        type: bitfield
+        name: description
+        mapping:
+          - dps_val: 0
+            value: ok
+          - dps_val: 1
+            value: no_weight
+  - entity: binary_sensor
+    name: Fault
```
make-all, line 44: Name not required. It is better to let the class translation be used

custom_components/tuya_local/devices/cast_lb500c_litterbox.yaml
```diff
@@ -0,0 +1,206 @@
...
+    category: config
+    dps:
+      - id: 109
+        type: boolean
+        optional: true
+        name: button
+        mapping:
+          - value: "nowtocancle"
+  - entity: select
+    name: Cleaning delay
```
make-all, line 98: Use `translation_key: timer` to get options translated

custom_components/tuya_local/devices/cast_lb500c_litterbox.yaml
```diff
@@ -0,0 +1,206 @@
...
+    category: config
+    dps:
+      - id: 103
+        type: string
+        name: option
+        mapping:
+          - dps_val: tensecond
+            value: "10 s"
```
make-all: Remove the gap between number and unit to match the translation keys.

60s should be `1m`
120s should be `2m`

Currently 40s and 90s (or 1m30s) are missing from the translations, but keeping the same format should future proof them.

custom_components/tuya_local/devices/cast_lb500c_litterbox.yaml
```diff
@@ -0,0 +1,206 @@
...
+            value: "40 s"
+          - dps_val: sixthsecond
+            value: "60 s"
+          - dps_val: ninthsecond
+            value: "90 s"
+          - dps_val: onetwosecond
+            value: "120 s"
+  - entity: select
+    name: Valid toilet time
+    category: config
```
make-all, line 123: Use `translation_key: timer` again, and remove spaces

make-all (changes requested):

PR #5456, opened 2026-06-30
Add Varin VA-E002 ceiling fan light

gkdimitris:
### Device
Varin® VA-E002 , CCT-only ceiling fan with light (no RGB), 6-speed fan.
Tuya WiFi, protocol 3.5, product category `fsd`.
Product id: vfqiwthdnlkynebk

### Why a new config rather than reusing amari_ceiling_fanlight
The DP family matches Amari (which is auto-selected), but three mappings
are wrong for this device:

| DP | Amari | VA-E002 | Fix |
|----|-------|---------|-----|
| 23 color_temp | `invert: true` | warm = low K natively | invert removed |
| 62 fan speed | range 1–100 | 6 discrete levels | range 1–6 |
| 61 preset_mode | WakeUp/Natural/Normal/Sleep | sends raw `fresh`, no app modes | omitted |

The VA-E002 also does not report dp 26 (timer) or dp 28 (light control),
so those entities are dropped to avoid permanently-unavailable entities.

### DIFF
--- a/custom_components/tuya_local/devices/varin_va_e002_fan_light.yaml
+++ b/custom_components/tuya_local/devices/varin_va_e002_fan_light.yaml
@@ -1,0 +1,62 @@
+name: Fan light
+entities:
+  - entity: light
+    name: Main Light
+    dps:
+      - id: 20
+        type: boolean
+        name: switch
+      - id: 22
+        type: integer
+        name: brightness
+        optional: true
+        range:
+          min: 10
+          max: 1000
+      - id: 23
+        type: integer
+        name: color_temp
+        optional: true
+        range:
+          min: 0
+          max: 1000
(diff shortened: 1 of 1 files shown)


gkdimitris:
Duplicate of #5455 (same product id vfqiwthdnlkynebk), which was opened first and also updates DEVICES.md, closing mine in favour of it.
One thing worth checking for #5455: on my unit the preset_mode (dp 61) was non-functional, the SmartLife app exposes no mode control and the device reports a static fresh, so I dropped it. If your device genuinely switches between fresh/nature, keep it; otherwise it may surface a preset selector that does nothing. Also FWIW I read color_temp as warm↔cool over the same 2700–6500 range you used.

Issue #5164, opened 2026-05-27
Request support for lidl-tronic-solarspeicher


Hofyyy:
### Log message

```text
Dieser Fehler stammt von einer benutzerdefinierten Integration

Logger: custom_components.tuya_local.config_flow
Quelle: custom_components/tuya_local/config_flow.py:502
Integration: Tuya Local (Dokumentation, Probleme)
Erstmals aufgetreten: 26. Mai 2026 um 17:24:24 (3 Vorkommnisse)
Zuletzt protokolliert: 20:40:26

Device matches zigbee_zb10wt_temphumid_sensor with quality of 20%. LOCAL DPS: {"updated_at": 1779809059.843034, "1": 100, "2": 21504, "4": 0, "10": 32, "37": 907, "102": 750, "103": 683, "104": 684, "105": "charge_first", "107": 80, "108": 80, "109": "standby", "111": true, "113": 0, "115": 600}
Device matches zigbee_zb10wt_temphumid_sensor with quality of 20%. LOCAL DPS: {"updated_at": 1779809126.206429, "1": 100, "2": 21504, "4": 0, "10": 32, "37": 907, "102": 750, "103": 683, "104": 684, "105": "charge_first", "107": 80, "108": 80, "109": "standby", "111": true, "113": 0, "115": 600}
Device matches zigbee_zb10wt_temphumid_sensor with quality of 21%. LOCAL DPS: {"updated_at": 1779907220.9757745, "1": 99, "2": 21228, "4": 0, "10": 25, "37": 1368, "102": 911, "103": 856, "104": 1133, "105": "charge_first", "107": 80, "109": "charge", "111": true, "113": 0, "115": 600}


Dieser Fehler stammt von einer benutzerdefinierten Integration

Logger: custom_components.tuya_local.config_flow
Quelle: custom_components/tuya_local/config_flow.py:502
Integration: Tuya Local (Dokumentation, Probleme)
Erstmals aufgetreten: 26. Mai 2026 um 17:24:24 (4 Vorkommnisse)
Zuletzt protokolliert: 20:47:14

Device matches zigbee_zb10wt_temphumid_sensor with quality of 20%. LOCAL DPS: {"updated_at": 1779809059.843034, "1": 100, "2": 21504, "4": 0, "10": 32, "37": 907, "102": 750, "103": 683, "104": 684, "105": "charge_first", "107": 80, "108": 80, "109": "standby", "111": true, "113": 0, "115": 600}
[...]

PR #5457, opened 2026-06-30
Add device config for Lidl Tronic Solarstromspeicher (Marstek Saturn B2500)

Maztah:
Adds local support for the Lidl Tronic Solarstromspeicher (2.2 kWh balcony
battery storage, hardware-identical to the Marstek Saturn B2500), Tuya
category `bxsdy`, product id `fpi7g58s9hoxlvng`, protocol 3.3.

Refs #5164
Refs Maztah/tronic-speicher-tuya-local#13

What's included:
- Main sensors: battery %, remaining time, status, charge/discharge
  power/current/voltage (decoded from base64 DP 3)
- PV input: PV1/PV2 power/voltage/current (decoded from base64 DP 101,
  only present while actively producing), total PV yield
- DC outputs: power/voltage/current (decoded from base64 DP 33)
- Lifetime energy counters: charged/discharged/consumed
- Controls: operating mode, temperature unit, inverter output power limit
  (DP 115), discharge power for slot 1 (DP 106, masked base64 write,
  80-800W), depth of discharge
- Diagnostics: fault code, pack count, standby threshold, inverter max
  power (decoded from DP 114)
- A button to force an immediate full DP push (DP 111)
- A button to write the device's default discharge-schedule blob directly
  to DP 106, used to recover after restart/reconnect when the device
  hasn't pushed DP 106 yet (masked writes fail with "Cannot mask unknown
  current value" until then)

Entity names use translation_key throughout instead of hardcoded strings,
with new keys added to translations/en.json only (translations to other
languages submitted separately, per the contributing guidelines).

Tested against a real device for several weeks.

### DIFF
--- a/custom_components/tuya_local/devices/lidl_tronic_solar_storage.yaml
+++ b/custom_components/tuya_local/devices/lidl_tronic_solar_storage.yaml
@@ -1,0 +1,547 @@
+# Lidl Tronic Solarstromspeicher, identical hardware to Marstek Saturn
+# B2500, Tuya category bxsdy, protocol Tuya 3.3
+# DP source: https://github.com/Maztah/tronic-speicher-tuya-local
+
+name: Battery storage
+
+products:
+  - id: fpi7g58s9hoxlvng
+    manufacturer: Lidl
+    model: Tronic Solarstromspeicher
+
+entities:
+
+  - entity: sensor
+    class: battery
+    dps:
+      - id: 1
+        type: integer
+        name: sensor
+        class: measurement
+        unit: "%"
+
+  # Values: charge / discharge / standby
+  - entity: sensor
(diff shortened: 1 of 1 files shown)


make-all:
Duplicate of #5368

Maztah:
> Duplicate of #5368

Sure? It seems to be only an inverter. My PR is a battery storage with including an inverter.

make-all:
Please read the instructions for PRs before submitting.
Bundling translation changes with new devices just creates work for the maintainers. Please file the device config separately, then we can have a discussion about which entities are worthwhile translating.

Maztah:
Sorry @make-all and thanks for your feedback. Split translation changes out so this PR now only touches the device config file.

PR #5463, opened 2026-07-01
Create tuya_smart_blind_wifi.yaml

holl595:

### DIFF
--- a/custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
+++ b/custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
@@ -115,4 +115,4 @@ entities:
     dps:
       - id: 104
         type: integer
-        name: sensor 
+        name: sensor
--- a/custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
+++ b/custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
@@ -1,7 +1,7 @@
 name: Blind robot
 legacy_type: tuya_smart_blind_wifi
 products:
-  - id: LY-1690
+  - id: Tuya-LY-1690
     manufacturer: Tuya
     model: Smart Blind Wifi
 entities:
--- a/custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
+++ b/custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
@@ -1,0 +1,118 @@
+name: Blind robot
+legacy_type: tuya_smart_blind_wifi
+products:
+  - id: LY-1690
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
```diff
@@ -0,0 +1,118 @@
+name: Blind robot
+legacy_type: tuya_smart_blind_wifi
```
make-all, line 2: There is no legacy file by that name.

custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
```diff
@@ -0,0 +1,118 @@
...
+products:
+  - id: Tuya-LY-1690
```
make-all, line 4: This is not a product id for matching by. It is the model

custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
```diff
@@ -0,0 +1,118 @@
...
+    manufacturer: Tuya
```
make-all, line 5: Tuya does not manuyfacture end devices, they only make the WiFi modules. This manufacturer is incorrect.

custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
```diff
@@ -0,0 +1,118 @@
...
+            value: "30°C"
+          - dps_val: 40
+            value: "40°C"
+          - dps_val: 50
+            value: "50°C"
+          - dps_val: 60
+            value: "60°C"
+
+  - entity: sensor
+    name: Ignored 104
```
make-all, line 112: Unknown dps should not be added as sensors, as that just creates a maintenance headache for later when migrations will be needed to correct the name if the actual function is discovered.
Add it as an attribute on another entity instead.

custom_components/tuya_local/devices/tuya_smart_blind_wifi.yaml
```diff
@@ -0,0 +1,118 @@
...
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 12
+        type: bitfield
+        name: fault_code
+  - entity: sensor
+    class: battery
```
make-all, line 82: Battery powered WiFi devices are not supported by this integration, and will not work reliably.

The presence of this sensor makes me think this will be a waste of time to take further, and should probably just be closed.

make-all (changes requested):

PR #5468, opened 2026-07-01
Fail fast on device offline to improve HA startup

SamJongenelen:
Fail fast when tuya device is offline. This reduces HA startup time for me to 4 secs (up from multiple minutes). 

My use case is my floor heater: it's off fully in the summer. This is my only Tuya device. 

When the device protocol is set to 'auto', a loop would retry different protocols an HA startup would be delayed and error logs kept happening. 

It is AI generated code (please squash), but the comments wrt error codes and the importof asyncio make sense to me. 

 Tested by me using my fork as repo source in HACS using only one Tuya device.

### DIFF
--- a/custom_components/tuya_local/__init__.py
+++ b/custom_components/tuya_local/__init__.py
@@ -7,6 +7,7 @@ investigation into Goldair's tuyapi statuses
 https://github.com/codetheweb/tuyapi/issues/31.
 """
 
+import asyncio
 import logging
 
 from homeassistant.config_entries import ConfigEntry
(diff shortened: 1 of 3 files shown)


make-all:
Don't use auto, you will get the same behaviour without any code change

make-all:
Changes like this have very high chance of causing regressions on some of the thousands of other devices out there, so I cannot accept this.

This integration was very stable until earlier this year when every man and his dog picked up AI to "fix" their pet niggles like this and started flooding the maintainer with PRs.

SamJongenelen:
I was thinking the tests would cover it. Nvm then

SamJongenelen:
I can make an issue if you want? 

I think there are a lot of seasonal devices out there and every IP based tuya device wil have those low level socket exceptions so I expect no regression.

Setting it to not auto will still have a default of 5sec which is too long imo

PR #5469, opened 2026-07-01
Manufactor rename and refresh interval fix.

m0bua:
Fixed value stuck (screenshot example):
зображення

Also renamed manufactor, GHpower is wrong naming.


andrea271988:
You probably removed the units of measurement by mistake; I find all the values ​​without them.

PR #5470, opened 2026-07-01
Add Create Windcalm S ceiling fan

ac-uy:
Adds configuration for the Create Windcalm S ceiling fan.

**Device:** Create Windcalm S ceiling fan
**Protocol:** Tuya local, protocol 3.5

**DPs:**
- DP 60 (boolean): power switch
- DP 62 (integer 1-6): fan speed, mapped to discrete percentage steps (17/33/50/67/83/100%)
- DP 63 (string): direction (forward/reverse)
- DP 64 (integer 0-540): countdown timer in minutes

Note: DP 62 uses explicit dps_val mappings rather than a 
ange, as the integer values 1-6 map to discrete speed steps and a linear range produces incorrect percentage values.

### DIFF
--- a/custom_components/tuya_local/devices/create_windcalm_s_fan.yaml
+++ b/custom_components/tuya_local/devices/create_windcalm_s_fan.yaml
@@ -1,4 +1,8 @@
 name: Ceiling fan
+products:
+  - id: bf8a0d2c5527605aca0hdv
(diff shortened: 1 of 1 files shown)


make-all:
This is already adequately covered by windcalm_fan.yaml

PR #5430, opened 2026-06-28
Add Emerio PAC-132905 AirCon

smoki3:
Add Emerio PAC-132905 Air Conditioner

https://github.com/make-all/tuya-local/issues/5426#event-27294864212

PR #5471, opened 2026-07-02
Improve user handling klarstein_iceblock_airconditioner

smoki3:
This will add Mode "off" to the hvac_mode, so that you are able to turn the AC off via the climate card. 
Some minor rename of the normal mode: use "comfort" instead of "none"

Otherwise you don't have the possibility to turn it off:

image

### DIFF
--- a/custom_components/tuya_local/devices/klarstein_iceblock_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/klarstein_iceblock_airconditioner.yaml
@@ -9,6 +9,21 @@ products:
 entities:
   - entity: climate
     dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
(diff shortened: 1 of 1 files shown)


make-all:
This does not appear to work reliably for Klarstein IceBlock Ecosmart devices: https://github.com/make-all/tuya-local/pull/2001#discussion_r1670551196

smoki3:
Can we then split the emerio and klarstein again? Working reliable for the Emerio device.

https://github.com/make-all/tuya-local/pull/5430

make-all:
With over 1600 device configs already, I am reluctant to add more for devices that are basically 100% compatible

smoki3:
So this PR is adding the Power button to HVAC Modes in addition to the Power button. So maybe then a compromise to have both on board?

make-all:
No, adding the power off to hvac_mode will break the other devices that cannot handle multiple dps being set together.

smoki3:
alright no worris, I will switch to my fork :-)

Issue #4214, opened 2025-12-28
Request support for EcoNour 42-Inch Smart Tower Fan


shopsD:
### Log message

```text
2025-12-28 18:53:16.311 WARNING (MainThread) [custom_components.tuya_local.config_flow] Adding Eco_TowerF_42_Fan device with product id rxhpjq5pk6kha69e
2025-12-28 18:53:16.494 WARNING (MainThread) [custom_components.tuya_local.config_flow] Cloud device spec:
[{"id": 1, "name": "switch", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 2, "name": "mode", "type": "Enum", "format": "{\"range\":[\"sleep\"]}", "enumMap": {}}, {"id": 3, "name": "fan_speed", "type": "Integer", "format": "{\"unit\":\"\",\"min\":1,\"max\":3,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 5, "name": "switch_horizontal", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 15, "name": "light", "type": "Boolean", "format": "{}", "enumMap": {}}, {"id": 21, "name": "temp_current", "type": "Integer", "format": "{\"unit\":\"\u2103\",\"min\":0,\"max\":99,\"scale\":0,\"step\":1}", "enumMap": {}}, {"id": 22, "name": "countdown_set", "type": "Enum", "format": "{\"range\":[\"cancel\",\"1h\",\"2h\",\"3h\",\"4h\",\"5h\",\"6h\"]}", "enumMap": {}}]
2025-12-28 18:53:16.497 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches flinq_breeze_pedestalfan with quality of 86%. DPS: {"updated_at": 1766947986.7104971, "1": false, "2": "normal", "3": 1, "5": false, "15": false, "21": 18, "22": "cancel"}
```

### Product ID

rxhpjq5pk6kha69e

### Product Name

EcoNour 42-Inch Smart Tower Fan

### DPS information

<details>
```text
{
  "result": {
    "properties": [
      {
        "code": "switch",
        "custom_name": "",
        "dp_id": 1,
        "time": 1766947840296,
        "type": "bool",
        "value": false
      },
      {
        "code": "mode",
        "custom_name": "",
        "dp_id": 2,
        "time": 1766947840300,
        "type": "enum",
        "value": "normal"
      },
      {
        "code": "fan_speed",
[...]

PR #5474, opened 2026-07-02
Add config for device EcoNour 42" tower fan HTS-F305A (new)

saravanabalagi:
Added config for EcoNour 42" tower fan
There seem to be 2 other versions of the device sold under the same name earlier #4214 
This is the newer version ATTOW (2026), added model number in the file name for distinction.

### DIFF
--- a/custom_components/tuya_local/devices/econour_towerfan_hts-f305a.yaml
+++ b/custom_components/tuya_local/devices/econour_towerfan_hts-f305a.yaml
@@ -116,12 +116,3 @@ entities:
             value: celsius
           - dps_val: f
             value: fahrenheit
-  - entity: sensor
-    name: Temperature F
-    class: temperature
-    dps:
-      - id: 101
-        type: integer
-        name: sensor
-        class: measurement
-        unit: F
--- a/custom_components/tuya_local/devices/econour_towerfan_hts-f305a.yaml
+++ b/custom_components/tuya_local/devices/econour_towerfan_hts-f305a.yaml
@@ -1,0 +1,127 @@
+name: Fan
+products:
+  - id: 0kcxic8phbllew01
+    manufacturer: EcoNour
+    model: HTS-F305A
+entities:
+  - entity: fan
+    translation_only_key: fan_with_presets
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+      - id: 2
+        name: preset_mode
+        type: string
+        mapping:
+          - dps_val: fresh
+            value: normal
+          - dps_val: nature
+            value: nature
+          - dps_val: sleep
+            value: sleep
+          - dps_val: auto
+            value: auto
+      - id: 3
+        name: speed
+        type: integer
+        range:
+          min: 1
+          max: 12
+      - id: 5
+        name: oscillate
+        type: boolean
+  - entity: select
+    name: Oscillation angle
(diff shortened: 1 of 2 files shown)


saravanabalagi:
In DEVICES.md file, please change the old entry as well for distinction. Also, I suggest adding speed and mode settings for users to distinguish. I could not find the model number for the old device.

- EcoNour 42" tower fan (3 speed, 2 modes)
- EcoNour 42" HTS-F305A tower fan (12 speed, 4 modes)

custom_components/tuya_local/devices/econour_towerfan_hts-f305a.yaml
```diff
@@ -0,0 +1,127 @@
...
+    dps:
+      - id: 28
+        type: string
+        name: option
+        mapping:
+          - dps_val: c
+            value: celsius
+          - dps_val: f
+            value: fahrenheit
+  - entity: sensor
```
make-all: This should be merged with the temperature sensor above. There is no need to have separate sensors for C anf F as they are measuring the same temperature.

make-all (changes requested):

custom_components/tuya_local/devices/econour_towerfan_hts-f305a.yaml
saravanabalagi: Initially thought of not adding in favour of hass conversions. But the device itself converts it internally and shows it on the physical display (when switched between C/F) which might be slightly off (eg from rounding) from the hass converted value. Hence I thought its worth adding the exposed converted value as a separate entity just to avoid any mismatch between the value on the display and the one shown on hass. What do you think?

make-all: That is exactly why they should be merged to display the appropriate value depending on the temperature unit setting. Then if HA and device settings match, the exact same values will display, and there will not be a second displayed value in HA that is potentially different because of rounding.

saravanabalagi: Just checked an example:
On device temp is 27C / 80F i.e. shown in the display.
If we only use one, say 27C, it would be 80.6F -> 81F after conversion and rounding.
i.e. on HA I would see 81F, but on the device display it would be 80F.

To avoid this difference, we could expose 2 entries that would exactly match what is being currently shown on the display.

Without the 2nd entry, we wont know what's displayed physically on the device. Basically both temp entries are always rounded and correspond to what's displayed on device, not the original temperature reading from the sensor (which is not exposed). Let me know if I misunderstood.

make-all: As far as I can see from the images online, the device only shows one temperature, and so should HA. There is no need to show two temperatures, which will both get converted to the user's preferred units by HA, and one will be rounded, the other from the actual device.

When the HA and device setting is matching, then only the device temperature with matching units should be shown.

make-all (changes requested):

custom_components/tuya_local/devices/econour_towerfan_hts-f305a.yaml
saravanabalagi: Ok. I have now removed the separate entity for fahrenheit.

PR #5485, opened 2026-07-03
Add Temprium PAC011 12,000 BTU Air Conditioner

tomgidden:
Identical to suntec_coolfix_airconditioner, except Dyr → Dry, and no sign of Heat mode, at least in the version on Amazon UK.

Diagnostics in dps 22 untested.

Note, this device has a bug confusing ºC and ºF, even in the official app. It'll say "81ºC" (!) then when switched to ºF, "28ºF", then a second later, "28ºC". When the temperature changes, it'll then go back to "82ºC", and so on. This behavior occurs in the API too, so I can't see a solution here that wouldn't be a hack.  Version: V.1.0.0(3.1.69), MCU: V1.0.0, no updates available.

### DIFF
--- a/custom_components/tuya_local/devices/temprium_pac011_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/temprium_pac011_airconditioner.yaml
@@ -21,6 +21,9 @@ entities:
                 value: dry
               - dps_val: Fan
                 value: fan_only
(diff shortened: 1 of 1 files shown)


make-all:
Identical to Meaco Cirro model that was added recently

PR #5487, opened 2026-07-03
Make loonas_curtain position invert conditional on motor direction

### LINKED ISSUE #5486: Curtain motor (loonas_curtain): open/closed state is reversed in HA when motor direction is set to "back"
### Description

I have a Tuya WiFi tubular blinds motor (YH-002) using the loonas_curtain config (tuya-local 2026.6.4). Because of how the blinds are mounted, the motor direction is set to "back" in the Tuya app (the same setting is exposed by tuya-local as the "Reverse" configuration switch).

With this setting, Home Assistant shows the open/closed state and position backwards: when the blinds are fully closed, HA shows them as fully open (100%) and vice versa. The Tuya app shows the state correctly.

Because of the swapped state, the cover buttons stop working at the end positions: with the blinds fully closed, HA believes they are fully open, so the Open button is greyed out, and pressing Close does nothing since the motor is already there. As soon as I move the blinds a little (so the position is somewhere in between), both buttons become active and work normally.

The open/close/stop commands themselves work in the correct direction — only the reported position and open/closed state are backwards.

Issue #3878 looks like the same underlying problem (a Nordice user with a reverse-mounted motor on this same config, who had to swap open/close manually after every update).

### Steps to reproduce

1. Add a curtain/blinds motor using the loonas_curtain config.
2. Turn on the "Reverse" switch (motor direction = "back" in the Tuya app) and recalibrate the limits in the Tuya app.
3. Fully close the blinds. The Tuya app shows closed; HA shows the cover as fully open, and the Open button is disabled.

### Expected behaviour

HA should show the same open/closed state and position as the Tuya app regardless of the motor direction setting.

### Additional context

_No response_

### Confirmation

galekseev:
Fixes the issue #5486 

loonas_curtain.yaml unconditionally inverts the position dps (2 and 3). However, when the motor direction (dp 5, control_back_mode, exposed as the "Reverse" config switch) is set to back, the device already reports position percentages flipped, so the unconditional invert becomes a double inversion: HA shows open as closed and disables the wrong cover buttons at the extremes. Commands via dp 1 (open/close/stop) are unaffected.

When the Reverse setting (dp 5, control_back_mode) is set to "back", the device reports position percentages already flipped, so the unconditional invert on dps 2 and 3 caused a double inversion: HA showed open as closed and disabled the wrong cover buttons at the extremes.

Invert the position dps only when dp 5 is "forward" or not reported (preserving existing behaviour for devices without dp 5), and skip the inversion when dp 5 is "back", matching what the Tuya app shows.

### DIFF
--- a/custom_components/tuya_local/devices/loonas_curtain.yaml
+++ b/custom_components/tuya_local/devices/loonas_curtain.yaml
@@ -43,6 +43,7 @@ entities:
           max: 100
         mapping:
           - invert: true
+          - available: reversed
       - id: 3
         name: current_position
         type: integer
@@ -52,6 +53,16 @@ entities:
           max: 100
         mapping:
           - invert: true
+          - available: reversed
+      - id: 5
+        name: reversed
+        type: string
+        optional: true
+        hidden: true
+        mapping:
+          - dps_val: back
+            value: true
+          - value: false
       - id: 7
         name: work_state
         type: string


make-all:
The syntax used is incorrect, so I can only conclude this PR is untested AI garbage. A proper fix has been committed.

galekseev:
Hi, sorry, forgot to add AI disclosure.
Fixed with AI, tested manully on my device (YH-002). Worked for me.
But anyway, thank you for paying attention to the issue and commiting the proper fix.

Issue #1840, opened 2024-04-19
Leo's Loo Too not detected


Jster42:
**Describe the bug**
Leo's Loo Too not detected

**To Reproduce**
Configure integration and put in device id and local key, a few devices are shown, accepted asc_wifi_circuit_breaker to see what logs would be created.

manual tinytuya output:
```
New base   Product ID = 1badzyvwh1e1hrog  [Valid Broadcast]:
    Address = xxx.xxx.xxx.xxx   Device ID = xxxxxxxxxx (len:22)  Local Key =xxxxxxxxxxxxxx  Version = 3.3  Type = default, MAC = xxxxxxxxxxxxxxx
    Status: {'1': True, '2': 'auto_clean', '5': 426, '6': 33, '7': 1, '9': False, '10': False, '13': False, '21': 0, '22': 0, '101': 15, '109': 30, '111': True, '112': False, '113': True, '115': False}
```

**Expected behavior**
I would have expected it to be matched properly

**Additional context**
Just for fun I removed all other yaml devices to see if it would get picked up, no matches

If the bug involves a device, then please include device diagnostics from
Settings / Devices & Services / Tuya Local (your device) / "1 device"

Please check if there are messages from Tuya Local in the Home
Assistant log (Settings / System / Logs), and include those too.
```
Device matches asc_wifi_circuit_breaker with quality of 25%. DPS: {"updated_at": 1713488128.8003364, "1": true, "2": "auto_clean", "5": 426, "6": 33, "7": 1, "9": false, "10": false, "13": false, "21": 0, "22": 0, "101": 15, "109": 30, "111": true, "112": false, "113": true, "115": false}
```
Not sure if this is related... but am including it
```
Detected blocking call to sleep inside the event loop by custom integration 'tuya_local' at custom_components/tuya_local/device.py, line 77: self._api = tinytuya.Device(dev_id, address, local_key), 
```

![image](https://github.com/make-all/tuya-local/assets/146975706/817198e7-1a8e-4342-a456-444f7ee09bcc)

![image](https://github.com/make-all/tuya-local/assets/146975706/5d85ccb0-b999-4674-9d5b-d0de1e304ddc)

Jster42:
I have tested with the optional flag on dps 16, the device is detected with this change.

Thank you!

PR #5488, opened 2026-07-04
Leo's Loo Too / MEET: fix weight scale, decode notifications, add MEET variant

algrym:
## Leo's Loo Too / MEET self-cleaning cat toilet — corrected config + notification decode

This corrects and extends the `leos_loo_too.yaml` config (Tuya category `msp`)
against live data from **four physical units** — three "Leo's Loo Too"
(`1badzyvwh1e1hrog`) and one rebadged "MEET" (`kw4ztiq3kxmcf8dx`) on identical
hardware — plus the Tuya cloud values as a reference oracle, and a multi-day
passive capture of the device's async report stream that decoded the DP21
notification bitfield.

### Changes

| Change | Confidence | Evidence |
| --- | --- | --- |
| **Weight DP6 scale `0.1` → `10`** | high | Raw is tenths of a kg (42 → 4.2 kg, matches cloud). The old `0.1` produced 420 kg. Consistent with the other weight litterbox configs (`duoqu`, `vevor`, `homall` all use `scale: 10`). |
| **Add the `kw4ztiq3kxmcf8dx` product id** | high | One unit enrols under this second product id. Owner-confirmed it is the *same* physical Leo's Loo Too hardware carrying the MEET / Chinese-market firmware profile — not a separate MEET product — so it stays under the `Leo's Loo` brand with model `Too (MEET firmware)`. |
| **DP2 (mode) / DP5 (delay) → `optional`** | high | Absent on that firmware variant (reports neither). |
| **DP22 fault → problem binary_sensor** | high | Same DP/pattern as the sibling `sailesi_litterbox` config. Also carries the raw DP21 notification code as a `notification_code` attribute (per review — see below). |
| **Decode DP21 notification bitfield** | high | See bit table below — confirmed by correlating the async report stream against observed device behaviour across all four units over several days. |
| **Keep DP7 "times used today", hidden** | high | Dead constant (`=1`) on every observed unit; never updates locally (also `1` in #1840's dump), so the real per-use count is not exposed on the LAN. Per review, rather than removing it (which would orphan the entity in existing installs) it is kept with an `available` null-test + `hidden: unavailable`, so it hides only where the DP is absent. |
| **Keep DP16 "light", hidden** | high | Never reported on the local protocol, so it only stranded an `unavailable` entity — but per review it is kept with `hidden: true` rather than removed, to avoid orphaning it in existing registries. |

### DP21 notification bits (decoded)

| Bit (value) | Meaning | Exposed as |
| --- | --- | --- |
| bit0 (1) | **Bin full / needs emptying** — box goes amber and *latches* this bit until the drawer is emptied and the reset button pressed | `binary_sensor` "Bin full" (plain on/off state) |
| bit1 (2) | Manual / commanded clean rotation active | folded into "Cleaning" (class `running`, mask 6) |
| bit2 (4) | Auto-clean rotation active | folded into "Cleaning" (class `running`, mask 6) |
| bit6 (64) | UV sterilisation phase active | `binary_sensor` "UV sterilizing" (class `running`) |

The raw notification code is retained as a `notification_code` attribute on the
DP22 Fault sensor (per review — not its own entity) so any remaining undecoded
bits stay observable. These are exposed as plain device **states** —
"bin full" is routine operation, not a malfunction, so no `problem`/alert
semantics are baked in; alerting is left to the user's own automations. (The
separate DP22 **Fault** binary_sensor keeps `class: problem`, as a fault is a
genuine device malfunction.)

The **bit0 "bin full"** decode was ground-truthed: an owner found the physical
amber "needs emptying" ring on a box while DP21 read exactly `1`; pressing the
reset button cleared it (`1 → 2 → …`), all signalled through DP21.

The **bit1 vs bit2** distinction was ground-truthed with a controlled test: a
timed manual-clean button press drove `DP21 0 → 2`, and **`2` held for 71 s** —
the full clean rotation, watched on the device — before clearing to `0`, with
the UV phase (`64`) following 6 s later. That is the identical lifecycle to a
bit2 auto-clean, just signalled on bit1, so **both bits mean "a clean rotation
is active"** and "Cleaning" masks them together (mask 6). Note: the device's own
Tuya data model labels bit1 `excessive_clean_times`, but the sustained 71-second
hold shows it is a clean-active state on this firmware, not a momentary warning —
so this decode follows the observed behaviour, not that label. (The DP22 fault
bit labels `motor`/`program`/`g_sensor` from the same model are left unverified,
as no fault fired on any unit; only the raw fault code is exposed.)

### Notes for the maintainer
- New binary_sensors use `name:` rather than `translation_key:` — happy to
  switch to translation keys (with `strings.json` additions) if preferred.
- Prior art: #1840 has another owner's raw dump corroborating DP7=1 and DP101=15.

### DIFF
--- a/custom_components/tuya_local/devices/leos_loo_too.yaml
+++ b/custom_components/tuya_local/devices/leos_loo_too.yaml
@@ -13,22 +13,34 @@
 #   high   - add the "MEET" product id; mark DP2 (mode) and DP5 (delay) optional
 #            because they are absent on some units (MEET reports neither).
 #   high   - DP22 fault bitfield as a problem binary_sensor (same pattern and DP
-#            as sailesi_litterbox). Tuya fault labels (from the device data
-#            model): bit0 motor_fault, bit1 program_fault, bit2 g_sensor_fault.
-#   high   - decode DP21 notification bitfield into binary_sensors. Bit labels
-#            come from the device's own Tuya data model (the `notification`
-#            status_range: garbage_box_full, excessive_clean_times, cleaning,
-#            no_weight) and were cross-checked against observed behaviour over
-#            several days on all four units:
-#              bit0 (1)  = garbage_box_full -> "Bin full" (ground-truthed: the
-#                          box goes amber and latches this bit until the drawer
-#                          is emptied and the reset button pressed)
-#              bit2 (4)  = cleaning         -> "Cleaning" (auto-clean rotation)
-#              bit6 (64) = UV phase         -> "UV sterilizing" (an undocumented
-#                          device extension, not in the 4-bit documented model)
-#            bit1 (excessive_clean_times) and bit3 (no_weight) are left in the
-#            raw code sensor undecoded. All exposed as plain device states;
-#            alerting is left to the user's own automations.
+#            as sailesi_litterbox). Per the Tuya data model the bits are bit0
+#            motor_fault, bit1 program_fault, bit2 g_sensor_fault - UNVERIFIED
+#            here (no fault fired on any unit), so only the raw code is decoded,
+#            not per-bit sensors.
+#   high   - decode DP21 notification bitfield into binary_sensors. Bit meanings
+#            were established by OBSERVING the device (a multi-day capture of
+#            the async report stream plus a controlled manual-clean test), then
+#            checked against the Tuya data model. Observed behaviour is the
+#            source of truth; where the model disagrees, the device wins:
+#              bit0 (1)  = bin full -> "Bin full". Ground-truthed: the box goes
+#                          amber and latches this bit until the drawer is
+#                          emptied and reset. Matches model "garbage_box_full".
+#              bit1 (2)  = manual/commanded clean rotation active -> "Cleaning".
+#                          Held for the full ~71 s rotation on a timed
+#                          manual-clean press, then UV followed - the same
+#                          lifecycle as bit2. NOTE: the Tuya model labels this
+#                          bit "excessive_clean_times", which is WRONG for this
+#                          firmware: it is a sustained clean-active state, not a
+#                          momentary warning.
+#              bit2 (4)  = auto-clean rotation active -> "Cleaning". Same ~70 s
+#                          rotation lifecycle, self-triggered. Matches
+#                          "cleaning".
+#              bit6 (64) = UV sterilisation phase -> "UV sterilizing". Not in
+#                          the documented 4-bit model at all (a device add-on).
+#            "Cleaning" therefore matches bits 1|2 (mask 6) so it covers both
+#            manual and auto cleans. bit3 ("no_weight") was never observed and
+#            is left in the raw code sensor undecoded. All exposed as plain
+#            device states; alerting is left to the user's own automations.
 #   high   - remove DP7 "times used today" (code excretion_times_day): on the
 #            local protocol it always reads 1. It is a per-use event pulse that
 #            the Tuya cloud accumulates server-side (status_range report_type
@@ -122,8 +134,9 @@ entities:
       - id: 9
         type: boolean
         name: button
-  # DP22 fault bitfield. Tuya labels: bit0 motor_fault, bit1 program_fault,
-  # bit2 g_sensor_fault. Any non-zero bit -> problem; raw code kept too.
+  # DP22 fault bitfield. Any non-zero bit -> problem; raw fault_code kept so the
+  # specific bit can be decoded when one actually fires. (Tuya model labels,
+  # unverified here: bit0 motor, bit1 program, bit2 g_sensor.)
   - entity: binary_sensor
     class: problem
     category: diagnostic
@@ -163,8 +176,11 @@ entities:
         name: sensor
         optional: true
         mapping:
-          # bit2 (4) = "cleaning": auto-clean rotation in progress.
-          - dps_val: 4
+          # bits 1|2 (mask 6): a clean rotation is active. bit2 = auto-clean,
+          # bit1 = manual/commanded clean (both observed holding for the full
+          # ~70 s rotation). The device's own model mislabels bit1 as
+          # "excessive_clean_times"; observation shows it is clean-active.
+          - dps_val: 6
             value: true
           - value: false
   - entity: binary_sensor
--- a/custom_components/tuya_local/devices/leos_loo_too.yaml
+++ b/custom_components/tuya_local/devices/leos_loo_too.yaml
@@ -1,16 +1,45 @@
-name: Pet toilet
+# Leo's Loo Too / MEET self-cleaning cat toilet  (Tuya category "msp")
+#
+# Corrected against live tinytuya DP dumps from four physical units (three
+# "Leo's Loo Too", one rebadged "MEET") plus the Tuya cloud values as a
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/leos_loo_too.yaml
```diff
@@ -69,12 +109,84 @@ entities:
...
+        type: bitfield
+        name: sensor
+        optional: true
+        mapping:
+          - dps_val: 64
+            value: true
+          - value: false
+  # Raw notification code retained for any bits not yet decoded.
+  - entity: sensor
+    name: Notification code
```
make-all: This is not really useful, so can just be an attribute on another entity (I suggest the problem binary sensor, as attaching it to one of the notification flags will be less obvious if you ever want to fetch it.)

custom_components/tuya_local/devices/leos_loo_too.yaml
```diff
@@ -1,16 +1,51 @@
...
+name: Leo's Loo Too
 products:
   - id: 1badzyvwh1e1hrog
     manufacturer: "Leo's Loo"
     model: Too
+  - id: kw4ztiq3kxmcf8dx
+    manufacturer: "Leo's Loo"
+    model: Too (MEET)
 entities:
-  - entity: sensor
```
make-all, line 7: Removing an entity from a released config will leave a zombie entity in the registry.

It may be better to keep it and add `hidden: true` instead.

custom_components/tuya_local/devices/leos_loo_too.yaml
```diff
@@ -1,16 +1,51 @@
...
+#            but is never reported on the local protocol, so it only ever
+#            stranded a permanently-unavailable entity.
+name: Leo's Loo Too
 products:
   - id: 1badzyvwh1e1hrog
     manufacturer: "Leo's Loo"
     model: Too
+  - id: kw4ztiq3kxmcf8dx
+    manufacturer: "Leo's Loo"
+    model: Too (MEET)
```
make-all: The brand for this one is MEET right? Not Leo's Loo.

custom_components/tuya_local/devices/leos_loo_too.yaml
```diff
@@ -40,13 +77,14 @@ entities:
       - id: 13
         name: switch
         type: boolean
-  - entity: light
+  - entity: switch
+    translation_key: sleep
     category: config
+    class: switch
     dps:
-      - id: 16
```
make-all, line 46: The original author put this in there, so their variant may be supporting this. Rather than removing it, add an available attribute that tests if the dps_val is null, and add `hidden: unavailable` to the entity

make-all (changes requested):

custom_components/tuya_local/devices/leos_loo_too.yaml
algrym: Dropped the standalone sensor and moved the raw DP21 code onto the Fault (problem) binary_sensor as a `notification_code` attribute, so it stays available for decoding any leftover bits without adding an entity.

custom_components/tuya_local/devices/leos_loo_too.yaml
algrym: Kept the DP16 `light` entity and added `hidden: true` rather than removing it. Agreed on not orphaning entities in existing installs.

custom_components/tuya_local/devices/leos_loo_too.yaml
algrym: Some Leo's Loo Too units come with the MEET (Chinese-market) firmware, which enrols under the separate `kw4ztiq3kxmcf8dx` product id. Same physical hardware, not a distinct MEET product. I've kept it under manufacturer "Leo's Loo" with model "Too (MEET firmware)" and added a comment explaining it. Happy to relabel if you'd rather it track the product-id branding instead.

custom_components/tuya_local/devices/leos_loo_too.yaml
algrym: Kept DP7 with an `available` dp that tests the raw value for null plus `hidden: unavailable`, so it stays visible on variants that report it and hides itself where the dp is absent. On my units it reads a constant `1`, but this keeps the door open for variants where it carries a real count.

custom_components/tuya_local/devices/leos_loo_too.yaml
```diff
@@ -1,16 +1,68 @@
-name: Pet toilet
+# Leo's Loo Too / MEET self-cleaning cat toilet  (Tuya category "msp")
```
make-all: The added comments are excessive. You should not be explaining what you have done in file comments, that is what commit comments are for.

custom_components/tuya_local/devices/leos_loo_too.yaml
```diff
@@ -1,16 +1,68 @@
...
+#   high   - DP7 "times used today" (excretion_times_day) reads a constant 1 on
+#            the local protocol: a per-use pulse the cloud sums server-side, so
+#            the local value never changes. Rather than remove it (orphaning the
+#            entity in existing installs) it is kept, and hidden when the dp is
+#            absent (available null-test + hidden: unavailable). #1840's dump
+#            shows =1 too.
+#   medium - DP16 "light" is never reported on the local protocol here. It
+#            shipped in the released config, so it is kept but marked
+#            hidden: true rather than removed (removal would orphan the entity).
+name: Leo's Loo Too
```
make-all: This change is incorrect. It will affect the name shown for other devices that happen to match the same config

custom_components/tuya_local/devices/leos_loo_too.yaml
```diff
@@ -69,12 +137,106 @@ entities:
...
+        class: measurement
+        mapping:
+          # Raw value is in tenths of a kg (42 -> 4.2 kg, cloud-confirmed).
+          - scale: 10
+  # DP7 "times used today" (excretion_times_day) reads a constant 1 on the local
+  # protocol here - a per-use pulse the cloud sums server-side, so the local
+  # value never changes. It shipped in the released config and another variant
+  # may report it, so rather than remove it (which would orphan the entity in
+  # existing installs) it is kept, and hidden when the dp is absent.
+  - entity: sensor
```
make-all: Please move this back to where it was to minimize changes and make it clearer for review what you actually changed here.

make-all (changes requested):

custom_components/tuya_local/devices/leos_loo_too.yaml
algrym: Done — dropped the header changelog and trimmed the inline comments to bare device facts; the rationale lives in the PR description and commit history now.

custom_components/tuya_local/devices/leos_loo_too.yaml
algrym: Reverted — the top-level name is back to "Pet toilet".

custom_components/tuya_local/devices/leos_loo_too.yaml
algrym: Moved "Times used today" back to its original position; the diff there is now just the added hidden/available DPs in place.

PR #5491, opened 2026-07-04
Add support for Nere Smart ceiling fan

irikuo:
Add support for Nere Smart ceiling fan.


make-all:
Merged with windcalm_fan_with_light.yaml

PR #5493, opened 2026-07-04
feat(temprium_pac011_airconditioner) fahrenheit/celsius confusion bug

tomgidden:
Following up on the bug mentioned in PR #5485 :

The Temprium does work with the `meaco_cirro_airconditioner` config as noted in response to that PR, but HASS still shows my room as being at 81ºC (ie. 177ºF) 🔥

While `dps-3` does correctly report the Celsius temperature for a couple of seconds when the units are updated, the rest of the time it's reliably showing the Fahrenheit temperature and `dps-24` is showing the Celsius temperature.  This is true whether the configured unit (dps-19) is 'c' or 'f'.  So, effectively swapping `dps-3` and `dps-24` seems to fix the behavior in HASS.

Rather than risking breaking `meaco_cirro_airconditioner` trying to support both, I've recreated `temprium_pac011_airconditioner` with this fix.

### DIFF
--- a/custom_components/tuya_local/devices/meaco_cirro_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/meaco_cirro_airconditioner.yaml
@@ -8,11 +8,6 @@ products:
     manufacturer: Meaco
     model: Cirro+ 14000 BTU Inverter
     model_id: CIRRO14000PRO
-## Temprium PAC011 appears identical apart from a celsius/fahrenheit confusion bug;
-## temprium_pac011_airconditioner.yaml has a workaround.
-#  - id: kzdx8i4g0wms1vta
-#    manufacturer: Temprium
-#    model: PAC011
 entities:
   - entity: climate
     dps:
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/meaco_cirro_airconditioner.yaml
```diff
@@ -8,9 +8,11 @@ products:
     manufacturer: Meaco
     model: Cirro+ 14000 BTU Inverter
     model_id: CIRRO14000PRO
-  - id: kzdx8i4g0wms1vta
-    manufacturer: Temprium
-    model: PAC011
+## Temprium PAC011 appears identical apart from a celsius/fahrenheit confusion bug;
+## temprium_pac011_airconditioner.yaml has a workaround.
```
make-all: You don't need a story here, just remove the product details for the Temprium

make-all (changes requested):

PR #5506, opened 2026-07-06
The NHZS fan speed was incorrectly displayed as 20% by default, even though it was 100% in Home Assistant.

witanage:
The NHZS fan speed was incorrectly displayed as 20% by default, even though it was 100% in Home Assistant. This has now been fixed.


make-all:
Is your product id matching cwwc6ga2gz8mvrqo?

It would be disruptive to other users to change the speed range of a fan they are already using just to suit your specific device. The original author set the range at 1-20, so I assume that was matching their device.

witanage:
Apologies. Would it be possible to introduce this as a new device instead? The existing ranges don't seem to work for my device, probably some other devices as well.

make-all:
NHZS is similar to Yidi fan light switch, but without the light. Yidi does have raw percentage steps though, and also reports the same "NHZS-风扇灯开关" product name through the Tuya cloud API, so likely the same manufacturer.

witanage:
I could be wrong, But it didn't gave me a choice of "Yidi fan light switch" while I'm integrating the device though.

make-all:
It should do now

PR #5521, opened 2026-07-08
Add Celsius variant of TCL AirMax heat pump (Nordis Orion EVO)

virtualas10:
The existing `daizuki_heatpump.yaml` (product id `hw50w7qvxluhslkk`) assumes the
device reports Fahrenheit: DP 2 is `unit: F`, range 610-880, and DP 3 carries a
`target_range` mapping (-20..100 to -4..212).

The same product id is used on export/EU-market units (Tuya cloud
`product_name` is "TCL外销分体机" - "TCL export split unit", sold in the EU as
Nordis Orion EVO among others), which report Celsius x10 on DP 2. The device
has no C/F setting and exposes no unit DP, so a conditional mapping within one
config is not possible.

With the current config, a 24 °C setpoint (DP 2 = 240) is displayed as -4 °C
in HA. This PR adds a sibling config with `unit: C`, setpoint range 16.0-31.0,
and the DP 3 conversion mapping removed. All other entities are unchanged from
the original config.

DPS dump from my unit with panel at 24 °C, room 24 °C:

    {'1': False, '2': 240, '3': 24, '4': 'cold', '5': 'low', '18': 0, '20': 0,
     '101': 0, '105': 'off', '110': 135724, '113': '0', '114': '0', '119': '0',
     '120': 'off', '123': '0000', '125': 'great', '126': '5', '127': '1',
     '128': '0', '129': '1', '130': 26, '131': False, '132': False, '133': '0',
     '134': '{"t":...,"s":false,"clr":true}'}

Tested live on a Nordis Orion EVO: setpoint read/write, HVAC modes and fan
speeds all working correctly with this config.

### DIFF
--- a/custom_components/tuya_local/devices/daizuki_heatpump_c.yaml
+++ b/custom_components/tuya_local/devices/daizuki_heatpump_c.yaml
@@ -1,0 +1,308 @@
+name: Heat pump (Celsius)
+products:
+  - id: hw50w7qvxluhslkk
+    manufacturer: TCL
+    model: AirMax (export, °C)
+  - id: hw50w7qvxluhslkk
+    manufacturer: Nordis
+    model: Orion EVO
+entities:
+  - entity: climate
+    translation_only_key: aircon_extra
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: auto
+                value: heat_cool
+              - dps_val: cold
+                value: cool
+              - dps_val: hot
+                value: heat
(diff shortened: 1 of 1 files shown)


make-all:
There are already a group of devices using the same dp layout as daizuki_heatpump, probably one of the existing configs will already work well with your device.

fisher_summer_air_conditioner matched daizuki_heatpump 100%
rotenso_ronix_heatpump matched daizuki_heatpump 76%
starlight_heatpump matched daizuki_heatpump 92%
hellnar_heatpump matched daizuki_heatpump 92%
fujicool_yuzu_heatpump matched daizuki_heatpump 100%

make-all:
Merged with daizuki_heatpump so the matching product id does not cause confusion

Issue #5347, opened 2026-06-17
Devices connected with protocol 3.4 randomly become Unavailable - 'Error': 'Check device key or version', 'Err': '914', 'Payload': None


yyoshii:
### Description

Hi,

i already made comments at the "old" ticket:
https://github.com/make-all/tuya-local/issues/5136#issuecomment-4723952535

bolderbast closed the old ticket with comment: "...So we'll re-test when Tuya Local 2026.6.2 is released, and open a new issue if it's not resolved..."

System:
Installation method: Home Assistant OS
Core: 2026.6.3
Supervisor: 2026.06.1
Operating System: 17.3
Frontend: 20260527.6
Tuya Local Version: 2026.6.2

I have got 6 Shutter devices (SMARTERCURRY Wlan Rolladenschalter).
4 Devices goes offline the first 12 hours after installing Tuya Local Version 2026.6.2.

To be sure that the problem is reproducible i powered off each device and powered the device on again yesterday afternoon.
In additional i restarted HomeAssistant.
Over night the entities of 3 devices are unavailable again.
"Rollade Wohnzimmer Rechts" - unavailable 
"Rollade WC" - unavailable 
"Rollade Treppenhaus" - Entities were unavailable and i powered off and on the device this morning and this should be visible in the logs too

The device can still be used by the official "SmartLife" / Tuya app.
Only Tuya Local is not working

If i read the log correct the device "Rollade WC" is working "2026-06-17 07:32:42.209" and get offline "2026-06-17 07:32:42.538":
Here is one example, in additional i added the logfile.

[HA_Logs_Local_Tuya.zip](https://github.com/user-attachments/files/29035060/HA_Logs_Local_Tuya.zip)

<details><summary>Details</summary>
<p>

```
2026-06-17 07:32:42.209 DEBUG (MainThread) [custom_components.tuya_local.device] Rollade WC received {"1": "stop", "2": 100, "7": false, "8": "forward", "full_poll": true}
2026-06-17 07:32:42.210 DEBUG (MainThread) [custom_components.tuya_local.helpers.device_config] Roller shutter: Mapped dps 3 value from None to Normal operation
2026-06-17 07:32:42.211 DEBUG (MainThread) [custom_components.tuya_local.helpers.device_config] Roller shutter: Mapped dps 8 value from forward to Forward
2026-06-17 07:32:42.264 DEBUG (SyncWorker_36) [tinytuya.core.XenonDevice] sending payload quick
[...]

Issue #4862, opened 2026-04-09
My Feyree EV charger stopped working (Failed to refresh device state for Feyree EV charger)


istaveren:
### Description

Sinds yesterday the EV charger is not displayed anymore in HA.

I have tried all kind of things like downgrade tuya-local and HA core. But didn't seem to help.

### Steps to reproduce

1. It was already installed
2. Look at the entities at Tuya Local
3. Configuration failed

### Expected behaviour

1. It is configured
2. Shows the values of the device like voltage, charger state etc.

### Additional context

I enabled debug level logging.
This is the result.

```
2026-04-09 19:58:21.742 DEBUG (MainThread) [custom_components.tuya_local] Setting up entry for device: bfcf358e06278286e9mz7j
2026-04-09 19:58:21.743 INFO (SyncWorker_1) [custom_components.tuya_local.device] Creating device: bfcf358e06278286e9mz7j
2026-04-09 19:58:21.744 DEBUG (MainThread) [custom_components.tuya_local.device] Refreshing device state for Feyree EV charger
2026-04-09 19:58:21.745 DEBUG (MainThread) [custom_components.tuya_local.device] Setting protocol version for Feyree EV charger to 3.4
2026-04-09 19:58:21.747 DEBUG (SyncWorker_6) [tinytuya.core.XenonDevice] status() entry (dev_type is default)
2026-04-09 19:58:21.747 DEBUG (SyncWorker_6) [tinytuya.core.XenonDevice] final payload_dict for 'bfcf358e06278286e9mz7j' ('v3.4'/'default'): {1: {'command': {'gwId': '', 'devId': '', 'uid': '', 't': ''}}, 7: {'command': {'protocol': 5, 't': 'int', 'data': {}}, 'command_override': 13}, 8: {'command': {'gwId': '', 'devId': ''}}, 9: {'command': {'gwId': '', 'devId': ''}}, 10: {'command': {}, 'command_override': 16}, 13: {'command': {'protocol': 5, 't': 'int', 'data': {}}}, 16: {'command': {}}, 18: {'command': {'dpId': [18, 19, 20]}}, 64: {'command': {'reqType': '', 'data': {}}}}
2026-04-09 19:58:21.747 DEBUG (SyncWorker_6) [tinytuya.core.XenonDevice] building command 10 payload=b'{}'
2026-04-09 19:58:21.754 DEBUG (SyncWorker_6) [tinytuya.core.XenonDevice] sending payload quick
[...]

PR #5525, opened 2026-07-09
Include the device error code in connection failure logs

com6056:
### Problem

When a device connection fails, `_retry_on_failed_connection` captures the error tinytuya returned but then logs only the generic message:

```
ERROR ... Failed to refresh device state for Bedroom Fan/Light.
```

The error code is dropped. Enabling debug logging doesn't recover it either — the code is never formatted into any message, so the only way to see it today is to reproduce the call outside Home Assistant with tinytuya directly.

This matters most for **error 914**. tinytuya returns 914 for *any* failure to negotiate a session, but its message — "Check device key or version" — names only two of the causes. A device that has stopped accepting local connections and needs a power cycle returns the same 914 with a perfectly correct key and protocol version.

That case is common: #5347, #5136, #4862 and [jasonacox/tinytuya#581](https://github.com/jasonacox/tinytuya/issues/581) are all the same shape — the device works fine in the Tuya app, tinytuya says 914, and only a power cycle fixes it. Because the log says nothing, every one of those reports starts by re-checking the local key, which is not the problem.

I hit this on a Treatlife DS03. The stored key was byte-identical to the one in the cloud, and the device still returned 914 on every protocol version. The log gave no hint that "check your key" was a dead end.

### Change

Keep the error code and the device's own message, and log them with the existing text. For 914, spell out the causes it can stand for:

```
ERROR ... Failed to refresh device state for Bedroom Fan/Light. Device reported error
914: Check device key or version (an incorrect local key or protocol version, or a
device that has stopped accepting local connections and needs to be power cycled)
```

Failures that aren't device errors (socket errors, timeouts) log exactly as before.

The existing carve-out that keeps 914 at debug level while auto-detecting the protocol version is untouched — 914 is expected there, and that path stays quiet.

### Tests

Three tests added to `tests/test_device.py`:

- `test_refresh_error_reports_device_error_code` — the code, the message and the 914 hint appear in the error log. Fails without this change.
- `test_refresh_error_without_error_code_is_unchanged` — a plain exception still logs the bare message, with no "Device reported error" suffix.
- `test_refresh_error_914_stays_quiet_when_rotating_protocols` — 914 during protocol auto-detection still does not log at error level.

`pytest tests/test_device.py` → 43 passed. `ruff check` and `ruff format --check` clean.

### Note

This only makes the failure legible; it doesn't fix it. The underlying device-side fault belongs upstream in tinytuya, where 914 conflates causes that are distinguishable on the wire. I'm investigating that separately and will report it there.

### DIFF
--- a/custom_components/tuya_local/device.py
+++ b/custom_components/tuya_local/device.py
@@ -34,6 +34,17 @@ from .helpers.log import log_json
 
 _LOGGER = logging.getLogger(__name__)
 
+# Extra context for tinytuya error codes whose message does not fully describe
+# the possible causes.  Error 914 in particular is reported for any failure to
+# negotiate a session, which includes a device that is refusing connections
+# until it is power cycled, not just a misconfigured key or protocol version.
+_ERROR_HINTS = {
+    "914": (
+        " (an incorrect local key or protocol version, or a device that has "
+        "stopped accepting local connections and needs to be power cycled)"
+    ),
+}
+
 
 def _collect_possible_matches(cached_state, product_ids):
     """Collect possible matches from generator into an array."""
@@ -658,12 +669,14 @@ class TuyaLocalDevice(object):
         )
 
         last_err_code = None
+        last_err_msg = None
         for i in range(connections):
             try:
                 if not self._hass.is_stopping:
                     retval = await self._hass.async_add_executor_job(func)
                     if isinstance(retval, dict) and "Error" in retval:
                         last_err_code = retval.get("Err")
+                        last_err_msg = retval.get("Error")
                         if last_err_code == "900":
                             # Some devices (e.g. IR/RF remotes) never return
                             # status data; error 900 is their normal response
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/device.py
```diff
@@ -34,6 +34,17 @@
 
 _LOGGER = logging.getLogger(__name__)
 
+# Extra context for tinytuya error codes whose message does not fully describe
+# the possible causes.  Error 914 in particular is reported for any failure to
+# negotiate a session, which includes a device that is refusing connections
+# until it is power cycled, not just a misconfigured key or protocol version.
+_ERROR_HINTS = {
+    "914": (
+        " (an incorrect local key or protocol version, or a device that has "
```
make-all: Is this supposed to be a standalone message to replace the tinytuya one, or a snippet to append to the end?
I see it is in parentheses so probably the latter, and that seems to be how it is formatted. In that case you don't need to repeat the information that is already in the tinytuya message, and don't deemphasize it by putting it in parentheses.

I suggest " Alternatively the device may have stopped accepting local connections and needs to be power cycled."

com6056: Append, yes. It's concatenated onto tinytuya's message, so I didn't want to fully replace it. You're right that the parens buried the important part. Your wording reads better than mine, thanks for taking it.

I pushed a small `ruff format` commit on top. Dropping the second string left the parens wrapped across two lines, so `format --check` was failing CI. Should be green now.

PR #5527, opened 2026-07-09
Create weau_pool_spapump_v2.yaml

johannesdeboer:
New device/device line: W'eau Hybride spa heatpomp.

https://www.toppy.nl/product/23912/w-eau-hybride-spa-warmtepomp-5-2-kw.html


make-all:
This seems to be almost identical to poolsana_inverprestige_pro_heatpump.yaml

The only difference is in the temperature range which is shifted up by 5 degrees. But information for this model I can find online gives a heating range of 15 - 40, and cooling range 8 - 28, which does not match the values here either.

PR #5531, opened 2026-07-09
Support Trident Air Dome 60

nicholasng99:
Similar to the existing AirDome-70, but with 6 fan speeds instead of 3, plus three extra controls the 70 doesn't expose: a countdown timer, a fault/problem sensor, and a second filter-reset button.

DP map (from `QueryThingsDataModel`):

| dp | code | type | range |
|----|------|------|-------|
| 1 | switch | bool | |
| 2 | pm25 | value | 0-999 |
| 3 | mode | enum | auto, manual, sleep |
| 4 | fan_speed_enum | enum | speed1-speed6 |
| 7 | child_lock | bool | |
| 8 | light | bool | |
| 18 | countdown_set | enum | cancel, 1h, 2h, 4h, 6h, 8h, 12h |
| 21 | air_quality | enum, ro | great, good, medium, severe |
| 22 | fault | bitmap, ro | e1=clean filter 2, e2=replace filter 3 |
| 101 | filter2_reset | bool | |
| 102 | replacefilter3_reset | bool | |
| 103 | flow_angle | enum | 30, 60, 90 |

### DIFF
--- a/custom_components/tuya_local/devices/trident_airdome60_airpurifier.yaml
+++ b/custom_components/tuya_local/devices/trident_airdome60_airpurifier.yaml
@@ -1,0 +1,163 @@
+name: Air purifier
+products:
+  - id: xkorr6vvmwjthxnl
+    manufacturer: Trident
+    model: AirDome-60
+entities:
+  - entity: fan
+    translation_key: air_purifier
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+      - id: 3
+        name: preset_mode
+        type: string
+        mapping:
+          - dps_val: auto
+            value: auto
+          - dps_val: manual
+            value: manual
+          - dps_val: sleep
+            value: sleep
+      - id: 4
+        name: speed
+        type: string
+        mapping:
+          - dps_val: speed1
+            value: 17
+          - dps_val: speed2
+            value: 33
+          - dps_val: speed3
+            value: 50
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/trident_airdome60_airpurifier.yaml
```diff
@@ -0,0 +1,163 @@
...
+        mapping:
+          - dps_val: 0
+            value: ok
+          - dps_val: 1
+            value: Clean filter 2
+          - dps_val: 2
+            value: Replace filter 3
+
+  - entity: button
+    name: Filter 2 reset
```
make-all: Am I missing something here? There doesn't seem to be a "Filter 1"?

Probably it is better to name the filters descriptively. eg "HEPA filter", "Mesh filter", "Carbon filter" or whatever the manual describes them as.

make-all (changes requested):

custom_components/tuya_local/devices/trident_airdome60_airpurifier.yaml
nicholasng99: there is no filter 1 in the dp map, filter 2 is the washable filter that the reminder will ask to clean, this button resets the reminder. Filter 3 i believe is a component the manufacturer needs to replace so this shouldn't be triggered by the user.

PR #5538, opened 2026-07-10
feat(devices): add support for Tuya M18 Smart Video Doorbell

hudsonbrendon:
## Summary

Adds local support for the **Tuya Smart Video Doorbell M18** (product `is5yumkqdhcfxskv`, category `sp`), used as “Video Porteiro”.

- Based on live cloud status + the [Tuya low-power doorbell DP standard](https://developer.tuya.com/en/docs/iot-device-dev/doorbell_solution?id=Kayamyivh15ox).
- Maps battery / power source / awake, record switch & mode, PIR sensitivity, night vision, flip, indicator, SD card, chime pairing, siren, volume, human filter, work mode, motion and doorbell snapshot events.

### Device details (from cloud)

| Field | Value |
| --- | --- |
| Name | Video Porteiro |
| Model | M18 / Smart Video Doorbell |
| Product ID | `is5yumkqdhcfxskv` |
| Category | `sp` |
| Device ID | `ebc4bc5e24ce1b28a78k79` |
| Connection | Wi‑Fi (battery / low-power) |

### Important limitation

**tuya-local does not provide live video streaming.**  
This config covers sensors, events (ring / motion snapshots) and settings. For live view, use ONVIF/WebRTC if the hardware exposes a local stream, or keep the official Tuya integration for camera only.

Battery doorbells also sleep most of the time; local updates only arrive when the device wakes (ring, motion, manual wake).

## Test plan

- [ ] Add device in tuya-local with host / device_id / local_key (wake the doorbell first if needed)
- [ ] Confirm type **Video doorbell / M18** is offered
- [ ] Battery, awake, record, night vision, volume entities appear
- [ ] Doorbell ring event fires when button is pressed
- [ ] Motion event / snapshot DPs update when motion is detected
- [ ] PIR sensitivity and work mode can be changed


make-all:
This is close enough to hunterbk_hf6602t_doorbell that the extra features should just be merged with that.

PR #5540, opened 2026-07-10
Add horizontal swing (DPS 106) to Teknopoint Idra Skiv

romainsc:
## Summary

Adds DPS 106 (horizontal vane position) as a `select` entity to the Teknopoint Idra Skiv air conditioner profile.

## What it does

Controls the horizontal left/right swing of the indoor unit vane:
- `Off` — no horizontal swing
- `Swing` — horizontal sweep active (dps_val: `same`)

The RA1A1 remote control has a dedicated "Left&right swing" button for this function.

## Testing

- **Hardware**: 4× Teknopoint EVO-09 indoor units with IDRA iSKV4-28C9 water-cooled condenser
- **Integration**: tuya-local, Home Assistant 2026.6.3
- **Discovery**: DPS 106 found via `tuya_local` debug logging (`full_poll` responses show `"106": "same"` or `"106": "off"`)
- **Values observed**: `"off"` (no swing) and `"same"` (swing active) across 4 units

## Notes

- DPS 106 was already documented in the project but not exposed in the device profile
- The EVO-09 catalogue confirms horizontal swing support
- Additional fixed positions may exist (the remote cycles through multiple directions) but only `off` and `same` have been observed so far — additional values can be added as they are discovered

### DIFF
--- a/custom_components/tuya_local/devices/idra_skiv_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/idra_skiv_airconditioner.yaml
@@ -251,6 +251,19 @@ entities:
             value: Slightly down
           - dps_val: "5"
             value: Bottom
+  - entity: select
+    name: Horizontal position
+    category: config
+    icon: "mdi:arrow-left-right"
+    dps:
+      - id: 106
+        type: string
+        name: option
+        mapping:
+          - dps_val: "off"
+            value: "Off"
+          - dps_val: "same"
+            value: "Swing"
   - entity: switch
     translation_key: sleep
     category: config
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/idra_skiv_airconditioner.yaml
```diff
@@ -251,6 +251,19 @@ entities:
             value: Slightly down
           - dps_val: "5"
             value: Bottom
+  - entity: select
+    name: Horizontal position
```
make-all: This should be `swing_horizontal_mode` on the climate entity rather than a separate entity.

make-all (changes requested):

romainsc:
Updated: moved DPS 106 from a separate `select` entity to `swing_horizontal_mode` on the climate entity as requested. Also added `optional: true` since the original contributor may not have had this DP.

PR #5548, opened 2026-07-11
Add Aqua Plus APAIO270 heat pump water heater

adfhogan:
## Summary

Adds device support for the Aqua Plus APAIO270 heat pump hot water system.

This device had no existing config and was being fuzzy-matched onto an unrelated reversible pool/spa heat pump config, producing a `climate` entity with wrong HVAC modes and a continuous `Unrecognised HVAC Mode of True ignored` warning.

- Modelled as a `water_heater` entity (`product_id: zcuzgqkr3aef1hap`) with the device's three real operation modes: Standard → `heat_pump`, Booster → `performance`, E-Heater → `electric`.
- Adds diagnostic sensors for tank/coil/discharge/suction temperatures, expansion valve position, fan speed, defrost/compressor/four-way-valve/water-pump/electric-heating-element status, high/low pressure switches, water flow switch, phase sequence module, circulation mode, energy/current/voltage/power, and compressor runtime before defrost.
- The full DP map (including several DPs not present in the standard schema, and English `code` strings that didn't match what they actually measured) was recovered via the Tuya Cloud API `/v2.0/cloud/thing/{id}/model` and `/shadow/properties` endpoints, then cross-checked against the manufacturer's own app.
- All three operation modes were verified live end-to-end: toggling Standard/Booster/E-Heater in the Tuya app was confirmed to transition `water_heater.hot_water_heat_pump`'s `current_operation` correctly in Home Assistant.
- DP labels/polarities for the pressure switches, water-flow switch, and several other diagnostic fields were empirically confirmed against the app's own "Temp. Query"/"Status Query" screens rather than assumed.

## Test plan

- [x] `uv run yamllint custom_components/tuya_local/devices/aquaplus_apaio270_waterheater.yaml` passes
- [x] `uv run ruff check .` / `uv run ruff check --select I .` / `uv run ruff format --check .` pass
- [x] `uv run pytest tests/test_device_config.py` passes (32/32)
- [x] No `product_id` collision with existing device configs
- [x] Deployed to a live production HA instance and verified: all three modes transition correctly, temperatures track live readings, no HVAC-mode warnings or entity errors since deployment

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -216,6 +216,7 @@
 
 - Apricus heat pump water heater
 - A.O. Smith HeatBot 15L electric water heater
+- Aqua Plus APAIO270 heat pump water heater
 - Aquatech Rapid/X6 heat pump water heater
 - Aquaviva AVH15S combo air-water heat pump
 - Arçelik AHPH-MM series combo air-water heat pump
--- a/custom_components/tuya_local/devices/aquaplus_apaio270_waterheater.yaml
+++ b/custom_components/tuya_local/devices/aquaplus_apaio270_waterheater.yaml
@@ -1,0 +1,312 @@
+name: Heat pump water heater
+products:
+  - id: zcuzgqkr3aef1hap
+    manufacturer: Aqua Plus
+    model: APAIO270
+entities:
+  - entity: water_heater
+    dps:
+      - id: 1
+        type: boolean
+        name: operation_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: Stand
+                value: heat_pump
+              - dps_val: Boost
+                value: performance
+              - dps_val: ELE
+                value: electric
+      - id: 2
+        type: string
+        name: mode
+        hidden: true
+      - id: 4
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 15
+          max: 75
+      - id: 21
+        type: integer
+        name: current_temperature
+        unit: C
+  - entity: binary_sensor
+    translation_key: defrost
+    category: diagnostic
+    dps:
+      - id: 7
+        type: boolean
+        name: sensor
+  - entity: sensor
+    name: Expansion valve position
+    icon: "mdi:pipe-valve"
+    category: diagnostic
+    dps:
+      - id: 14
+        type: integer
+        name: sensor
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 15
+        type: bitfield
+        name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 15
+        type: bitfield
+        name: fault_code
+  - entity: sensor
+    name: Phase sequence module
+    class: enum
+    category: diagnostic
+    dps:
+      - id: 17
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: "NO"
+            value: Nothing
+          # "normal" and "ERR" are unconfirmed guesses - never observed live
+          - dps_val: normal
+            value: Normal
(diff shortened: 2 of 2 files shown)


adfhogan:
Generated with Claude Code Sonnet 5. Settings verified between "Tuya" android app and HA. It looks like Tuya config provides power monitoring, but the hardware doesn't implement it (no current clamp etc.) as values are always zero for voltage, power, energy. Other units based around same Tuya config, however, may implement power monitoring.

adfhogan:
Initially closest fuzzy match was its_45hd_heatpump.yaml, but it didn't display/allow selection of the correct different modes available (it showed Heat and Cool), such as Electric (resistive immersion), heat pump, performance (both) and Off.

adfhogan:
Added additional update whilst PR pending.. Exposed the set and current water temperature DPs as standalone sensors as well as part of the water_heater entity, such that they can benefit from home assistant's long term logging and not be wiped after ~10 days.

make-all:
Extra sensors are not needed. You can use template sensors if you really want them.

adfhogan:
@make-all the device definition I started out with, its_45hd_heatpump.yaml also duplicates DP21 though not the setpoint DP4 so there is existing precedent for at least the water temp being duplicated.

Was this what you meant by extra sensors should be templates, or did you mean all the extra diagnostic stuff?

adfhogan:
To check whether "Inlet temperature" duplicating `current_temperature` in `its_45hd_heatpump.yaml` was a one-off or an actual convention, I scanned all 1,632 device configs in this repo for the pattern: a DP used by a primary entity (`climate`/`water_heater`/`humidifier`/etc.) that's *also* exposed as a standalone `sensor`/`binary_sensor`.

Excluded two categories of false positive: DPs shared only via a `temperature_unit` toggle feeding the `unit` field of unrelated diagnostic sensors, and composite base64/bitmask DPs (e.g. `powerworld_pw58410_*`) where different `mask`s decode genuinely different values from the same DP id — that's decomposition, not duplication.

**14 configs duplicate a current-value DP as a standalone sensor:**
- `its_45hd_heatpump.yaml` — "Inlet temperature" (the original fuzzy-match for this device)
- `aquastrong_hex_heatpump.yaml`, `solareast_aurora2_heatpump.yaml` — "Inlet temperature"
- `dr_heater_dr008_thermostat.yaml`, `drexma_et7aw_thermostat.yaml`, `jiahong_et61plus_thermostat.yaml`, `jiahong_et72w_thermostat.yaml`, `eurom_sani_bathroom_towel_radiator.yaml` — "Room temperature"
- `elko_cfmtb_thermostat.yaml` — "ambient_temperature"
- `atorch_s1tw_thermostat.yaml`, `giatsu_vaw2_waterheater.yaml`, `gotzejensen_kt975k_kettle.yaml` — unnamed sensor
- `alecoair_pu55_humino.yaml`, `electriq_cd12pw_dehumidifier.yaml` — current_humidity/current_temperature, unnamed

In every one of these, only the **current/measured** value is duplicated. None duplicate a **target/setpoint** temperature as a standalone sensor — that holds across all 1,632 files, not just this one.

Given that, would you be OK with keeping the "Tank temperature" (current, DP21) sensor since it matches this established convention, and I drop the "Target temperature" (setpoint, DP4) sensor which has no precedent anywhere in the repo?

make-all:
No, duplicate sensors should not be added.  Of the 14 / 1632 files you have identified that appear to go against convention, one was already deprecated, and 6 are not exactly duplicating the current_temperature, as there are complex conditions that cause different temperature to be displayed by the climate entity depending on the mode of the device. The other 7 have been deprecated now.

adfhogan:
Precedent deprecated.. got it.. I do honestly appreciate the clarification and will submit a fresh PR without the sensor duplication for the Aqua Plus APAIO270.

PR #5550, opened 2026-07-11
feat(devices): add support for Tower Fan Yoer Verto TF01

chrRtg:
Device Support for Yoer Verto TF01 tower fan. (https://yoer.pl/en/tower-fan-yoer-verto-tf01w-2178.html)

All functions tested.


make-all:
Seems mostly compatible with brandson_a307362x3_fan, only the ionizer was missing from that config

Issue #5270, opened 2026-06-07
power usage for ev charger Afyeev 16A


wallb35:
### Description

power should be around 1100W, I see 0.001 kW in the picture below.
is it possible to fix this ?
regards
Yannig

<img width="1143" height="1012" alt="Image" src="https://github.com/user-attachments/assets/706f256b-f128-4789-bb63-f0a794ad8538" />

### Additional context

in file tuya_local/devices/afyeev_16a_evcharger.yaml, if I change the following :

  - entity: sensor
    class: power
    dps:
      - id: 9
        type: integer
        name: sensor
        unit: kW
        class: measurement
        optional: true
        mapping:
          - scale: 1000

to 

  - entity: sensor
    class: power
    dps:
      - id: 9
        type: integer
        name: sensor
        unit: kW
        class: measurement
        optional: true
        mapping:
          - scale: 1

the result is better but I’ve got only steps of 1000W.

### Confirmation

- [ ] I am sure this is a bug or improvement, that is well enough described that it can be implemented.
- [x] This report is for make-all/tuya-local, **not** for localtuya, a modified tuya-local not from make-all, the Tuya app or cloud.
- [x] The device this issue is about is externally powered, or connecting via a hub (battery powered WiFi devices are NOT supported by this integration).
- [x] I have checked the latest released version of tuya-local
- [ ] If this is for a specific device, the device config filename is mentioned above, or diagnostics are included.

- [ ] Log messages or diagnostics relevant to the issue are included.

- [x] This is not requesting addition of a new device.

make-all:
The original submission did not have full info from platform.tuya.com, so I can't say whether the current scaling is correct or not, but it seems highly unlikely that the device is only reporting 1kW granularity on a charger that should only go to 7kW or 11kW max.

If you can, please provide Query Things Data Model results for the device from the API explorer.

wallb35:
Query Things Data Model results :

{
  "result": {
[...]

PR #5552, opened 2026-07-11
Added delta sensor with total and average mode

camel1cz:
## Summary

This PR adds support for delta-based sensor processing.

Some Tuya devices report energy consumption as interval/delta values instead of a continuously increasing total. This change allows such values to be handled in two ways:

* `delta_function: total` accumulates interval deltas into a `total_increasing` sensor.
* `delta_function: average` converts interval energy deltas into an average power value over a configurable time window.

The default behavior remains unchanged: delta sensors use `delta_function: total` unless configured otherwise.

## Related issues
The following issues may be related, and this PR could improve processing for the mentioned devices. I have not changed their YAML files because I do not have those devices available for testing.

* [#4623 Request support for Pioneer Quantum Wall AC - WYT-25 Quantum Series](https://github.com/make-all/tuya-local/issues/4623)
* [#4634 Request support for Pioneer Quantum Wall AC](https://github.com/make-all/tuya-local/issues/4634)
* [#5270 power usage for ev charger Afyeev 16A](https://github.com/make-all/tuya-local/issues/5270)

## Testing environment

Tested with:

* Kaisai Pro Heat+ AC
* Device config: `kaisai_proheat_airconditioner.yaml`

## Example

```yaml
- entity: sensor
  mode: delta
  delta_function: total
  name: Total energy
  class: energy
  dps:
    - id: 127
      name: sensor
      type: integer
      unit: kWh
      class: total_increasing
      mapping:
        - scale: 100
    - id: 128
      name: delta_start_key
      type: integer
    - id: 129
      name: delta_key
      type: integer

- entity: sensor
  mode: delta
  delta_function: average
  averaging_window: 600
  name: Power
  class: power
  dps:
    - id: 127
      name: sensor
      type: integer
      unit: W
      class: measurement
      mapping:
        - scale: 100
    - id: 128
      name: delta_start_key
      type: integer
    - id: 129
      name: delta_key
      type: integer
```

For the `average` function, the source DP is still an interval energy value, but the resulting sensor value is average power, so the entity uses `unit: W` and `class: power`.

## Testing

Added unit tests covering:

* total accumulation from delta intervals
* duplicate `delta_key` protection
* default `delta_function: total`
* rolling average power calculation
* `averaging_window: 0` behavior
* fallback for invalid `delta_function`

Tested with:

```bash
pytest -vv tests/test_sensor_delta.py
```

Result:

```text
6 passed
```

### DIFF
--- a/custom_components/tuya_local/devices/kaisai_proheat_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/kaisai_proheat_airconditioner.yaml
@@ -153,6 +153,49 @@ entities:
           - scale: 100
 
   - entity: sensor
+    mode: delta
+    delta_function: total
+    name: Total energy
+    class: energy
+    dps:
+      - id: 127
+        name: sensor
+        type: integer
+        unit: kWh
+        class: total_increasing
+        mapping:
+          - scale: 100
+      - id: 128
+        name: delta_start_key
+        type: integer
+      - id: 129
+        name: delta_key
+        type: integer
+
+  - entity: sensor
+    mode: delta
+    delta_function: average
+    averaging_window: 400
+    name: Power
+    class: power
+    dps:
+      - id: 127
+        name: sensor
+        type: integer
+        unit: W
+        class: measurement
+        mapping:
+          - scale: 100
+
+      - id: 128
+        name: delta_start_key
+        type: integer
+
+      - id: 129
+        name: delta_key
+        type: integer
+
+  - entity: sensor
     class: energy
     dps:
       - id: 127
--- a/custom_components/tuya_local/sensor.py
+++ b/custom_components/tuya_local/sensor.py
@@ -3,12 +3,14 @@ Setup for different kinds of Tuya sensors
 """
 
 import logging
+from collections import deque
 
 from homeassistant.components.sensor import (
     STATE_CLASSES,
     SensorDeviceClass,
     SensorEntity,
 )
+from homeassistant.helpers.restore_state import RestoreEntity
 
 from .device import TuyaLocalDevice
 from .entity import TuyaLocalEntity, unit_from_ascii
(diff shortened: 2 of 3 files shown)


make-all:
This request goes beyond the scope of this integration, which is to bridge the data coming over Tuya local protocol into Home Assistant.

Supporting this integration already takes more time than there is available from the maintainer, so adding additional features to clean up Tuya data by tracking it long term is just adding support burden from the point of view of the maintainer.

camel1cz:
Thanks for the explanation. I understand the concern about long-term maintenance burden and I agree that the integration should primarily expose data from the Tuya local protocol rather than grow into a data-cleanup layer.

My motivation here was that some devices appear to expose energy consumption only as interval-based values, together with start/end timestamps, so the raw local data is not directly usable as a Home Assistant energy sensor without minimal state tracking.

That said, I understand if the generic `delta_function: average` part is considered out of scope. Would a smaller PR be acceptable if it only handled the accumulated total energy case?

In other words:

* keep `mode: delta`
* support only accumulating interval deltas into a `total_increasing` energy sensor
* remove the average power calculation
* remove `averaging_window`
* keep the implementation and tests focused only on avoiding duplicate counting for the same `delta_key`

If even that is outside the intended scope of the integration, I completely understand and will maintain this as a local patch/fork instead.

make-all:
No, accumulation is also maintaining state in the integration.
Does this not work already by using the start time as last_reset attribute on the entity, and using `class: total` on the sensor dp?

camel1cz:
I thought more about this case and about the scope you described.

I decided to move the accumulation and average power calculation into a small separate custom integration instead.

Would you be open to a tiny YAML-only PR for the Kaisai Pro Heat+ AC device that only exposes the missing raw interval timestamp DPs as regular sensors?

make-all:
Yaml only pr is fine, but since the timestamps are linked to a specific sensor value, it might be better to put them as attributes there.

camel1cz:
I’ve opened a follow-up PR with the suggested changes: #5579.

PR #5559, opened 2026-07-12
Add Woods MRD20GW dehumidifier device config

AlexNaga:
The Woods MRD20GW is a variant of the MRD25GW with an extended humidity range (40-70% vs 40-50%).

This adds a separate device config for the MRD20GW with:
- Humidity target range 40-70%
- Dedicated tank full binary sensor (DP19 bit 4, confirmed on hardware: fault_code=16 when tank is full)

Product ID: `g5nx6i7qxfale9t7`

Tested on a live MRD20GW unit.


make-all:
Most likely the same as MRD25GW.
The information for both models says "Mellan 30% rh och 80% rh", and the tank indicator had just not been identified by the contributor for the MRD25GW.

PR #5562, opened 2026-07-12
Add IMAGINE 3+1 WiFi Switch device config

gurmeherchawla:
Adds support for the IMAGINE 3+1 WiFi Switch (model IT31LFW, product_id haqyilpinwntqgzj), a 3-gang switchboard with an integrated fan speed regulator.

Dp mapping confirmed directly from the device's live diagnostics dump (not guessed): dps 1, 2 and 3 are the switch relays (boolean); dps 9, 10 and 11 are the matching per-switch auto-off timers in seconds; dp 101 is fan power (boolean); dp 102 is fan speed as a string enum (off/level_1/level_2/level_3/level_4), mapped here to 0/25/50/75/100; dp 103 is the fan's own auto-off timer in seconds.

Single device, single branch, no changes to DEVICES.md or ACKNOWLEDGEMENTS.md.

### DIFF
--- a/custom_components/tuya_local/devices/imagine3+1_switch.yaml
+++ b/custom_components/tuya_local/devices/imagine3+1_switch.yaml
@@ -3,7 +3,6 @@ products:
   - id: haqyilpinwntqgzj
     manufacturer: Imagine
     model: IT31LFW
-    name: IMAGINE 3+1 WiFi Switch
 
 # Built from real diagnostics dump:
 #   1,2,3   = switch relays (boolean)
@@ -68,7 +67,7 @@ entities:
         name: second
         range:
           min: 0
-          max: 86400
+          max: 86399
 
   - entity: time
     category: config
@@ -80,7 +79,7 @@ entities:
         name: second
         range:
           min: 0
-          max: 86400
+          max: 86399
 
   - entity: time
     category: config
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/imagine3+1_switch.yaml
```diff
@@ -0,0 +1,106 @@
+name: Switchboard with fan
+products:
+  - id: haqyilpinwntqgzj
+    manufacturer: Imagine
+    model: IT31LFW
+    name: IMAGINE 3+1 WiFi Switch
```
make-all: name is not needed here

custom_components/tuya_local/devices/imagine3+1_switch.yaml
```diff
@@ -0,0 +1,106 @@
...
+  - entity: time
+    category: config
+    name: Fan timer
+    dps:
+      - id: 103
+        type: integer
+        name: second
+        range:
+          min: 0
+          max: 86400
```
make-all: Technically time entities can only go to 86399 (23:59:59, as they wrap to 0:00:00 at 86400)

make-all (changes requested):

Issue #3241, opened 2025-05-14
Request support for Smart Sprinkler Controller 16 zones


LuisSanz98:
### Log message

```text
"data": {
    "endpoint": "https://apigw.tuyaeu.com",
    "terminal_id": "1743327412391oZJwZg",
    "mqtt_connected": true,
    "disabled_by": null,
    "disabled_polling": false,
    "id": "",
    "name": "Smart Sprinkler Controller",
    "category": "ggq",
    "product_id": "ajg67sq0tpjxgd40",
    "product_name": "Smart Sprinkler Controller",
    "online": true,
    "sub": false,
    "time_zone": "+02:00",
    "active_time": "2025-05-14T06:36:29+00:00",
    "create_time": "2025-05-14T06:36:29+00:00",
    "update_time": "2025-05-14T06:36:29+00:00",
    "function": {
      "switch_1": {
        "type": "Boolean",
        "value": {}
      },
      "switch_2": {
        "type": "Boolean",
        "value": {}
      },
      "switch_3": {
        "type": "Boolean",
        "value": {}
      },
      "switch_4": {
        "type": "Boolean",
        "value": {}
      },
      "switch_5": {
        "type": "Boolean",
        "value": {}
      },
      "switch_6": {
        "type": "Boolean",
        "value": {}
      },
      "switch_7": {
        "type": "Boolean",
        "value": {}
      },
      "switch_8": {
        "type": "Boolean",
        "value": {}
      },
      "switch_9": {
        "type": "Boolean",
        "value": {}
      },
      "switch_10": {
        "type": "Boolean",
        "value": {}
      },
      "switch_11": {
        "type": "Boolean",
        "value": {}
      },
      "switch_12": {
        "type": "Boolean",
        "value": {}
      },
      "weather_delay": {
        "type": "Enum",
        "value": {
          "range": [
            "24h",
            "48h",
            "72h",
            "cancel"
          ]
        }
      },
      "normal_timer": {
        "type": "String",
        "value": {
          "maxlen": 255
        }
      },
      "water_control": {
        "type": "Enum",
        "value": {
          "range": [
            "on",
            "off"
          ]
        }
      },
      "smart_weather": {
        "type": "Enum",
        "value": {
          "range": [
            "sunny",
            "cloudy",
            "rainy",
            "snowy"
          ]
        }
      },
      "weather_switch": {
        "type": "Boolean",
        "value": {}
      },
      "control_skip": {
        "type": "Boolean",
[...]

PR #5563, opened 2026-07-12
Add NEO Coolcam 16-zone sprinkler controller

msemenkin:
Adds support for the NEO Coolcam Smart Sprinkler Controller v1.1.

Related support request: #3241

The value `1743327412391oZJwZg` shown in the issue's "Product ID"
section is actually the `terminal_id` from the submitted diagnostics.

The actual device Product ID in those diagnostics is:

`ajg67sq0tpjxgd40`

This matches the Product ID tested with this configuration. The issue
device also has the same Tuya category (`ggq`), product name, standard
DP set, and 16-zone hardware behaviour.

Vendor product page:
https://www.szneo.com/en/products/show.php?id=272

Brand information:
https://www.szneo.com/en/about/index.php?id=9

The vendor identifies NEO Coolcam as the product brand. The company website
also uses the names NEO Smart Technology (Guangdong) Co., Ltd. and
Shenzhen NEO Electronic Co., Ltd.

Tested environment:

- Home Assistant 2026.7.2
- Tuya Local 2026.7.1
- TinyTuya 1.20.0
- Local Tuya protocol 3.4

Product identification:

- Product ID / productKey: ajg67sq0tpjxgd40
- Tuya category: ggq
- Product name: Smart Sprinkler Controller
- Hardware variant: 16 zones

Verified local DP mapping:

- DP 1-12: valves 1-12
- DP 101-104: valves 13-16
- DP 37: weather delay
- DP 38: irrigation schedule, raw/hidden
- DP 39: work state
- DP 42: weather switch
- DP 43: skip control
- DP 44: irrigation mode
- DP 46: irrigation restriction schedule, raw/hidden
- DP 105: countdown
- DP 106: irrigation use time
- DP 107: valve fault event

DP 105 and DP 106 contain four Base64-encoded bytes:

- byte 1: valve number
- bytes 2-4: duration in seconds, big-endian

Examples verified against actual operation:

- BQAABw== -> valve 5, 7 seconds
- CAAAOw== -> valve 8, 59 seconds remaining
- CAAABQ== -> valve 8, 5 seconds used

Valves 1-12 were tested with physically connected irrigation valves.
DP 101-104 were verified as valves 13-16; unconnected outputs produced
the expected alram13-alram16 fault events.

Known device behaviour:

The controller physically energizes only one valve at a time, but when
multiple valves are commanded directly, some boolean DP states may remain true. This configuration reports the DP states provided by the device.

Neo1
Neo2
Neo3

### DIFF
--- a/custom_components/tuya_local/devices/neo_coolcam_16zone_sprinkler_controller.yaml
+++ b/custom_components/tuya_local/devices/neo_coolcam_16zone_sprinkler_controller.yaml
@@ -1,4 +1,4 @@
-name: 16-zone sprinkler controller
+name: Sprinkler controller
 products:
   - id: ajg67sq0tpjxgd40
     manufacturer: NEO Coolcam
@@ -166,6 +166,7 @@ entities:
 
   - entity: select
     name: Weather delay
+    translation_key: timer
     icon: "mdi:weather-pouring"
     category: config
     dps:
@@ -238,9 +239,9 @@ entities:
         optional: true
         mapping:
           - dps_val: Order1
-            value: Mode 1
+            value: Independent valves
           - dps_val: Order2
-            value: Mode 2
+            value: Master valve
 
   - entity: text
     name: Irrigation restriction schedule
@@ -298,18 +299,17 @@ entities:
         class: measurement
         optional: true
 
-  - entity: sensor
+  - entity: event
     name: Fault
-    class: enum
     category: diagnostic
     dps:
       - id: 107
         type: string
-        name: sensor
+        name: event
         optional: true
         mapping:
           - dps_val: normal
-            value: Clear
+            value: clear
           - dps_val: alram1
             value: Valve 1
           - dps_val: alram2
--- a/custom_components/tuya_local/devices/neo_coolcam_16zone_sprinkler_controller.yaml
+++ b/custom_components/tuya_local/devices/neo_coolcam_16zone_sprinkler_controller.yaml
@@ -1,4 +1,4 @@
-name: 16 zone sprinkler controller
+name: 16-zone sprinkler controller
 products:
   - id: ajg67sq0tpjxgd40
     manufacturer: NEO Coolcam
@@ -167,7 +167,6 @@ entities:
   - entity: select
     name: Weather delay
     icon: "mdi:weather-pouring"
-    translation_key: timer
     category: config
     dps:
       - id: 37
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/neo_coolcam_16zone_sprinkler_controller.yaml
```diff
@@ -0,0 +1,344 @@
+name: 16-zone sprinkler controller
```
make-all: It is better to avoid starting names with numbers, as users can use their HA in various contexts including external systems, where we don't know if this will cause problems.

Just "Sprinkler controller" is probably enough for the device name, as these should also be concise for when they get prepended to entity names.

custom_components/tuya_local/devices/neo_coolcam_16zone_sprinkler_controller.yaml
```diff
@@ -0,0 +1,344 @@
...
+      x: "16"
+    dps:
+      - id: 104
+        type: boolean
+        name: valve
+
+  - entity: select
+    name: Weather delay
+    icon: "mdi:weather-pouring"
+    category: config
```
make-all, line 171: Use `translation_key: timer` to get translations for the options below

custom_components/tuya_local/devices/neo_coolcam_16zone_sprinkler_controller.yaml
```diff
@@ -0,0 +1,344 @@
...
+    icon: "mdi:sprinkler-variant"
+    category: config
+    dps:
+      - id: 44
+        type: string
+        name: option
+        optional: true
+        mapping:
+          - dps_val: Order1
+            value: Mode 1
```
make-all: It would be better to use descriptive names for the modes

custom_components/tuya_local/devices/neo_coolcam_16zone_sprinkler_controller.yaml
```diff
@@ -0,0 +1,344 @@
...
+    dps:
+      - id: 106
+        type: base64
+        mask: "00FFFFFF"
+        name: sensor
+        unit: s
+        class: measurement
+        optional: true
+
+  - entity: sensor
```
make-all: Probably this is better as an event (with normal mapped to `clear`)

PR #5565, opened 2026-07-12
Add KGM 12000-58 Portable Air Conditioner (OBI)

mrhund:
Added KGM 12000-58 Portable Air Conditioner which is sold by OBI in DACH region: 
https://www.obi.at/p/3795747/mobiles-klimageraet-kgm-12000-58-inkl-fernbedienung-mit-wifi-und-heizfunktion
Internal Name from iot.tuya.com: JWBR2S-5V-CH

### DIFF
--- a/custom_components/tuya_local/devices/KGM_12000-58_OBI_portable_air_conditioner.yaml
+++ b/custom_components/tuya_local/devices/KGM_12000-58_OBI_portable_air_conditioner.yaml
@@ -2,7 +2,7 @@ name: Air conditioner
 products:
   - id: JWBR2S-5V-CH
     manufacturer: OBI
-    model: JWBR2S-5V-CH
+    model: KGM 12000-58
     name: Portable air conditioner
 entities:
   - entity: climate
--- a/custom_components/tuya_local/devices/KGM_12000-58_OBI_portable_air_conditioner.yaml
+++ b/custom_components/tuya_local/devices/KGM_12000-58_OBI_portable_air_conditioner.yaml
@@ -46,13 +46,13 @@ entities:
             conditions:
               - dps_val: true
                 value_redirect: temp_current_f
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/KGM_12000-58_OBI_portable_air_conditioner.yaml
```diff
@@ -0,0 +1,126 @@
+name: Air conditioner
+products:
+  - id: JWBR2S-5V-CH
```
make-all: This is not a valid product id. You can find the actual one in the Query Device Details cloud API call, or by running tinytuya scan locally (may require that the device is not connected locally)

make-all (commented):
Are there any other dps reported by the device? In particular 110 and 123 have been used by other similar devices to indicate supported features.

I am wondering if it might be possible to merge this with the existing klarstein_iceblock_airconditioner config, which appears identical except for the absence of heat mode. Often there is a flag to indicate whether a model is a heat/cool model or a cool-only model for air conditioners, so that several models can share the same WiFi module.

make-all (changes requested):

mrhund:
@make-all Thanks you for the feedback. 

Available DPs are: 
{'1': False, '2': 19, '3': 25, '17': 32, '20': 0, '101': '1', '103': False, '104': '1', '105': 0, '106': False, '109': False, '110': 66, '111': 75, '112': 32}

...so 126 is not present. 106 is present, but the device has no "swing" funcionality, so I left it out. Also there is no dp to find out if heating functionality is there or not, so similarity to the Klarstein air conditioner is imho limited.

ID will be updated.

make-all:
Actually electriq_airflex15w_heatpump appears to be the best match. It has heat mode, but includes the missing humidity and swing mode controls as well as an "auto" mode (mapped to heat_cool hvac_mode), which will have to be ignored for this model.

PR #5560, opened 2026-07-12
Add support for AirSensor MT13W multi-function air quality sensor

LinusIta:
This adds support for a Tuya-based multi-function air quality sensor, sold as "AirSensor MT13W" (product_id: `tf5vafuyq3hiibs4`).
The device is a desktop air quality monitor with an LCD display showing temperature, humidity, CO2, PM2.5, PM10, HCHO (formaldehyde), TVOC, and an overall air quality level indicator.
Testing:
This config has been tested and is working correctly on real hardware.
DPS values were confirmed by cross-referencing the raw DPS dump against the values shown on the device's physical LCD display side by side.
Mapped and verified DPS:
`1`  -> overall air quality level (enum: level_1-4, mapped to Good/Slight/Moderate/Serious)
`2`  -> temperature (°C, direct value)
`3`  -> humidity (%, direct value)
`4`  -> CO2 (ppm, direct value)
`5`  -> formaldehyde/HCHO (mg/m³, scale 1000)
`7`  -> PM2.5 (µg/m³, direct value)
`9`  -> PM10 (µg/m³, direct value)
`23` -> display on/off (boolean)
`101` -> TVOC (mg/m³, scale 1000)

Note: DPS `1` (air quality level) required `force: true` since the device does not proactively resend this value once connected if it stays stable, causing the entity to remain unavailable after a fresh connection/restart until explicitly polled.
Other DPS observed in the raw dump but not yet confidently identified (left unmapped for now): `102`, `104`-`107`, `112`-`115`. Happy to look into these further if useful.
Test file:
I don't currently have a test file for this config — I wasn't able to confirm the exact test harness pattern used in the current version of the repo. Happy to add one if you can point me to a recent example to follow, or feel free to add it during review.
Let me know if any changes are needed!

PR #5570, opened 2026-07-12
Add support for ZN-MT13 air quality monitor

mirchik:
## **Summary**

Adds support for the ZN-MT13 air quality monitor.

## **Testing**

Tested with a physical ZN-MT13 device using protocol version 3.5.
The configuration is based on a tuya-local diagnostics dump.


make-all:
Duplicate of #5560

PR #5575, opened 2026-07-13
feat(devices): Add support for IllumiX Outdoor Eaves Lights

TN-1:
This PR adds support for IlluminX Outdoor Eaves Lights: https://www.giftbox.com.au/smart-outd-eaves-string-light-18-bulb
Implements all effects that come out of the box, However if they are edited in the Tuya app then the effect will be inaccesible.

### DIFF
--- a/custom_components/tuya_local/devices/illuminx_outdoor_eaves_lights.yaml
+++ b/custom_components/tuya_local/devices/illuminx_outdoor_eaves_lights.yaml
@@ -70,8 +70,7 @@ entities:
         unit: s
 
   - entity: select
-    name: Scene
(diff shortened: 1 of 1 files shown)


TN-1:
@make-all I dont think the testing failure is relating to my code, I think the CI has bugged out, seems to be happening a bunch recently. I have the config working on my local HA instance

Otherwise, Im not sure how to identify where the error is because the logs doesnt point to anything.

make-all:
Merged with existing lscsmartconnect_partystring_light.yaml which has the same string handling dps.

TN-1:
Thanks!

PR #5579, opened 2026-07-13
Kaisai energy attribs

camel1cz:
Follow-up to #5552, based on the discussion there.

I’ve moved the timestamps to attributes of the energy sensor, as suggested. This YAML-only change also renames the sensor to `Add electricity` for consistency with other devices, sets it as diagnostic and hidden by default, and removes the `total_increasing` state class because the raw value is difficult to use directly.

The related `delta_energy` integration is available here: https://github.com/camel1cz/delta_energy


make-all:
Please try the version committed. If that can work, it is a better solution I think

camel1cz:
I can confirm your solution does work as expected with sensor like below:

```
utility_meter:
  clima_energy_total:
    source: sensor.xxx_energy
    name: Clima energy total
    delta_values: true
    periodically_resetting: true
    always_available: true
```

PR #5589, opened 2026-07-14
Add CONOW CBE2000 Pro Solar Battery device definition file

marc-schieferdecker:
# Summary
* Add [CONOW CBE2000 Pro](https://www.conow.com/products/cbe-2000-pro-solar-storage) Solar Battery to Tuya Local.

# Why
* The standard Tuya integration of Home Assistant is very limited for the CBE2000 Pro, many datapoints missing and you can't even set the base load

# Testing
* Tested with the CBE2000 Pro I have and Tuya Local Version 2026.7.1
  * All inputs/selects tested ✓
  * Loading the battery from grid via automation ✓
  * Feeding the grid by automation ✓

# Screenshots

image

image

image

image

image

### DIFF
--- a/custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
+++ b/custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
@@ -149,7 +149,7 @@ entities:
     class: power
     dps:
       - id: 20
-        type: float
+        type: integer
         name: sensor
         unit: W
   - entity: sensor
@@ -157,7 +157,7 @@ entities:
     class: power
     dps:
       - id: 21
-        type: float
+        type: integer
         name: sensor
         unit: W
   - entity: sensor
@@ -165,7 +165,7 @@ entities:
     class: power
     dps:
       - id: 22
-        type: float
+        type: integer
         name: sensor
         unit: W
   - entity: sensor
@@ -173,7 +173,7 @@ entities:
     class: power
     dps:
       - id: 36
-        type: float
+        type: integer
         name: sensor
         unit: W
   - entity: sensor
@@ -181,7 +181,7 @@ entities:
     class: power
     dps:
       - id: 37
-        type: float
+        type: integer
         name: sensor
         unit: W
   - entity: sensor
--- a/custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
+++ b/custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
@@ -20,6 +20,7 @@ entities:
     class: problem
     dps:
       - id: 11
+        optional: true
         type: bitfield
         name: sensor
         mapping:
@@ -30,9 +31,11 @@ entities:
         type: string
         name: error_code
       - id: 11
+        optional: true
         type: bitfield
         name: fault_code
       - id: 11
+        optional: true
         type: bitfield
         name: description
         mapping:
@@ -145,6 +148,7 @@ entities:
     class: occupancy
     dps:
       - id: 18
+        optional: true
         type: boolean
         name: sensor
   - entity: sensor
@@ -242,6 +246,7 @@ entities:
     class: power
     dps:
       - id: 28
+        optional: true
         type: integer
         name: sensor
         unit: W
@@ -250,6 +255,7 @@ entities:
     class: power
     dps:
       - id: 29
+        optional: true
         type: integer
         name: sensor
         unit: W
@@ -397,6 +403,7 @@ entities:
     class: energy
     dps:
       - id: 47
+        optional: true
         type: integer
         name: sensor
         unit: kWh
@@ -530,6 +537,7 @@ entities:
     category: config
     dps:
       - id: 65
+        optional: true
         type: string
         name: option
         mapping:
@@ -555,6 +563,7 @@ entities:
     category: config
     dps:
       - id: 69
+        optional: true
         type: integer
         name: value
         unit: W
@@ -566,6 +575,7 @@ entities:
     category: config
     dps:
       - id: 70
+        optional: true
         type: boolean
         name: switch
   - entity: select
@@ -573,6 +583,7 @@ entities:
     category: config
     dps:
       - id: 73
+        optional: true
         type: string
         name: option
         mapping:
@@ -587,6 +598,7 @@ entities:
     category: config
     dps:
       - id: 78
+        optional: true
         type: string
         name: option
         mapping:
@@ -599,6 +611,7 @@ entities:
     category: diagnostic
     dps:
       - id: 84
+        optional: true
         type: integer
         name: sensor
         unit: W
@@ -611,6 +624,7 @@ entities:
     hidden: true
     dps:
       - id: 85
+        optional: true
         type: boolean
         name: switch
   - entity: number
@@ -619,6 +633,7 @@ entities:
     category: config
     dps:
       - id: 91
+        optional: true
         type: integer
         name: value
         unit: W
--- a/custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
+++ b/custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
@@ -7,6 +7,7 @@ entities:
   - entity: sensor
     category: diagnostic
     name: Battery capacity
+    class: energy_storage
     dps:
       - id: 2
         type: integer
@@ -15,7 +16,6 @@ entities:
         mapping:
           - scale: 1000
   - entity: binary_sensor
-    name: Fault
     category: diagnostic
     class: problem
     dps:
--- a/custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
+++ b/custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
@@ -5,15 +5,8 @@ products:
     model: CBE2000 Pro
 entities:
   - entity: sensor
-    name: Serial Number
     category: diagnostic
-    dps:
-      - id: 1
-        type: string
-        name: sensor
-  - entity: sensor
-    category: diagnostic
-    name: Battery Capacity
+    name: Battery capacity
     dps:
       - id: 2
         type: integer
@@ -21,21 +14,27 @@ entities:
         unit: kWh
         mapping:
           - scale: 1000
-  - entity: sensor
-    name: Error Code
-    category: diagnostic
-    dps:
-      - id: 10
-        type: string
-        name: sensor
-  - entity: sensor
+  - entity: binary_sensor
     name: Fault
     category: diagnostic
-    class: enum
+    class: problem
     dps:
       - id: 11
         type: bitfield
         name: sensor
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+      - id: 10
+        type: string
+        name: error_code
+      - id: 11
+        type: bitfield
+        name: fault_code
+      - id: 11
+        type: bitfield
+        name: description
         mapping:
           - dps_val: 0
             value: No Fault
@@ -83,8 +82,11 @@ entities:
             value: meter_comm_failure
           - dps_val: 2097152
             value: system_temp_diff_protection
+      - id: 1
+        type: string
+        name: serial_number
   - entity: sensor
-    name: Connection State
+    name: Connection state
     category: diagnostic
     class: enum
     hidden: true
@@ -100,7 +102,7 @@ entities:
           - dps_val: paired_uncon
             value: Paired and not connected
   - entity: sensor
-    name: Signal Strength
+    class: signal_strength
     category: diagnostic
     hidden: true
     dps:
@@ -109,7 +111,7 @@ entities:
         name: sensor
         unit: dBm
   - entity: sensor
-    name: Hardware Version
+    name: Hardware version
     category: diagnostic
     hidden: true
     dps:
@@ -117,35 +119,36 @@ entities:
         type: string
         name: sensor
   - entity: sensor
-    name: System Version
+    name: System version
     category: diagnostic
     hidden: true
     dps:
       - id: 15
         type: string
         name: sensor
-  - entity: sensor
-    name: Cell Heating
+  - entity: binary_sensor
+    name: Cell heating
     category: diagnostic
     dps:
       - id: 16
         type: boolean
         name: sensor
   - entity: sensor
-    name: Expansion Pack Count
+    name: Expansion pack count
     category: diagnostic
     dps:
       - id: 17
         type: integer
         name: sensor
-  - entity: sensor
-    name: Human Presence
+  - entity: binary_sensor
+    name: Human presence
+    class: occupancy
     dps:
       - id: 18
         type: boolean
         name: sensor
   - entity: sensor
-    name: PV Power Total
+    name: PV power total
     class: power
     dps:
       - id: 20
@@ -153,7 +156,7 @@ entities:
         name: sensor
         unit: W
   - entity: sensor
-    name: PV Power Port 1
+    name: PV power port 1
     class: power
     dps:
       - id: 21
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
+name: Battery storage
+products:
+  - id: c4ilzd7aybycece9
+    manufacturer: Conow
+    model: CBE2000 Pro
+entities:
+  - entity: sensor
+    name: Serial Number
```
make-all: Static info like serial and version numbers do not need to be entities. Just add them as attributes on another entity.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+          - scale: 1000
+  - entity: sensor
+    name: Error Code
+    category: diagnostic
+    dps:
+      - id: 10
+        type: string
+        name: sensor
+  - entity: sensor
+    name: Fault
```
make-all: Standard pattern for this type of sensor is to make it a `binary_sensor` with `class: problem`, and a mapping of 0 to false, any other value to true, with a `fault_code` attribute providing the raw value, and optionally `description` providing the textual representation. Most other devices have an example of the first two, but description is only in maybe 1 in 10 configs.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+    name: Battery Capacity
+    dps:
+      - id: 2
+        type: integer
+        name: sensor
+        unit: kWh
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: Error Code
```
make-all: I'm not sure how useful this is, maybe it can be an attribute under the below `problem` sensor.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+  - entity: sensor
+    name: Serial Number
+    category: diagnostic
+    dps:
+      - id: 1
+        type: string
+        name: sensor
+  - entity: sensor
+    category: diagnostic
+    name: Battery Capacity
```
make-all: HA naming convention: device and entity names should only have the first word capitalized (unless there is a good reason like abbreviations or proper names)

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+        name: sensor
+        mapping:
+          - dps_val: paired_connected
+            value: Paired and connected
+          - dps_val: unpaired_uncon
+            value: Unpaired and not connected
+          - dps_val: paired_uncon
+            value: Paired and not connected
+  - entity: sensor
+    name: Signal Strength
```
make-all: Use `class: signal_strength` and no name.

custom_components/tuya_local/devices/conow_cbe2000pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+    dps:
+      - id: 15
+        type: string
+        name: sensor
+  - entity: sensor
+    name: Cell Heating
+    category: diagnostic
+    dps:
+      - id: 16
+        type: boolean
```
make-all, line 127: This should be a binary_sensor. Maybe `class: heat` would fit.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+        name: sensor
+  - entity: sensor
+    name: Expansion Pack Count
+    category: diagnostic
+    dps:
+      - id: 17
+        type: integer
+        name: sensor
+  - entity: sensor
+    name: Human Presence
```
make-all: This should also be a binary_sensor. Normally human presence sensors indicate `class: occupancy`, but I'm not sure whether that fits in this case, it may be closer to `class: tamper`

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+        type: integer
+        name: value
+        unit: "%"
+        range:
+          min: 0
+          max: 100
+        mapping:
+          - step: 1
+  - entity: select
+    name: Work Mode
```
make-all: Use `translation_key: mode` instead of this name

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+    category: config
+    dps:
+      - id: 51
+        type: string
+        name: option
+        mapping:
+          - dps_val: self_powered
+            value: Self Powered
+          - dps_val: time_of_use
+            value: Time of Use
```
make-all: `schedule` from the existing translations probably matches this option.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+          - dps_val: manual
+            value: Manual
```
make-all: `manual` (all lower case) matches the existing translation key.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+  - entity: sensor
+    name: Country Code
+    category: diagnostic
+    hidden: true
+    dps:
+      - id: 56
+        type: integer
+        name: sensor
+  - entity: select
+    name: Indicator Light Mode
```
make-all: `translation_key: light_mode` with values `auto`, `on` and `off` to use existing translations. If `auto` does not fit the `eco` mode, then leave it as `eco`

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: none
+            value: None
+          - dps_val: ty_lan
+            value: Lan
+          - dps_val: ty_rs485
+            value: RS485
+  - entity: switch
```
make-all: May be better as a button, as this seems like a one-off operation.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+    category: config
+    dps:
+      - id: 91
+        type: integer
+        name: value
+        unit: W
+        range:
+          min: 0
+          max: 99999
+        mapping:
```
make-all: step: 1 is default, so this mapping is not needed

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+    category: config
+    dps:
+      - id: 69
+        type: integer
+        name: value
+        unit: W
+        range:
+          min: 0
+          max: 2500
+        mapping:
```
make-all: step 1 is default, so this mapping is not needed

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+    category: config
+    dps:
+      - id: 64
+        type: integer
+        name: value
+        unit: W
+        range:
+          min: 0
+          max: 30000
+        mapping:
```
make-all: step 1 is default, this mapping is not needed

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+    category: config
+    dps:
+      - id: 55
+        type: integer
+        name: value
+        unit: W
+        range:
+          min: 0
+          max: 100000
+        mapping:
```
make-all: Step 1 is default, this mapping is not needed

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+    category: config
+    dps:
+      - id: 53
+        type: integer
+        name: value
+        unit: W
+        range:
+          min: 0
+          max: 100000
+        mapping:
```
make-all: step 1 is default, this mapping is not needed

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,640 @@
...
+    category: config
+    dps:
+      - id: 50
+        type: integer
+        name: value
+        unit: "%"
+        range:
+          min: 0
+          max: 100
+        mapping:
```
make-all: step 1 is default, this mapping is not needed

make-all (commented):
There are no optional dps here. On a device with this many dps it is very unusual for the device to always send them all. Does the config match if the product id is removed?

make-all (changes requested):

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
marc-schieferdecker: Moved to the fault entity as attribute.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
marc-schieferdecker: Fixed all.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
marc-schieferdecker: Moved to entity "Fault" as attribute.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
marc-schieferdecker: Thank you, I did not understand the mechanics because this is my first contribution to the project. Now I understand it better and it is very nice and I even added the descriptions.

custom_components/tuya_local/devices/conow_cbe2000pro_battery.yaml
marc-schieferdecker: Tested class "heat" but it does not fit really well. This is just a sensor which tells if the battery heating is "on" or "off" (if the system is located outside in the winter for example) and I think bool fits best.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
marc-schieferdecker: Occupancy fits well.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
marc-schieferdecker: "Auto" is okay for what it does. Fixed.

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
marc-schieferdecker: Button is indeed a better choice for that entity.

marc-schieferdecker:
All change requests are now implemented. Thank you for your advice @make-all :-)

marc-schieferdecker:
> There are no optional dps here. On a device with this many dps it is very unusual for the device to always send them all. Does the config match if the product id is removed?

Without the product id it does not match but I do not know which dps are optional or not. I queried the the device info and device properties info on the Tuya IOT portal and in the responses is no info about whats optional and what not so I added the entities that I think are usefull to have in Home Assistant (which is more what you have in den Conow App or SmartLife).

make-all:
You won't find this info in the cloud, you figure it out from the log message when you add the device. Any dps that are missing from the log message need to be optional.

marc-schieferdecker:
> You won't find this info in the cloud, you figure it out from the log message when you add the device. Any dps that are missing from the log message need to be optional.

Got it. I set all dps that are not present in the "Partial cloud device spec" json log message to be optional.

I addded the json data of the partial cloud spec for reference.

[CBE2000Pro-Partial-Cloud-Spec.json](https://github.com/user-attachments/files/30148327/CBE2000Pro-Partial-Cloud-Spec.json)

custom_components/tuya_local/devices/conow_cbe2000pro_battery.yaml
```diff
@@ -0,0 +1,642 @@
+name: Battery storage
+products:
+  - id: c4ilzd7aybycece9
+    manufacturer: Conow
+    model: CBE2000 Pro
+entities:
+  - entity: sensor
+    category: diagnostic
+    name: Battery capacity
```
make-all, line 9: Add `class: energy_storage`

custom_components/tuya_local/devices/conow_cbe2000_pro_battery.yaml
```diff
@@ -0,0 +1,642 @@
...
+    name: Battery capacity
+    dps:
+      - id: 2
+        type: integer
+        name: sensor
+        unit: kWh
+        mapping:
+          - scale: 1000
+  - entity: binary_sensor
+    name: Fault
```
make-all: remove this name, the HA built in translation for `problem` is better than a hard-coded name

make-all:
> Got it. I set all dps that are not present in the "Partial cloud device spec" json log message to be optional.

That is not really correct either.

You need to capture the log of what the device actually sends. Either the warning level log containing `LOCAL DPS` that appears when you add the device, or a debug level log containing `full_poll=true` captured during normal device operation.

This can be different from the basic info the cloud returns from the API calls accessible from a QR code login.

marc-schieferdecker:
> > Got it. I set all dps that are not present in the "Partial cloud device spec" json log message to be optional.
> 
> That is not really correct either.
> 
> You need to capture the log of what the device actually sends. Either the warning level log containing `LOCAL DPS` that appears when you add the device, or a debug level log containing `full_poll=true` captured during normal device operation.
> 
> This can be different from the basic info the cloud returns from the API calls accessible from a QR code login.

Okay I understand. I got this from the log:

```text
2026-07-20 10:51:51.276 DEBUG (MainThread) [custom_components.tuya_local.device] CBE2000 received {"1": "-- serial --", "2": 2048, "10": "0000", "11": 0, "12": "unpaired_uncon", "16": false, "17": 0, "18": false, "20": 0, "21": 0, "22": 0, "23": 7, "24": "idle", "25": 0, "26": 0, "27": 0, "30": 17, "31": 0, "32": 0, "33": 0, "34": 0, "35": 0, "36": 0, "37": 0, "38": 0, "40": 3760, "41": 16714, "42": 12116, "43": 12175, "45": 12701, "46": 0, "47": 0, "50": 5, "51": "diy", "53": 200, "54": false, "55": 0, "56": "1", "57": "eco", "59": false, "60": false, "63": false, "64": 800, "65": "scene", "68": "backup", "69": 325, "70": false, "73": "led_bypass", "78": "self_powered", "81": "1784537330", "82": 20, "83": true, "84": 800, "91": 0, "full_poll": true}
```

I will use this to set the optional dps later this day. Thank you for your advice on that.

marc-schieferdecker:
I updated the optional DPS based on the debug message above. Should now be correct.

PR #5593, opened 2026-07-15
Add Baxi Smart Link thermostat

msemenkin:
## Summary

- add a Tuya Local device profile for the Baxi Smart Link OpenTherm thermostat
- expose thermostat control, heating and domestic hot water state, boiler diagnostics, and supported configuration datapoints
- map Tuya bitmap faults and OpenTherm connectivity
- apply datapoint ranges and scaling from the Tuya device data model

## Why

The standard Tuya integration exposes only a limited subset of this thermostat's capabilities. This profile provides local access to heating, domestic hot water, modulation, temperature, and diagnostic datapoints.

The mappings were checked against the Tuya data model and live captures in idle, central heating, and domestic hot water operation. DP3 distinguishes central heating from domestic hot water, while DP108 reports burner modulation.

## Validation

- `ruff check .`
- `ruff check --select I .`
- `ruff format --check .`
- `yamllint custom_components/tuya_local/devices`
- `pytest tests/test_device_config.py` — 32 passed
image
image
image
image

### DIFF
--- a/custom_components/tuya_local/devices/baxi_smart_link_thermostat.yaml
+++ b/custom_components/tuya_local/devices/baxi_smart_link_thermostat.yaml
@@ -473,9 +473,19 @@ entities:
         type: integer
         name: sensor
 
-  - entity: water_heater
+  - entity: climate
     name: Domestic hot water
+    category: config
+    icon: "mdi:water-boiler"
     dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: heat
       - id: 121
         type: integer
         name: temperature
--- a/custom_components/tuya_local/devices/baxi_smart_link_thermostat.yaml
+++ b/custom_components/tuya_local/devices/baxi_smart_link_thermostat.yaml
@@ -71,7 +71,6 @@ entities:
 
   - entity: select
     name: Heating season
-    category: config
     icon: "mdi:sun-snowflake-variant"
     dps:
       - id: 28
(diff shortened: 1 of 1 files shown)


msemenkin:
> 👮‍♀️⚠️ Your PR seems to be larger than normal. Please check that you did not accidentally submit something you did not intend to. If you are submitting multiple new devices, please consider splitting them into separate PRs.

This PR intentionally contains a single device profile. The file is relatively large because the thermostat exposes more than 30 Tuya datapoints covering climate control, DHW, OpenTherm diagnostics, and configuration. No unrelated files are included.

custom_components/tuya_local/devices/baxi_smart_link_thermostat.yaml
```diff
@@ -0,0 +1,542 @@
...
+        name: max_temperature
+        range:
+          min: 200
+          max: 800
+        mapping:
+          - scale: 10
+            step: 5
+
+  - entity: select
+    name: Heating season
```
make-all, line 62: I'm not sure why you removed `category: config` in the last commit. This definitely seems like a config option that rarely needs changing (twice a year).

custom_components/tuya_local/devices/baxi_smart_link_thermostat.yaml
```diff
@@ -0,0 +1,542 @@
...
+    name: OpenTherm diagnostic value
+    category: diagnostic
+    hidden: true
+    dps:
+      - id: 120
+        type: integer
+        name: sensor
+
+  - entity: number
+    name: Domestic hot water target temperature
```
make-all: This could be part of a water_heater entity, along with some of the other dps

make-all (changes requested):

custom_components/tuya_local/devices/baxi_smart_link_thermostat.yaml
msemenkin: Thanks. I grouped DP121 (dhw_settemp) and DP122 (dhw_curtemp) into a climate entity instead. A temperature-only water_heater reports unknown (see first screenshot), while the device has no independent DHW operation mode: DP1 controls the whole thermostat, and DHW remains active in both summer and winter modes.
I kept it in the configuration category because the DHW setpoint is changed infrequently and was already categorized as configuration.

User attachment

image
image

custom_components/tuya_local/devices/baxi_smart_link_thermostat.yaml
msemenkin: You're right. I removed it because the season is shown as an operating mode on the physical panel, but it is still an infrequently changed setting. I've restored category: config.

Issue #4314, opened 2026-01-12
Request support for dé EV Charger


J14111:
### Log message

```text
Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:436
integration: Tuya Local (documentation, issues)
First occurred: 10:14:05 PM (1 occurrence)
Last logged: 10:14:05 PM

Device matches abalon_bcm700d_curtain with quality of 18%. DPS: {"updated_at": 1768256040.9380865, "101": 101, "102": "{\"L1\":[2360,0,0],\"L2\":[2360,0,0],\"L3\":[2360,0,0],\"t\":280,\"p\":0,\"d\":0,\"e\":0}", "106": "{\"r\":\"Type B, AC 30mA + DC 6mA\",\"fv\":\"1.6.2\",\"cp\":\"13.7\",\"t\":\"0\",\"e\":\"0\"}", "107": "[6, 8, 10, 13]", "150": 13, "151": "{\"m\":0,\"dt\":0,\"ss\":\"00:00\",\"se\":\"08:00\"}", "152": 13, "155": false, "156": false, "157": 1, "188": false}
Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:436
integration: Tuya Local (documentation, issues)
First occurred: 10:14:05 PM (1 occurrence)
Last logged: 10:14:05 PM

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:316
integration: Tuya Local (documentation, issues)
First occurred: 10:12:53 PM (1 occurrence)
Last logged: 10:12:53 PM

Could not find device: bf23bbdd4b1f0cba3adwic

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:412
integration: Tuya Local (documentation, issues)
First occurred: 10:14:05 PM (1 occurrence)
Last logged: 10:14:05 PM

Adding dé EV Charger device with product id gxrtu5vljdthtd3g

Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:436
integration: Tuya Local (documentation, issues)
First occurred: 10:14:05 PM (1 occurrence)
Last logged: 10:14:05 PM
```

### Product ID

gxrtu5vljdthtd3g

### Product Name

dé EV Charger gd version

### DPS information

```text
{
  "result": {
[...]

Issue #3152, opened 2025-04-16
Request support for dé Mobile Wallbox (1-phase 11kW)


Frostdead:
### Log message

```text
Device matches abalon_bcm700d_curtain with quality of 20%. DPS: {"updated_at": 1744808638.7530944, "101": 300, "102": "{\"L1\":[2270,61,13],\"L2\":[2270,0,0],\"L3\":[2270,0,0],\"t\":280,\"p\":13,\"d\":35550,\"e\":16}", "106": "{\"r\":\"Type B, AC 30mA + DC 6mA\",\"fv\":\"1.0.1\",\"cp\":\"6.0\",\"t\":\"0\",\"e\":\"0\"}", "107": "[6, 8, 10, 13, 16]", "109": "WORKING", "150": 6, "151": "{\"m\":0,\"dt\":0,\"ss\":\"00:00\",\"se\":\"08:00\"}", "152": 16, "155": false, "156": false}
Device matches abalon_bcm700d_curtain with quality of 20%. DPS: {"updated_at": 1744808788.7882693, "101": 300, "102": "{\"L1\":[2270,61,13],\"L2\":[2270,0,0],\"L3\":[2270,0,0],\"t\":280,\"p\":13,\"d\":37050,\"e\":17}", "106": "{\"r\":\"Type B, AC 30mA + DC 6mA\",\"fv\":\"1.0.1\",\"cp\":\"6.0\",\"t\":\"0\",\"e\":\"0\"}", "107": "[6, 8, 10, 13, 16]", "109": "WORKING", "150": 6, "151": "{\"m\":0,\"dt\":0,\"ss\":\"00:00\",\"se\":\"08:00\"}", "152": 16, "155": false, "156": false}
Device matches None with quality of 0%. DPS: {"updated_at": 1744809024.7897608, "102": "{\"L1\":[2280,61,14],\"L2\":[2280,0,0],\"L3\":[2280,0,0],\"t\":280,\"p\":14,\"d\":39430,\"e\":18}"}
[...]

Issue #3120, opened 2025-04-06
Request support for dé Wall Charger (3-phase 22Kw)


MchlLyn:
### Log message

```text
2025-04-06 10:05:01.952 WARNING (MainThread) [custom_components.tuya_local.config_flow] Adding dé EV Charger device with product id witok7vhhjohtr02
2025-04-06 10:05:02.059 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches abalon_bcm700d_curtain with quality of 20%. DPS: {"updated_at": 1743926698.4034767, "101": 101, "102": "{\"L1\":[2370,0,0],\"L2\":[2370,0,0],\"L3\":[2370,0,0],\"t\":280,\"p\":0,\"d\":0,\"e\":0}", "106": "{\"r\":\"Type B, AC 30mA + DC 6mA\",\"fv\":\"1.0.0\",\"cp\":\"12.5\",\"t\":\"0\",\"e\":\"0\"}", "107": "[6, 8, 10, 13, 16, 20, 25, 32]", "109": "SLEEP", "150": 32, "151": "{\"m\":0,\"dt\":0,\"ss\":\"00:00\",\"se\":\"08:00\"}", "152": 32, "155": true, "156": false}
```

### Product ID

witok7vhhjohtr02

### Product Name

dé EV Charger

### DPS information

```text
{
  "result": {
    "model": "{\"modelId\":\"e1krltgk\",\"services\":[{\"actions\":[],\"code\":\"\",\"description\":\"\",\"events\":[],\"name\":\"\",\"properties\":[
{\"abilityId\":101,\"accessMode\":\"rw\",\"code\":\"x_work_state\",\"description\":\"表示设备的工作状态，1000以内\",\"name\":\"工作状态\",\"typeSpec\":{\"type\":\"value\",\"max\":10000,\"min\":1,\"scale\":0,\"step\":1,\"unit\":\"\"}},
{\"abilityId\":102,\"accessMode\":\"ro\",\"code\":\"x_metrics\",\"description\":\"json字符串\",\"name\":\"指标信息\",\"typeSpec\":{\"type\":\"string\",\"maxlen\":255}},
{\"abilityId\":103,\"accessMode\":\"ro\",\"code\":\"x_selftest\",\"description\":\"json字符串\",\"name\":\"开机自检结果\",\"typeSpec\":{\"type\":\"string\",\"maxlen\":255}},
[...]

PR #5594, opened 2026-07-15
fix(dewall_evcharger): keep refresh available

szkocot:
## Summary

- treat a missing/null DP188 availability value as available for the hidden Refresh button
- fixes newer dewall_evcharger firmware where DP188 is listed in the cloud specification but omitted from the initial local DPS snapshot
- allows the existing Refresh button to request updated DP102 electrical metrics

## Compatibility note

This is intentionally a draft. Reports #3120 and #3152 use the same product ID but do not list DP188, while #4314 and the tested firmware do. Because product ID alone cannot distinguish these variants, maintainer guidance is requested on whether always exposing the hidden optional Refresh button is acceptable or whether a stronger discriminator is needed.

Tested device: product witok7vhhjohtr02, firmware 9.7.5. Before this change the enabled Refresh entity was unavailable; after it, button.press completes without an availability error.

## Validation

- uv run pytest tests/test_device_config.py — 32 passed
- uv run ruff check .
- uv run ruff check --select I .
- uv run ruff format --check .
- uv run yamllint custom_components/tuya_local/devices/dewall_evcharger.yaml

### DIFF
--- a/custom_components/tuya_local/devices/dewall_evcharger.yaml
+++ b/custom_components/tuya_local/devices/dewall_evcharger.yaml
@@ -214,5 +214,5 @@ entities:
         name: available
         mapping:
           - dps_val: null
-            value: false
+            value: true
           - value: true


custom_components/tuya_local/devices/dewall_evcharger.yaml
```diff
@@ -214,5 +214,5 @@ entities:
         name: available
         mapping:
           - dps_val: null
-            value: false
+            value: true
```
make-all, line 217: Just remove the available attribute completely and the hidden: unavailable

PR #5595, opened 2026-07-15
Add AmperePoint Q Series EV charger profile

amperepoint:
## Summary

- Add a single capability-based AmperePoint Q Series EV charger profile.
- Match all ten product IDs from the AmperePoint Tuya product export:
  - `kvldga0omutrnify`
  - `anmnwqqojlg9og2t`
  - `pt6e5u40nasunnsp`
  - `g5pkcqekcb7hhnm4`
  - `1qu8ca6etdubbigj`
  - `c2sgesr3zo1met7q`
  - `3axf0fkgiop0ukhb`
  - `cu111poj2mtikvls`
  - `jhlyzpk5nfk28nrh`
  - `fdfjiphjxtc9qyhd`
- Expose charging status, energy, current limit, total power, fault state, vehicle connection state, charging mode, target energy, start/stop and temperature.
- Keep optional schedule, last-session, system-version and raw mode data available when the device reports the related DPS.
- Keep raw phase-derived entities disabled by default because phase payloads are not reported in every state and have not been validated on every Q Series generation.
- Use optional DPS so products that report a smaller runtime set still expose their stable capabilities without failing configuration.

## Hardware validation

The profile was tested with an AmperePoint Q11 OTA charger using product ID `3axf0fkgiop0ukhb` and local protocol 3.5.

- The config flow selected the profile with a 101% product/DPS match.
- Initial local status reported DP 1, 3, 4, 9, 10, 13, 14, 18 and 24.
- DP 17 was reported after setting a target through Home Assistant.
- Readback-confirmed current-limit changes included 12 -> 14 -> 15 -> 12 A.
- Immediate, fixed-energy and scheduled charging modes were confirmed through Home Assistant.
- A 10 kWh target-energy value was written and read back.
- Start/stop state transitions were confirmed.
- After a Home Assistant restart, the charger reconnected automatically in approximately 15 seconds and restored the reported current limit, target energy, charging mode, connection state and switch state.
- No Tuya Local errors were logged during the restart and control tests.

AmperePoint confirms that the OTA variants use the same DP layout. The remaining product IDs are included from the manufacturer product export, including products marked `NOT_in_use`, the NO_OTA product and the VE product. The profile therefore keeps generation-dependent DPS optional and phase-derived entities disabled by default. A charger may omit optional DPS from an idle status response and report them only after a write or relevant state change.

The live hardware validation above applies to the OTA generation. Older, NO_OTA and VE variants are catalogued for matching but should be refined if their live payload semantics differ.

## Testing

- `uv run pytest tests/test_device_config.py`: 32 passed
- Full Linux test suite: 393 passed
- `uv run ruff check custom_components tests util`: passed
- `uv run ruff check --select I custom_components tests util`: passed
- `uv run ruff format --check custom_components tests util`: passed
- `uv run yamllint custom_components/tuya_local/devices`: passed

## Follow-up validation

DP 6/7/8 phase payloads, DP 19 schedule payload, DP 23 system version, DP 25 last-session energy and DP 33 raw mode payload remain optional. They can be enabled or refined when trustworthy payloads are captured from the relevant operating state.

### DIFF
--- a/custom_components/tuya_local/devices/amperepoint_q_series_evcharger.yaml
+++ b/custom_components/tuya_local/devices/amperepoint_q_series_evcharger.yaml
@@ -1,0 +1,403 @@
+# AmperePoint Q Series capability profile.
+# Product IDs are sourced from the AmperePoint Tuya product export. Optional
+# entities are exposed only when their datapoints are reported by the device.
+name: EV charger
+products:
+  - id: kvldga0omutrnify
+    manufacturer: AmperePoint
+    model: Q Series
+  - id: anmnwqqojlg9og2t
+    manufacturer: AmperePoint
+    model: Q Series
+  - id: pt6e5u40nasunnsp
+    manufacturer: AmperePoint
+    model: Q Series
+  - id: g5pkcqekcb7hhnm4
+    manufacturer: AmperePoint
+    model: Q Series
+  - id: 1qu8ca6etdubbigj
+    manufacturer: AmperePoint
+    model: Q Series
+  - id: c2sgesr3zo1met7q
+    manufacturer: AmperePoint
+    model: Q Series
+  - id: 3axf0fkgiop0ukhb
+    manufacturer: AmperePoint
+    model: Q Series
+  - id: cu111poj2mtikvls
+    manufacturer: AmperePoint
+    model: Q Series
+  - id: jhlyzpk5nfk28nrh
+    manufacturer: AmperePoint
+    model: Q Series
+  - id: fdfjiphjxtc9qyhd
+    manufacturer: AmperePoint
+    model: Q Series
+entities:
+  - entity: sensor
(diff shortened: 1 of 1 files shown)


make-all:
Only confirmed product ids should be listed. But judging by your username, you may have access to information that would confirm that they are in fact compatible.

amperepoint:
Thanks for checking! We’re AmperePoint, and all of these PIDs come directly from our company’s Tuya Developer account. Some are from older product generations, but they use the same DP logic and are still supported by us.

Would it be okay to keep the full PID list in this profile? We’ll be happy to provide any additional confirmation if needed.

make-all:
You sounded a bit unsure about some of them in the original request. If you know they all use the same dps layout then we can include all

amperepoint:
Thank you :) 
Have a great week!

Issue #5526, opened 2026-07-09
Background LAN rediscovery: auto-update a device's IP after a DHCP change


neneonline:
## What / why

When DHCP moves a device to a new IP (router reboot or replacement), `tuya_local`
keeps using the stale `CONF_HOST`; the device goes `unavailable` until it's
manually reconfigured or removed/re-added (which forces a cloud login to re-fetch
a local key the user already has). For history logging this means multi-day gaps.

This overlaps with the README "Contributing" list item 3 (background discovery),
so I wanted to check your preferred direction before proposing a PR.

## A finding that might be useful regardless

I first tried a **passive** broadcast listener (bind UDP 6666/6667). On real
hardware it received **nothing** — ~3.5 min of silence across 6666/6667/6669/7000.
The devices are protocol **3.5**, and 3.4/3.5 devices don't broadcast unsolicited:
they only reply to a discovery **request** on UDP **7000** (tinytuya's scanner
sends it — `BROADCASTTIME` comment: "get v3.5 devices to send us their info").
Replaying that probe got 10–12 replies per device within 18 s. So any local
discovery here has to **actively probe**, not just listen. (Consistent with how
`localtuya` and your own `scan_for_device` already work.)

## Proposed approach (implemented + production-validated)

A small background sweeper that, on a fixed interval, finds **unreachable**
configured devices and looks up each one's current IP by `gwId` via
`tinytuya.find_device(...)`, then updates `CONF_HOST` in place — the existing
update-listener reload reconnects the device. Keyed on `gwId` (already stored as
`CONF_DEVICE_ID`/`unique_id`), fully local, no MAC, no cloud, no new dependency.

- Reachable devices are never scanned → **no idle network traffic**.
- HA `dhcp` discovery can't be the primary mechanism: `registered_devices`
  matching needs a `CONNECTION_NETWORK_MAC` the integration never learns (the Tuya
  payload and cloud API both omit MAC). It could be a later fast-path trigger.
- Extends naturally toward the rest of item 3 (`find_device`/`deviceScan` also
  return `productKey` and surface *new* devices).

Validated on a live HAOS install: a real router swap moved 4 thermostats to a new
IP block and they self-healed unattended (all `unavailable` ~12:02 → recovered
~12:10, stored hosts updated to the new IPs, no restart, no manual action).

I have this implemented with unit tests (mirrors `test_config_flow.py`
conventions) and `ruff`/`pytest` green, ready to submit as a PR.

## Questions before I open a PR

[...]

PR #5605, opened 2026-07-16
feat: discover new Tuya devices on the LAN

neneonline:
Third increment for #5526 (follow-up you invited when closing it), building on the merged #5532 and #5573. This adds the **new-device discovery** piece.

### What it does
- Turns the periodic (~10&nbsp;min) pass into a single `tinytuya.deviceScan` that does two things from one scan:
  1. **Configured devices** — the product-id check from #5573 (WARN once/device/run when the reported `productKey` isn't listed in the config's `products:`), now fed by the shared scan instead of a per-device `find_device`.
  2. **Unconfigured devices** — raises an `integration_discovery` flow so each unknown Tuya device on the LAN surfaces in Home Assistant for one-click setup.
- Adds `ConfigFlowHandler.async_step_integration_discovery`: it sets the unique id to the device's `gwId`, **aborts if already configured or ignored**, and reuses the existing manual-setup prefill (id / ip / version from the scan; the user still supplies the local key). The built-in per-`unique_id` ignore therefore works out of the box.
- The fast 60&nbsp;s IP-rediscovery sweep from #5532 is untouched; both tasks share a scan guard so they never run concurrently.

### Notes
- No manifest matchers needed — discovery is raised programmatically via `SOURCE_INTEGRATION_DISCOVERY`, so it uses the LAN scan (not DHCP/zeroconf), consistent with the active-only direction agreed on #5526.
- A local set stops the same new device being re-offered every scan within one run; across restarts, HA's unique-id abort keeps configured/ignored devices from reappearing.
- I couldn't exercise the live `deviceScan` from my side (needs the LAN), but it rides on the same scanner already proven in production by #5532/#5573; the new HA glue (flow init + discovery step) is unit-tested. Happy to adjust the discovery UX (e.g. a confirm step, or a startup trigger) to your preference.

### Tests
`tests/test_discovery.py`: product-id warn (fed by the scan), new-device flow raised for unconfigured devices, configured devices skipped, deduped per run, empty scan is a no-op. `tests/test_config_flow.py`: discovery step advances to the local form; aborts when already configured. `uv run pytest` (403 passed) and `ruff check` / `ruff format --check` green.

### DIFF
--- a/custom_components/tuya_local/config_flow.py
+++ b/custom_components/tuya_local/config_flow.py
@@ -70,6 +70,30 @@ class ConfigFlowHandler(ConfigFlow, domain=DOMAIN):
         if self.cloud is None:
             self.cloud = Cloud(self.hass)
 
+    async def async_step_integration_discovery(self, discovery_info):
+        """Handle a device found on the LAN by the background scanner.
+
+        Pre-fills the manual setup form with the discovered id/ip/version; the
+        user still supplies the local key. Aborts if the device is already
+        configured or has been ignored.
+        """
+        device_id = discovery_info.get(CONF_DEVICE_ID)
+        await self.async_set_unique_id(device_id)
+        self._abort_if_unique_id_configured()
+        # Reuse the cloud-device plumbing that async_step_local reads for its
+        # form defaults; the local key is not known from discovery.
+        self.__cloud_device = {
+            "id": device_id,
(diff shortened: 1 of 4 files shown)


make-all:
Thanks, this is a clean addition, there doesn't seem to be any risk of regression to other functionality, so should be OK to go.

PR #5607, opened 2026-07-17
Add Aquaviva Model 12 ON/OFF pool heat pump

msemenkin:
Adds support for the Aquaviva Model 12 ON/OFF pool heat pump
(AVM-ON12RW), tested with product ID `cbnlx1l9nyal6vk4`.

A separate profile is needed even though this device reports the same
Tuya product ID and model ID as the existing Turbro 75000 BTU profile:

- Aquaviva reports DP 5 and uses it to select Heating, Cooling and Auto.
  DP 5 was absent from the cached state submitted for the Turbro profile.
- Aquaviva reports `unit_type: 0`, while the Turbro diagnostics reported
  `unit_type: 1`.
- Aquaviva actively reports the Celsius temperature DPs. Its Fahrenheit
  DPs return the inactive value `-19000`, while the Turbro profile uses
  the Fahrenheit variants.
- The Aquaviva Smart Life interface does not offer the DP 2 operating
  presets or the DP 117 Silent/Smart/Powerful selector. During testing
  these DPs remained fixed at `Auto` and `Smart`. The Turbro device
  reported DP 117 as `Powerful` and its profile exposes these controls.
- The manufacturers direct users to different mobile applications:
  Aquaviva uses Smart Life, while Turbro documents AIR.ai/Airia for the
  Beluga series. These applications expose different controls, labels
  and workflows, creating different user expectations for entity names
  and available features.

The new profile therefore follows the controls and terminology observed
on the tested Aquaviva device in Smart Life. Reusing or modifying the
Turbro profile could change the behaviour experienced by existing Turbro
users, so the existing Turbro YAML is left unchanged.

Verified on the physical device and through Tuya DP Instruction Mode:

- separate DP 1 power control;
- DP 5 Heating, Cooling and Auto modes;
- mode-specific target temperatures on DPs 101, 102 and 104;
- 1 °C target temperature step;
- inlet, outlet, ambient, exhaust, suction and coil temperatures;
- work state, fault bitmaps and expansion valve opening.

### DIFF
--- a/custom_components/tuya_local/devices/aquaviva_model_12_onoff_pool_heatpump.yaml
+++ b/custom_components/tuya_local/devices/aquaviva_model_12_onoff_pool_heatpump.yaml
@@ -200,7 +200,7 @@ entities:
             value: fahrenheit
 
   - entity: sensor
-    name: Status
+    translation_key: status
     class: enum
     category: diagnostic
     dps:
--- a/custom_components/tuya_local/devices/aquaviva_model_12_onoff_pool_heatpump.yaml
+++ b/custom_components/tuya_local/devices/aquaviva_model_12_onoff_pool_heatpump.yaml
@@ -1,0 +1,472 @@
+name: Pool heat pump
+products:
+  - id: cbnlx1l9nyal6vk4
+    manufacturer: Aquaviva
+    model: Model 12 ON/OFF (AVM-ON12RW)
+entities:
+  - entity: switch
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+
+  - entity: climate
+    translation_key: pool_heatpump
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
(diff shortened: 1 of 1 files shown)


make-all:
Merged with the Turbro config, since it is better to have a single product id match. The config has become more complicated as a result, so there is high chance that something may be wrong but I will do my best to fix any issues.

msemenkin:
> Merged with the Turbro config, since it is better to have a single product id match. The config has become more complicated as a result, so there is high chance that something may be wrong but I will do my best to fix any issues.

Following up on the merged AquaViva support from this PR: I found an issue affecting AquaViva devices that do not report DP 2 locally.

The target temperature was displayed as the raw value `290 °C` instead of `29.0 °C`, and cooling action detection could also fall back to heating.

I opened #6093 with a minimal fallback for devices without DP 2. It has been tested on a physical AquaViva Model 12 ON/OFF: temperature changes in 1 °C steps and updates are reflected correctly in Smart Life. The existing TURBRO paths remain unchanged.

PR #5611, opened 2026-07-17
Add Lyfco 9kW Pool Heatpump device config

mkuoppa:
Adds device config support for the Lyfco 9kW Pool Heatpump (product ID zwmuxensvaoxc8n5).
- Now PR with the correct file, running and verified in my system

Entities:
climate:
- heat/cool/auto modes via dps 1 (on/off) + dps 2 (mode)
- target temperature via dps 4 (10–40°C, step 1°C)
- current temperature via dps 16
- Smart/Silence/Boost modes via dps 5

binary_sensor (diagnostic, problem class):
- fault detection via dps 15, a 37-bit bitfield covering all documented error codes (E01–E37, e.g. IPM protection, voltage/current faults, sensor errors, water flow protection, etc.)

Testing:
Configured and tested locally against a physical unit running HAOS under Proxmox. Mode switching, temperature set/read, and fault sensor reporting all behave as expected.

### DIFF
--- a/custom_components/tuya_local/devices/lyfco_9kw_heatpump.yaml
+++ b/custom_components/tuya_local/devices/lyfco_9kw_heatpump.yaml
@@ -129,7 +129,8 @@ entities:
           - dps_val: 1048576
             value: "E21 Evaporator Temperature (T2) Low Protection"
           - dps_val: 2097152
-            value: "E22 Communication Error between Wireless Control and Mainboard"
+            value: "E22 Communication Error between Wireless Control and\
+            Mainboard"
           - dps_val: 4194304
             value: "E23 Phase Missing Protection"
           - dps_val: 8388608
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/lyfco_9kw_heatpump.yaml
```diff
@@ -0,0 +1,165 @@
...
+          - dps_val: true
+            constraint: preset_mode
+            conditions:
+              - dps_val: "Cooling"
+                value: cool
+              - dps_val: "Heating"
+                value: heat
+              - dps_val: "Auto"
+                value: auto
+      - id: 2
```
make-all, line 28: hvac_mode should not in general be duplicated as preset_mode. (there are a few configs that do it because of limitations of devices, but not in general)

make-all: Actually what I think you want here is to rename this to mode and change the condition above. There is no need for a mapping here, as it is already mapped above.

custom_components/tuya_local/devices/lyfco_9kw_heatpump.yaml
```diff
@@ -0,0 +1,165 @@
...
+        name: preset_mode
+        type: string
+        mapping:
+          - dps_val: "Smart"
+            value: auto
+          - dps_val: "Silence"
+            value: quiet
+          - dps_val: "Boost"
+            value: boost
+      - id: 2
```
make-all: You should not reuse the hvac_mode selector for hvac_action. If the device does not have an indicator of whether it is actually working or idle, then just leave hvac_action out.

make-all (changes requested):

make-all:
Merged with poolsana_heatpump, as only the hvac_action and fault descriptions are missing from that, no other differences.

PR #5617, opened 2026-07-18
chore(doc): Add documentation for creating device descriptor assisted by AI Agent

mathieupotier:
A documentation about creating device descriptor assisted by AI.
This doc reflects how I generated those devices :
 - https://github.com/make-all/tuya-local/pull/5616
 - https://github.com/make-all/tuya-local/pull/5615
 - https://github.com/make-all/tuya-local/pull/5614

Hope it could help.
I wrote this document in the perspective of being used by both human and agents.

### DIFF
--- a/CREATING_DEVICES_WITH_AGENT.md
+++ b/CREATING_DEVICES_WITH_AGENT.md
@@ -1,0 +1,251 @@
+# Procedure for using AI to create new device configs
+
+## Procedure followed by the developer to gather information from the Tuya Cloud API
+
+**Note: these API calls require an authenticated session on the [Tuya IoT
+Developer Platform](https://iot.tuya.com) (Cloud > API Explorer). An AI
+agent cannot perform them on its own - a human must run them manually and
+drop the resulting JSON files in a folder accessible to the agent, named
+`<device>.json`, `<device>_dps.json` and optionally `<device>_functions.json`.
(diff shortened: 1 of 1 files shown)


CREATING_DEVICES_WITH_AGENT.md
```diff
@@ -0,0 +1,251 @@
...
+    "update_time": 1784320266,
+    "uuid": "<uuid>"
+  },
+  "success": true,
+  "t": 1784320266,
+  "tid": "<transaction_id>"
+}
+```
+
+### Get Device Properties (`Device Control > Query Properties` in Api Explorer)
```
make-all, line 54: This is poor quality information and often incorrect and incomplete. Users should not be directed to use this function. The only information needed apart from the connection info above is the product_id from above, the local log message from HA log, and the Query Things Data Model result from the API Explorer.

make-all (commented):
It's a good idea, but the instructions are a bit off. It might be better if I rewrote them myself (I will still give you credit)

mathieupotier:
Help yourself, if you think you could better document. The final goal is to ease the descriptor generation. I tried to put in there what I used for those pull requests.

Contribution welcome (and honestly I don't mind being credited 😉, if it helps others)

I wrote the main course and use the context agent to fill the blanks from its point of view.

That why you can see some hints really specific from the issue I got during the development process.

PR #5621, opened 2026-07-18
Add  profile for the Powerworld PW030-KZXYC-E pool heat pump

toolson:
Tested profile for the Powerworld PW030PW030-KZXYC-E pool heat pump (pw58474).

### DIFF
--- a/custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
+++ b/custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
@@ -1,9 +1,9 @@
 name: Powerworld PW030 Pool Heatpump - pw58474
 products:
-  - id: pw58474
+  - id: powerworld_pw58474_waterheater
     manufacturer: Powerworld
     model: PW030
-    
+
 entities:
   - entity: switch
     name: power switch
@@ -53,7 +53,7 @@ entities:
           max: 40
         mapping:
           - constraint: mode
-            condition: cool
+            value: cool
             value_redirect: cool_temp
       - id: 152
         type: integer
@@ -136,7 +136,7 @@ entities:
         name: unit
         mapping:
           - value: "°C"
-          
+
   - entity: sensor
     name: return gas temperature
     category: diagnostic
@@ -193,7 +193,7 @@ entities:
         name: unit
         mapping:
           - value: "°C"
-          
+
   - entity: sensor
     name: refrigerant gas concentration
     category: diagnostic
--- a/custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
+++ b/custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
@@ -1,0 +1,321 @@
+name: Powerworld PW030 Pool Heatpump - pw58474
+products:
+  - id: pw58474
+    manufacturer: Powerworld
+    model: PW030
+    
+entities:
+  - entity: switch
+    name: power switch
+    icon: mdi:power
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+        mapping:
+          - dps_val: false
+            value: false # Schaltet die Wärmepumpe AUS
+          - dps_val: true
+            value: true  # Schaltet die Wärmepumpe AN
+
+  - entity: climate
+    dps:
+      - id: 2
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: strong
+            value: boost
+          - dps_val: smart
+            value: comfort
+          - dps_val: silent
+            value: eco
+      - id: 5
+        type: string
+        name: hvac_mode
+        mapping:
+          - dps_val: heat
+            value: heat
+          - dps_val: cool
+            value: cool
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
```diff
@@ -0,0 +1,370 @@
+name: Powerworld PW030 Pool Heatpump - pw58474
```
make-all: Top level name should be generic, as other devices may match this config also. Branding belongs under products

custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
```diff
@@ -0,0 +1,370 @@
...
+products:
+  - id: amnpmrzcbrisfq0r
+    name: PW030-KZXYC-E
```
make-all: name not required here, it is anyway just a more cryptic version of model in this case.

custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
```diff
@@ -0,0 +1,370 @@
...
+    manufacturer: Powerworld
+    model: PW030
+
+entities:
+  - entity: switch
+    name: power switch
```
make-all: Entity names should be human readable, but in this case, not needed, as the power switch is implied as the main switch of a device.

custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
```diff
@@ -0,0 +1,370 @@
...
+    icon: mdi:power
```
make-all: Icons where provided (usually they are unnecessary, and may override user themes when specified like this) must be enclosed in double-quotes, as the yaml parser gets confused by the `:`

custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
```diff
@@ -0,0 +1,370 @@
...
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+        mapping:
```
make-all: This mapping is unneccesary

custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
```diff
@@ -0,0 +1,370 @@
...
+          - dps_val: true
+            value: true
+
+  - entity: climate
+    translation_key: pool_heatpump
+    dps:
+      - id: 2
+        type: string
+        name: preset_mode
+        mapping:
```
make-all, line 21: The pool_heatpump translation key also enables the use of "quick", "smart", "quiet" as preset names, which are commonly used in pool heatpumps (quick often as boost or performance, quiet often as silent or mute, smart sometimes as auto)

custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
```diff
@@ -0,0 +1,370 @@
...
+      - id: 152
+        type: integer
+        name: cool_temp
+        optional: true
+        hidden: true
+        range:
+          min: 10
+          max: 40
+
+  - entity: sensor
```
make-all, line 72: One of these temperatures (that defines either the pool water temperature, or if that is unavailable the temperature of the water going into the pool) should be moved to the current_temperature of the climate entity

custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
```diff
@@ -0,0 +1,370 @@
...
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+
+  - entity: binary_sensor
+    translation_key: problem
+    category: diagnostic
+    dps:
+      - id: 15
```
make-all, line 363: The raw value of dp 15 should be retained as a `fault_code` attribute, as with most `problem` sensors in other configs.

make-all (changes requested):

custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
```diff
@@ -0,0 +1,376 @@
...
+               00000000FFFFFFFF00000000000000000000000000000000\
+               000000000000000000000000000000000000000000000000\
+               0000000000000000"
+        mapping:
+          - dps_val: 0
+            value: false
+          - value: true
+
+  - entity: binary_sensor
+    translation_key: problem
```
make-all: Should be class, not translation_key here.

make-all (changes requested):

custom_components/tuya_local/devices/powerworld_pw58474_waterheater.yaml
make-all: Human readable means the first word of each entity or device name should be capitalised to follow HA naming convention.

PR #5628, opened 2026-07-19
Added support for Yale YMC 420W Zigbee Smart Lock

victoraraujo01:
The Yale YMC 420W Zigbee Smart Lock also works with Tuya Zigbee hubs, not just the proprietary Yale Smart Connect Hub. 

This file sets it up as a smart lock on Home Assistant with the Tuya Local integration, with support to lock/unlock functions, battery indicator and activity logs.


make-all:
Non Tuya Zigbee devices are out of scope for this integration.

There is more than enough work for maintainers already without bringing in more Zigbee devices from other brands.

PR #5630, opened 2026-07-19
Vevor EV charger: add missing class for Phase A power

alexandrebelloni:
The measurement class is needed to be able to add the current power to the energy dashboard.


alexandrebelloni:
Sorry, I pushed the second commit after opening the pull request, so the tests did run without it but this should not cause any issues as then both files are the same apart from the device id and the selectable intensity

PR #5634, opened 2026-07-20
Add Lumary C2 ceiling fan light

digitalvir:
Adds a device config for the Lumary C2 ceiling fan light.

Device details:
- Product id: egakknayghmwfrpv
- Manufacturer: Lumary
- Model: C2
- Model id: L-CFL19C2
- Tested locally with Tuya protocol 3.5 across six installed fixtures

This initial config covers the core fan, main light, nightlight, timer, and settings DPs. Scene/effect presets are intentionally left for a future follow-up after they can be captured and named accurately.

Validation run:
- uv run yamllint --format parsable custom_components/tuya_local/devices/lumary_c2_ceilingfanlight.yaml
- uv run pytest tests/test_device_config.py
- uv run yamllint --format parsable custom_components/tuya_local/devices
- git diff --check

### DIFF
--- a/custom_components/tuya_local/devices/lumary_c2_ceilingfanlight.yaml
+++ b/custom_components/tuya_local/devices/lumary_c2_ceilingfanlight.yaml
@@ -1,0 +1,301 @@
+name: Ceiling fan light
+products:
+  - id: egakknayghmwfrpv
+    manufacturer: Lumary
+    model: C2
+    model_id: L-CFL19C2
+entities:
+  - entity: light
+    icon: mdi:ceiling-fan-light
+    dps:
+      - id: 20
+        type: boolean
+        name: switch
+      - id: 21
+        name: color_mode
+        type: string
+        mapping:
+          - dps_val: white
+            value: color_temp
+          - dps_val: colour
+            value: hs
+          - dps_val: scene
+            value: Scene
+          - dps_val: music
+            value: Music
+      - id: 22
+        name: brightness
+        type: integer
+        range:
+          min: 10
+          max: 1000
+      - id: 23
+        name: color_temp
+        type: integer
+        range:
+          min: 0
+          max: 1000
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/lumary_c2_ceilingfanlight.yaml
```diff
@@ -0,0 +1,301 @@
+name: Ceiling fan light
+products:
+  - id: egakknayghmwfrpv
+    manufacturer: Lumary
+    model: C2
+    model_id: L-CFL19C2
+entities:
+  - entity: light
+    icon: mdi:ceiling-fan-light
```
make-all: Icons require double quotes, as the yaml parser gets confused by the `:`

custom_components/tuya_local/devices/lumary_c2_ceilingfanlight.yaml
```diff
@@ -0,0 +1,301 @@
...
+    category: config
+    dps:
+      - id: 66
+        type: boolean
+        optional: true
+        name: switch
+        mapping:
+          - dps_val: null
+            value_redirect: alt
+      - id: 103
```
make-all, line 175: Is this alternate dp really needed, or was this cargo culted from another config?

make-all (changes requested):

digitalvir:
Addressed the review feedback in 0fafb63:

- Quoted all `mdi:*` icon strings in the new Lumary C2 config.
- Removed the DPS 66 fallback from the sound switch. The local probe data for this device showed DPS 103 present and DPS 66 absent, so DPS 103 is now used directly as the switch.

Validation run locally:
- `uv run yamllint custom_components/tuya_local/devices/lumary_c2_ceilingfanlight.yaml`
- `uv run pytest tests/test_device_config.py`

PR #5637, opened 2026-07-20
[Kibernetik 12K Airconditioner] added heating mode, constrained temperature range to fit the air conditioner

Louisbertelsmann:
Added support for heating mode for this device (supports it, can confirm) and fixed the homeassistant climate entity not being limited to the devices capabilities, causing usability problems (device supports 16°C to 30°C, not the standard 7°C to 35°C).

Updated temperature range for the air conditioner and added HEAT mode.

### DIFF
--- a/custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/kibernetik_12k_airconditioner.yaml
@@ -18,26 +18,18 @@ entities:
             conditions:
               - dps_val: COOL
                 value: cool
+              - dps_val: HEAT
+                value: heat
               - dps_val: FAN
                 value: fan_only
-              # dp4 also lists HEAT, but this is a cool-only model (see dp110).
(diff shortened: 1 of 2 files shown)


make-all:
Based on the comments in the file, the original submission was a cool-only model, so any change also needs to work with that. The comment mentions dp 110, but that was not implemented in the file by the submitter. Maybe you can confirm on your model that dp 110 is present and indicates the heat mode as expected.

make-all (changes requested):
This change completely ignores the comment in the original file that the submitter had a cool only model. The change cannot be accepted without feature detection based on model (if possible).

Louisbertelsmann:
I saw no way to only allow heating with the heat & cool model due to the DP110 always being 0 apparently, and the product id is also the same, so I added a second configuration file to choose during setup.

make-all:
#5715 was already added with a suitable config for this variant

PR #5647, opened 2026-07-21
Add Daolar F098 EV charger

sev3ran:
Adds support for the Daolar F098 EV charger.

Add a new device entry for the following EV charger:

- name: EV charger
- product id: dkzx0axhp6zzcyp4
- manufacturer: Daolar
- model: F098

The configuration has been tested and verified on a physical device.


make-all:
Likely a match for afyeev_16a_evcharger

sev3ran:
Yes, they are very similar, but I found differences on the Daolar physical device:

1. DP4 only accepts 6, 8, 10, 12, 14 and 16 A, so this device requires step: 2. 

2. The packed DP6 layout is different. The Daolar uses 24-bit fields for current and power:

```yaml
current: "0000FFFFFF000000"
power:   "0000000000FFFFFF"
```

while the Afyeev configuration uses 16-bit masks:

```yaml
current: "000000FFFF000000"
power:   "000000000000FFFF"
```

I’m not sure whether these differences are acceptable from a functionality and compatibility perspective. If you think this device should be added to the existing afyeev_16a_evcharger configuration, could you please confirm that these differences are acceptable?

make-all:
The mask difference probably applies to the original config, but it is a theoretical issue only, because I don't think the device should ever go beyond 65.5A or 65.5kW. I'm not sure if a step size of 2 would be an issue for the charging current. There have been requests in the past to open up devices to go lower than the advertised current so users can manage their charging to avoid mains use when solar output is low, but maybe the steps in between are not so important, and it is possible that the Afyeev devices also have this limitation but nobody commented on it before.

sev3ran:
Thanks for explaining your reasoning and sharing your perspective. Based on that I tested the Daolar F098 with the existing afyeev_16a_evcharger configuration.

Voltage, current and power are reported correctly. Unsupported charging current values, such as 7A, are simply ignored by the device while supported values work correctly.

I have therefore updated the PR to add the Daolar product ID to the existing configuration and removed the separate device configuration.

PR #5653, opened 2026-07-21
Add dedicated config for Parkside PMRDA 20-Li A1 mower with integer DP 121 type

nima-1102:
## 🔍 Problem
The Parkside PMRDA 20-Li A1 mower was previously included in the shared `parkside_bluetooth_mower.yaml` config file, but newer Parkside mower models (e.g., PAMRC 250) send DP 121 as a JSON boolean (`true`/`false`), while the PMRDA 20-Li A1 sends it as an integer (`0`/`1`).

This type mismatch caused:
- ❌ The PMRDA 20-Li A1 device not appearing in the GUI device selection
- ❌ Incompatibility with the boolean-based configs of newer models

## ✅ Solution
- **Created dedicated config file**: `parkside_pmrda_mower.yaml` for the PMRDA 20-Li A1
- **Removed the device from shared config**: Removed PMRDA 20-Li A1 (icw5sal7xfcevsve) from `parkside_bluetooth_mower.yaml`
- **Preserved integer type for DP 121**: The new config correctly defines DP 121 as `type: integer` with proper mappings (0 → false, 1 → true)
- **Backward compatible**: No breaking changes for other Parkside mower models

## 📋 Changes
- **New file**: `parkside_pmrda_mower.yaml` (399 lines) – complete config for PMRDA 20-Li A1
- **Modified**: `parkside_bluetooth_mower.yaml` – removed PMRDA 20-Li A1 device entry

## 🎯 Impact
- ✓ PMRDA 20-Li A1 is now selectable in the GUI
- ✓ Newer Parkside models with boolean DP 121 continue to work correctly
- ✓ No regressions for other devices

### DIFF
--- a/custom_components/tuya_local/devices/parkside_bluetooth_mower.yaml
+++ b/custom_components/tuya_local/devices/parkside_bluetooth_mower.yaml
@@ -7,9 +7,6 @@ products:
   - id: 7yr5iwga
     manufacturer: Parkside
     model: PMRC 250 Ai Bluetooth
-  - id: icw5sal7xfcevsve
-    manufacturer: Parkside
-    model: PMRDA 20-Li A1
 entities:
   - entity: lawn_mower
     dps:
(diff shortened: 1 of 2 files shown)


make-all:
Moved the product details back to moebot_s_mower.yaml as an alternative to expanding the number of configs unnecessarily.

PR #5654, opened 2026-07-21
Add Hantech 9000 BTU portable air conditioner

grinser:
Adds a device profile for the Hantech 9000 BTU mobile air conditioner (Tuya category `ydkt`, product id `av6ejal3j9epuw3e`).

The official Tuya integration cannot map this device: the cloud returns an empty function/status schema (DP Instruction mode) and category `ydkt` is not covered. It shares the OEM base of the Cecotec ForceClima Soundless, but without heat mode and without the swing DP (110).

All DP behaviour was verified locally over the LAN with tinytuya (protocol 3.5) and confirmed working in Home Assistant via tuya-local:

- Power (1), target temperature (2, 16-30 C), current temperature (3)
- Mode (4): `cold`->cool, `wet`->dry, `wind`->fan_only (no heat/auto)
- Fan (5): low / high only
- Sleep switch (103), fault (20, bitfield), temperature F mirror (107, hidden)
- dp109 is read-only (always 0)

The `problem` and `tank_full` fault mappings on dp20 are inherited from the ForceClima OEM base; dp20 read 0 throughout testing, so the other fault codes were not individually exercised.

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -138,6 +138,7 @@
 - Fujicool Yuzu heat pump
 - Goldair GCPAC350W portable air conditioner
 - Haier Airmart wall air conditioner
+- Hantech 9000 BTU portable air conditioner
 - Hokkaido HKEDM 263 ZL air conditioner
 - HTW IX75B air conditioner
 - Idea Heating Belt (with CS1 USB dongle)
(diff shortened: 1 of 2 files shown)


make-all:
Seems to match the costway_portable_ac.yaml config (with sleep, temp_set_f and problem sensor identified.

It would be better to make the changes in that file.

grinser:
Thanks for the feedback. Moved it into costway_portable_ac.yaml and identified the shared DPs there (103 sleep, 107 temp_set_f, 20 fault).

PR #5655, opened 2026-07-21
Add water chiller (fre1l4)

klobuczek:
Adds a device config for a Tuya-based water chiller (product id
tmwafkxho3xaqhzt, model fre1l4, category wk).

Exposes: climate (power, cool/heat/auto, target and current temperature),
plus diagnostic sensors for the refrigeration circuit (fault code, ambient/
return/outlet/suction/discharge/coil temperatures, expansion valve, compressor
current) and running-state binary sensors (compressor, fan, pump, defrost,
anti-freeze, aux heat). Defrost/refrigeration tuning parameters are included as
config numbers, hidden by default. An ozone/oxygenation control is exposed on
dp 125.

DP map obtained from the Tuya "query things data model" API and verified
against a live device. dps 126-129 are present in the data model but inert in
this unit's firmware, so they are omitted.

### DIFF
--- a/custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
+++ b/custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
@@ -1,21 +1,10 @@
-# Tuya Local device profile — Water chiller
-# Product: tmwafkxho3xaqhzt (model fre1l4), Tuya category "wk"
-# Protocol 3.4. DP behavior verified by live experiment 2026-07:
-#   - DP 125 acts as a DIRECT ozone/oxygenation on/off (despite its
-#     "timer enable" name). Firmware enforces: 1-hour auto-off,
-#     circulation-pump interlock, no persistence across power cycles.
-#   - DPs 126-129 (daily timer window) are declared in the data model
-#     but INERT in this MCU firmware — intentionally omitted.
-#
-# Ozone is modelled as a climate PRESET (Normal / Disinfect) rather than
-# a standalone switch, so it lives inside the chiller card. DP 124 remains
-# a read-only "Ozone running" binary_sensor for state feedback.
+# Water chiller (fre1l4). dp 125 ozone auto-offs after 1h and interlocks on
+# the circulation pump (dp 120). dps 126-129 are inert on this unit.
 name: Water chiller
 products:
   - id: tmwafkxho3xaqhzt
     manufacturer: Unbranded
-    model: fre1l4 water chiller
-
+    model: fre1l4
 entities:
   - entity: climate
     dps:
@@ -41,70 +30,67 @@ entities:
       - id: 3
         type: integer
         name: temperature
+        mapping:
+          - constraint: temperature_unit
+            conditions:
+              - dps_val: f
+                range:
+                  min: 32
+                  max: 105
         range:
           min: 0
           max: 41
-      - id: 24
-        type: integer
-        name: current_temperature
       - id: 23
         type: string
         name: temperature_unit
-        optional: true
         mapping:
           - dps_val: c
             value: C
           - dps_val: f
             value: F
-      # Ozone / oxygenation as a preset overlay (independent of hvac_mode).
-      # Requires the circulation pump running; firmware auto-offs after 1h.
+      - id: 24
+        type: integer
+        name: current_temperature
       - id: 125
         type: boolean
         name: preset_mode
         mapping:
           - dps_val: false
-            value: Ozone Off
+            value: normal
           - dps_val: true
-            value: Ozone On
-
-  # Read-only ozone running state (the actual pump, not the enable bit)
+            value: ozone
   - entity: binary_sensor
-    name: Ozone running
+    name: Ozone
+    class: running
     category: diagnostic
-    icon: "mdi:air-purifier"
     dps:
       - id: 124
         type: boolean
         name: sensor
-
-  # --- Running-state diagnostics ---
   - entity: binary_sensor
     name: Compressor
-    category: diagnostic
     class: running
+    category: diagnostic
     dps:
       - id: 118
         type: boolean
         name: sensor
-
   - entity: binary_sensor
     name: Fan
-    category: diagnostic
     class: running
+    category: diagnostic
     dps:
       - id: 119
         type: boolean
         name: sensor
-
   - entity: binary_sensor
     name: Circulation pump
-    category: diagnostic
     class: running
+    category: diagnostic
     dps:
       - id: 120
         type: boolean
         name: sensor
-
   - entity: binary_sensor
     name: Defrosting
     category: diagnostic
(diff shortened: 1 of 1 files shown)


klobuczek:
Pushed a fix: diagnostic temperature sensors are fixed °C (verified on-device — toggling the C/F display DP doesn't change their values); only the setpoint follows the display unit.

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
+# Water chiller (fre1l4). dp 125 ozone auto-offs after 1h and interlocks on
+# the circulation pump (dp 120). dps 126-129 are inert on this unit.
+name: Water chiller
+products:
+  - id: tmwafkxho3xaqhzt
+    manufacturer: Unbranded
```
make-all: This is not a real brand. Leave this attribute out if it cannot be found.

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,321 @@
...
+      # Ozone / oxygenation as a preset overlay (independent of hvac_mode).
+      # Requires the circulation pump running; firmware auto-offs after 1h.
+      - id: 125
+        type: boolean
+        name: preset_mode
+        mapping:
+          - dps_val: false
+            value: Ozone Off
+          - dps_val: true
+            value: Ozone On
```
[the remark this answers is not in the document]
make-all: Don't listen to the clueless AI bot. 

Presets should either match an existing translation key, or the first word should be capitalised.

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+                  max: 105
+        range:
+          min: 0
+          max: 41
+      - id: 23
+        type: string
+        name: temperature_unit
+        mapping:
+          - dps_val: c
+            value: C
```
make-all: It is generally better to have C as the default here (no dps_val required to match), so there is a fallback for the unit as HA will complain if there is no unit declared

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
+# Water chiller (fre1l4). dp 125 ozone auto-offs after 1h and interlocks on
+# the circulation pump (dp 120). dps 126-129 are inert on this unit.
+name: Water chiller
+products:
+  - id: tmwafkxho3xaqhzt
+    manufacturer: Unbranded
+    model: fre1l4
+entities:
+  - entity: climate
+    dps:
```
make-all, line 10: add `translation_key: cold_water` here

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+          - dps_val: c
+            value: C
+          - dps_val: f
+            value: F
+      - id: 24
+        type: integer
+        name: current_temperature
+      - id: 125
+        type: boolean
+        name: preset_mode
```
make-all: I don't think this really fits as a preset mode. This is like a sterilization control, which is separate from the cooling function, so I think a separate switch entity would be more appropriate.
Most likely this is using UV light to generate Ozone from the Oxygen already in the water, so `translation_key: uv_sterilization` might be considered a match.

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+  - entity: binary_sensor
+    name: Circulation pump
+    class: running
+    category: diagnostic
+    dps:
+      - id: 120
+        type: boolean
+        name: sensor
+  - entity: binary_sensor
+    name: Defrosting
```
make-all: `translation_key: defrost` can be used here instead of the name

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+        name: sensor
+  - entity: binary_sensor
+    name: Electric auxiliary heat
+    class: heat
+    category: diagnostic
+    dps:
+      - id: 123
+        type: boolean
+        name: sensor
+  - entity: sensor
```
make-all: Normal pattern used here is a `binary_sensor` with `class: problem`. Around 90% of configs have this so you should be able to find a reference. Normally the dp will be a bitfield (documented as "bitmap" in Tuya docs) so defining it like that in the config will provide bitwise matching.

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+        unit: A
+  - entity: number
+    name: Defrost interval
+    category: config
+    hidden: true
+    dps:
+      - id: 101
+        type: integer
+        name: value
+        range:
```
make-all, line 215: unit? Maybe minutes (`min`)?

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+    dps:
+      - id: 117
+        type: integer
+        name: sensor
+        class: measurement
+        unit: A
+  - entity: number
+    name: Defrost interval
+    category: config
+    hidden: true
```
make-all, line 209: can be class: duration if unit is defined

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+    dps:
+      - id: 102
+        type: integer
+        name: value
+        range:
+          min: 1
+          max: 30
+  - entity: number
+    name: Defrost exit temperature
+    category: config
```
make-all, line 233: class: temperature, or temperature_delta?

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+    hidden: true
+    dps:
+      - id: 103
+        type: integer
+        name: value
+        range:
```
make-all, line 241: unit: C?

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+    dps:
+      - id: 104
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 60
+  - entity: number
+    name: Defrost entry differential 1
+    category: config
```
make-all, line 259: class: temperature_delta

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+    hidden: true
+    dps:
+      - id: 105
+        type: integer
+        name: value
```
make-all, line 265: unit: C?

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+      - id: 106
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 60
+  - entity: number
+    name: Superheat compensation
+    category: config
+    hidden: true
```
make-all, line 286: class: temperature_delta?

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+    dps:
+      - id: 107
+        type: integer
+        name: value
+        range:
```
make-all, line 291: unit: C?

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+    category: config
+    hidden: true
+    dps:
+      - id: 108
+        type: integer
+        name: value
+        range:
+          min: 0
+          max: 150
+  - entity: number
```
make-all: This and the following can be combined into a `time` entity.

custom_components/tuya_local/devices/water_chiller_fre1l4.yaml
```diff
@@ -0,0 +1,336 @@
...
+    dps:
+      - id: 127
+        type: integer
+        optional: true
+        name: value
+        range:
+          min: 0
+          max: 59
+  - entity: number
+    name: Ozone timer off hour
```
make-all: Can be combined with the following into a `time` entity

make-all (changes requested):

PR #5658, opened 2026-07-22
Oscal PowerMax2400Pro (EU) fixes and additions

### LINKED ISSUE #5657: Oscal PowerMax2400Pro (EU) incorrect DPs and missing functionality
### Description

In the newest version support for this power station was added with the oscal_powermax2400pro_portablepower.yaml file

There are some slight mis-assignments and some missing DPS for this device

- dp 9: "breath" option missing
- dp 24: missing dps_val for centigrade
- dps 108,37,127: Misleading labeling, that's days/hours/minutes of the remaining time (charging/discharging)
- dp 110: status assignments are incorrect
- dp 114: That's a toggle for the bluetooth speaker, not "Slow charge"
- dp 115: missing: Toggle for the Grid feed-in function
- dp 119: This isn't the pack id, it's the power limit for the primary feed-in period
- dp 128: This is the AC charging limit and it goes up to 1400, not 1200
- dp 140: missing name "DC Standby Time"
- dps 143-146: missing: Primary feed-in period starting and ending times
- dp 147: missing: Secondary feed-in period power limit
- dps 148-151: missing: Secondary feed-in period starting and ending times

### Steps to reproduce

1. Add a Oscal PowerMax2400 Pro as a Device
2. Verify the above mentioned misassignments/missing functionality

### Expected behaviour

The functionality should be available and correctly labeled

### Additional context

_No response_

### Confirmation

linedot:
Corresponding issue: https://github.com/make-all/tuya-local/issues/5657

- dp 9: Added "breath" option, which is a slow pulse of the light
- dp 24: missing dps_val for centigrade
- dp 25: that's not just "sound", it's a toggle that makes the box beep until toggled off, renamed it
- dps 108,37,127: That's days/hours/minutes of the remaining time (charging/discharging)
- dp 110: status assignments were incorrect
- dp 114: That's a toggle for the bluetooth speaker, not "Slow charge"
- dp 115: Toggle for the Grid feed-in function
- dp 119: This isn't the pack id, it's the power limit for the primary feed-in period
- dp 128: This is the AC charging limit and it goes up to 1400, not 1200. Also renamed it to "AC charging limit"
- dp 139: renamed from AC timer to "AC Standby Time"
- dp 140: added name "DC Standby Time"
- dps 143-146: Primary feed-in period starting and ending times
- dp 147: Secondary feed-in period power limit
- dps 148-151: Secondary feed-in period starting and ending times

### DIFF
--- a/custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
+++ b/custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
@@ -64,10 +64,10 @@ entities:
         type: string
         name: unit
         mapping:
-          - dps_val: f
-            value: F
           - dps_val: c
             value: C
+          - dps_val: f
+            value: F
   - entity: switch
     translation_key: keytone
     category: config
@@ -94,23 +94,27 @@ entities:
           - dps_val: f
             value: fahrenheit
   - entity: switch
-    name: Buzzer (Beep)
+    translation_key: sound
     category: config
-    icon: "mdi:volume-high"
     dps:
       - id: 25
         type: boolean
         name: switch
   - entity: sensor
-    name: Time Remaining (Hours)
-    icon: "mdi:timer-sand"
+    translation_key: time_remaining
     category: diagnostic
+    class: duration
     dps:
+      - id: 108
+        type: integer
+        name: day
       - id: 37
         type: integer
         name: sensor
         unit: h
-        class: measurement
+      - id: 127
+        type: integer
+        name: minute
   - entity: switch
     name: Notifications
     category: config
@@ -172,15 +176,6 @@ entities:
         range:
           min: 0
           max: 30
-  - entity: sensor
-    name: Time Remaining (Days)
-    icon: "mdi:calendar-clock"
-    category: diagnostic
-    dps:
-      - id: 108
-        type: integer
-        name: sensor
-        unit: d
   - entity: select
     name: Display timeout
     icon: "mdi:television-ambient-light"
@@ -202,8 +197,8 @@ entities:
           - dps_val: "never_close"
             value: "cancel"
   - entity: sensor
-    name: Device Status
     class: enum
+    translation_key: status
     category: diagnostic
     dps:
       - id: 110
@@ -371,19 +366,11 @@ entities:
         name: sensor
         unit: W
         class: measurement
-  - entity: sensor
-    name: Time Remaining (Minutes)
-    icon: "mdi:timer-sand"
-    category: diagnostic
-    dps:
-      - id: 127
-        type: integer
-        name: sensor
-        unit: min
   - entity: number
     name: AC charging limit
     icon: "mdi:battery-charging-90"
     category: config
+    class: power
     dps:
       - id: 128
         type: integer
@@ -557,47 +544,37 @@ entities:
             value: "12h"
           - dps_val: "24hour"
             value: "24h"
-  - entity: number
-    name: Primary feed-in start hour
+  - entity: time
+    name: Primary feed-in time period start
     category: config
     icon: "mdi:clock-start"
     dps:
       - id: 143
         type: integer
-        name: value
+        name: hour
         range:
           min: 0
           max: 23
-  - entity: number
-    name: Primary feed-in start minute
-    category: config
-    icon: "mdi:clock-start"
-    dps:
       - id: 144
         type: integer
-        name: value
+        name: minute
         range:
           min: 0
           max: 59
-  - entity: number
-    name: Primary feed-in end hour
+  - entity: time
+    name: Primary feed-in time period end
     category: config
     icon: "mdi:clock-end"
     dps:
       - id: 145
         type: integer
-        name: value
+        name: hour
         range:
           min: 0
           max: 23
-  - entity: number
-    name: Primary feed-in end minute
-    category: config
-    icon: "mdi:clock-end"
-    dps:
       - id: 146
         type: integer
-        name: value
+        name: minute
         range:
           min: 0
           max: 59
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
```diff
@@ -64,7 +66,8 @@ entities:
         mapping:
           - dps_val: f
             value: F
-          - value: C
+          - dps_val: c
```
make-all: It is better to leave this as it was, where C is the default.

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
```diff
@@ -91,15 +94,16 @@ entities:
           - dps_val: f
             value: fahrenheit
   - entity: switch
-    translation_key: sound
+    name: Buzzer (Beep)
```
make-all: It is better to leave this as a translation, not a hardcoded string, especially one with punctuation characters in it that might not translate well if users export their entities into other systems (Apple/Google/Alexa/NodeRed etc)

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
```diff
@@ -91,15 +94,16 @@ entities:
...
     category: config
+    icon: "mdi:volume-high"
```
make-all: Icon is not needed if the existing translation is used

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
```diff
@@ -198,20 +202,20 @@ entities:
           - dps_val: "never_close"
             value: "cancel"
   - entity: sensor
+    name: Device Status
```
make-all: Translation key is better, and required for the values to be translated.

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
```diff
@@ -352,28 +372,26 @@ entities:
...
         type: integer
         name: sensor
         unit: min
-        class: measurement
   - entity: number
-    name: Slow charge
+    name: AC charging limit
+    icon: "mdi:battery-charging-90"
     category: config
-    class: power
```
make-all, line 367: Why remove the class: power? If it is not power, then should the unit also be changed?

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
```diff
@@ -91,15 +94,16 @@ entities:
...
-    translation_key: sound
+    name: Buzzer (Beep)
     category: config
+    icon: "mdi:volume-high"
     dps:
       - id: 25
         type: boolean
         name: switch
   - entity: sensor
-    class: duration
```
make-all, line 101: why remove the class: duration?

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
```diff
@@ -169,8 +173,8 @@ entities:
           min: 0
           max: 30
   - entity: sensor
-    name: Estimated battery life
-    class: duration
+    name: Time Remaining (Days)
```
make-all: I think these should all be combined into a single entity for time remaining, instead of 3 different entities that need to be combined externally anyway.

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
```diff
@@ -536,3 +557,104 @@ entities:
             value: "12h"
           - dps_val: "24hour"
             value: "24h"
+  - entity: number
```
make-all: These 2 entities can be combined to a `time` entity.

Same applies to the "end" and secondary ones.

make-all (changes requested):

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
linedot: I'm somewhat struggling with this. Using `entity: time` makes it an editable field and doesn't seem to support days. Making it a sensor IIUC requires one dps to act as a sensor, but then only this one is displayed. Maybe something like this:

```
  - entity: sensor
    category: diagnostic
    class: duration
    dps:
      - id: 108
        type: integer
        name: day
      - id: 37
        type: integer
        name: sensor
        unit: h
      - id: 127
        type: integer
        name: minute
```
And then combine them in HA?

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
linedot: agree, changed

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
linedot: agree, changed

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
linedot: agree, changed

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
linedot: agree, changed

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
linedot: agree, added power class back

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
linedot: added class: duration to the combined sensor

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
linedot: agree, combined both primary and secondary start and end times into "entity: time"

linedot:
I've added most of the requested changes, but I don't know how to combine the remaining time. I made an attempt

custom_components/tuya_local/devices/oscal_powermax2400pro_portablepower.yaml
make-all: This is why I did not suggest it for this one, only the others that were editable and had no days. There is datetime also, but this is also editable.

What I meant for this one is to make a sensor that shows days, and add hours and minutes as attrubutes.

linedot: Changed it so that days is the sensor and hours and minutes are attributes.

linedot:
I changed time remaining as suggested. I am now using this template in HA to show the combined string:
```
template:
  - sensor:
      - name: "PowerMax 2400 Pro time remaining"
        unique_id: tuya_pm2400pro_time_remaining
        icon: mdi:timer-outline
        state: >
          {% set days = states('sensor.pm2400pro_time_remaining') | int(0) %}
          {% set hours = state_attr('sensor.pm2400pro_time_remaining', 'hour') | int(0) %}
          {% set minutes = state_attr('sensor.pm2400pro_time_remaining', 'minute') | int(0) %}
          {{ days }} d {{ hours }} h {{ minutes }} min
```
I have not found a way to make it show up in along with other diagnostics in the device card.

PR #5663, opened 2026-07-23
fix(petoneer_fresco_ezgo_petfountain): map status code 32

K900:
This is not documented on the API, but it is reported when water level is low.

### DIFF
--- a/custom_components/tuya_local/devices/petoneer_fresco_ezgo_petfountain.yaml
+++ b/custom_components/tuya_local/devices/petoneer_fresco_ezgo_petfountain.yaml
@@ -118,6 +118,8 @@ entities:
             value: "Pump clogged"
           - dps_val: 16
             value: "Pump not detected"
+          - dps_val: 32
+            value: "Water level low"
   - entity: binary_sensor
     name: Geyser mode status
     category: diagnostic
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/petoneer_fresco_ezgo_petfountain.yaml
```diff
@@ -118,6 +118,8 @@ entities:
             value: "Pump clogged"
           - dps_val: 16
             value: "Pump not detected"
+          - dps_val: 32
```
make-all, line 123: It might be better to split this one into it's own binary sensor (translation_key: tank_empty), as it is not so much a fault as something that will regularly come up in normal usage.

K900: There's a separate sensor for water level already, it's reporting both ways for some reason.

make-all: Water level is multi stage. This is more of an alert (binary sensor)

K900: Maybe it should just be ignored then, given the other sensor already provides the same info? I'm not entirely sure why the vendor decided to do it like this to begin with.

make-all: It's a bit strange that the water level also duplicates pump faults, but it is quite normal to have a level sensor reporting multiple levels and a binary sensor giving an alert when the tank needs filling.

I have made the change.

PR #5672, opened 2026-07-23
feat(devices): add Cast LB500 series self-cleaning cat litter box

samclark2015:
Adds support for the Cast LB500 series self-cleaning cat litter box (LB500A, LB500C, also sold under the UPFAS brand).

Product ID: `frho2xmkraxldtwo`

Entities:
- Status sensor (idle/smoothing/cleaning)
- Cat weight, last visit weight, last visit duration (all in lb / seconds)
- Visits today (live) and visits yesterday (end-of-day snapshot)
- Fault binary_sensor with descriptions (drum not installed, overload)
- No-cat-detected binary_sensor (24h notification)
- Clean now and Cancel buttons
- Auto-clean switch
- Config selects: clean delay (10–120 s), minimum visit time (10–30 s), cleaning cycles (1–2), minimum cat weight (3.3–3.9 lb)

The device reports weights only in pounds (there is no unit selector DP). DPS 105 ("timerclean", labeled "Regular Clean" in the app) was tested and triggers the same clean cycle as DPS 102, so it is intentionally not exposed as a separate button.

### DIFF
--- a/custom_components/tuya_local/devices/cast_lb500_litterbox.yaml
+++ b/custom_components/tuya_local/devices/cast_lb500_litterbox.yaml
@@ -1,0 +1,220 @@
+## Cast LB500 series self-cleaning cat litter box (LB500A, LB500C, etc.)
+## Also sold under the UPFAS brand.
+##
+## Notes:
+## - Weights are reported in pounds (raw value / 10). Device firmware has no
+##   kg/lb unit selector.
+## - DPS 101 "isnowmode" only reports idle/levelling/cleaning; there is no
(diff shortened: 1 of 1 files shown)


make-all:
Sorry, I cannot accept submissions of devices that are exclusively benefitting the US market. For some time now I have been mulling over whether the US actions in the Hormuz strait that affect all of us meet the same standard as I have used to refuse requests benefitting only Russia and Israel, but the sanctioning of ICC justices is the last straw, and I now believe that retaliatory sanctions are appropriate until such time the populace wakes up and fixes its governance.

PR #5676, opened 2026-07-25
feat(devices): add support for Elspet Orbitie litter box

sthomas727:
Adds support for Elspet Orbitie litter box

Product ID:
nbkqubgxxmanacsv

Entities:
- Status sensor
- Manual clean
- Clean delay
- Indicator light toggle
- Sound toggle
- Kitten mode toggle
- Automatic homing toggle
- Sleep mode toggle
- Sleep start and finish time
- Problem indicator
- Clean count
- Litter level
- Reset litter level
- Full reset

### DIFF
--- a/custom_components/tuya_local/devices/elspet_orbitie_litterbox.yaml
+++ b/custom_components/tuya_local/devices/elspet_orbitie_litterbox.yaml
@@ -54,7 +54,7 @@ entities:
   - entity: switch
     name: Sound
     category: config
-    trnaslation_key: sound
+    translation_key: sound
     dps:
       - id: 117
         type: boolean
--- a/custom_components/tuya_local/devices/elspet_orbitie_litterbox.yaml
+++ b/custom_components/tuya_local/devices/elspet_orbitie_litterbox.yaml
@@ -1,0 +1,172 @@
+name: Cat litter box
+products:
+  - id: nbkqubgxxmanacsv
+    manufacturer: Elspet
+    model: Orbitie
+entities:
+  - entity: sensor
+    translation_key: status
+    category: diagnostic
+    class: enum
+    dps:
+      - id: 109
+        type: string
+        name: sensor
+        mapping:
+          - dps_val: satnd_by
+            value: standby
+          - dps_val: caking
+            value: caking
+          - dps_val: cleaning
+            value: cleaning
+          - dps_val: sleeping
+            value: sleep
+          - dps_val: UV
+            value: sterilizing
+  - entity: button
+    name: Manual clean
+    category: config
+    icon: "mdi:shimmer"
+    dps:
+      - id: 107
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/elspet_orbitie_litterbox.yaml
```diff
@@ -52,7 +52,7 @@ entities:
         type: boolean
         name: switch
   - entity: switch
-    name: Sound
+    name: Indicator sound
```
make-all: Use the translation_key (by removing the name entirely) instead of changing the name to get around the CI error.

sthomas727: Thanks for your feedback. I have updated accordingly, my apologies.

custom_components/tuya_local/devices/elspet_orbitie_litterbox.yaml
```diff
@@ -0,0 +1,170 @@
...
+    dps:
+      - id: 101
+        type: integer
+        name: value
+        unit: s
+        range:
+          min: 0
+          max: 1800
+  - entity: light
+    name: Indicator light
```
make-all: Use `translation_key: indicator` instead of this name

custom_components/tuya_local/devices/elspet_orbitie_litterbox.yaml
```diff
@@ -0,0 +1,170 @@
...
+        name: switch
+  - entity: switch
+    category: config
+    translation_key: sound
+    dps:
+      - id: 117
+        type: boolean
+        name: switch
+  - entity: switch
+    name: Kitten (manual) mode
```
make-all: Just use "Kitten mode" here, parentheses do not translate well into entity ids among other potential problems.

custom_components/tuya_local/devices/elspet_orbitie_litterbox.yaml
```diff
@@ -0,0 +1,170 @@
...
+  - entity: button
+    name: Reset litter
+    icon: "mdi:delete-empty"
+    category: config
+    dps:
+      - id: 113
+        type: boolean
+        name: button
+  - entity: button
+    name: Full reset
```
make-all: I think this should be `translation_key: factory_reset` without the name and class. Generally I mark these `hidden: true` because they are the type of function that users do not want to hit by accident, so for many it would be better not to create the entity by default.

custom_components/tuya_local/devices/elspet_orbitie_litterbox.yaml
sthomas727: I have added this change, along with your other comments. Thanks for the feedback

PR #5690, opened 2026-07-26
Add Energy Saver pool controller YAML

matisere38:
BWT frequency converter for controlled pool pumps ref : PI-03902


custom_components/tuya_local/devices/bfffba1300a3373525gyzj.yaml
```diff
@@ -0,0 +1,380 @@
+name: Energy Saver
+entities:
+  # === COMMANDES DE BASE ===
+  - entity: switch
+    name: Pompe Piscine
```
make-all: This project serves users all over the world. Please use English entity names.
Also note that HA naming convention specifies that only the first word of a device or entity name should be capitalised.

custom_components/tuya_local/devices/bfffba1300a3373525gyzj.yaml
```diff
@@ -0,0 +1,380 @@
...
+entities:
+  # === COMMANDES DE BASE ===
+  - entity: switch
+    name: Pompe Piscine
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+  - entity: select
+    name: Mode Pompe
```
make-all: You can use `translation_key: mode` here for a more general translated name.

custom_components/tuya_local/devices/bfffba1300a3373525gyzj.yaml
```diff
@@ -0,0 +1,380 @@
...
+        name: option
+        mapping:
+          - dps_val: "day"
+            value: "day"
+          - dps_val: "night"
+            value: "night"
+          - dps_val: "backwash"
+            value: "backwash"
+  - entity: sensor
+    name: Puissance
```
make-all: Remove this name to let the class translations be used

custom_components/tuya_local/devices/bfffba1300a3373525gyzj.yaml
```diff
@@ -0,0 +1,380 @@
...
+
+  # === PROGRAMMATION DES PLAGES HORAIRES ===
+  # Prog 1
+  - entity: switch
+    name: Activation Prog 1
+    dps:
+      - id: 133
+        type: boolean
+        name: switch
+  - entity: number
```
make-all: This can be combined with the next entity into a `time` entity.

make-all (changes requested):

matisere38:
Hello, I have taken all the suggestions into account.

make-all:
I still have a couple of concerns about this.

- some entities have the name "Di2 speed", "Di3 speed" etc. There is no Di1, but they come immediately after "Minimum speed" and have the same range. Maybe they are related, and also related to the programs 1-4?

- There are no optional dps. Is this config really matching?

matisere38:
Di2, Di3, and Di4 correspond to the speeds used when the Di2, Di3, and Di4 inputs are connected to COM; Di1, meanwhile, switches the system to OFF.
The minimum speed setting prevents speed 1 (night mode) from dropping below that level.
I can send scans of the documentation.
Here, we have all the DPS values ​​reported by the device; in my case, I only use "=== BASE COMMANDS ===", "=== MAIN SPEEDS ===", and "=== READ-ONLY SENSORS ===".

matisere38:
[CFSpeedRegulator.pdf](https://github.com/user-attachments/files/30870092/CFSpeedRegulator.pdf)

PR #5693, opened 2026-07-26
New dimmer with backlight and 1000 max brightness

RamonJardim:
Hello, working on this new device to cover a new dimmer switch I bought. The current "Simple Dimmer" goes from 10 to 255 brightness, while this one has a maximum of 1000, so the current max on home assistant is actually only about 25% brightness of my device. I'm also adding the backlight configuration for the dimmer which is not available on "Simple Dimmer"

I'm still trying to test it, for some reason I wasn't able to make it show up for me as an option on HA, even though I see the file on the tuya-local folder. Opening the PR as draft for now while I try to fix it.


RamonJardim:
Managed to test it and got my subscription extension on tuya developer to finally find out the DP Id of the backlight. All working now.

RamonJardim:
hey, @make-all. Anything I could work on to move this forward?

PR #5695, opened 2026-07-26
Add Clean Air Optima CA-707

Tweebloesem:
Add support for the Clean Air Optima CA-707 Smart dehumidifier.

Tested functions:

Power
Operating modes
Target humidity
Fan speed
Ionizer
Sleep mode
Temperature
Current humidity
Tank full
Defrost
The timer (DP12) is included with a note because the device firmware only reports the last commanded value instead of the actual remaining time.

### DIFF
--- a/custom_components/tuya_local/devices/cleanair_ca707_dehumidifier.yaml
+++ b/custom_components/tuya_local/devices/cleanair_ca707_dehumidifier.yaml
@@ -1,0 +1,225 @@
+name: Clean Air Optima CA-707
+
+products:
+  - id: gynthidZEhTUCoMN
+    manufacturer: Clean Air Optima
+    model: CA-707 Smart
+
+entities:
+  #
+  # Main
+  #
+  - entity: humidifier
+    class: dehumidifier
+    translation_only_key: extended
+    dps:
+      - id: 1
+        name: switch
+        type: boolean
+
+      - id: 2
+        name: mode
+        type: string
+        mapping:
+          - dps_val: "0"
+            value: normal
+          - dps_val: "1"
+            value: continuous
+          - dps_val: "2"
+            value: laundry
+          - dps_val: "3"
+            value: purify
+
+      - id: 4
+        name: humidity
+        type: integer
+        range:
+          min: 30
+          max: 80
+        mapping:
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/cleanair_ca707_dehumidifier.yaml
```diff
@@ -0,0 +1,225 @@
...
+          min: 30
+          max: 80
+        mapping:
+          - step: 5
+
+      - id: 104
+        name: current_humidity
+        type: integer
+
+      - id: 102
```
make-all: This does not need to be here, as there is already a switch below, and nothing in the humidifier depends on it.

custom_components/tuya_local/devices/cleanair_ca707_dehumidifier.yaml
```diff
@@ -0,0 +1,225 @@
+name: Clean Air Optima CA-707
```
make-all: Top level name should be generic in case other products also match the config. Branding should only be under products.

custom_components/tuya_local/devices/cleanair_ca707_dehumidifier.yaml
```diff
@@ -0,0 +1,225 @@
...
+        type: integer
+
+      - id: 102
+        name: sleep_mode
+        type: boolean
+
+  #
+  # Fan speed
+  #
+  - entity: select
```
make-all: Use a fan entity.

custom_components/tuya_local/devices/cleanair_ca707_dehumidifier.yaml
```diff
@@ -0,0 +1,225 @@
...
+      - id: 103
+        name: sensor
+        type: integer
+        unit: "°C"
+        class: measurement
+
+  #
+  # Current humidity
+  #
+  - entity: sensor
```
make-all: Duplicate of the humidifier current_humidity, not needed as a separate sensor

make-all (changes requested):

Tweebloesem (commented):
proposed changes made

PR #3709, opened 2025-09-08
Add new device tongou towsmr1 20 rcbo breaker

uspass:
<details>
#### Device Management - Query Device Details

{
  "result": {
    "active_time": 1757028663,
    "bind_space_id": "2794571",
    "category": "dlq",
    "create_time": 1757028663,
    "custom_name": "",
    "icon": "smart/icon/bay16570761439434Ehc/bcc0c94725b1ae2c62596ab6695860df.jpg",
    "id": "bf25a33a835f1607faef8b",
    "ip": "188.25.27.249",
    "is_online": false,
    "lat": "44.5168",
    "local_key": "g=x]K9f9KHxxxxxxx",
    "lon": "26.1303",
    "model": "TOSMR1-20JWTR",
    "name": "Breaker",
    "product_id": "hktsllupwmoii1it",
    "product_name": "tongou 20A Wi-Fi Smart Reclosing RCBO",
    "sub": false,
    "time_zone": "+03:00",
    "update_time": 1757028666,
    "uuid": "60d285161d80f9bc"
  },
  "success": true,
  "t": 1757055716596,
  "tid": "35d229128a2611f08e5dda57e8f5d0ed"
}

#### Device Management - Get Update Information

{
  "result": [
    {
      "channel": 0,
      "control_type": 0,
      "current_version": "3.1.28",
      "dev_type": 0,
      "firmware_key": "key83r5jq9qqeaxt",
      "last_upgrade_time": 0,
      "timeout": 0,
      "type": 0,
      "type_desc": "Main Module",
      "upgrade_status": 0
    },
    {
      "channel": 9,
      "control_type": 0,
      "current_version": "1.0.0",
      "dev_type": 0,
      "firmware_key": "",
      "last_upgrade_time": 0,
      "timeout": 0,
      "type": 9,
      "type_desc": "MCU Module",
      "upgrade_status": 0
    }
  ],
  "success": true,
  "t": 1757061151052,
  "tid": "dd0b9f8a8a3211f0805d96c8b66247b7"
}

#### Device Control - Query Properties

{
  "result": {
    "properties": [
      {
        "code": "forward_energy_total",
        "custom_name": "",
        "dp_id": 1,
        "time": 1757028666441,
        "type": "value",
        "value": 1
      },
      {
        "code": "phase_a",
        "custom_name": "",
        "dp_id": 6,
        "time": 1757029788181,
        "type": "raw",
        "value": "CQcAAAAAAAA="
      },
      {
        "code": "leakage_current",
[...]

PR #5704, opened 2026-07-27
Add Tongou TOSMR1-40 reclosing RCBO (40 A variant)

fuentesdelafuente:
Adds support for the 40 A variant of the Tongou TOSMR1 Wi-Fi reclosing RCBO.

### Why a new config instead of a `products` entry on the existing one

`uv run duplicates` reports a **100% structural match** with
`tongou_towsmr1-20_rcbo_breaker`, so per AGENTS.md here is the explanation of the
differences:

The two configs are identical in structure — same dps, same entities. They differ only in:

1. **Product identity** — my device reports product id `f1q6h4xhddjnopxu`
   (`category: dlq`, product name "Breaker"), not `hktsllupwmoii1it`.
2. **Overcurrent threshold range (dps 114)** — `max: 20` → `max: 40`.

Adding my product id to the existing config would have capped the overcurrent threshold at
20 A on a 40 A breaker. In practice this is not cosmetic: with the 20 A config the threshold
gets constrained and the breaker trips when the load exceeds 20 A. My device is protecting an
EV charger circuit that draws up to 32 A.

Raising `max` to 40 on the shared config did not seem right either, as it would allow a 40 A
overcurrent threshold to be set on 20 A hardware.

If you prefer a single config with the product id added and a wider range, I'm happy to
rework it that way — see the note on the manufacturer's table below, which suggests the
firmware may accept the same range on both models.

### Device details

- **Product ID**: `f1q6h4xhddjnopxu`
- **Category**: `dlq`
- **Manufacturer / model**: Tongou TOSMR1-40 (Wi-Fi variant)
- **Discovered via**: cloud assisted setup

### Threshold ranges verified

All six thresholds in the config were checked against the manufacturer's manual
([Smart-Circuit-Breaker-TOSMR1-Series.pdf](https://elcb.net/wp-content/uploads/2024/04/Smart-Circuit-Breaker-TOSMR1-Series.pdf))
and against what the SmartLife app offers for this unit:

| Setting | dps | Existing 20 A config | Manual / app | Change |
|---|---|---|---|---|
| Overcurrent | 114 | 1–20 A | TOSMR1-40: 20–40 A · app offers 1–40 A | **20 → 40** |
| Overvoltage | 115 | 245–295 V | 245–295 V (default 280) | none |
| Undervoltage | 116 | 145–220 V | 145–220 V (default 165) | none |
| Leakage | 117 | 30–100 mA | app offers 30–100 mA | none |
| Temperature | 118 | −25–100 °C | — | none |
| Overpower | 119 | 5–24995 W | — | none |

Note: the manual's table reads
`TOSMR1-40: 20 - 40A (default 40A)` and `TOSMR1-20: 1 - 40A (default 20A)`, which looks like
it may contain a typo. The SmartLife app on my TOSMR1-40 offers 1–40 A, so I used `min: 1`
to stay consistent with the existing config.

### Cross-check against the cloud data model

The Tuya cloud only exposes 6 dps for this device
(1 `forward_energy_total`, 6 `phase_a`, 15 `leakage_current`, 16 `switch`,
109 `online_state`, 142 `relay_status`). All of them match what the existing config expects,
which confirms the mapping. The protection threshold dps (114–119) are not exposed by the
cloud at all — they are local-only, which is exactly why this integration is valuable here.

### Checks run

- `uv run yamllint custom_components/tuya_local/devices` — clean
- `uv run duplicates …` — 100% match with the 20 A config, explained above


make-all:
The details on #3709 show a range of 1-20 for that model, so in the absense of any way to identify which model it is from the dps, a separate config is probably best.

PR #5712, opened 2026-07-28
Add support for Tesla Smart Pet Sofa PS300

Koty97:
Add support for the Tesla Smart Pet Sofa PS300 (product ID `13oghytketwcoz1z`).
Entities
* Climate
  * HVAC mode (Off / Heat_Cool)
  * Target temperature
  * Current temperature
  * Preset mode (Auto, Manual, Smart)
  * HVAC action (Idle, Heating, Cooling)
  
* Ambient temperature sensor
* Fault sensor
* Schedule (text)
* Push state (diagnostic)

Testing
Tested against a physical Tesla Smart Pet Sofa PS300 using protocol 3.3.

### DIFF
--- a/custom_components/tuya_local/devices/tesla_smart_sofa.yaml
+++ b/custom_components/tuya_local/devices/tesla_smart_sofa.yaml
@@ -1,0 +1,95 @@
+name: Tesla Smart Pet Sofa PS300
+
+products:
+  - id: 13oghytketwcoz1z
+    manufacturer: Tesla
+    model: Smart Pet Sofa PS300
+
+entities:
+  - entity: climate
+    icon: mdi:sofa
+    dps:
+      - id: 1
+        name: hvac_mode
+        type: boolean
+        mapping:
+          - dps_val: true
+            value: "heat_cool"
+          - dps_val: false
+            value: "off"
+
+      - id: 8
+        name: temperature
+        type: integer
+        unit: C
+        range:
+          min: 0
+          max: 45
+          step: 1
+
+      - id: 16
+        name: current_temperature
+        type: integer
+
+      - id: 101
+        name: hvac_action
+        type: integer
+        mapping:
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/tesla_smart_sofa.yaml
```diff
@@ -0,0 +1,79 @@
+name: Tesla Smart Pet Sofa PS300
+
+products:
+  - id: 13oghytketwcoz1z
+    manufacturer: Tesla
+    model: Smart Pet Sofa PS300
+
+entities:
+  - entity: climate
+    icon: mdi:sofa
```
make-all: icons require quotes around them to ensure the : does not cause problems for the yaml parser

custom_components/tuya_local/devices/tesla_smart_sofa.yaml
```diff
@@ -0,0 +1,79 @@
...
+  - entity: sensor
+    translation_key: ambient_temperature
+    class: temperature
+    dps:
+      - id: 102
+        type: integer
+        name: sensor
+        unit: "C"
+
+  - entity: sensor
```
make-all: Standard pattern here is to use a binary_sensor with `class: problem`. About 90% of configs have an example you can follow.

custom_components/tuya_local/devices/tesla_smart_sofa.yaml
```diff
@@ -0,0 +1,79 @@
...
+    name: Fault
+    category: diagnostic
+    dps:
+      - id: 22
+        type: integer
+        name: sensor
+
+  - entity: sensor
+    name: Push state
```
make-all: It isn't clear what this sensor is for. It may be better as an attribute if it is nothing important.

make-all (changes requested):

PR #5720, opened 2026-07-28
feat(devices): add FrankEver position-controlled water valve

larstobi:
Adds support for the FrankEver Smart Water Valve with product ID
`nzx0kku9d6eq59nt`.

The device uses DP 101 for the target position, DP 102 for position
feedback, and DP 1 as the enable/off control. This also adds valve support
for separate target and current-position DPs so Home Assistant reports the
position correctly.

Countdown on DP 9 and power-on state on DP 38 are retained.

`duplicates` reports a 100% DPS match with the Tellur TL331501 profile.
The difference is that the Tellur profile treats DP 102 as the writable
position and DP 101 as a separate maximum setting, while this device uses
DP 101 as the writable target and DP 102 as read-only feedback.

Tested against a physical protocol 3.5 device. Commands and DP feedback
were confirmed. Automated tests cover 0%, 20%, 50%, and 100%, including
open, close, target position, current position, and the DP 1 interaction.

Validation:
- 418 tests passed
- Ruff checks and formatting passed
- YAML lint passed

The two reported test warnings are pre-existing AsyncMock warnings in the
humidifier tests and are unrelated to this change.


custom_components/tuya_local/devices/frankever_watervalve.yaml
```diff
@@ -0,0 +1,51 @@
...
+        name: valve
+        range:
+          min: 0
+          max: 100
+        mapping:
+          - step: 10
+      - id: 102
+        type: integer
+        name: current_position
+        readonly: true
```
make-all: readonly is not needed in contexts where the attribute is implicitly read-only

custom_components/tuya_local/devices/README.md
```diff
@@ -809,6 +809,7 @@ to use it for other length timers.
 ### `valve`
 - **valve** (required, boolean or integer): a dp that reports the current state of the valve, and if not readonly, can also be used to set the state.  If a number, it should be a percentage between 0 and 100 indicating how far open the valve is.  If a boolean, it should indicate open (true) or closed (false).
 - **switch** (optional, boolean): if the valve dp is an integer, the valve may also have a boolean switch dp for closing and opening the valve without affecting the open valve position.
+- **current_position** (optional, number 0-100): a read-only dp that reports the actual position when the writable **valve** dp is only a target position. If a separate **switch** is also present, position commands set both the target and switch so that 0 closes the valve and non-zero positions enable it. Open and close commands set 100 and 0 respectively.
```
make-all: I don't understand why you see a need to change the behaviour of the open and close commands because of the presence of this dp.

custom_components/tuya_local/devices/frankever_watervalve.yaml
larstobi: Done, thanks!

custom_components/tuya_local/devices/README.md
larstobi: I initially used combined writes because I assumed DP 1 was an enable switch that had to be kept in sync with the target position on DP 101. After your comment, I tested them separately on the physical device. The device handles the relationship itself: closing through DP 1 resets the position DPs to 0, and setting a non-zero target on DP 101 enables DP 1 automatically. I have therefore removed the combined-write behavior. Thanks for the guidance!

PR #5722, opened 2026-07-28
Add support for Friedrich Floating Air conditioner

Lionwo:
## Description

This PR adds support for the Friedrich Floating Air Wi-Fi mini-split air conditioner.

### Product ID
- iufq71jwsputqsct

### Supported features
- Power
- HVAC modes
- Target temperature
- Current temperature
- Fan modes
- Vertical swing
- Horizontal swing
- Eco mode
- Sleep mode
- Quiet mode
- Surge mode
- 46°F Heat mode
- Fahrenheit/Celsius

This profile was created by identifying and validating the device DPS mappings on the physical hardware.

### DIFF
--- a/custom_components/tuya_local/devices/friedrich_floating_air.yaml
+++ b/custom_components/tuya_local/devices/friedrich_floating_air.yaml
@@ -108,7 +108,7 @@ entities:
         name: switch
 
   - entity: switch
-    name: Sleep
+    translation_key: sleep
     icon: mdi:sleep
     category: config
     dps:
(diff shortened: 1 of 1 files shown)


make-all:
Sorry, I cannot accept submissions of devices that are exclusively benefitting the US market. For some time now I have been mulling over whether the US actions in the Hormuz strait that affect all of us meet the same standard as I have used to refuse requests benefitting only Russia and Israel, but the sanctioning of ICC justices is the last straw, and I now believe that retaliatory sanctions are appropriate until such time the populace wakes up and fixes its governance.

PR #5723, opened 2026-07-29
Add support for Kogan heater fan

SiwatS:
Adds support for the Kogan heater fan.

It is a fan heater with `heat` and `fan_only` modes, oscillation, a countdown
timer, sound toggle and a 0-2 heat level. dp 1 is the main power switch and
dp 101 selects between heating and fan-only, so `hvac_mode` is mapped from
dp 1 with a condition on dp 101.

Tested against a physical device on protocol 3.4. All 12 dps the device
reports are covered by the config, and it matches at 100% quality.

I have not included a `products:` block as I do not have a reliable product id
for this device; it matches on dps.

### DIFF
--- a/custom_components/tuya_local/devices/kogan_heater_fan.yaml
+++ b/custom_components/tuya_local/devices/kogan_heater_fan.yaml
@@ -1,0 +1,153 @@
+name: Heater fan
+entities:
+  - entity: climate
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: heater
+            conditions:
+              - dps_val: true
+                value: heat
+              - dps_val: false
+                value: fan_only
+      - id: 101
+        type: boolean
+        name: heater
+        hidden: true
+      - id: 20
+        type: integer
+        name: temperature
+        range:
+          min: 15
+          max: 35
+        unit: C
+      - id: 21
+        type: integer
+        name: current_temperature
+      - id: 3
+        type: integer
+        name: fan_mode
+        mapping:
+          - dps_val: 1
+            value: "1"
+          - dps_val: 2
+            value: "2"
+          - dps_val: 3
+            value: "3"
+          - dps_val: 4
+            value: "4"
+          - dps_val: 5
+            value: "5"
+          - dps_val: 6
+            value: "6"
+          - dps_val: 7
+            value: "7"
+          - dps_val: 8
+            value: "8"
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/kogan_heater_fan.yaml
```diff
@@ -0,0 +1,131 @@
...
+        name: current_temperature
+      - id: 2
+        type: string
+        name: fan_mode
+        mapping:
+          - dps_val: strong
+            value: high
+          - constraint: speed
+            conditions:
+              - dps_val: 1
```
make-all: Why are the 9 speeds here mapped to only two distinct values?

SiwatS: Good catch. That collapse was a workaround for fitting a 9-step integer speed into climate's `fan_mode`, not something the device requires, so I've replaced it.

dp 3 is now mapped to `fan_mode` `"1"` through `"9"` one-to-one, matching `hyd_weair_plus_fan.yaml`. dp 2 isn't a speed at all. It's a turbo mode that drives the fan harder than speed 9 and ignores dp 3 while active, so it's now a `boost` `preset_mode` rather than being folded into the speed scale.

I've also added a `fan` entity (switch dp 1, speed dp 3 as a 1 to 9 range auto-scaled to a percentage, the same boost preset, oscillate dp 5), so the speed is reachable as a real fan percentage and voice assistants handle "turn the fan up/down". Sharing dps between `climate` and `fan` follows `breville_airrounderplus_heater.yaml`.

Rebased onto current main.

make-all (commented):
If you are going to add the fan entity, it might be better to completely separate the fan from the heating (climate). So remove fan_mode and preset_mode from the climate entity, and simplify the hvac_mode to only toggle between heat and "off", with the fan controlling the main switch. The heat level could then be brought into the climate entity as preset_mode, using the standard HA "eco", "comfort", "boost" preset names, or from the heater translation_key in this integration you could use "low", "medium", "high" instead.

I haven't seen exactly what type of product this is, but if it is primarily a fan with heating as an additional feature, then completely separating them like this might make sense.

But if it is primarily a heater, with a fan-only mode, then keeping the hvac_mode as it is will probably be better even though it is duplicating the fan switch to some extent.

make-all:
Speculatively made the change to make fan the primary function, on assumption the device is something like this:
https://help.kogan.com/s/article/KoganSmarterHomeBladelessPurifierFanHeaterKABLFAN3RSAManualsandSupport

Can change later if you disagree.

PR #5726, opened 2026-07-29
Add support for Aquaforte SC9xx Series Pool Heatpump

Bukem7:
This adds support for the Aquaforte SC9xx Series Full DC Inverter Pool Heatpump.

Tested with:
- Aquaforte SC981
- Tuya protocol 3.3
- Home Assistant tuya_local 2026.7.2

Supported DPS:
- 1: power state
- 2: target temperature
- 3: current temperature
- 4: operating mode
- 9: fault code

The Tuya Product ID could not be retrieved. The device was successfully configured using Device ID, Local Key, IP address and protocol version.

### DIFF
--- a/custom_components/tuya_local/devices/Aquaforte_SC9xx_series_Pool_Heatpump
+++ b/custom_components/tuya_local/devices/Aquaforte_SC9xx_series_Pool_Heatpump
@@ -1,0 +1,62 @@
+name: Aquaforte SC9xx Series Pool Heatpump
+
+# Tested with:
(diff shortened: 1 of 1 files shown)


Bukem7:
Screenshot 2026-07-29 at 16 30 02

make-all:
This is essentially the same as remora_heatpump (the lower temperature range differs, but this is minor, and I don't think the 5-8 degree range is going to matter to users in practice).

Since the product id is not supplied, I cannot make the change to prioritise the remora config over others with the same dps layout that use different preset names.

PR #5732, opened 2026-07-30
Add Nedis WIFIWEST500WT weather station configuration

Jampie85:
Adds support for the Nedis WIFIWEST500WT weather station.

Tested locally on my own device. All available datapoints are included.

### DIFF
--- a/custom_components/tuya_local/devices/e1mpm0mc_weather_station.yaml
+++ b/custom_components/tuya_local/devices/e1mpm0mc_weather_station.yaml
@@ -1,0 +1,126 @@
+name: Nedis WIFIWEST500WT Weather station
+products:
+  - id: d3cxdxqogwqvlwax
+    manufacturer: Nedis
+    model: WIFIWEST500WT
+
+entities:
+  - entity: sensor
+    name: Indoor temperature
+    class: temperature
+    dps:
+      - id: 101
+        type: integer
+        name: sensor
+        class: measurement
+        unit: "°C"
+        mapping:
+          - scale: 10
+
+  - entity: sensor
+    name: Indoor humidity
+    class: humidity
+    dps:
+      - id: 102
+        type: integer
+        name: sensor
+        class: measurement
+        unit: "%"
+        mapping:
+          - scale: 10
+
+  - entity: sensor
+    name: Outdoor temperature
+    class: temperature
+    dps:
+      - id: 103
+        type: integer
+        name: sensor
+        class: measurement
+        unit: "°C"
+        mapping:
+          - scale: 10
+
+  - entity: sensor
+    name: Outdoor humidity
+    class: humidity
+    dps:
+      - id: 104
+        type: integer
+        name: sensor
+        class: measurement
+        unit: "%"
+        mapping:
+          - scale: 10
+
+  - entity: sensor
+    name: Pressure
+    class: atmospheric_pressure
+    dps:
+      - id: 109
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/e1mpm0mc_weather_station.yaml
```diff
@@ -0,0 +1,126 @@
+name: Nedis WIFIWEST500WT Weather station
```
make-all, line 1: Top level name should be generic, as other products may also match the same config

custom_components/tuya_local/devices/e1mpm0mc_weather_station.yaml
```diff
@@ -0,0 +1,126 @@
...
+      - id: 104
+        type: integer
+        name: sensor
+        class: measurement
+        unit: "%"
+        mapping:
+          - scale: 10
+
+  - entity: sensor
+    name: Pressure
```
make-all, line 57: name not neccesary

custom_components/tuya_local/devices/e1mpm0mc_weather_station.yaml
```diff
@@ -0,0 +1,126 @@
...
+      - id: 109
+        type: integer
+        name: sensor
+        class: measurement
+        unit: "hPa"
+        mapping:
+          - scale: 10
+
+  - entity: sensor
+    name: Wind speed (avg)
```
make-all, line 69: Name not necessary. Puntuation should be avoided in entity names, as they can cause problems when exporting the entity to systems outside HA control

custom_components/tuya_local/devices/e1mpm0mc_weather_station.yaml
```diff
@@ -0,0 +1,126 @@
...
+    class: wind_speed
+    dps:
+      - id: 111
+        type: integer
+        name: sensor
+        class: measurement
+        unit: "km/h"
+
+  - entity: sensor
+    name: Wind direction
```
make-all, line 89: Use `class: wind_direction`, and a mapping to map output values to the expected compass degrees.

custom_components/tuya_local/devices/e1mpm0mc_weather_station.yaml
```diff
@@ -0,0 +1,126 @@
...
+    icon: "mdi:compass"
+    dps:
+      - id: 112
+        type: string
+        name: sensor
+
+  - entity: sensor
+    name: Rain last hour
+    class: precipitation
```
make-all, line 98: Use `class: precipitation_intensity` (no name required to allow translations to be used).

Unit should then be mm/h

custom_components/tuya_local/devices/e1mpm0mc_weather_station.yaml
```diff
@@ -0,0 +1,126 @@
...
+      - id: 113
+        type: integer
+        name: sensor
+        class: measurement
+        unit: "mm"
+        mapping:
+          - scale: 100
+
+  - entity: sensor
+    name: Rain today
```
make-all, line 109: No name required, to let translations be used

make-all (changes requested):

make-all:
Seems to be a match for the existing em3390tf_weatherstation

If there is some specific difference that prevents that config from being used, please reply, or this PR will be closed.

PR #5739, opened 2026-07-31
Add generic "PRO V2 W/Wi-Fi & Bluetooth LE" electric fireplace (v4r5dzjgkhnagnqc)

mabrews:
New device config for an electric fireplace, product ID `v4r5dzjgkhnagnqc`. No branding found on the unit or in the Tuya app beyond the generic name "PRO V2 W/Wi-Fi & Bluetooth LE" - seems to be a whitelabeled OEM product.

DP info came from the Tuya Cloud `/v2.0/cloud/thing/{id}/model` endpoint, cross-checked against a local `status()` poll of the real device. That data model gave descriptions (in Chinese) for all 20 DPs, including the ones outside the standard "specification" schema (the RGB/mode/custom-scene ones).

What's mapped:
- `climate`: power, target/current temp (C and F), heater level (off/low/high)
- `select`: temp unit, timer, shared light effect mode, flame motor speed
- `sensor`: timer remaining
- `light` x3 (flame/ember/downlight): each has its own DP encoding RGB+HSV as a 7-byte hex string (R,G,B,H(2b),S,V) - only H/S/V are used, R/G/B are redundant duplicates per the device's own docs, so they're pulled in as unused spare fields to keep the byte offsets right
- `switch`: sync-lights toggle
- 3 custom-scene DPs + weekly-schedule DP are exposed as hidden/opaque text fields - their packed formats are variable-length free-form byte sequences (documented in the cloud data model) that don't fit the fixed-format decoding the format: option supports, so I didn't try to force them into real controls

`pytest tests/test_device_config.py`, `untranslated_entities`, and `yamllint` all pass locally. I haven't yet confirmed entity behavior against the live device through the HA UI (pending a restart on my end) - will follow up in this PR if anything needs adjusting once I've done that.

### DIFF
--- a/custom_components/tuya_local/devices/fireplace_pro_v2.yaml
+++ b/custom_components/tuya_local/devices/fireplace_pro_v2.yaml
@@ -3,6 +3,8 @@ name: Fireplace
 # "PRO V2 W/Wi-Fi & Bluetooth LE"
 products:
   - id: v4r5dzjgkhnagnqc
+    manufacturer: Modern Flames
+    model: Landscape Pro Slim
 entities:
   - entity: climate
     icon: "mdi:fireplace"
@@ -23,6 +25,7 @@ entities:
                 value: heat
               - dps_val: "3"
                 value: heat
+                hidden: true
       - id: 2
         name: temperature
         type: integer
@@ -60,8 +63,7 @@ entities:
         name: temperature_unit
         type: string
         mapping:
-          - dps_val: c
-            value: C
+          - value: C
           - dps_val: f
             value: F
       - id: 20
@@ -126,7 +128,7 @@ entities:
         name: sensor
         unit: min
   - entity: select
-    name: Light effect
+    name: Light preset
     icon: "mdi:palette"
     category: config
     dps:
@@ -167,11 +169,11 @@ entities:
         name: rgbhsv
         type: hex
         format:
-          - name: spare_r
+          - name: r
             bytes: 1
-          - name: spare_g
+          - name: g
             bytes: 1
-          - name: spare_b
+          - name: b
             bytes: 1
           - name: h
             bytes: 2
@@ -199,11 +201,11 @@ entities:
         name: rgbhsv
         type: hex
         format:
-          - name: spare_r
+          - name: r
             bytes: 1
-          - name: spare_g
+          - name: g
             bytes: 1
-          - name: spare_b
+          - name: b
             bytes: 1
           - name: h
             bytes: 2
@@ -232,11 +234,11 @@ entities:
         name: rgbhsv
         type: hex
         format:
-          - name: spare_r
+          - name: r
             bytes: 1
-          - name: spare_g
+          - name: g
             bytes: 1
-          - name: spare_b
+          - name: b
             bytes: 1
           - name: h
             bytes: 2
@@ -319,7 +321,7 @@ entities:
         name: value
         optional: true
   - entity: text
-    name: Weekly schedule
+    translation_key: schedule
     icon: "mdi:calendar-clock"
     category: config
     hidden: true
--- a/custom_components/tuya_local/devices/fireplace_pro_v2.yaml
+++ b/custom_components/tuya_local/devices/fireplace_pro_v2.yaml
@@ -1,0 +1,321 @@
+name: Fireplace
+# manufacturer/model unknown - sold under multiple brands as
+# "PRO V2 W/Wi-Fi & Bluetooth LE"
+products:
+  - id: v4r5dzjgkhnagnqc
+entities:
+  - entity: climate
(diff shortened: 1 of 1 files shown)


make-all:
Closed accidentally, sorry

mabrews:
No problem! This has now been fully tested on the device, and I made a small change to the config as a result. Thanks!

custom_components/tuya_local/devices/fireplace_pro_v2.yaml
```diff
@@ -0,0 +1,330 @@
+name: Fireplace
+# manufacturer/model unknown - sold under multiple brands as
+# "PRO V2 W/Wi-Fi & Bluetooth LE"
+products:
+  - id: v4r5dzjgkhnagnqc
```
make-all, line 5: Include the `model` here

custom_components/tuya_local/devices/fireplace_pro_v2.yaml
```diff
@@ -0,0 +1,330 @@
...
+            value: "off"
+          - dps_val: true
+            value: fan_only
+            constraint: fan_mode
+            conditions:
+              - dps_val: "1"
+                value: fan_only
+              - dps_val: "2"
+                value: heat
+              - dps_val: "3"
```
make-all, line 26: one of these needs to be marked `hidden: true` to make selecting "heat" deterministic.

custom_components/tuya_local/devices/fireplace_pro_v2.yaml
```diff
@@ -0,0 +1,330 @@
...
+            value: "off"
+          - dps_val: "2"
+            value: low
+          - dps_val: "3"
+            value: high
+      - id: 19
+        name: temperature_unit
+        type: string
+        mapping:
+          - dps_val: c
```
make-all: It is better to make C the default (no dps_val filter), so the climate entity still reports a valid unit to HA in case something goes wrong (like this dp not reported by the device)

custom_components/tuya_local/devices/fireplace_pro_v2.yaml
```diff
@@ -0,0 +1,330 @@
...
+    translation_key: time_remaining
+    class: duration
+    category: diagnostic
+    dps:
+      - id: 12
+        type: integer
+        name: sensor
+        unit: min
+  - entity: select
+    name: Light effect
```
make-all: This can be the `effect` attribute of a light entity.

custom_components/tuya_local/devices/fireplace_pro_v2.yaml
```diff
@@ -0,0 +1,330 @@
...
+    category: config
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+      - id: 101
+        name: rgbhsv
+        type: hex
+        format:
+          - name: spare_r
```
make-all: I think it is better to use "r", "g", "b" here, as the integration will then handle keeping them in sync with hsv changes. I don't know what the device will do if rgb conflicts with hsv when sending new values. Maybe it always prioritises hsv so it acts ok, but there isn't really a way of knowing there isn't some edge case where it leads to buggy behaviour.

custom_components/tuya_local/devices/fireplace_pro_v2.yaml
```diff
@@ -0,0 +1,330 @@
...
+    icon: "mdi:eye"
+    category: diagnostic
+    hidden: true
+    dps:
+      - id: 108
+        type: string
+        name: value
+        optional: true
+  - entity: text
+    name: Weekly schedule
```
make-all: Use `translation_key: schedule` instead of a hardcoded name

make-all (changes requested):

custom_components/tuya_local/devices/fireplace_pro_v2.yaml
mabrews: "Light effect" was not the best choice for a name, dp4 is more of a preset list that controls the flame light, ember light, and downlight simultaneously. Rather than add to a single light entity I'm going to leave it as an independent select entity and rename it "Light preset."

make-all: Maybe use `translation_key: scene` instead

mabrews:
Made all requested changes except making dp4 the `effect` of a light entity. See comment, but I'm also happy for it to be an effect of the Flame entity if that better matches existing configs. Thanks!

custom_components/tuya_local/devices/fireplace_pro_v2.yaml
mabrews: Yeah that works, thanks for the suggestion.

PR #5754, opened 2026-08-01
Add Eberg Qubo Q40NE AC

Ferszus:
Adds a separate device configuration for the Eberg Qubo Q40NE portable AC.

The device is similar to the existing Q40HD profile, but DPS 101 incorrectly reports `heat_s` while the unit is cooling. This causes Home Assistant to display `heating`.

For the Q40NE profile, `hvac_action` is mapped from DPS 4:

- `cold` -> `cooling`
- `hot` -> `heating`
- `dehumidify` -> `drying`

- Maps HVAC action from DPS 4 because DPS 101 reports an incorrect state.
- Maps DPS 30 as horizontal swing instead of vertical swing.
- 

`DPS 4 represents the selected mode rather than the real-time compressor state.`

Tested locally on an Eberg Qubo Q40NE with cooling, heating and drying modes.

obraz
obraz
obraz

### DIFF
--- a/custom_components/tuya_local/devices/eberg_qubo_q40ne_airconditioner.yaml
+++ b/custom_components/tuya_local/devices/eberg_qubo_q40ne_airconditioner.yaml
@@ -1,4 +1,4 @@
-name: Portable heat pump
+name: Portable air conditioner
 # products: Eberg Qubo Q40NE
 entities:
   - entity: climate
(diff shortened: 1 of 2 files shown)


make-all:
This is unnecessary, as these are issues that can be fixed in the existing config

PR #5763, opened 2026-08-02
2 Bug fixes and New Calibration Feature for "ANY" Device

cataseven:
# 1. Theme-independent QR code in the cloud login flow

## Summary

Renders the Smart Life login QR code server side as a fixed black-on-white PNG and embeds it in the `scan` step description, replacing the frontend `QrCodeSelector`. The QR code is now scannable with every theme.

## Motivation

The frontend `ha-qr-code` component derives its colors from the theme variables (`--rgb-primary-text-color`, `--rgb-card-background-color`). Many custom themes do not define the `--rgb-*` variants, in which case the QR code renders as a blank/white image and cloud onboarding becomes impossible. Embedding a `data:image` URI is not an option either, because the markdown XSS filter strips `data:` sources — so the image is served by the integration itself.

## Changes

- `config_flow.py`
  - New `TuyaLocalQrView` (`/api/tuya_local/qr/{token}`, `requires_auth`): serves the PNG stored per flow in `hass.data`. Access is via **signed URLs** (`async_sign_path`, 15 min expiry), since `` requests carry no auth header.
  - `_qr_code_url()` renders the QR with `segno` (error level Q, fixed black on white), stores it keyed by `flow_id` (oldest-first eviction beyond 16 entries) and returns the signed URL.
  - `async_step_scan` shows the image via a `{qr_url}` description placeholder in both the initial and the retry-after-failure paths; `async_remove` deletes the stored image when the flow ends.
- `manifest.json`: added `segno~=1.6.6` requirement and `http` dependency (also mirrored in `pyproject.toml`).
- Translations: all 23 language files embed `QR` in `config.step.scan.description`. Also added the previously **missing** `config.error.login_error` string to all 23 files — the code already supplied `{msg}`/`{code}` placeholders, but users saw the raw `login_error` key on login failures.

## Backward compatibility

The flow steps and stored entry data are unchanged; only the presentation of the QR code moved from a form selector into the step description.

## Validation

Signed-path auth, view registration, `async_remove` invocation on all flow termination paths and empty-schema submit behavior were verified against the HA 2026.7 source. All translation files pass a recursive key-parity check against `en.json`. `ruff check` / `ruff format --check` clean.

---

# 2. Offline devices no longer fail setup / flag "Needs attention"

## Summary

A configured device that is offline at startup no longer raises `ConfigEntryNotReady`. This is like integration is failed entirely but in fact it is not, it is only a offline device! The entry loads normally, a single WARNING is logged, its entities show as unavailable, and everything comes to life automatically when the device first responds.

## Motivation

Devices are not expected to be powered 24/7, so the permanent "Failed setup, will retry: tuya-local device offline" banner on the integrations page was noise. Users who want an offline notification can already build one with a simple automation on the entities becoming unavailable, and the logged warning already provides sufficient information.

## Changes

- `__init__.py`
  - When the initial refresh returns no state, setup continues instead of raising; a WARNING is logged once.
  - An entry-scoped watcher polls every 30 s and schedules an entry reload (`async_schedule_reload`) after the device answers for the first time — some entity features are derived from live dp values at construction time, so the entities are rebuilt once real state exists. `async_schedule_reload` is used instead of awaiting `async_reload`, because the entry-scoped task would be cancelled by the unload that a direct await triggers.
  - The update listener is now registered via `entry.async_on_unload(entry.add_update_listener(...))` and the listener reloads through the config-entries manager instead of calling unload/setup directly. This also fixes a pre-existing bug where every options change stacked an additional listener, causing concurrent unload/setup races.
- `device.py`
  - New `has_received_state` property: true only after an actual device reply (`updated_at > 0`), unlike `has_returned_state` which is also satisfied by locally queued, unacknowledged commands. The watcher uses it to avoid spurious reloads triggered by service calls to an offline device.
  - The receive loop now backs off exponentially (5 s → 300 s cap) while a device has never answered, instead of hammering the network with continuous connection attempts that keep an executor thread busy. The interval resets to normal as soon as the device responds.

## Trade-offs

- A device that has been offline for a long time reconnects within at most ~5 minutes of coming back (backoff cap), plus instantly on IP change via the existing LAN rediscovery sweep.
- Genuine setup errors (exceptions during device creation) still raise `ConfigEntryNotReady` as before.

## Validation

Watcher/reload cancellation semantics, listener deduplication and background-task lifecycle were verified against the HA 2026.7 source (`config_entries.py`). An adversarial review pass surfaced the duplicate-listener bug, the pending-update false-positive and the executor-hammering issue; all are addressed in this PR. `ruff` clean.

---

# Per-device measurement calibration (`tuya_local.set_calibration`)

## Summary

Adds integration-level, per-device calibration for devices that have no native calibration dp. A new `tuya_local.set_calibration` action stores a persistent offset for an entity's reading; the offset is added to the value reported by the device.

## Motivation

Identical devices share one config file, so a single inaccurate unit cannot be corrected through YAML. Example: three identical thermometers side by side, one reads 0.4° too high. With this PR:

```yaml
action: tuya_local.set_calibration
data:
  entity_id: sensor.thermometer_3_temperature
  offset: -0.4
```

only that device's reading is shown 0.4 lower — as its real value. `offset: 0` removes the calibration.

## Changes

- `services.py` — new `set_calibration` action (`entity_id`, `offset`, optional `attribute`). Resolves the entity to its config entry via the entity registry, validates the dp exists and is numeric, and stores the offset in the entry **options** (`options["calibration"]["<entity_config_id>/<dp_name>"]`). The options update reloads the entry through the existing update listener (with an `async_schedule_reload` fallback if an update lands mid-reload). Errors are raised as user-friendly `ServiceValidationError`s.
- `__init__.py` — the action is registered in `async_setup` (`CONFIG_SCHEMA = config_entry_only_config_schema`), as it needs no entry state.
- `device.py` — `TuyaLocalDevice` carries the calibration map (`get_calibration`, `has_calibration`); `setup_device` passes it in from the merged entry config.
- `helpers/device_config.py` — `TuyaDpsConfig._calibrated()` adds the offset at the end of `get_value()` (after mapping/scale), for numeric values only.
- `entity.py` — active offsets are surfaced as `<dp_name>_calibration` attributes on the entity.
- `config_flow.py` — the options flow now merges over existing options when saving, so editing connection settings does not wipe calibration.
- Translations: `services.set_calibration` strings in all 23 language files, `services.yaml` definition, `mdi:tune` icon in `icons.json`.
- `README.md` — new "Sensor calibration" section.

## Design decisions

- **Calibration is read-only.** Settable dps (target temperatures, numbers) are deliberately rejected: HA core validates writes against the *uncalibrated* entity bounds (an offset-wide band would fail with errors), step snapping would make writes read back differently, and constraint condition co-writes in device configs operate on raw device values. Allowed targets are therefore measurement dps only (`CALIBRATABLE_DPS`): `sensor` value, climate/water_heater `current_temperature`, climate/humidifier `current_humidity`. `bitfield` and `unixtime` raw types are rejected.
- **Stored in entry options, not YAML** — devices sharing a config file are corrected individually, the setting survives restarts, and config files remain untouched.

## Backward compatibility

No behavior change for devices without calibration; no new dependency; entity ids and unique_ids are unchanged.

## Validation

- `ruff check` + `ruff format --check` clean; modules imported against a real HA 2026.7 environment.
- All 23 translation files pass a recursive key-parity check.
- End-to-end unit verification: raw 215 → scale 10 → 21.5; with -0.4 offset → 21.1; settable dp writes unaffected by stray offsets; readonly dp writes return empty.
- An adversarial review pass (value pipeline, service/options lifecycle, translation/packaging consistency) surfaced write-path interactions, a reload race, service registration timing and an options-wipe bug — all fixed within this PR.


make-all:
Nobody asked for this.

cataseven:
Thanks for considering and the vision. Cheers

PR #5764, opened 2026-08-02
Add Trevi InnovAqua variable speed pool heat pump

x409no1:
Adds a device config for the Trevi InnovAqua variable speed pool heat pump (Tuya product id `bqfay9re2z3hfhoq`, local protocol 3.5).

All datapoints were confirmed from a full local dump of my own unit: on/off (dp1), Celsius setpoint (dp2) and inlet water temperature (dp3), mode silent/powerful/heat/cool (dp4), fault bitfield (dp9), display unit (dp17), and the native Fahrenheit setpoint (dp25) and inlet water temperature (dp26).

The climate uses the Celsius datapoints. The device's native Fahrenheit setpoint and water temperature are exposed as separate `number` and `sensor` entities. I also tried modelling the climate directly on the Fahrenheit datapoints (dp25/dp26), but tuya-local does not create the climate entity for it on this device, so Celsius is used for the climate.

Validated against `device_config_schema.json` and `yamllint`.

### DIFF
--- a/custom_components/tuya_local/devices/trevi_innovaqua_heatpump.yaml
+++ b/custom_components/tuya_local/devices/trevi_innovaqua_heatpump.yaml
@@ -6,9 +6,7 @@ products:
 # Mapping validated on Trevi InnovAqua variable speed over local protocol 3.5.
 # The Celsius setpoint (dp2) and inlet water temp (dp3) drive the climate.
 # The device also exposes an integer Fahrenheit setpoint (dp25) and inlet
-# water temperature (dp26), provided as separate number and sensor entities.
-# A Fahrenheit climate (dp25/dp26) was tried, but tuya-local does not create
-# the climate entity for it on this device, so Celsius is used for climate.
+# water temperature (dp26).
 entities:
   - entity: climate
     translation_key: pool_heatpump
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/trevi_innovaqua_heatpump.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - entity: climate
+    translation_key: pool_heatpump
+    dps:
+      - id: 1
+        name: hvac_mode
+        type: boolean
+        mapping:
+          - dps_val: false
+            value: "off"
+            icon: "mdi:hvac-off"
```
make-all: Don't override default icons without good reason

custom_components/tuya_local/devices/trevi_innovaqua_heatpump.yaml
```diff
@@ -0,0 +1,97 @@
+name: Pool heat pump
+products:
+  - id: bqfay9re2z3hfhoq
+    manufacturer: Trevi
+    model: InnovAqua Variable Speed
+# Mapping validated on Trevi InnovAqua variable speed over local protocol 3.5.
+# The Celsius setpoint (dp2) and inlet water temp (dp3) drive the climate.
+# The device also exposes an integer Fahrenheit setpoint (dp25) and inlet
+# water temperature (dp26), provided as separate number and sensor entities.
```
make-all: That is not the way to handle parallel C and F controls. They should be merged into the climate entity and used conditionally based on temperature_unit.

See poolex_q7_heatpump.yaml for an example

make-all (changes requested):

PR #5766, opened 2026-08-02
Add Grillfürst Independence 530/530E gas grill

SinedOhtit:
Adds support for the Grillfürst Independence 530 and 530E smart gas grills.
Both models share the same electronics and thing model; they differ only in
the grate material (cast iron vs stainless steel).

The grill uses vendor specific dps only (101-132). DP details were taken from
"Query Things Data Model" on the Tuya IoT platform.

### Cooking modes

Each of the five modes has its own setpoint dp:

| Mode | dp | Range |
|---|---|---|
| D - Direct grilling | 110 | 200-360 °C |
| I - Indirect grilling | 111 | 120-200 °C |
| L - Low & slow | 112 | 95-115 °C |
| S - Step mode | 113 | burner level 1-5 |
| M - Manual grilling | - | no setpoint |

The climate entity exposes dp 110 as the target temperature and redirects it
to dp 111 or 112 depending on the selected preset, with the range adjusted
accordingly. In step and manual mode the target temperature is marked invalid.
Step mode has a burner level rather than a temperature, so it is exposed as a
separate number entity.

### Verified on the device

- Mode switching and target temperature redirection, including the range change
- Target temperature correctly rejected in step and manual modes
- All four timer buttons and the alarm acknowledge button
- Gas bottle scale, lighting and knob light

### Note

`tests/test_device.py::test_api_protocol_version_is_rotated_with_each_failure`
fails on my machine, but it also fails on an unmodified checkout of main, so it
appears unrelated to this change. All other tests pass, as do ruff and yamllint.

### DIFF
--- a/custom_components/tuya_local/devices/grillfuerst_independence530_bbq.yaml
+++ b/custom_components/tuya_local/devices/grillfuerst_independence530_bbq.yaml
@@ -1,0 +1,504 @@
+# Grillfuerst Smart Grill Independence 530 / 530E
+#
+# The 530 and 530E differ only in the grate material (cast iron vs stainless
+# steel); electronics, firmware and Tuya thing model are identical.
+#
+# All datapoints are vendor specific (101-132), the device does not use a
+# standard Tuya instruction set. DP ids and ranges were taken from the thing
+# model in the Tuya IoT platform (abilityId == dp id).
+#
+# The grill has five cooking modes, each with its own setpoint dp:
+#   D  direct grilling   dp 110, 200-360 C
+#   I  indirect grilling dp 111, 120-200 C
+#   L  low & slow        dp 112,  95-115 C
+#   S  step mode         dp 113, level 1-5 (not a temperature)
+#   M  manual grilling   no setpoint, burners are controlled by hand
+# The climate entity redirects the target temperature to the dp matching the
+# currently selected mode, and rejects it in the two modes that have none.
+#
+# Most dps are marked optional so the device still matches when it is powered
+# off and not reporting every value at the time it is added.
+
+name: Gas grill
+products:
+  - id: eo4tfylu1iue4ytt
+    manufacturer: Grillfürst
+    model: Independence 530
+
+entities:
+  # --- Main entity: cook chamber ------------------------------------------
+  - entity: climate
+    dps:
+      - id: 131
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: heat
+
+      - id: 101
+        type: integer
+        name: current_temperature
+        unit: C
+
+      - id: 114
+        type: integer
+        name: preset_mode
+        optional: true
+        mapping:
+          - dps_val: 0
+            value: "D - Direct grilling"
+          - dps_val: 1
+            value: "S - Step mode"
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/grillfuerst_independence530_bbq.yaml
```diff
@@ -0,0 +1,504 @@
...
+          - dps_val: true
+            value: heat
+
+      - id: 101
+        type: integer
+        name: current_temperature
+        unit: C
+
+      - id: 114
+        type: integer
```
make-all, line 47: Is it really an integer? Usually "enum" dps in Tuya protocol are implemented as strings, even if they are carrying numbers.

custom_components/tuya_local/devices/grillfuerst_independence530_bbq.yaml
```diff
@@ -0,0 +1,504 @@
...
+    class: switch
+    name: Cook chamber
+    dps:
+      - id: 129
+        type: boolean
+        name: switch
+        optional: true
+
+  # Projects the brand logo onto the floor in front of the grill.
+  - entity: switch
```
make-all, line 424: light entity might be more appropriate

custom_components/tuya_local/devices/grillfuerst_independence530_bbq.yaml
```diff
@@ -0,0 +1,504 @@
...
+    name: Logo projector
+    dps:
+      - id: 128
+        type: boolean
+        name: switch
+        optional: true
+
+  - entity: switch
+    category: config
+    name: Buzzer
```
make-all, line 435: `translation_key: sound` is intended for this purpose

custom_components/tuya_local/devices/grillfuerst_independence530_bbq.yaml
```diff
@@ -0,0 +1,504 @@
...
+    dps:
+      - id: 130
+        type: boolean
+        name: switch
+        optional: true
+
+  # --- Illuminated control knob -------------------------------------------
+  - entity: select
```
make-all, line 443: This can be a light entity `effect` dp. The "off" value (all lower case) will be used to turn off the light, and marking one of the below as `default: true` will make it the default when someone turns on the light without specifying an effect.

Again, check if this is not really a string, as that would be more usual for a dp with a fixed list of discrete options.

custom_components/tuya_local/devices/grillfuerst_independence530_bbq.yaml
```diff
@@ -0,0 +1,504 @@
...
+          - dps_val: 2
+            value: "Chamber temp"
+          - dps_val: 3
+            value: "Fixed"
+          - dps_val: 4
+            value: "Core temp"
+
+  - entity: select
+    category: config
+    name: Knob light colour
```
make-all, line 465: This can be the `named_color` attribute of the light above. Colors should be mapped to colors from HA's [COLORS](https://github.com/home-assistant/core/blob/dev/homeassistant/util/color.py#L26) list (mostly keeping their original names I think, but there are some quirks, like "green" in HA is a darker green, so "lime" may be a better fit if all the other colors are fully saturated basic colors).

make-all (changes requested):

PR #5775, opened 2026-08-03
fix: Reload config entry after option changes

### LINKED ISSUE #5740: Config entry has already been setup
### Description

Even though I setup router DHCP to give specific IP to the device it decided to give it another one today
The integration obviously stopped working as it expects old IP
I've updated the integration settings for the device with the new IP and now I get this error in logs
I didn't do any other changes, just updated the IP

Image

Image

### Steps to reproduce

_No response_

### Expected behaviour

_No response_

### Additional context

_No response_

### Confirmation

- [ ] If this is for a specific device, the device config filename is mentioned above, or diagnostics are included.

mvanhorn:
## Summary

Changing a configured device's IP through the integration options now reloads the config entry via `hass.config_entries.async_reload(entry.entry_id)` instead of calling the integration's unload and setup functions directly.

## Why this matters

The options update listener called `async_unload_entry` / `async_setup_entry` itself while Home Assistant still tracked the entry as loaded, so the reload logged that the config entry had already been set up and left the entry in an invalid state. Letting Home Assistant own the unload, state transition, and setup sequence keeps its config-entry bookkeeping correct. The same listener handles LAN rediscovery host updates, so that path is fixed too.

## Changes

- `custom_components/tuya_local/__init__.py`: the options update listener delegates to `async_reload`.
- `tests/test_config_flow.py`: regression test asserting the listener requests exactly one reload for the affected entry and no longer drives unload/setup directly.

Fixes #5740

AI was used for assistance.

### DIFF
--- a/custom_components/tuya_local/__init__.py
+++ b/custom_components/tuya_local/__init__.py
@@ -1079,5 +1079,4 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry):
 
 async def async_update_entry(hass: HomeAssistant, entry: ConfigEntry):
     _LOGGER.debug("Updating entry for device: %s", get_device_id(entry.data))
-    await async_unload_entry(hass, entry)
-    await async_setup_entry(hass, entry)
+    hass.config_entries.async_schedule_reload(entry.entry_id)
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thanks for merging the config reload fix, @make-all. Option changes taking effect immediately is a nice quality of life win.

PR #5777, opened 2026-08-03
3x Arlec lights supported (CHR370HA, CHR352HA, ALP018HA)

doctordarko:
Added support for 3 Arlec branded lights:

Smart 5m Neon Flex - CHR370HA
Smart Desk Light - CHR352HA
Smart Sound Sync Mood Lamp - ALP018HA

Unified colour mode and added sound sync mode configurations across 2 devices; dreamlight_led_strip.yaml and dreamlight_rgb_strip.yaml.

### DIFF
--- a/custom_components/tuya_local/devices/dreamlight_led_strip.yaml
+++ b/custom_components/tuya_local/devices/dreamlight_led_strip.yaml
@@ -14,7 +14,6 @@ entities:
     dps:
       - id: 20
         type: boolean
-        optional: true
         name: switch
       - id: 21
         type: string
@@ -115,23 +114,6 @@ entities:
         range:
           min: 1
           max: 1000
-  - entity: select
-    translation_key: mode
-    category: config
-    dps:
-      - id: 21
-        type: string
-        name: option
-        optional: true
-        mapping:
-          - dps_val: "white"
-            value: "White"
-          - dps_val: "colour"
-            value: "Color"
-          - dps_val: "scene"
-            value: "Scene"
-          - dps_val: "music"
-            value: "Music"
   - entity: select
     translation_key: scene
     category: config
--- a/custom_components/tuya_local/devices/dreamlight_led_strip.yaml
+++ b/custom_components/tuya_local/devices/dreamlight_led_strip.yaml
@@ -6,11 +6,15 @@ products:
   - id: rqrkvuxebofiaifp
     manufacturer: Fancy LED
     model: Fancy Lightstrip  3
+  - id: m6abhkeubvtdfjgw
+    manufacturer: Arlec
+    model: Smart Sound Sync Mood Lamp - ALP018HA
 entities:
   - entity: light
     dps:
       - id: 20
         type: boolean
+        optional: true
         name: switch
       - id: 21
         type: string
@@ -72,9 +76,11 @@ entities:
       - id: 46
         name: length_cm
         type: integer
+        optional: true
       - id: 47
         name: pixels
         type: integer
+        optional: true
       - id: 51
         name: dreamlight_scene
         type: string
@@ -110,7 +116,7 @@ entities:
           min: 1
           max: 1000
   - entity: select
-    name: Color mode
+    name: Mode
     icon: "mdi:palette"
     category: config
     dps:
@@ -119,7 +125,9 @@ entities:
         name: option
         optional: true
         mapping:
-          - dps_val: "color"
+          - dps_val: "white"
+            value: "White"
+          - dps_val: "colour"
             value: "Color"
           - dps_val: "scene"
             value: "Scene"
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/dreamlight_led_strip.yaml
```diff
@@ -110,16 +116,17 @@ entities:
           min: 1
           max: 1000
   - entity: select
-    name: Color mode
-    icon: "mdi:palette"
+    translation_key: mode
```
make-all: This change will modify entity ids of existing installations, so would require a migration.
Since the options are not covered by the existing translations, and other translation keys may be more suited, I suggest not to change it at this time.

make-all: Actually this is already covered by the light color_mode, so is not really needed and should probably be removed if anything.

custom_components/tuya_local/devices/dreamlight_rgb_strip.yaml
```diff
@@ -83,6 +90,21 @@ entities:
           - dps_val: null
             value: false
           - value: true
+  - entity: select
```
make-all: This is not needed, as dp 21 is covered in the light entity.

custom_components/tuya_local/devices/dreamlight_led_strip.yaml
```diff
@@ -222,3 +229,61 @@ entities:
             value: "Summer wind"
           - dps_val: "AToCXV3gAABNALReARxkAOhJAMZf"
             value: "Planet journey"
+  - entity: select
```
make-all: Every light has a slightly different set of preset scenes and music modes, so my preference is to just leave this as a text entity and let the user create their own list using helpers.

custom_components/tuya_local/devices/dreamlight_rgb_strip.yaml
```diff
@@ -180,6 +202,64 @@ entities:
             value: "Summer wind"
           - dps_val: "AToCXV3gAABNALReARxkAOhJAMZf"
             value: "Planet journey"
+  - entity: select
```
make-all: as above

custom_components/tuya_local/devices/dreamlight_led_strip.yaml
```diff
@@ -6,11 +6,15 @@ products:
...
     model: Fancy Lightstrip  3
+  - id: m6abhkeubvtdfjgw
+    manufacturer: Arlec
+    model: Smart Sound Sync Mood Lamp - ALP018HA
 entities:
   - entity: light
     dps:
       - id: 20
         type: boolean
+        optional: true
```
make-all: Having non-optional dps helps to prevent false matches when the device is matched by dps. Since the product id is being added here, it will no longer be needed to match this device. I prefer not to mark too many dps as optional for this config.

make-all (changes requested):

custom_components/tuya_local/devices/dreamlight_led_strip.yaml
make-all: Also lacking this dp previously was one of the distinctions between these two configs. Are you sure the device you have added here is not a match for the other config?

custom_components/tuya_local/devices/dreamlight_led_strip.yaml
doctordarko: Makes total sense, I'll remove 'optional: true' for DP 20 and push an update to this PR.

custom_components/tuya_local/devices/dreamlight_led_strip.yaml
doctordarko: Since DP 21 is already handled natively by the light entity's `color_mode`, I will completely remove the redundant DP 21.

custom_components/tuya_local/devices/dreamlight_rgb_strip.yaml
doctordarko: Since DP 21 is already handled natively by the light entity's `color_mode`, I will completely remove the redundant DP 21.

custom_components/tuya_local/devices/dreamlight_led_strip.yaml
doctordarko: I have triple checked the native app options for all 3 devices mentioned, and they have matching Scene and Music modes. The only difference noted was the white channel on `dreamlight_led` yaml.

The fact Scene mode already had dropdown selection as well as text string entity, is the reason why I matched the logic for Music sync mode across both yaml.

custom_components/tuya_local/devices/dreamlight_led_strip.yaml
doctordarko: Fair point, will remove optional tag.

PR #5784, opened 2026-08-04
Add Rhino King smart AC controller

zhuguoqing123:
Adds device configuration for the Rhino King smart AC controller (product ID `y6lcryonnh5klkmo`).

The configuration is in use on the contributor’s Home Assistant instance. Upstream device-config parsing and untranslated-entity checks pass locally.

AI assistance was used to align the submitted YAML with the current schema and run the project tests; the device mapping itself was supplied and tested by the contributor.


custom_components/tuya_local/devices/rhino_king_ac.yaml
```diff
@@ -0,0 +1,119 @@
...
+      - id: 2
+        type: string
+        name: mode
+        hidden: true
+      - id: 2
+        type: string
+        name: preset_mode
+        mapping:
+          - dps_val: wind
+            value: "出风"
```
make-all: This project is used by users from around the world. Please use English for values and entity names if there are not suitable translation keys.

custom_components/tuya_local/devices/rhino_king_ac.yaml
```diff
@@ -0,0 +1,119 @@
...
+          - dps_val: floor_heat
+            value: "地暖"
+          - dps_val: heat_floorheat
+            value: "制热+地暖"
+          - dps_val: cold_floorcold
+            value: "制冷+地冷"
+      - id: 16
+        type: integer
+        name: temperature
+        range: {min: 150, max: 400}
```
make-all: The point of using yaml as a config format is that it can be made more human readable by using multi-line representations over syntax like curly and square braces that need programming knowledge to decipher. Please follow the convention set down by the rest of the configs.

custom_components/tuya_local/devices/rhino_king_ac.yaml
```diff
@@ -0,0 +1,119 @@
...
+            value: "标准"
+          - dps_val: "5"
+            value: "强"
+      # 关键修改：使用 current_humidity 属性名
+      - id: 34
+        type: integer
+        name: current_humidity
+
+  # 独立传感器依然保留，因为这是最稳妥的显示方式
+  - entity: sensor
```
make-all: Duplicate low level entities are not required

make-all (changes requested):

custom_components/tuya_local/devices/rhino_king_ac.yaml
zhuguoqing123: Updated all preset and fan-mode values to English in 790b1e9d.

custom_components/tuya_local/devices/rhino_king_ac.yaml
zhuguoqing123: Expanded the inline range and mapping objects to the project-style multi-line YAML format in 790b1e9d.

custom_components/tuya_local/devices/rhino_king_ac.yaml
zhuguoqing123: Removed the duplicate temperature, humidity, and operating-mode entities; these values remain available through the climate entity in 790b1e9d.

custom_components/tuya_local/devices/rhinoking_ac_thermostat.yaml
```diff
@@ -0,0 +1,81 @@
+name: AC controller
+products:
+  - id: y6lcryonnh5klkmo
+    manufacturer: Rhino King
```
make-all, line 4: It would be useful to have some model info here, as you are submitting 3 similar devices at once here, all of the same brand - users need to know which is which.

custom_components/tuya_local/devices/rhinoking_ac_thermostat.yaml
```diff
@@ -0,0 +1,81 @@
...
+                value: cool
+              - dps_val: hot
+                value: heat
+              - dps_val: wind
+                value: fan_only
+              - dps_val: floor_heat
+                value: heat
+              - dps_val: heat_floorheat
+                value: heat
+              - dps_val: cold_floorcold
```
make-all, line 28: There is no "floor_cold" for floor cooling only?

zhuguoqing123: This device does not support standalone floor cooling. The only floor-cooling mode is `cold_floorcold`, which combines cooling and floor cooling.

custom_components/tuya_local/devices/rhinoking_ac_thermostat.yaml
zhuguoqing123: Added `model: Smart Thermostat` in 7e229282 to distinguish this smart thermostat from the H7 and three-speed variants.

PR #5786, opened 2026-08-04
Add Rhino King three-speed AC controller

zhuguoqing123:
Adds a device configuration for the Rhino King XNW-AC three-speed controller.

The two listed identifiers (`p4ykyk6ljdcoktyq` and `z7acaw0j7l7xg4to`) represent the same controller model through its available Tuya product identifiers. The separate Rhino King smart controller uses a different product ID and is submitted in #5784.

Upstream device-config parsing and untranslated-entity checks pass locally.

AI assistance was used to separate the supplied variant from the smart-controller config, align it with the current schema, and run project tests; the device mapping was supplied by the contributor.


custom_components/tuya_local/devices/rhinoking_3speed_airconditioner.yaml
```diff
@@ -0,0 +1,86 @@
...
+            constraint: mode
+            conditions:
+              - dps_val: cold
+                value: cool
+              - dps_val: hot
+                value: heat
+              - dps_val: wind
+                value: fan_only
+              - dps_val: floor_heat
+                value: heat
```
make-all, line 28: Needs to be `hidden: true` to prevent confusion over which heat mode to select when user presses "heat"

custom_components/tuya_local/devices/rhino_king_three_speed_ac.yaml
```diff
@@ -0,0 +1,86 @@
...
+            value: medium
+          - dps_val: "4"
+            value: medium_high
+          - dps_val: "5"
+            value: high
+      - id: 34
+        type: integer
+        name: current_humidity
+
+  - entity: sensor
```
make-all: Duplicate sensors not needed

custom_components/tuya_local/devices/rhinoking_3speed_airconditioner.yaml
make-all: Because this is inaccessible through the HA climate API, there needs to be another way to select it, such as renaming "mode" to "preset_mode", and make the full list of modes available there, or using the same `hidden: true` trick to restrict the list the user sees to `air` and `floor`, with all other modes mapped to `air` and all hidden except for the "hot" and "floor_heat" mappings.

make-all (changes requested):

custom_components/tuya_local/devices/rhinoking_3speed_airconditioner.yaml
zhuguoqing123: Replaced the raw  attribute with , exposed all four device modes there, and made  constrain against  in 439dbfee.

custom_components/tuya_local/devices/rhino_king_three_speed_ac.yaml
zhuguoqing123: Removed the duplicate temperature and humidity sensor entities in 5ea54436; both values remain on the climate entity.

custom_components/tuya_local/devices/rhinoking_3speed_airconditioner.yaml
zhuguoqing123: Marked the  HVAC mapping as hidden in 5ea54436, so the standard heat control does not expose two ambiguous choices.

PR #5791, opened 2026-08-04
Add FP10351GB-WH portable air conditioner

VulcanB2XH558:
### Adds a new device profile for the FP10351GB-WH portable air conditioner.

Product ID: `asgmtdjvhragxx24`

I am submitting this as a separate profile because its supported functions differ from the existing Costway portable air-conditioner profile.

### Tested functionality

The following functions have been tested locally and are working:

- Power
- Target temperature
- Current temperature
- Cooling
- Heating
- Dry mode
- Fan-only mode
- Low fan speed
- High fan speed
- Sleep mode
- Fault reporting

Sleep mode is exposed as a climate preset using DP103.

 ### Intentionally excluded datapoints

The Tuya product schema advertises several additional datapoints that are not implemented by this hardware variant:

- DP11: anion
- DP15: vertical swing
- DP104: TimerOn
- DP105: TimerOff
- DP110: horizontal swing

These datapoints retain stale/default cloud values and are not reported as live states through the local protocol.

DP104 and DP105 were tested as number entities, but remained `unknown` in Home Assistant and did not accept commands. The timer controls were also not available as functional commands through the Tuya cloud API.

They have therefore been excluded from this profile so users of this product ID do not receive non-functional entities.

### Relationship to existing Costway profile

This device is similar to `costway_portable_ac.yaml`, but differs in the following ways:

- Supports heat mode through DP4 value `hot`
- Supports low and high fan speeds only
- Supports sleep through DP103
- Does not support the advertised ioniser, swing or remote timer datapoints

I have submitted it as a separate profile so those differences do not alter behaviour for existing Costway and Hantech users.

If the maintainers prefer to combine this variant into the existing Costway profile, the required product entry is:

```yaml
  - id: asgmtdjvhragxx24
    model: FP10351GB-WH

also requiring:
              - dps_val: hot
                value: heat

### DIFF
--- a/custom_components/tuya_local/devices/fp10351gb_wh_portable_ac.yaml
+++ b/custom_components/tuya_local/devices/fp10351gb_wh_portable_ac.yaml
@@ -1,0 +1,107 @@
+name: Portable air conditioner
+
+products:
+  - id: asgmtdjvhragxx24
+    model: FP10351GB-WH
+
+entities:
+  - entity: climate
+    dps:
+      # Main power and operating mode
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            constraint: mode
+            conditions:
+              - dps_val: cold
+                value: cool
+              - dps_val: hot
+                value: heat
+              - dps_val: wind
+                value: fan_only
+              - dps_val: wet
+                value: dry
+
+      # Target temperature
+      - id: 2
+        type: integer
+        name: temperature
+        range:
+          min: 16
+          max: 31
+        unit: C
+
+      # Current room temperature
+      - id: 3
+        type: integer
+        name: current_temperature
+        unit: C
+
+      # Raw mode used by the conditional power mapping
+      - id: 4
+        type: string
+        name: mode
+        hidden: true
+
+      # This model supports low and high fan speeds only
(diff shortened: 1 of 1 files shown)


make-all:
What is dp 109 reporting on this device? Previous models marked this as unknown, with a reported value of 0 on Hantech.
Likely this also holds the F/C switch but hopefully also a flag indicating that heat mode is supported.

VulcanB2XH558:
I have not identified DP 109, reporting under funcTag and the two api dumps showed it reporting updates, but the value having not changed. Reported timestamps extremely close to DP20 and so I would agree it also holds the F/C. Reports 8 on this unit.

Dumps:
First capture
DP20  Fault   1785787743928
DP109 funcTag 1785787744335

Second capture
DP20  Fault   1785873079891
DP109 funcTag 1785873080301

{

        "code": "funcTag",

        "custom_name": "",

        "dp_id": 109,

        "time": 1785873080301,

        "type": "bitmap",

        "value": 8

      },

make-all:
8 is consistent with a bitfield indicating one bit difference from the Hantech value of 0, which could be the heat support bit. This would make it possible to merge with the costway config.

make-all:
Also did a search on the model number and came up with forum posts about repairs on a Costway portable air conditioner with that same model code. So I think merging with the Costway should be the right thing here.

PR #5794, opened 2026-08-05
Wd 01 gde irrigation controller

Daniel-Boluda:
Hi team! I'm adding support for the WD-01GDE Dual Irrigation Controller

The basic controls for watering are fully operational, but I'm encountering a couple of edge cases regarding the countdown timer behavior and the schedule DP formatting for Home Assistant user experience.

Device Info & Basic Setup
Product ID: cgz1qpr1aeladx2t
Model: WD-01GDE (Dual Zone / 2-way valve)

Supported Features (Working):
Manual Valve control (switch_1 / switch_2)
Protection Mode (switch_3)
Pump Force selector (power_level_a / power_level_b, values: 1 to 3)
Device logs & Fault indicators (run_log, switch_log_a/b, fault)
Timer irrigation (seting seconds to irrigate before stoping)

Decoded Schedule Binary Format (timer1 / timer2, DPs 101/102)
I have successfully reverse-engineered the 12-byte raw/base64 binary schedule payload:
Byte 0 : Enable switch (0x01 = On, 0x00 = Off)
Byte 1 : Unknown / Entry Index (Always 0x00)
Byte 2 : Day Mask + Repeat flag (Bit 7: Mon ... Bit 1: Sun | Bit 0: Repeat)
Byte 3 : Start Hour (0-23)
Byte 4 : Start Minute (0-59)
Byte 5 : Unknown / Padding (Always 0x00)
Bytes 6-7 : Duration in minutes
Bytes 8-9 : Duration in seconds
Bytes 10-11: Padding (Always 0x00)
Example Decoded Payload:

AQAACAEAAgAAAAAA → 01 00 00 08 01 00 02 00 00 00 00 00 (Enabled, Once, 08:01, 2s)

Current Issues & Need Guidance / Help
Issue A: Countdown Timer (woter_timer1 / woter_timer2, DPs 103/104)
No State Reset: When setting a duration (e.g., 120 seconds), the device opens the valve and automatically stops after 120 seconds. However, DP 103/104 remains at 120 in the device state.
Because the entity state doesn't reset to 0, Home Assistant won't let users trigger another 120-second cycle without setting a different value first.

Value 0 Behavior: Setting woter_timer = 0 makes the device water indefinitely instead of stopping the active cycle or clearing the timer.

Question: How should we handle this in tuya-local? Should woter_timer be exposed strictly as a number entity with a range starting at 1 (to prevent accidental infinite watering), or is there a standard pattern in tuya-local to send a transient duration command that resets after completion?

Issue B: Setting Schedules from Home Assistant
The schedules (timer1/timer2) work fine if configured via raw Base64 string through a text entity using type: base64.

However, this is not user-friendly for end users in HA.

Question: Is there any built-in mechanism or recommended integration pattern in tuya-local to map a 12-byte raw payload into user-editable HA UI components (like separate time, select for days, and number entities)? Or is exposing the raw/base64 text entity as category: config the current limitation for raw-schedule Tuya DPs?

Issue C: Unknown DP Functionality
DP 107 (set_cut_time): Tested setting 1 while watering for 5s; it did not limit or cut off the watering duration. Has anyone seen this DP in other Tuya irrigation controllers?

Any feedback or suggestions on how to best structure these DPs in the YAML profile would be greatly appreciated!


make-all:
It is beyond the scope of this integration to deal with complex packed data structures in a user friendly manner, the integration just provides them as a text entity so another solution can be built on top (see the "meal plan" referenced in the README for an example).

It is also beyond the scope of this integration to provide complex workarounds for device behaviour.

So I think what you have now is OK.

Daniel-Boluda:
> It is beyond the scope of this integration to deal with complex packed data structures in a user friendly manner, the integration just provides them as a text entity so another solution can be built on top (see the "meal plan" referenced in the README for an example).
> 
> It is also beyond the scope of this integration to provide complex workarounds for device behaviour.
> 
> So I think what you have now is OK.

Great! so then i can handle the rest with automations on the hass side, good for me.

Thanks @make-all !

custom_components/tuya_local/devices/wd01gde_irrigation_controller.yaml
```diff
@@ -0,0 +1,160 @@
+name: WD-01GDE Irrigation Controller
```
make-all: Top level name should be generic in case other products match the same config.
HA naming convention for devices and entities require capitalisation only on the first word of a name.

custom_components/tuya_local/devices/wd01gde_irrigation_controller.yaml
```diff
@@ -0,0 +1,160 @@
...
+        unit: s
+  - entity: sensor
+    name: Run log
+    category: diagnostic
+    dps:
+      - id: 105
+        type: string
+        name: sensor
+        optional: true
+  - entity: sensor
```
make-all: Follow the pattern used in 90% of other configs.

binary_sensor with `class: problem`, 0 mapped to false, default (all other values) mapping to true.
Attribute `fault_code` to hold the original raw value. 
Optionally an attribute `description`, but since only OK is mapped here, it should be left out as it doesn't add any value to just map that.

make-all (changes requested):

custom_components/tuya_local/devices/wd01gde_irrigation_controller.yaml
Daniel-Boluda: @make-all  i can rename it to "Irrigation Controller", im not sure if it is matching the config for other products though,  do you preffer this name?

make-all: Yes, that will be fine. Matches are no problem, the user can rename devices after installation if they have multiple.

custom_components/tuya_local/devices/wd01gde_irrigation_controller.yaml
```diff
@@ -0,0 +1,164 @@
...
+  - entity: binary_sensor
+    class: problem
+    category: diagnostic
+    dps:
+      - id: 106
+        type: bitfield
+        name: sensor
+        optional: true
+        mapping:
+          - dps_val: 0
```
make-all, line 84: When the dp is optional, you generally also need to map `dps_val: null` to `value: false` also.

ffais:
The 12-byte `timer1`/`timer2` schedule format described here looks like the same structure that was reverse-engineered for the WD-01ADE dual pump in #5895 — just mislabeled, because the one worked example in this PR happens to be consistent with both readings.

Your example: `AQAACAEAAgAAAAAA` → `01 00 00 08 01 00 02 00 00 00 00 00`, annotated "Enabled, Once, 08:01, 2s".

| idx | byte | this PR's label | #5895's label (WD-01ADE) |
|---|---|---|---|
| 0 | `01` | Enable switch (On) | list header (non-empty) |
| 1 | `00` | Unknown, always 0x00 | mode (0x00 = Custom) |
| 2 | `00` | Day mask + repeat flag | flags (bit0 = enabled, bits7-1 = weekdays Mon..Sun) |
| 3 | `08` | Start hour | Start hour |
| 4 | `01` | Start minute | Start minute |
| 5 | `00` | padding, always 0x00 | duration hi byte |
| 6 | `02` | duration-minutes hi | duration lo byte |
| 7 | `00` | duration-minutes lo | interval days |
| 8 | `00` | duration-seconds hi | interval hours |
| 9 | `00` | duration-seconds lo | padding |
| 10-11 | `00 00` | padding | padding |

Bytes 3-4 (hour/minute) match under both readings. But the annotated "2s" duration only comes out right under the #5895 layout: a 2-byte big-endian **seconds** value at bytes 5-6 (`00 02` = 2). Under this PR's own stated layout, duration lives 4 bytes later (bytes 6-9, minutes then seconds), which can't be verified from this example since that whole region is zero — it just doesn't happen to contradict it either.

Given that, my guess is the WD-01GDE uses the identical structure as the WD-01ADE (header / mode / flags(enabled+weekdays) / hour / minute / duration-as-uint16-BE-seconds / interval-days / interval-hours / padding), and the layout above is an artifact of only having one degenerate test case: a single non-repeating entry, no interval, sub-256s duration — which is consistent with either interpretation by coincidence.

Worth double-checking before locking in the DP mapping, since it also affects the "byte 0 = enable switch" assumption — in the ADE device byte 0 is a list header, and it only reads `0x01` here because there's exactly one enabled entry. A couple more dumps would settle it:
- a duration ≥ 256s (would disambiguate the 2-byte-at-5/6 vs 4-byte-at-6/9 question)
- a disabled entry (to see whether byte 0 or byte 2 bit0 changes)
- a repeating/Interval-mode entry (to see whether byte 1 changes)

Happy to help decode further dumps if you post them.

Shagrat2:
Thanks @ffais for the cross-reference — I am the WD-01ADE owner from #5895. I have dumps that answer all three of your questions, so here they are in case they help lock in the mapping.

All from a **WD-01ADE, product_id `x7uchb3hpkpptpni`, model_id `000004sa7s`, protocol 3.4** — a different revision from both of yours, so treat this as supporting evidence rather than proof for the GDE. Each line was read back from the device and cross-checked against what the Smart Life app displayed.

**1. Duration ≥ 256 s** — settles the 2-byte vs 4-byte question:

```
AQH+DAABAAMAAAAA  →  01 | 01 FE 0C 00 01 00 03 00 00 00 00
                          app showed: 12:00 / 04:16 / 03Day00H
```
`04:16` = 256 s, and it appears as `01 00` in bytes 4-5 of the entry — a 2-byte big-endian **seconds** value. A 4-byte minutes+seconds field cannot produce this: the app would have had to show `01:00`. Also note the app's own label is `min : s`, which is what makes this easy to misread.

**2. Disabled entry** — settles "byte 0 = enable" vs "bit0 of flags = enable". I toggled the switch in the app and re-read; **only one bit moved**:

```
toggle ON   →  entry flags FE → FF     (list header stayed 01)
toggle OFF  →  entry flags FF → FE     (list header stayed 01)
```

**3. Interval vs Custom mode** — settles byte 1 (entry byte 0). Changing only the mode in the app:

```
Custom   → entry byte0 = 00   (app: "Repeat: Mon Wed Sat", interval bytes zero)
Interval → entry byte0 = 01   (app: "Interval duration 00Day01H", weekday bits ignored)
```

**Two more observations that rule out "byte 0 = enable switch" independently:**

- With no schedules at all, the whole dp value on my unit is a **single byte** (`/w==` = `0xFF`). An enable flag for an entry cannot exist when there is no entry — so byte 0 belongs to the list, not to a record.
- With two entries in one channel the value is **23 bytes** = `1 + 11 + 11`. One header, then fixed 11-byte records. I also wrote a deliberately wrong header (`0x05`) with two entries and the app still displayed exactly two — so it is not a count either, just "non-empty".

For the weekday bits, on my unit they are the top 7 bits of the flags byte, `bit7 = Mon … bit1 = Sun`: `0x40` = Tue only, `0xA4` = Mon+Wed+Sat, `0xFE` = all seven.

One caveat worth stating plainly: the empty-list marker differs between revisions — `0xFF` on mine, `0x00` on @ffais's — so the family clearly drifts between firmwares. Whatever you settle on for the GDE is worth confirming with one dump of a ≥256 s duration on that actual device.

PR #6056, opened 2026-09-03
Add LEDVANCE PTZ floodlight camera config

mcdax:
Add PTZ floodlight camera config for LEDVANCE Wifi SMART+ WALL CAMERA CONTROL (product ID derumwjnpzfopvn0, model 4099854474972).

All DPS were verified against a live device using local protocol queries. Entities: floodlight light (DP 138/158), siren (159), motion detection (134), motion event (115), motion sensitivity (106), PTZ controls (116/119), flip image (103), watermark (104), status LED (101), night vision (108), SD status/capacity/format (110/109/111/117), camera record (150), recording mode (151), PIR sensitivity (152), PIR lux threshold (231), PIR light duration (232).

The closest existing config is emos_ip300_camera (86% DPS overlap), but it diverges in PTZ handling, night vision, PIR entities and lacks the LEDVANCE-specific DP set, so a separate file is justified.

The unrelated LEDVANCE SMART+ Floodlight Camera (w8uuhywphjjcwosc) is submitted separately as required.


mcdax:
Follow-up in 06df164: removed DP 231 from the WACAMCON profile as well. Its previous "PIR lux threshold" mapping was not independently validated. The corresponding DP on the Floodlight Camera returns `0` despite daylight and does not persist non-zero local writes, so I did not assume identical semantics across the two variants. It is omitted until the WACAMCON-specific behaviour is directly established.

PR #6059, opened 2026-09-03
Add LEDVANCE floodlight camera config

mcdax:
Add floodlight camera config for LEDVANCE SMART+ Floodlight Camera (product ID w8uuhywphjjcwosc).

All DPS were verified against a live device using local protocol queries (brightness DP 158 confirmed by a controlled write/read-back test). Entities: floodlight light (DP 138/158), sirens (159/120), camera privacy (105), motion detection (134), flip image (103), watermark (104), status LED (101), SD status/capacity/format (110/109/111/117), camera record (150), recording mode (151), PIR lux threshold (231).

The closest existing config is emos_ip300_camera (81% DPS overlap), but it diverges in entity structure (camera/siren entities, different mappings) and lacks privacy mode and the dual siren, so a separate file is justified.

The PTZ variant (LEDVANCE Wifi SMART+ WALL CAMERA CONTROL, derumwjnpzfopvn0) is submitted separately as required.

### DIFF
--- a/custom_components/tuya_local/devices/ledvance_floodlight_camera.yaml
+++ b/custom_components/tuya_local/devices/ledvance_floodlight_camera.yaml
@@ -1,0 +1,170 @@
+name: Floodlight camera
+products:
+  - id: w8uuhywphjjcwosc
+    manufacturer: LEDVANCE
+    model: SMART+ Floodlight Camera
+entities:
+  - entity: light
+    name: Floodlight
+    dps:
+      - id: 138
+        type: boolean
+        name: switch
+      - id: 158
+        type: integer
+        name: brightness
+        optional: true
+        range:
+          min: 1
+          max: 100
+  - entity: switch
+    name: Siren
+    category: config
+    dps:
+      - id: 159
+        type: boolean
+        name: switch
+        optional: true
+        mapping:
+          - dps_val: null
+            value: false
+            hidden: true
+  - entity: switch
+    name: Siren 2
+    category: config
+    dps:
+      - id: 120
+        type: boolean
+        name: switch
+        optional: true
+        mapping:
+          - dps_val: null
+            value: false
+            hidden: true
+  - entity: switch
+    name: Camera privacy
+    category: config
+    dps:
+      - id: 105
+        type: boolean
+        name: switch
+        optional: true
+  - entity: switch
+    translation_key: motion_detection
+    category: config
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/ledvance_floodlight_camera.yaml
```diff
@@ -0,0 +1,170 @@
...
+    dps:
+      - id: 104
+        type: boolean
+        name: switch
+        optional: true
+        mapping:
+          - dps_val: null
+            value: false
+            hidden: true
+  - entity: switch
```
make-all: use a light entity with `translation_key: indicator`

custom_components/tuya_local/devices/ledvance_floodlight_camera.yaml
```diff
@@ -0,0 +1,170 @@
...
+        optional: true
+  - entity: button
+    name: SD format
+    category: config
+    dps:
+      - id: 111
+        type: boolean
+        name: button
+        optional: true
+        persist: false
```
make-all: There is no need to make a button dp persist: false, it has no status.

custom_components/tuya_local/devices/ledvance_floodlight_camera.yaml
```diff
@@ -0,0 +1,170 @@
...
+        type: string
+        name: option
+        optional: true
+        mapping:
+          - dps_val: "1"
+            value: event
+          - dps_val: "2"
+            value: continuous
+  - entity: sensor
+    name: PIR lux threshold
```
make-all: If this is a sensor, use `class: illuminance` instead of this name.
If it is truely a threshold rather than a current measurement, then it should be a number entity so it can be adjusted.

make-all (changes requested):

mcdax:
Follow-up validation in cfceb5f:

- DP 101 is now a config `light` using `translation_key: indicator`.
- Removed redundant `persist: false` entries from the SD-format button DPS.
- Removed DP 231 from the profile. Three fresh reads returned `0` despite daylight; writing `100` only echoed the request and a new connection still read `0`; the upper-bound candidate `500` returned no write echo and also read back as `0`. It is neither a verified adjustable configuration nor a reliable illuminance measurement, so it is omitted until its semantics are independently established.

PR #6064, opened 2026-09-03
add Medion LIFE P901 climater

sreichholf:
Adds the Medion Life P901.
I was using this for a few weeks already (like 6 or so).
It's pretty much a slightly weaker P1002 without swing mode and a different ID.


make-all:
Seems the same as monzana_klimaanlage_air_conditioner, but with additional heat mode (P1002 also is missing heat mode)

Many models based off the same firmware but with different features like this often provide a features bitfield that can be used to determine which features are available. Are there any more unidentified dps sent out by the device?

make-all:
Merged with beltax_aircondition, which already has handling for the heat and cool-only models.

PR #6076, opened 2026-09-05
Trevi InnovAqua: re-add Fahrenheit setpoint and water temperature entities

x409no1:
Follow-up to #5764. The device natively exposes an integer Fahrenheit setpoint (dp25) and inlet water temperature (dp26).

Context: in North America, pool water temperature is universally expressed in **whole degrees Fahrenheit** — never in Celsius with decimals. That is what the heat pump's own display and its Tuya app show, and what pool owners expect to read and set.

The refactored climate reads dp25/dp26 via `value_redirect`, but Home Assistant converts climate values to the system temperature unit. On a Celsius HA install (common in Canada, where the rest of the home is metric but the pool is still Fahrenheit), the native Fahrenheit reading is lost to rounding — e.g. 70 °F → 21 °C → 69.8 °F — so the value shown in HA never matches the device.

This adds dp25/dp26 back as separate `number`/`sensor` entities (config/diagnostic, no `device_class` so HA does not re-convert), alongside the existing climate, which is unchanged. Purely additive (+25 lines).

Tested on my unit with 2026.9.0: climate and both entities work together.

### DIFF
--- a/custom_components/tuya_local/devices/trevi_innovaqua_heatpump.yaml
+++ b/custom_components/tuya_local/devices/trevi_innovaqua_heatpump.yaml
@@ -7,6 +7,10 @@ products:
 # The Celsius setpoint (dp2) and inlet water temp (dp3) drive the climate.
 # The device also exposes an integer Fahrenheit setpoint (dp25) and inlet
 # water temperature (dp26).
+# dp25/dp26 are additionally exposed as separate number/sensor entities: HA
+# converts climate values to the system unit, so on a Celsius install the
+# native Fahrenheit reading (as shown by the device and its app) is otherwise
+# lost to rounding.
 entities:
   - entity: climate
     translation_key: pool_heatpump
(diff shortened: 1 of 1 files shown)


make-all:
This is unnecessary, the redirection to appropriate dps is already handled in the climate entity

x409no1:
Thanks for looking. To clarify — I'm not disputing the redirection: the climate does read dp25/dp26. The issue is downstream: Home Assistant converts all climate temperatures to the *system* unit, and unlike `sensor`/`number` entities, a climate has no per-entity unit override. So on a Celsius HA install the climate shows 70 °F as 21 °C, and reading it back gives 69.8 °F — it can never show the whole-degree Fahrenheit value the device and its app display (the norm for pools in North America). The two extra entities are the only way, within HA, to expose that native value unconverted. Would you be open to them as diagnostic/config entities (or disabled by default), or is there another approach you'd prefer?

make-all:
You should match the device's unit setting to your HA unit setting, then you will not have problems with unit conversion.

Issue #3357, opened 2025-06-10
Request support for Nedis SmartLife Pill Dispenser


theqz:
### Log message

```text
2025-06-10 15:12:33.048 WARNING (MainThread) [custom_components.tuya_local.config_flow] Adding Wi-Fi Pill Dispenser device with product id 8noakuggsvhniojb
2025-06-10 15:12:33.166 WARNING (MainThread) [custom_components.tuya_local.config_flow] Device matches None with quality of 0%. DPS: {"updated_at": 1749561144.5139554, "105": true, "107": true, "123": "1", "124": "0", "126": 30, "127": 24, "129": "1", "130": "3"}
2025-06-10 15:12:33.166 WARNING (MainThread) [custom_components.tuya_local.config_flow] Include the previous log messages with any new device request to https://github.com/make-all/tuya-local/issues/
```

### Product ID

bf03257821f39da477ee8a

### Product Name

Nedis SmartLife Pill Dispenser

### DPS information

```text
{
  "result": {
[...]

PR #6082, opened 2026-09-05
fix(nedis_pilldispenser): correct dp 102 masks, add alarm status sensors

MartinNuc:
**I have this device (Nedis WIFIPD10WT) and tested all of this against it.**

The dp 102 masks are the wrong width. The payload is 30 bytes but the masks are 32, so every field was read from the alarm-status bytes at the *end* of the payload instead of the status bytes at the start — everything decoded as 0 or off.

Real dp 102 payload from my device:

`00002b001c0805030000b302000000000000000000070000000000000000`

| field | before | after |
|---|---|---|
| total compartments | 0 | 28 |
| loaded compartments | 0 | 8 |
| current compartment | 0 | 5 |
| remaining compartments | 0 | 3 |
| battery | 0 | 43 % |
| plug / lid | off | correct |

I also added sensors for the per-alarm status bytes (B21–B29). On my device dp 103 and dp 120 never arrive over the LAN, and are not answered when requested explicitly, so these bytes are the only local way to see whether a dose was taken. Confirmed live: alarm 2 went `waiting` → `missed` → `taken_after_missing` across a real dose.

Finally, the battery binary_sensor: its constraint compares integers (`2`, `[0, 1, 3]`) against `battery_state`, but that dp maps its values to strings first, so nothing ever matched and the entity stayed `unknown`. It now compares the mapped names, and reads `off` with `battery_state: full` on my device.

The config was originally created from the cloud data model in #3357 without hardware to test against, which is presumably why the mask alignment was never caught.

Screenshot 2026-09-05 at 21 26 50
Screenshot 2026-09-05 at 21 26 37
Screenshot 2026-09-05 at 21 26 26
Screenshot 2026-09-05 at 21 26 17

### DIFF
--- a/custom_components/tuya_local/devices/nedis_pilldispenser.yaml
+++ b/custom_components/tuya_local/devices/nedis_pilldispenser.yaml
@@ -24,7 +24,7 @@ entities:
         type: base64
         optional: true
         mask: >-
-          0000000000000000000000000000000000000000000000000000000000000001
+          010000000000000000000000000000000000000000000000000000000000
         name: sensor
   - entity: binary_sensor
     class: battery
@@ -38,15 +38,15 @@ entities:
           - dps_val: null
             constraint: battery_state
             conditions:
-              - dps_val: 2
+              - dps_val: low
                 value: true
-              - dps_val: [0, 1, 3]
+              - dps_val: [normal, full, charging]
                 value: false
       - id: 102
         type: base64
         optional: true
         mask: >-
-          0000000000000000000000000000000000000000000000000000000000000700
+          000700000000000000000000000000000000000000000000000000000000
         name: battery_state
         mapping:
           - dps_val: 0
@@ -67,7 +67,7 @@ entities:
         type: base64
         optional: true
         mask: >-
-          0000000000000000000000000000000000000000000000000000000000000700
+          000700000000000000000000000000000000000000000000000000000000
         name: sensor
         mapping:
           - dps_val: 3
@@ -81,7 +81,7 @@ entities:
         type: base64
         optional: true
         mask: >-
-          0000000000000000000000000000000000000000000000000000000000FF0000
+          0000FF000000000000000000000000000000000000000000000000000000
         name: sensor
         unit: "%"
         class: measurement
@@ -94,7 +94,7 @@ entities:
         type: base64
         optional: true
         mask: >-
-          0000000000000000000000000000000000000000000000000000000001000000
+          000000010000000000000000000000000000000000000000000000000000
         name: sensor
   - entity: sensor
     name: Total compartments
@@ -105,7 +105,7 @@ entities:
         type: base64
         optional: true
         mask: >-
-          000000000000000000000000000000000000000000000000000000FF00000000
+          00000000FF00000000000000000000000000000000000000000000000000
         name: sensor
   - entity: sensor
     name: Loaded compartments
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/nedis_pilldispenser.yaml
```diff
@@ -38,15 +38,15 @@ entities:
           - dps_val: null
             constraint: battery_state
             conditions:
-              - dps_val: 2
+              - dps_val: low
```
make-all, line 41: This change does not seem correct. The battery_state is a 3 bits extracted from dp 102, but you are changing the comparison values from integers to strings.

make-all (changes requested):
Sorry for the delay, it seems this accidentally got moved to the stalled queue

custom_components/tuya_local/devices/nedis_pilldispenser.yaml
MartinNuc: Strings are needed with current code. For base64/hex constraints [`_active_condition`](https://github.com/make-all/tuya-local/blob/8dbf671139fc4f7068e32201440e009acc93cc3c/custom_components/tuya_local/helpers/device_config.py#L933-L953) uses `c_dps.get_value()` ([L944](https://github.com/make-all/tuya-local/blob/8dbf671139fc4f7068e32201440e009acc93cc3c/custom_components/tuya_local/helpers/device_config.py#L944)), which applies the mask and then the mapping ([L504](https://github.com/make-all/tuya-local/blob/8dbf671139fc4f7068e32201440e009acc93cc3c/custom_components/tuya_local/helpers/device_config.py#L504)). So `battery_state` comes in as `"low"`, not `2`, and `"low" == 2` never matches. Other dp types use raw `get_property()`, which is why ints work everywhere else.

Ran `device_config.py` locally with my payload:
- PR: low -> `on`, normal/full/charging -> `off`
- PR masks + int conditions: always `unknown`

On my device (running this PR since 12 Sept) all 22 dp 102 updates show `off` with `battery_state: full`, plug on, battery 100%. dp 104 never comes, so it's the `battery_state` fallback doing this.

Same issue in [`inkbird_ibs_m1s.yaml`](https://github.com/make-all/tuya-local/blob/8dbf671139fc4f7068e32201440e009acc93cc3c/custom_components/tuya_local/devices/inkbird_ibs_m1s.yaml#L43-L66): `dps_val: 65535` is compared to the scaled value (6553.5), so a missing probe shows 6553.5% and stays available. If you'd rather change `_active_condition` to compare the raw masked value, that fixes inkbird too and I'll switch back to ints.

Fixed the description: real cause is byte order, not mask width. Old masks + `endianness: little` would give the same result as this PR. Can switch to that if you prefer the smaller diff.

dps 121, 128, 132, 135 seem to have the same problem but I can't check them on my device, so I left them out.

MartinNuc: Switched to `endianness: little` with the original masks, alarm masks done the same way. Same output as before: checked on the device and in a local harness (identical on ~9k random payloads). Updated the description.

make-all:
Sorry, you are right. In order to get the mask taken into account, hex and base64 have a special case to use the resolved HA value instead of the raw dps_val in condition comparisons. I reviewed the usage of conditions against hex/base64 dps:

- cecotec_congaz100_vacuum: only compares with null, so is not affected
- nedis_pilldispenser: this PR
- owon_pc341_energymeter: the config this special case was originally added for, so is implemented correctly
- inkbird_ibs_m1s: as you describe, needs to have the scale applied to condition dps_val

PR #6083, opened 2026-09-05
Add config for Shrivee 4-valve (quad) RF water timer

joerg501:
Hi,

I'm not a developer and I don't speak English well, so this text was written with the help of an AI assistant (Claude).

I own a 4-valve variant of the Shrivee RF water timer - the same device family as the existing `shrivee_triple_rf_water_timer.yaml`, just with 4 valves instead of 3. I previously opened an issue asking if a config for the 4-valve version could be added, and understood that you don't have the time to build one yourself from scratch.

As an absolute beginner, I worked through this together with Claude, using `shrivee_triple_rf_water_timer.yaml` as a reference and the actual confirmed data points from my own device's diagnostics. The attached config has been running reliably on my own device.

These 4-valve units currently seem to be sold quite widely, so there is likely demand from other users with the same device as well.

Since this file currently lives outside the official project, it gets deleted every time Tuya Local updates, and I have to manually restore it. It would be great if this could be included officially so that stops happening - and so other owners of the same 4-valve device could benefit too.

Please feel free to rename it or adjust any details as needed - I understand the config may need changes before it can be merged. I don't have the exact product_id for this device, so I left it as a placeholder.

Thank you for maintaining this project!

### DIFF
--- a/custom_components/tuya_local/devices/shrivee_quad_rf_water_timer.yaml
+++ b/custom_components/tuya_local/devices/shrivee_quad_rf_water_timer.yaml
@@ -1,6 +1,6 @@
 name: Bewaesserungscomputer (4 Zonen)
 products:
-  - id: PLATZHALTER_PRODUKT_ID
+  - id: TODO_PRODUCT_ID
     manufacturer: Shrivee
     model: Bewaesserung Nebenanschluss (4 Ventile)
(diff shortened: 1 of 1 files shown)


make-all:
This config is not in a suitable state for contribution

PR #6088, opened 2026-09-06
Add device config for Arlec/Deta EV200D-HA EV charger

### LINKED ISSUE #6087: Arlec/Deta EV200D-HA EV charger (22kW/32A) - currently matches tera_w01_evcharger but is missing several DPs and has a wrong current range
### Log message

```text
This device already matches tera_w01_evcharger.yaml well enough that it was usable, so the config flow never showed the "no match" warning with the DPS dump. Instead here are two real local DPS captures from custom_components.tuya_local debug logging (not the config-flow warning, but the same underlying data):

Idle, plugged in, current limit set to 6A (charger_end):
{"1": 238433, "3": "charger_end", "4": 6, "5": 0, "9": 0, "10": 0, "13": "controlpi_9v", "14": "charge_now", "16": false, "18": false, "22": "3091002509060039", "23": "H0_ARLEC-V1.0.6", "24": 35, "25": 1830, "101": false, "102": 10, "103": "Disable", "104": 6, "105": 60}

Actively charging single-phase at 6A:
{"1": 238435, "3": "charger_charging", "4": 6, "5": 1397, "9": 1397, "10": 0, "13": "controlpi_6v_pwm", "14": "charge_now", "16": false, "18": true, "22": "3091002509060039", "23": "H0_ARLEC-V1.0.6", "24": 36, "25": 1, "101": false, "102": 10, "103": "Disable", "104": 6, "105": 60}

DPs 6/7/8 (the phase_a/b/c raw voltage+current+power blobs) are pushed on a slower, separate cadence than the rest and were captured separately, e.g. phase_a = "CT4AFwwAAdE=" (decodes cleanly as voltage/current/power using the exact same mask layout as tera_w01_evcharger's L1/L2/L3 entries).
```

### Product ID

svufayckw5hozqbz

### Product Name

Deta EV200D-HA (7/22kW wall-mount Grid Connect smart EV charger, 5m cable). Firmware self-identifies as "H0_ARLEC-V1.0.6" - looks to be the same hardware rebadged as Arlec in the Australian Grid Connect range.

### DPS information

```text
Not the dedicated Query Things Data Model function (no iot.tuya.com account), but the device owner pulled the cloud "thing" info and current properties directly:

GET https://openapi.tuyaeu.com/v2.0/cloud/thing/bfd5b45e6c1199b68dmnok
{
  "result": {
    "category": "qccdz",
    "custom_name": "EV Charger",
    "model": "EV200DHA",
    "name": "EV200D-HA",
    "product_id": "svufayckw5hozqbz",
    "product_name": "EV200D-HA"
  }
}

Properties (current values, from the cloud "properties" endpoint - so types are Tuya's own but ranges below are inferred, not authoritative):
| dp_id | code | type | value (sample) |
|---|---|---|---|
| 1 | forward_energy_total | value | 77428 |
| 3 | work_state | enum | charger_charging |
| 4 | charge_cur_set | value | 32 (device also accepts and echoes back 6 locally - see below) |
| 5 | sigle_phase_power [sic, Tuya's own typo] | value | 6782 |
| 6 | phase_a | raw | CK0Ad0wACNQ= |
| 7 | phase_b | raw | CSUAAAAACPM= |
| 8 | phase_c | raw | CUYAAAAACPM= |
| 9 | power_total | value | 6873 |
| 10 | fault | bitmap | 0 |
| 11 | alarm_set_1 | raw | AQIAIAMCCeIEAggWBQL//wcC//8= |
| 12 | alarm_set_2 | raw | AQIACQIC//8DAv//CAL//w== |
| 13 | connection_state | enum | controlpi_6v_pwm |
| 14 | work_mode | enum | charge_now |
| 15 | balance_energy | value | 200 |
| 16 | clear_energy | bool | true |
| 17 | energy_charge | value | 200 |
| 18 | switch | bool | true |
| 19 | local_timer | raw | AQA= |
| 22 | meter_id | string | 3091002509060039 |
| 23 | system_version | string | H0_ARLEC-V1.0.6 |
| 24 | temp_current | value | 54 |
| 25 | charge_energy_once | value | 780 |
| 27 | online_state | enum | online |
| 33 | mode_set | raw | AQABAQAAAAA= |
| 101 | dlm_switch | bool | false |
| 102 | home_max_current | value | 10 |
| 103 | solar_mode | enum | Disable |
| 104 | solar_current_from_grid | value | 6 |
| 105 | solar_stable_time | value | 60 |

Notes on what's confirmed locally vs still guesswork:
- DPs 1,3,4,6,7,8,9,10,13,14,18,23,24 are already covered by tera_w01_evcharger.yaml and confirmed working locally (identical dp ids/format to Tera's own device).
- DP 4's range in tera_w01_evcharger.yaml is min 8, max 32. This hardware accepts and echoes back 6A locally (confirmed by setting it and reading it back), and 6A is the IEC 61851/J1772 minimum, so 8A looks like it may be Tera-specific rather than universal to this dp.
- DPs 5, 22, 25, 101, 102, 104, 105 were confirmed present and sane locally (see the two captures above). 103 (solar_mode) was seen locally too but only ever as "Disable" - full enum set unconfirmed.
- DPs 11, 12, 15, 16, 17, 19, 27, 33 come only from the cloud properties dump above; not yet confirmed to be readable over the local protocol (tera_w01_evcharger.yaml doesn't declare them, so they've never been requested locally).
- DP 28 (timer_on / "Delay charge" in tera_w01_evcharger.yaml) does NOT appear in this product's DP list at all - it looks like a Tera-only feature.

Would be happy to run a local capture with 11/12/15/17/19/27/33 explicitly requested if that's useful for confirming which are actually exposed locally, and/or to get proper QueryThingsDataModel ranges if someone can point me to a way to do that without iot.tuya.com access (Tuya's trial access to that function is apparently now capped at one month, refreshable every 6 months).
```

Australian market, IEC 61851/J1772 Type 2, 22kW 3-phase (32A per phase), sold both as "Arlec" (Bunnings' house brand) and "Deta" (also an AU electrical brand) - the firmware itself self-identifies as ARLEC regardless of which retail name it's sold under, so these are very likely identical hardware.

I've built and tested a full local device config against the real hardware (all 19 dps tera_w01_evcharger.yaml already handles, plus the additional ones listed above, decoded and sane, current min lowered to 6A and confirmed). I know from the note above that PRs are currently paused, so I'm not opening one - happy to share the yaml here or via a gist if it's useful reference material once someone has bandwidth, otherwise I'm also happy to just leave this as a device report for whenever it's picked up.

Product page: https://www.arlec.com.au/collections/ev-charging (no exact model number listed publicly that I could find matching EV200D-HA - it may be discontinued/OEM-only under that code).

kbinish-ship-it:
Adds a dedicated device profile for the Arlec/Deta EV200D-HA EV charger (22kW/32A, 3-phase), which is currently unsupported and incorrectly matches `tera_w01_evcharger`: wrong min current (this hardware supports 6A, not 8A) and missing several DPs the Tera profile doesn't define.

Related issue: #6087, filed with the full cloud DP dump, local DPS captures, and notes on what's confirmed locally vs still guesswork.

This profile matches on product_id `svufayckw5hozqbz` (the device's own cloud "thing" dump; firmware self-identifies as `H0_ARLEC-V1.0.6`), corrects the charge current range to 6-32A, keeps all the Tera-profile entities this hardware also supports, and adds entities for the DPs this device exposes that Tera's profile doesn't: single-phase power (DP5), session energy (DP25), meter ID (DP22), dynamic load management switch (DP101), home max current (DP102), solar mode (DP103, read-only/diagnostic since the full enum isn't confirmed), solar current from grid (DP104), and solar stable time (DP105). DPs I couldn't confirm safely from local capture alone (11, 12, 15, 16, 17, 19, 27, 33) are deliberately left unmapped rather than guessed.

Testing: validated with the integration's own matching code (TuyaDeviceConfig.matches() / possible_matches()), where this profile scores 101 against the real device's DPS vs 59 for tera_w01_evcharger (today's fallback match). Also ran it live against the physical device: switched a running Home Assistant instance onto this profile, confirmed all entities (existing and new) report correct values, and confirmed the corrected 6A minimum actually works on the hardware.

I'm aware PRs are currently paused. Opening this anyway at the device owner's request, since the fix is tested and the corresponding issue is already filed. Happy to wait for review whenever bandwidth allows, and to make any changes requested.

https://claude.ai/code/session_01X3P3d2rdYSTW9tXCYvDTUa


custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+        range:
+          min: 0
+          max: 100
+  - entity: sensor
+    name: Solar mode
+    category: diagnostic
+    icon: "mdi:solar-power"
+    dps:
+      - id: 103  # solar_mode - enum, but full option set is unconfirmed, so
+                 # exposed read-only rather than guessing at a select mapping
```
make-all, line 316: The full list of options should be listed in the Query Things Data Model API result

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+      - id: 1  # forward_energy_total
+        type: integer
+        optional: true
+        name: sensor
+        class: total_increasing
+        unit: kWh
+        mapping:
+          - scale: 100
+  - entity: number
+    name: Max current
```
make-all, line 47: Configs for similar devices are using "Charging current". In general, abbreviations in names should be avoided.

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+    dps:
+      - id: 4  # charge_cur_set
+        type: integer
+        name: value
+        unit: A
+        range:
+          min: 6
+          max: 32
+  - entity: sensor
+    name: L1 voltage
```
make-all, line 60: Use the following instead of the explicit name
```
translation_key: voltage_x
translation_placeholders:
  x: L1
```

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+        type: base64
+        name: sensor
+        class: measurement
+        optional: true
+        unit: V
+        mask: "FFFF000000000000"
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: L1 current
```
make-all, line 73: As above, but `translation_key: current_x`

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+        type: base64
+        name: sensor
+        optional: true
+        class: measurement
+        unit: A
+        mask: "0000FFFFFF000000"
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: L1 power
```
make-all, line 86: As above, but `translation_key: power_x`

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+        type: base64
+        name: sensor
+        optional: true
+        unit: kW
+        class: measurement
+        mask: "0000000000FFFFFF"
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: L2 voltage
```
make-all, line 99: As above but `x: L2`

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+        type: base64
+        name: sensor
+        optional: true
+        class: measurement
+        unit: V
+        mask: "FFFF000000000000"
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: L2 current
```
make-all, line 112: See above

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+        type: base64
+        name: sensor
+        optional: true
+        unit: A
+        class: measurement
+        mask: "0000FFFFFF000000"
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: L2 power
```
make-all, line 125: See above

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+        type: base64
+        name: sensor
+        optional: true
+        class: measurement
+        unit: kW
+        mask: "0000000000FFFFFF"
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: L3 voltage
```
make-all, line 138: See above

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+        type: base64
+        name: sensor
+        optional: true
+        unit: V
+        class: measurement
+        mask: "FFFF000000000000"
+        mapping:
+          - scale: 10
+  - entity: sensor
+    name: L3 current
```
make-all, line 151: See above

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+        type: base64
+        name: sensor
+        optional: true
+        unit: A
+        class: measurement
+        mask: "0000FFFFFF000000"
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: L3 power
```
make-all, line 164: See above

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+    dps:
+      - id: 9  # power_total
+        type: integer
+        name: sensor
+        unit: kW
+        class: measurement
+        mapping:
+          - scale: 1000
+  - entity: sensor
+    name: Single phase power
```
make-all, line 187: What is this measuring? Probably it can be an attribute of the power sensor above, as it seems it will not be useful alongside the total power above and the L1 power further up.

custom_components/tuya_local/devices/arlec_ev200dha_evcharger.yaml
```diff
@@ -0,0 +1,347 @@
...
+    name: Session energy
+    class: energy
+    dps:
+      - id: 25  # charge_energy_once
+        type: integer
+        optional: true
+        name: sensor
+        unit: Wh
+  - entity: sensor
+    name: Meter ID
```
make-all, line 280: This is static info, so should just be an attribute on one of the other entities.

make-all (changes requested):

kbinish-ship-it:
Pushed 3 follow-up commits addressing all review feedback:

- Renamed "Max current" → "Charging current" to match the convention used by similar devices
- Switched the L1/L2/L3 voltage/current/power sensors to `translation_key: voltage_x/current_x/power_x` with `x: L1/L2/L3` instead of explicit names
- Folded "Meter ID" (dp22) into the status sensor as an attribute, alongside `system_version`
- Folded "Single phase power" (dp5) into the total power sensor as an attribute — confirmed from the local captures in #6087 that it duplicates `power_total` whenever charging single-phase (and reads 0 otherwise), so it isn't useful as its own entity
- solar_mode (dp103): the device owner pulled the Query Things Data Model API result directly — it's a genuine enum `["Disable", "ECO"]`, so this is now a proper `select` entity instead of a read-only sensor
- While in there, corrected a few DP ranges against the same Data Model response, which turned out wider than my original locally-inferred guesses: `home_max_current` (dp102) is 0-600A not 0-100A, `solar_current_from_grid` (dp104) is 6-32A not 0-32A, and `solar_stable_time` (dp105) is 10-3600s not 0-600s

(One of the intermediate commits accidentally duplicated the file content due to a tooling issue on my end — fixed in the commit right after, sorry for the noise in the history.)

Let me know if anything else needs adjusting.

kbinish-ship-it:
Thanks for the comment formatting fixes — I'll pull those into my branch before making any further changes. As far as I can tell everything from the review is addressed now, so nothing outstanding on my side. Happy to adjust anything else if needed.

PR #6089, opened 2026-09-06
Add Goldair GCT315/GCT425 2000W ceramic tower heater

### LINKED ISSUE #6038: Request support for Goldair 2000W Ceramic Tower Heater
### Log message

```text
Logger: custom_components.tuya_local.config_flow
Source: custom_components/tuya_local/config_flow.py:542
Integration: Tuya Local

Device matches goldair_gct315_heater with quality of 100%. LOCAL DPS: {"updated_at": 1788254797.6257222, "1": true, "2": 15, "3": 24, "4": "normal", "5": "level_1", "8": true, "19": "cancel"}

A second observation taken earlier with the heater off, via a local query:
{"1": false, "2": 15, "3": 20, "4": "auto", "5": "level_0", "8": false, "19": "cancel"}
```

### Product ID

8eapmvhb79pcxx7x

### Product Name

Goldair 2000W Ceramic Tower Heater (GCT315 / GCT425)

### DPS information

```text
{
  "result": {
    "category": "qn",
    "functions": [
      {
        "code": "switch",
        "desc": "{}",
        "name": "\u5f00\u5173",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "desc": "{\"unit\":\"\u2103\",\"min\":15,\"max\":45,\"scale\":0,\"step\":1}",
        "name": "\u8bbe\u7f6e\u6e29\u5ea6",
        "type": "Integer",
        "values": "{\"unit\":\"\u2103\",\"min\":15,\"max\":45,\"scale\":0,\"step\":1}"
      },
      {
        "code": "mode",
        "desc": "{\"range\":[\"auto\"]}",
        "name": "\u6a21\u5f0f",
        "type": "Enum",
        "values": "{\"range\":[\"auto\"]}"
      },
      {
        "code": "level",
        "desc": "{\"range\":[\"1\",\"2\"]}",
        "name": "\u6863\u4f4d",
        "type": "Enum",
        "values": "{\"range\":[\"1\",\"2\"]}"
      },
      {
        "code": "shake",
        "desc": "{}",
        "name": "\u6447\u5934",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "countdown_set",
        "desc": "{\"range\":[\"cancel\",\"1h\",\"2h\",\"3h\",\"4h\",\"5h\",\"6h\"]}",
        "name": "\u5012\u8ba1\u65f6\u8bbe\u7f6e",
        "type": "Enum",
        "values": "{\"range\":[\"cancel\",\"1h\",\"2h\",\"3h\",\"4h\",\"5h\",\"6h\"]}"
      }
    ],
    "status": [
      {
        "code": "switch",
        "name": "\u5f00\u5173",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "temp_set",
        "name": "\u76ee\u6807\u6e29\u5ea6C",
        "type": "Integer",
        "values": "{\"unit\":\"\u2103\",\"min\":15,\"max\":45,\"scale\":0,\"step\":1}"
      },
      {
        "code": "temp_current",
        "name": "\u5f53\u524d\u6e29\u5ea6C",
        "type": "Integer",
        "values": "{\"unit\":\"\u2103\",\"min\":0,\"max\":70,\"scale\":0,\"step\":1}"
      },
      {
        "code": "mode",
        "name": "\u6a21\u5f0f",
        "type": "Enum",
        "values": "{\"range\":[\"auto\"]}"
      },
      {
        "code": "level",
        "name": "\u6863\u4f4d",
        "type": "Enum",
        "values": "{\"range\":[\"1\",\"2\"]}"
      },
      {
        "code": "shake",
        "name": "\u6447\u5934",
        "type": "Boolean",
        "values": "{}"
      },
      {
        "code": "countdown_set",
        "name": "\u5012\u8ba1\u65f6\u8bbe\u7f6e",
        "type": "Enum",
        "values": "{\"range\":[\"cancel\",\"1h\",\"2h\",\"3h\",\"4h\",\"5h\",\"6h\"]}"
      }
    ]
  },
  "success": true,
  "t": 1788254240118,
  "tid": "eefbad89a5e511f1979312b1f8d51a63"
}
```

A 2000W ceramic tower heater with an oscillating base, sold in New Zealand and Australia. The product covers two model numbers, GCT315 and GCT425. The device speaks protocol 3.5.

```
1   boolean  power
2   integer  set temperature, 15-45 C, step 1
3   integer  current temperature
4   enum     mode
5   enum     heat level
8   boolean  oscillation
19  enum     countdown timer
```

The part that is not obvious from the data model is the relationship between dp 4 and dp 5. dp 4 selects *who* chooses the heat level: in `normal` the level on dp 5 is the user's choice, in `auto` the appliance picks it itself and dp 5 reflects whatever it has settled on rather than anything the user asked for. The front panel and remote reach auto by cycling past the highest level, so from the user's point of view it reads as a fourth position in the same list rather than a separate setting.

Setting auto has to write dp 4 on its own. Sending a level in the same message puts the appliance straight back to normal, so auto never takes effect. Setting a level does the reverse and is safe to write together with dp 4.

Two places where the data model above disagrees with the hardware, in case it matters when interpreting it. Both are observed values, not inference:

- `level` is documented with range `["1","2"]`. The device reports `level_0` and `level_1` in the two observations above. `level_0`, `level_1` and `level_2` correspond to the fan-only, low and high positions on the panel.
- `mode` is documented with the single value `auto`. The device reports both `auto` and `normal` in the observations above, `normal` being the value it holds whenever a level has been selected directly.

I have a config written and tested against the device, currently running in my own HA. I have deliberately not attached it, since I gather a submitted config is not the format that helps you most at the moment. Happy to send it, open a PR if the pause lifts, or just leave the data above with you, whichever is least work for you.

Manual: https://www.goldair.co.nz/

leashjay:
Closes #6038.

Adds support for the Goldair 2000W ceramic tower heater (GCT315 and GCT425), product `8eapmvhb79pcxx7x`. One new device config, nothing else changed.

Notes:

- dps read from my own heater over the LAN on protocol 3.5: `{"1": false, "2": 15, "3": 20, "4": "auto", "5": "level_0", "8": false, "19": "cancel"}`
- dp 4 chooses whether the heat level in dp 5 is set by the user or by the heater, so Auto sits on dp 4 with dp 5 as its constraint rather than being a value of dp 5. Auto writes dp 4 alone, a named level writes both. Sending a level together with Auto puts the heater back to normal.
- dp 5 reports `level_0` and dp 4 accepts `normal`, neither of which is in the Tuya data model for this product, so the values used here come from the device.
- `duplicates` reports no 100% match.
- yamllint, tests and ruff pass locally.

### DIFF
--- a/custom_components/tuya_local/devices/goldair_gct315_heater.yaml
+++ b/custom_components/tuya_local/devices/goldair_gct315_heater.yaml
@@ -1,0 +1,80 @@
+name: Heater
+products:
+  - id: 8eapmvhb79pcxx7x
+    manufacturer: Goldair
+    model: GCT315/GCT425
+    name: 2000W ceramic tower heater
+entities:
+  - entity: climate
+    translation_only_key: heater
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: "heat"
+      - id: 2
+        type: integer
+        name: temperature
+        unit: C
+        range:
+          min: 15
+          max: 45
+      - id: 3
+        type: integer
+        name: current_temperature
+      - id: 4
+        type: string
+        name: preset_mode
+        # dp 4 selects whether the heat level on dp 5 is chosen by the user or
+        # by the appliance. Auto is therefore not a value of dp 5, and setting
+        # it must write dp 4 alone - sending a level in the same message drops
+        # the appliance straight back to normal.
+        mapping:
+          - dps_val: auto
+            value: auto
+          - dps_val: normal
+            constraint: heat_level
+            conditions:
+              - dps_val: level_0
+                value: fan_only
+              - dps_val: level_1
+                value: low
+              - dps_val: level_2
+                value: high
+      - id: 5
+        type: string
+        name: heat_level
+      - id: 8
+        type: boolean
+        name: swing_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: horizontal
+  - entity: select
+    translation_key: timer
+    category: config
+    dps:
+      - id: 19
+        type: string
+        name: option
+        mapping:
+          - dps_val: cancel
+            value: cancel
+          - dps_val: "1h"
+            value: "1h"
+          - dps_val: "2h"
+            value: "2h"
+          - dps_val: "3h"
+            value: "3h"
+          - dps_val: "4h"
+            value: "4h"
+          - dps_val: "5h"
+            value: "5h"
+          - dps_val: "6h"
+            value: "6h"


custom_components/tuya_local/devices/goldair_gct315_heater.yaml
```diff
@@ -0,0 +1,80 @@
+name: Heater
+products:
+  - id: 8eapmvhb79pcxx7x
+    manufacturer: Goldair
+    model: GCT315/GCT425
+    name: 2000W ceramic tower heater
```
make-all, line 6: name has no effect here

custom_components/tuya_local/devices/goldair_gct315_heater.yaml
```diff
@@ -0,0 +1,80 @@
...
+        type: string
+        name: heat_level
+      - id: 8
+        type: boolean
+        name: swing_mode
+        mapping:
+          - dps_val: false
+            value: "off"
+          - dps_val: true
+            value: horizontal
```
make-all, line 58: Use "on" is better, older configs use "horizontal" or "vertical" because "on" was not an available standard option until the swing mode was split, but quite likely HA will remove the translations for that one day.

make-all (changes requested):

PR #6090, opened 2026-09-06
Add support for Madimack GT Freedom i80 pool cleaner

petermcg151:
Adds Tuya Local support for the Madimack GT Freedom i80 robotic pool cleaner.

Product ID: 7up9kxk6vx0e7pcm

The configuration has been tested locally with Tuya Local on the physical device.

### DIFF
--- a/custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
+++ b/custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
@@ -2,12 +2,13 @@ name: Pool cleaning robot
 
 products:
   - id: 7up9kxk6vx0e7pcm
-    manufacturer: Madimack
+  - manufacturer: Madimack
     model: GT Freedom i80
 
 entities:
   - entity: number
     name: Cleaning duration
+    class: duration
     category: config
     dps:
       - id: 103
@@ -27,15 +28,24 @@ entities:
         name: switch
 
   - entity: sensor
-    name: Battery level
+    class: battery
     category: diagnostic
     dps:
       - id: 107
-        type: string
+        type: integer
         name: sensor
+        unit: "%"
+        mapping:
+          - dps_val: 0
+            value: 33
+          - dps_val: 1
+            value: 67
+          - dps_val: 2
+            value: 100
 
   - entity: number
     name: Repeat interval
+    class: duration
     category: config
     dps:
       - id: 108
--- a/custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
+++ b/custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
@@ -1,0 +1,47 @@
+name: Pool cleaning robot
+
+products:
+  - id: 7up9kxk6vx0e7pcm
+    manufacturer: Madimack
+    model: GT Freedom i80
+
+entities:
+  - entity: number
+    name: Cleaning duration
+    category: config
+    dps:
+      - id: 103
+        type: integer
+        name: value
+        unit: min
+        range:
+          min: 30
+          max: 480
+
+  - entity: switch
+    name: Wall cleaning
+    category: config
+    dps:
+      - id: 104
+        type: boolean
+        name: switch
+
+  - entity: sensor
+    name: Battery level
+    category: diagnostic
+    dps:
+      - id: 107
+        type: string
+        name: sensor
+
+  - entity: number
+    name: Repeat interval
+    category: config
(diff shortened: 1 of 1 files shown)


custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - id: 104
+        type: boolean
+        name: switch
+
+  - entity: sensor
+    name: Battery level
+    category: diagnostic
+    dps:
+      - id: 107
+        type: string
```
make-all: This should be made into a standard class: battery sensor with range 0-100, as HA expects for battery sensors.

custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+products:
+  - id: 7up9kxk6vx0e7pcm
+    manufacturer: Madimack
+    model: GT Freedom i80
+
+entities:
+  - entity: number
+    name: Cleaning duration
+    category: config
```
make-all, line 12: class: duration

custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
```diff
@@ -0,0 +1,47 @@
...
+    name: Battery level
+    category: diagnostic
+    dps:
+      - id: 107
+        type: string
+        name: sensor
+
+  - entity: number
+    name: Repeat interval
+    category: config
```
make-all, line 49: class: duration

make-all (changes requested):

custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
petermcg151: Updated as requested and tested successfully on the physical device. DP 107’s three battery levels are mapped to 33%, 67%, and 100%.

custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
petermcg151: Updated

custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
petermcg151: Updated

petermcg151:
Updated as requested and tested successfully on the physical device in Home Assistant. DP 107 reports three battery levels, which are now mapped to 33%, 67%, and 100%.

custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
```diff
@@ -2,12 +2,13 @@ name: Pool cleaning robot
 
 products:
   - id: 7up9kxk6vx0e7pcm
-    manufacturer: Madimack
+  - manufacturer: Madimack
```
make-all: This new change is incorrect, manufacturer and model should be part of the same dict entry as the id.

custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
```diff
@@ -27,15 +28,24 @@ entities:
         name: switch
 
   - entity: sensor
-    name: Battery level
+    class: battery
     category: diagnostic
     dps:
       - id: 107
-        type: string
+        type: integer
```
make-all: If the values coming from the device are strings, then this should be a string, and the dps_val in the mapping should be quoted to ensure they are matched to strings.

custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
petermcg151: Typo mistake sorry. Fixed.

custom_components/tuya_local/devices/madimack_gtfreedomi80_poolcleaner.yaml
petermcg151: Updated DP 107 to use the string type with quoted source values. The mapped outputs remain numeric percentages for the standard battery class. Tested successfully in Home Assistant.

PR #6092, opened 2026-09-06
Add support for The Unclog AC condensate line cleaner (ir6qiyn5s0pbnbz7)

funkspoc:
Adds a device configuration for "The Unclog" — a Tuya-based AC condensate drain-line
cleaner (product_id ir6qiyn5s0pbnbz7, category cz, protocol 3.5).

The official Tuya cloud integration only surfaces the on/off switch (switch_1) because
the other datapoints have no standard instruction mapping. This config exposes the full
datapoint set via tuya-local:

- switch  (DP 1  / switch_1)  — run a cleaning cycle
- select  (DP 101 / scheduler) — auto-run frequency: Every 3 Days / Weekly / Monthly /
  Every 3 Months / Every 6 Months
- select  (DP 103 / duration)  — cycle length: 15 / 30 / 45 / 60 s
- binary_sensor (DP 102 / temp) — status flag (exact meaning unconfirmed; exposed as a
  diagnostic sensor)

Datapoints and mappings were captured live from a physical device over the local
protocol. Sample payload used for the test: {"1": false, "101": "3_days", "102": true,
"103": "15"}.

Notes for review:
- Duration is labeled in seconds based on an observed 60-second cycle and a maximum option of 60. 
- DP 102's meaning is "Status" is deliberately a diagnostic binary sensor. 
- DP 1 is momentary and does not self-reset. Toggle on or off triggers a cycle. 

Validation:
- `uv run pytest -q`: 432 passed, with two AsyncMock coroutine warnings in the existing humidifier tests.
- Ruff lint, import order, formatting, YAML lint, and `untranslated_entities` passed.
- `uv run duplicates the_unclog.yaml` reported no matches above 50%.
- Tests use the current `get_config`, `mock_device`, and `assert_device_properties_set` helpers; the former device-test base class and mixins are no longer on main.
- DEVICES.md is left to the maintainer, as requested by the PR template.

Prepared with AI assistance from the supplied physical-device diagnostics; the profile has been validated in tests and tested on the hardware successfully.

### DIFF
--- a/custom_components/tuya_local/devices/the_unclog.yaml
+++ b/custom_components/tuya_local/devices/the_unclog.yaml
@@ -1,0 +1,54 @@
+name: The Unclog
+products:
+  - id: ir6qiyn5s0pbnbz7
+    name: The Unclog
+entities:
+  - entity: switch
+    icon: "mdi:water-pump"
+    dps:
+      - id: 1
+        type: boolean
+        name: switch
+  - entity: select
+    name: Frequency
+    icon: "mdi:calendar-refresh"
+    category: config
+    dps:
+      - id: 101
+        type: string
+        name: option
+        mapping:
+          - dps_val: "3_days"
+            value: "Every 3 Days"
+          - dps_val: "Weekly"
+            value: "Weekly"
+          - dps_val: "Monthly"
+            value: "Monthly"
+          - dps_val: "3_Months"
+            value: "Every 3 Months"
+          - dps_val: "6_Months"
+            value: "Every 6 Months"
+  - entity: select
+    name: Duration
+    icon: "mdi:timer-cog"
+    category: config
+    dps:
+      - id: 103
+        type: string
+        name: option
+        mapping:
+          - dps_val: "15"
+            value: "15 s"
+          - dps_val: "30"
+            value: "30 s"
+          - dps_val: "45"
+            value: "45 s"
+          - dps_val: "60"
+            value: "60 s"
+  - entity: binary_sensor
+    name: Status
+    category: diagnostic
+    dps:
+      - id: 102
+        type: boolean
+        name: sensor
--- a/tests/devices/test_the_unclog.py
+++ b/tests/devices/test_the_unclog.py
@@ -1,0 +1,95 @@
+"""Tests for The Unclog using the current config and entity test helpers."""
+
+import pytest
+
+from custom_components.tuya_local.binary_sensor import TuyaLocalBinarySensor
+from custom_components.tuya_local.helpers.device_config import get_config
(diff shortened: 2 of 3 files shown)


tests/devices/test_the_unclog.py
make-all, line 1: While this might have been useful in developing the config, with thousands of configs now, it is not scalable to have full tests for every config, so I prefer to remove this and leave the testing to the generic tests.

custom_components/tuya_local/devices/the_unclog.yaml
```diff
@@ -0,0 +1,54 @@
+name: The Unclog
```
make-all, line 1: Top level name should be generic, branding belongs under products.

Filename should also contain some info about the type of device it is, to help users that may have a differently branded version of the device.

custom_components/tuya_local/devices/the_unclog.yaml
```diff
@@ -0,0 +1,54 @@
...
+products:
+  - id: ir6qiyn5s0pbnbz7
+    name: The Unclog
```
make-all, line 4: manufacturer and model should be used rather than name in the products section.

custom_components/tuya_local/devices/the_unclog.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - dps_val: "15"
+            value: "15 s"
+          - dps_val: "30"
+            value: "30 s"
+          - dps_val: "45"
+            value: "45 s"
+          - dps_val: "60"
+            value: "60 s"
+  - entity: binary_sensor
+    name: Status
```
make-all, line 49: Rather than the vague name "Status", it is probably better to use `class: running` here if I am guessing the purpose of this sensor correctly.

custom_components/tuya_local/devices/the_unclog.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - dps_val: "Monthly"
+            value: "Monthly"
+          - dps_val: "3_Months"
+            value: "Every 3 Months"
+          - dps_val: "6_Months"
+            value: "Every 6 Months"
+  - entity: select
+    name: Duration
+    icon: "mdi:timer-cog"
+    category: config
```
make-all, line 34: Use `translation_key: timer` here, to pull in translations for the durations (this can also be used above). Remove the space between the number and the unit to match the existing translation keys, and use "1m" instead of "60s".

The same translation_key may also be used for the Frequecy select above, but current translations are in days, so the values would have to map to "3d", "7d", "30d", "90d" and the non-existent "180d" which can be added to the translations.

make-all (changes requested):

PR #6095, opened 2026-09-06
Add support for Siguro TURBOVac Navigator vacuum

placekmilan:
## Device

Siguro TURBOVac Navigator (model SGR-RV-M300B)
Product ID: 6ewsmyude3artjun

## Config

Based on abir_x9_vacuum.yaml - the datapoint layout is nearly
identical. Removed the entities whose datapoints this model does
not report: dp 27 (resume cleaning), 45 (carpet boost),
48 (y-mopping), 102 (spiral cleaning), 103 (square spiral) and
40 (mop installed).

Confirmed datapoints from the device:
1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 13, 15, 16, 17, 19, 21, 23,
25, 26, 28, 29, 30, 31, 32, 33, 34, 35, 36, 39, 41, 101, 104

dp 104 reports a constant 0 on this model and is exposed as an
attribute for now.

## Testing

Tested on my own device. Start, pause, stop, return to base, fan
speed, mopping level, cleaning mode, volume and language all work.
Fault reporting works too - dp 28 correctly reported dust_box when
the bin was removed.

## Note

If you would rather merge this into abir_x9_vacuum.yaml by adding
the product id there and marking the missing datapoints optional,
that works for me too - happy to redo it that way.

### DIFF
--- a/custom_components/tuya_local/devices/siguro_turbovac_navigator.yaml
+++ b/custom_components/tuya_local/devices/siguro_turbovac_navigator.yaml
@@ -86,22 +86,18 @@ entities:
             value: sleep
           - dps_val: select_room
             value: selecting_room
-          - dps_val: seek_dust_bucket
-            value: seek_dust_bucket
+          - dps_val: in_trouble
+            value: error
+          - dps_val: mopping
+            value: mopping
           - dps_val: collecting_dust
             value: collecting_dust
           - dps_val: self_clean
             value: self_clean
-          - dps_val: mapping
-            value: mapping
-          - dps_val: mapping_done
-            value: mapping_done
       - id: 9
         type: string
         name: fan_speed
         mapping:
-          - dps_val: closed
-            value: "Off"
           - dps_val: gentle
             value: Low
           - dps_val: normal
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/siguro_turbovac_navigator.yaml
```diff
@@ -0,0 +1,480 @@
+name: Siguro TURBOVac Navigator
```
make-all: Top level name should be generic, in case of other devices matching the same config. Branding belongs under products.

custom_components/tuya_local/devices/siguro_turbovac_navigator.yaml
```diff
@@ -0,0 +1,480 @@
...
+    name: Custom room
+    icon: "mdi:floor-plan"
+    category: config
+    dps:
+      - id: 39
+        type: boolean
+        name: switch
+        optional: true
+  - entity: light
+    name: Light
```
make-all: name not needed

make-all (changes requested):

custom_components/tuya_local/devices/siguro_turbovac_navigator.yaml
placekmilan: Both fixed - thanks. Top level name is now the generic
"Robot vacuum" with the branding left in products, and the
redundant name on the light entity is gone.

PR #6096, opened 2026-09-06
Add support for ETA 2625 Sole Smart heater

placekmilan:
## Device

ETA 2625 Sole Smart (model ETA-YBX04-EW), a Czech brand panel heater.
Product ID: gshakqhah24xdssy

## Config

Datapoints taken from the device data model via the Tuya API:

  1    switch          bool
  2    temp_set        5-35 C
  4    mode            CF normal / AF antifreeze / ST weekly program
  7    child_lock      bool
  11   work_state      standby / working
  19   countdown_set   cancel / 1h .. 24h
  20   countdown_left  0-1440 min
  21   fault           bitmap: Err1 probe open, Err2 probe short
  26   week_program9   raw weekly temperature program
  101  fan_mode        H1 / H2 / H3 heating levels
  103  night_mode      bool

The device has no ambient temperature sensor, so the climate entity
exposes a target temperature only.

dp 101 is named fan_mode in the data model but the Chinese label
reads "heating mode" - it selects the heating power level. Exposed
as fan mode as the closest match in the climate platform.

## Note

Until now this device matched taurus_agadir_heater, which only
covers part of it: it misses the ST weekly program mode, night mode
and the weekly program datapoint, and it creates a display light
entity for a datapoint this device does not have.

## Testing

Tested on my own device.

### DIFF
--- a/custom_components/tuya_local/devices/eta_2625_sole_smart.yaml
+++ b/custom_components/tuya_local/devices/eta_2625_sole_smart.yaml
@@ -1,0 +1,208 @@
+name: ETA 2625 Sole Smart
+
+products:
+  - id: gshakqhah24xdssy
+    manufacturer: ETA
+    model: 2625 Sole Smart
+    model_id: ETA-YBX04-EW
+
+# Datapoints confirmed from the device data model via the Tuya API:
+#   1   switch          bool
+#   2   temp_set        5-35 C
+#   4   mode            CF (normal) / AF (antifreeze) / ST (weekly program)
+#   7   child_lock      bool
+#   11  work_state      standby / working
+#   19  countdown_set   cancel / 1h .. 24h
+#   20  countdown_left  0-1440 min
+#   21  fault           bitmap: Err1 (probe open), Err2 (probe short)
+#   26  week_program9   raw, weekly temperature program
+#   101 fan_mode        H1 / H2 / H3 heating levels
+#   103 night_mode      bool
+#
+# The device has no ambient temperature sensor, so the climate
+# entity reports a target temperature only.
+
+entities:
+  - entity: climate
+    dps:
+      - id: 1
+        type: boolean
+        name: hvac_mode
+        mapping:
+          - dps_val: true
+            value: heat
+          - dps_val: false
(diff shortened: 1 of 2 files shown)


custom_components/tuya_local/devices/eta_2625_sole_smart.yaml
```diff
@@ -0,0 +1,208 @@
+name: ETA 2625 Sole Smart
```
make-all: Top level name should be generic in case of other devices matching the config. Branding belongs under products.

custom_components/tuya_local/devices/eta_2625_sole_smart.yaml
```diff
@@ -0,0 +1,208 @@
...
+#   26  week_program9   raw, weekly temperature program
+#   101 fan_mode        H1 / H2 / H3 heating levels
+#   103 night_mode      bool
+#
+# The device has no ambient temperature sensor, so the climate
+# entity reports a target temperature only.
+
+entities:
+  - entity: climate
+    dps:
```
make-all, line 28: Use `translation_key: heater`

custom_components/tuya_local/devices/eta_2625_sole_smart.yaml
```diff
@@ -0,0 +1,208 @@
...
+      - id: 11
+        type: string
+        name: hvac_action
+        mapping:
+          - dps_val: standby
+            value: idle
+          - dps_val: working
+            value: heating
+
+      - id: 21
```
make-all: This is covered by the problem sensor below, not needed here.

placekmilan:
All three addressed:
- top level name is now the generic "Panel heater", ETA moved
  to products
- added translation_key: heater to the climate entity
- removed dp 21 from the climate entity, the problem sensor
  covers it

placekmilan:
Switched the remaining entities to translation keys as the
linter suggested - night_mode and schedule. Also added a note
that dp 103 only accepts being turned on over the local
protocol; turning it off is ignored by the device and only
works from the vendor app or the panel.

